PR #85, opened 2015-04-28
Bravura updated and simple license information added

rettinghaus:


lpugin:
Thanks. Is this a more recent version for Bravura?

rettinghaus:
It's 1.16, directly here from github. Check the link in my Readme.

PR #175, opened 2016-02-14
Correct obvious typos and improve wording of some comments.

maximumspatium:
Hello crews,

while studying Verovio's source code, I started to improve its comments by correcting obvious typos and rewriting some sentences. I share this work in the hope it can be useful for improving code quality and readability. Feel free to correct me if I'm wrong.

Best regards
Max


craigsapp:
All the corrections look good to me (a supposed native-English speaker).

The only thing to note is that there is a case where two lines are merged into one (so don't know if this was intentional or an accident):

```
 -     * Reset method reset all attribute classes.
 -     * No Reset() method required.
 +     * No Reset() method is required.
```

ahwitz:
One instance of nitpicking, since I was looking through at the comments I'd written: view_element.cpp:1109/1119 should actually be "expand it **by** a full radius" for clarity's sake rather than "expand it a full radius" or "expand it to the full radius". The (also clarifiable) "it" is the length of the ledger line in cases where the length needs to be doubled.

Other than that, thanks a ton for this.

lpugin:
Thanks! Could you integrate the changes suggested by @ahwitz and re-submit the pull request to the develop branch instead of master?

lpugin:
PS the line breaks in the comments are handled by the https://github.com/rism-ch/verovio/blob/develop/.clang-format parameters

maximumspatium:
Done. Resubmitted as https://github.com/rism-ch/verovio/pull/176
This one will be closed immediately.

PR #258, opened 2016-07-22
Added replacement for emscripten/build.sh.

craigsapp:
The file `emscripten/buildToolkit` is designed as a replacement for `emscripten/build.sh`.  Original options are the same, but new ones are:

* `-x`: exclude list of fonts, as in `./buildToolkit -x "Gootville, Bravura"`.
* `-D`, `-H`, `-P`, `-X`: exclude DARMS, Humdrum, PAE, MusicXML importers respectively. (`-P` and `-X` don't do anything yet).

Long option aliases are allowed.  For example, `--exclude "Gootville, Bravura"` is the same as `-x "Gootville, Bravura"`.

Example usage of new options: a minimal footprint version of the toolkit could be compiled as:

```bash
     ./buildToolkit -DHPX -v min -x "Gootville, Bravura"
```

(This compiles to a 3MB file as compared to full version at 4MB).

New script also automatically includes all *.cpp files in the `src` directory, so new files will not need to be manually added to the script (libmei files are still added manually).

### DIFF
--- a/emscripten/buildToolkit
+++ b/emscripten/buildToolkit
@@ -1,0 +1,264 @@
+#!/usr/bin/perl
+#
+# Description: Compile script for the JavaScript toolkit version of verovio.
+#
+# Changes from bash script version:
+#   Allow for excluding specific fonts.
+#      Example:      ./buildToolkit -x "Gootville,Bravura"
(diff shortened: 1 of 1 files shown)


craigsapp:
I made a temporary repository for the pull request, so I'll just delete this one and to back to the rism-ch one if everything is set.

Issue #130, opened 2015-07-20
Not able to compile python version under Windows


sonovice:
I am not able to compile the python version under Windows.
The error is pretty short: `../src/iopae.cpp:36:19: fatal error: regex.h: No such file or directory`

The regular regex libs are usually heavily POSIX related and thus not available for Windows. Is there any way to make verovio compilable under Windows?

I am using MinGW32.

lpugin:
Are you interested in using Plain and Easy import? We could make it excluded with a preprocessor option

craigsapp:
In the long term you can implement regular expressions according to the C++
2011 standard which should be cross platform:
     https://solarianprogrammer.com/2011/10/12/cpp-11-regex-tutorial

Currently available in clang, and is available in GCC v 4.9 and higher (but
note that GCC 4.8 is the most common linux compiler at the moment, so still
difficult to work with).

Note that C++ 2011 regular expressions are different than POSIX regular
expressions:
C++ 2011:
    #include <regex>
Posix (C) regular expressions:
    #include <regex.h>

It seems to be possible in Micrsoft settings:
    https://msdn.microsoft.com/en-us/magazine/Dn519920.aspx

-=+Craig

sonovice:
@lpugin Excluding P&E would be great as I solely need MEI. Thanks a lot, Laurent!
Greetings from Detmold

lpugin:
I pushed a modification to the develop branch. It should work for the Python toolkit with

```python setup.py build_ext --inplace --define NO_PAE_SUPPORT```

or for the C++ toolkit with

```cmake . -DNO_PAE_SUPPORT=OFF```

Keep me posted

**Comment by lpugin** [comment]
Thanks @craigsapp , but since C++11 regex is not easy to work with yet I went for the easy fix ;-)

**Comment by sonovice** [comment]
Thanks @lpugin !

**Comment by sonovice** [comment]
Hm, still getting some errors. The first one was a missing `../src/git_commit..h` that I created manually according to the `get_git_commit.sh` file.
No I have plenty of warnings and an error, logfile: http://pastebin.com/rf70jVgc

I am not a C or C++ programmer so I dont really have a clue what happens... :-/

**Comment by lpugin** [comment]
I think it would be best to try to compile the C++ version with cmake first. Could you try it?

**Comment by sonovice** [comment]
I've installed MinGW32 from scratch, as well as cmake. Still getting this error:
```
[...]

PR #282, opened 2016-08-27
Fixed Windows compatibility

sonovice:
Fix for #130

### DIFF
--- a/src/bboxdevicecontext.cpp
+++ b/src/bboxdevicecontext.cpp
@@ -10,6 +10,7 @@
 //----------------------------------------------------------------------------
 
 #include <assert.h>
+#define _USE_MATH_DEFINES // needed by Windows for math constants like "M_PI"
 #include <math.h>
 
 //----------------------------------------------------------------------------
--- a/src/midi/MidiMessage.cpp
+++ b/src/midi/MidiMessage.cpp
@@ -16,6 +16,10 @@
 #include <vector>
 #include <iostream>
 
+#ifdef _WIN32
+#include <iterator>
+#endif
+
 using namespace std;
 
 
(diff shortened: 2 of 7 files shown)


src/bboxdevicecontext.cpp
```diff
@@ -10,6 +10,7 @@
 //----------------------------------------------------------------------------
 
 #include <assert.h>
+#define _USE_MATH_DEFINES // needed by Windows for math constants like "M_PI"
```
lpugin: Can't we have this (and the following) in vrv.h within the #ifndef _WIN32 ?

src/midi/MidiMessage.cpp
```diff
@@ -16,6 +16,10 @@
 #include <vector>
 #include <iostream>
 
+#ifdef _WIN32
+#include <iterator>
```
lpugin: This needs do to be modified in https://github.com/craigsapp/midifile . Maybe @craigsapp will make it without a pull request?

src/bboxdevicecontext.cpp
sonovice: Depends on the include order. I will have a look...

sonovice: Nope, not working. The vrv.h is included after all the other header files.

src/midi/MidiMessage.cpp
sonovice: Would it be okay to leave it here until @craigsapp modifies his repo? I would take care of removing it once the midifile lib has been changed.

lpugin:
Could you try with this fix a27f9b85bffbe7ab858922566d7c6236d20295ea ?

sonovice:
Works :-)

lpugin:
Great. I merged it to develop. Could you check if it works after the merge before I delete the branch?

sonovice:
Already did that, no problems at all.

PR #300, opened 2016-09-22
Fixed compatibility with Visual Studio under Windows

sonovice:


lpugin:
Great! Could you make the pull request to rism-ch:develop?

craigsapp:
The changes to `iohumdrum.cpp` are good.

The changes to humlib are OK:

```cpp
+#ifdef _WIN32
 +    #include <cctype>
 +    #include <functional>
 +    #include <locale>
 +#endif
```

humlib is an external libray, so I will make the changes in humlib directly (which will overwrite these additions).  I will prefer not to have the `ifdef`, and perhaps it is not needed.  `cctype` is for defining character functions such as `tolower()` and `isdigit()`, and which seem to be included in gcc by default (maybe through stdlib.h).  I'll check on what `functional` and `locale` are doing, and remove the `ifdef` if they are also present in gcc.

sonovice:
All of these things are present in gcc, but not in Visual Studio by default, hence the conditional includes. I'm afraid I don't see any way around it.

craigsapp:
That will not be a problem: as the header files exist in gcc and they are standard C++ headers, I just remove the `if` statement and include the three header files regardless of if they are included automatically in gcc (and apple's emulation of gcc).  I will fix it in humlib after Laurent merges your pull request.

PR #311, opened 2016-09-28
Added some info

rettinghaus:


topic03.html
```diff
@@ -10,6 +10,7 @@
 $(".dynam" ).attr("fill", "#d00").attr("stroke", "#d00");
 {% endhighlight %}
 
+This one colors the second staff with color <var>Blue 2<var>.
```
lpugin: This needs to be in a ```<p>``` and needs to be escaped

rettinghaus: Going to try harder next time!

rettinghaus:
This one should be better!

Issue #285, opened 2016-09-06
Chords missing dots in Plain and Easy conversion


lpugin:
Example 8.{'G^D^,B6'D}6.3{EGFA}{GDEF}{GDAD}/{BD''C'D}{3.''D5C3'BAGFED}

rettinghaus:
@lpugin my commit fixes this import problem. 
However: 
In the generated MEI all notes within chords carry duration attributes, which isn't good style. Is this intended, or should I open another issue to address this?

rettinghaus:
this may be closed now

PR #324, opened 2016-10-04
fixed missing dots in chords

rettinghaus:
addresses #285


lpugin:
I think we can also remove the `@dur` and `@dots` on all chord notes here (after the chord is created). E.g

`mnote->SetDur(DURATION_NONE)`

Could you try it?

rettinghaus:
This does the trick. But I'm wondering if there wasn't a cleaner way to do it. Or are you fine with it @lpugin ?

lpugin:
Yes, this looks good. Thanks

PR #325, opened 2016-10-05
gh-pages: new order for features in sidenav

musicEnfanthen:
This PR proposes a new order of the feature examples ("How it works") as requested by @lpugin at @Edirom Summer School 2016. The new order follows that of the corresponding modules in the MEI Guidelines.


lpugin:
Thanks for this proposal. I think having a header corresponding to the guidelines structure is a good idea. However, since we have only a few chapters covered we can maybe remove the figures (1, 4, etc)? The gaps are a bit confusing. 

Styling: you should add list-group-item-menu in the verovio.scss instead of using the twb waring. A light gray (`#efefef`?) would be nicer, I think. We can also have the menu `span` having a reduced height as links `a` in the sub menu since it is already quite long. OK?

musicEnfanthen:
Tried to fix it as you suggested. Really looks better without twb warning.

lpugin:
Nice, thanks!

Issue #323, opened 2016-10-03
Overview


marie-ototoi:
Hi Laurent, 
You asked how I would see this overview I had the feeling was missing, here is a try. 
Is the interpretation correct ? and do you think it could be helpful ? 
(if so I could add it to gh-pages, this is only a screenshot)
Best regards,
Marie
![overview_20161003](https://cloud.githubusercontent.com/assets/1785024/19042772/a1f53a66-898d-11e6-9c50-3b5b6f31e635.png)

lpugin:
Hi Marie,
many thanks for this. This type of help is highly appreciated. A few comments.

Verovio isn't really a set of libraries, I think we can say that it is one library. 

Then, I would say that "Verovio is a C++ codebase that can be integrated into various environments. The C++ Python, Java toolkits are ...

Finally, I am less sure about the last example with the server-side and client-side integration. It looks that the JS toolkit can interact with SVG generated from the server, which is not really the case. One we might want to show here is that in some cases we can have server side processing on the MEI before it is processed by Verovio (server-side or client-side, probably client-side). This can for example be XSLT. What do you think? Other ideas?

I created a gh-pages-develop branch, feel free to make a pull request there (including the source files of the illustrations). I will actually leave this branch since it will be useful for people to make such proposals in the future.

marie-ototoi:
Bonjour Laurent,

Thank you for your feedback and clarification... will try again in the new branch !

A few questions to try and get the whole picture...

Do I understand correctly that people cannot install python or command line compiled versions without  building them themselves from C++ ?
So the audience would be : 
- javascript developers who need to display MEI documents (+ midi output)
- C++ developers who want to set a workflow for treatment and display of MEI files
Do you have an idea of the proportion ?
Can you think of other cases ?

DARMS format is mentioned on the command-line.xhtml page, but not in the Other Formats section, whereas Humdrum is only in the Other Formats... is there any difference in formats handled by one or the other version or is it just the doc that has to be fixed ?

I'm asking all this because I think it would be nice if the user can understand quickly if the lib is the right tool according to his needs (treating and/or displaying and/or "listening" from the score), constraints (formats, infrastructure) and skills (languages mastered)

++++

lpugin:
> Do I understand correctly that people cannot install python or command line compiled versions without building them themselves from C++ ?

[...]

PR #330, opened 2016-10-07
Add an Overview page

marie-ototoi:
Images exported from keynote (source included) - quick and easy tool for that kind of schema
If this is a problem for updates I could draw them in SVG

Did not include info about installation and compilation discussed in #323


craigsapp:
The diagrams all look ok.  "Command line toolkit" would better be termed "Command line interface":
http://www.catb.org/esr/writings/taouu/html/ch02s02.html

A toolkit would be used inside of a command-line interface.  Verovio is a single command (tool) in the CLI, but toolkits are collections of tools, so using the term "toolkit" for verovio in the CLI context does not quite make sense.   For the Java/Python/JavaScript it can be thought more as a toolkit since it is used as a collection of functions to mainpulate MEI data.

I am wondering how CSS is used to manipulate MIDI data :-)

marie-ototoi:
I can leave CSS out if you think there's a risk for misunderstanding.

craigsapp:
There won't be a chance of misunderstanding, just amusement since CSS can't be applied to MIDI data (at least I can't imagine how).

lpugin:
Hi Marie,

It looks pretty good, thanks for this. I'll have a closer look next week. I don't think the arrow to the MIDI is a problem. It is obvious that it refers to the JavaScript and not the CSS. We actually do this with the player.

One thing that we could add is some XSLT on the server, but maybe this is going to be too much and confusing.

craigsapp:
> One thing that we could add is some XSLT on the server, but maybe this is going to be too much and confusing.

Or XSLT processing could be done in the browser as well...

lpugin:
Yes, there is so many ways to use Verovio... ;-)

marie-ototoi:
Hi Laurent,
Sorry, XSLT processing got lost on the way, this was your first comment... I will add it !

lpugin:
I cannot open the keynote file because I have an older version (6.2) ... could you save it for this version? Otherwise we should have the figures as plain svg

marie-ototoi:
In an open world SVG makes more sense

PR #337, opened 2016-10-21
Adding AttColor to accid, note, and rend

rettinghaus:
`@color` on `<note>` is ignored by iomei.cpp for now, as it would color all subelements of note in the svg.


lpugin:
Coloring all note children seems to be the expected behavior. Could you add iomei?

PR #339, opened 2016-10-26
removing g containers in extracted glyphs

rettinghaus:
This update removes the unnecessary `<g>` container from the extracted glyphs putting the necessary `@transform` directly into the `<path>`, reducing the payload, resulting finally in smaller SVGs. 

The problem with the latest version of Bravura was, that it removed the needed `@font-family` from `<font-face>` and all `@glyph-name`.


lpugin:
Good. Have you tested the result with several browsers?

rettinghaus:
Yes. Chrome + Chromium, Firefox, IE, and other viewers. Works great!

lpugin:
I tried to re-generate the font and there is still the problem with the Bravura. Could you revert it? Or we need to fix the issue but I am not sure I got what was changed.

rettinghaus:
Coudn't figure the problem. Try Bravura 1.18, with this version it works fine (for me at least).

Issue #69, opened 2015-01-08
Change SVG style attribute to separate attributes


lpugin:
Instead of having one single style attribute in SVG /g, we could have separate attribute for each parameter (stroke, stroke-opacity, fill, etc). This would make the building of the SVG tree cleaner.

Alternatively, we could have a CSS, but this can be left for a later enhancement.

lpugin:
@kepper do you have any opinion on this?

kepper:
I wouldn't do CSS by default, as it requires people to load another file. Right now, using Verovio is bloody simple, and I'd recommend to keep it like that. Having separate attributes for stroke and fill seems reasonable and might be helpful when someone wants to use the verovio SVG output in other applications (SVG editors etc.). However, I'd say that this doesn't need a very high priority – browsers will render the SVG both ways…

PR #341, opened 2016-10-29
cleaning up svg output

rettinghaus:
This basically removes the Verovio font defaults and puts them into a single style element. That makes styling text elements with CSS much easier and more intuitive. Also it reduces and clears up the output, as things like `@font-family` don't have to be printed over and over again. 

Additionally all style attributes have been replaced by the proper properties, again for easier access via CSS. Also this is more as SVG (Tiny) is meant to be. 

At last, attributes that just carry the SVG default values (e.g. stroke-opacity="1") are no longer printed, resulting in a much lesser cluttered SVG. 

This reduces the size of the rendered Schubert_Lindenbaum.mei by 9% (22149 bytes) and the Chopin Etude by 5% (27039 bytes).

**This all doesn't change the look of the rendering!**

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -159,14 +159,6 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         m_currentNode.append_attribute("id") = gId.c_str();
     }
 
-    if (object->HasAttClass(ATT_VISIBILITY)) {
-        AttVisibility *att = dynamic_cast<AttVisibility *>(object);
-        assert(att);
-        if (att->GetVisible() == BOOLEAN_false) {
-            m_currentNode.append_attribute("visibility") = "hidden";
-        }
-    }
-
     if (object->HasAttClass(ATT_COLOR)) {
         AttColor *att = dynamic_cast<AttColor *>(object);
         assert(att);
@@ -176,6 +168,14 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         }
     }
 
+    if (object->HasAttClass(ATT_VISIBILITY)) {
+        AttVisibility *att = dynamic_cast<AttVisibility *>(object);
+        assert(att);
+        if (att->GetVisible() == BOOLEAN_false) {
+            m_currentNode.append_attribute("visibility") = "hidden";
+        }
+    }
+    
     // m_currentNode.append_attribute("style") = StringFormat("stroke: #%s; stroke-opacity: %f; fill: #%s; fill-opacity:
     // %f;",
     // GetColour(currentPen.GetColour()).c_str(), currentPen.GetOpacity(), GetColour(currentBrush.GetColour()).c_str(),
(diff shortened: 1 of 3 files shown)


src/svgdevicecontext.cpp
```diff
@@ -240,6 +240,11 @@ void SvgDeviceContext::StartPage()
     // Initialize the flag to false because we want to know if the font needs to be included in the SVG
     m_vrvTextFont = false;
 
+    // default styles
+    m_currentNode = m_currentNode.append_child("style");
+    m_currentNode.append_child(pugi::node_pcdata).set_value("g.page-margin{font-family:Times;} g.tempo{font-weight:bold;} g.dir, g.dynam {font-style:italic;}");
```
rettinghaus: Definitely they should! But I leave that to Laurent as for naming conventions and stuff.

src/svgdevicecontext.cpp
```diff
@@ -159,14 +159,6 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         m_currentNode.append_attribute("id") = gId.c_str();
     }
 
-    if (object->HasAttClass(ATT_VISIBILITY)) {
```
lpugin: Any reason why this was moved below?

src/svgdevicecontext.cpp
```diff
@@ -240,6 +240,12 @@ void SvgDeviceContext::StartPage()
     // Initialize the flag to false because we want to know if the font needs to be included in the SVG
     m_vrvTextFont = false;
 
+    // default styles
+    m_currentNode = m_currentNode.append_child("style");
+    m_currentNode.append_attribute("type") = "text/css";
+    m_currentNode.append_child(pugi::node_pcdata).set_value("g.page-margin{font-family:Times;} g.tempo{font-weight:bold;} g.dir, g.dynam {font-style:italic;}");
```
lpugin: Concerning @ahankinson comment about it: it would indeed be better, but having all of them here is already better than before. One step forward. Eventually, we should pass a `Style` object to `StartPage` and get them for there. I am fine leaving this like this for now.

src/svgdevicecontext.cpp
```diff
@@ -253,7 +259,7 @@ void SvgDeviceContext::StartPage()
     m_currentNode.append_attribute("class") = "page-margin";
     m_currentNode.append_attribute("transform")
         = StringFormat("translate(%d, %d)", (int)((double)m_originX), (int)((double)m_originY)).c_str();
-    m_currentNode.append_attribute("style") = "stroke: #000; stroke-opacity: 1.0; fill: #000; fill-opacity: 1.0";
+    m_currentNode.append_attribute("style") = "stroke: #000; fill: #000;";
```
lpugin: Is stroke and fill opacity assumed to be 1.0 by default?

rettinghaus: `fill-opacity`, `stroke-width` and `stroke-opacity` have all an initial value of 1 and they get inherited. See [here](https://www.w3.org/TR/SVG/painting.html).

src/svgdevicecontext.cpp
rettinghaus: I thought it would be nicer and a cleaner approach to have the attribute classes here in alphabetical order just like everywhere else, too. 
… and of course the svg attributes, as I tried to achieve.

src/svgdevicecontext.cpp
lpugin: OK, perfect.

rettinghaus:
Just noticed that this mainly addresses #69.

lpugin:
Indeed. Thanks!

Issue #278, opened 2016-08-20
Staff and Ledger line groupings in SVG renderings


craigsapp:
It would be useful for staff lines and ledger lines to be grouped in SVG renderings.  Currently they are not:

![screen shot 2016-08-20 at 9 31 27 am](https://cloud.githubusercontent.com/assets/3487289/17829741/e51046d0-66b8-11e6-8ee5-1b7e3464225c.png)

``` xml
  <g class="staff" id="staff-000000057062299">
      <path d="M0 1145 L3245 1145" style="stroke-width: 20;" />
      <path d="M0 1325 L3245 1325" style="stroke-width: 20;" />
      <path d="M0 1505 L3245 1505" style="stroke-width: 20;" />
      <path d="M0 1685 L3245 1685" style="stroke-width: 20;" />
      <path d="M0 1865 L3245 1865" style="stroke-width: 20;" />
      <g class="layer" id="layer-000000040212117">
         <g class="clef" id="clef-000000010423556">
            <use xlink:href="#E050" x="117" y="1685" height="720px" width="720px" />
         </g>           
         <g class="meterSig" id="msig-000000031143706">
            <use xlink:href="#E081" x="750" y="1325" height="720px" width="720px" />
            <use xlink:href="#E084" x="717" y="1685" height="720px" width="720px" />
         </g>           
         <g class="note" id="note-L6F1">
            <path d="M1658 2045 L1982 2045" style="stroke-width: 20;" />
            <path d="M1658 2225 L1982 2225" style="stroke-width: 20;" />
            <path d="M1658 2405 L1982 2405" style="stroke-width: 20;" />
            <use xlink:href="#E0A4" x="1707" y="2495" height="720px" width="720px" />
            <rect x="1913" y="1505" width="20" height="968" rx="0" style="stroke-width: 0;" />
         </g>           
      </g>      
   </g> 
```

For staff lines:

``` xml
      <path d="M0 1145 L3245 1145" style="stroke-width: 20;" />
      <path d="M0 1325 L3245 1325" style="stroke-width: 20;" />
      <path d="M0 1505 L3245 1505" style="stroke-width: 20;" />
      <path d="M0 1685 L3245 1685" style="stroke-width: 20;" />
      <path d="M0 1865 L3245 1865" style="stroke-width: 20;" />
```

should become:

``` xml
       <g class="staffLines">
          <path d="M0 1145 L3245 1145" style="stroke-width: 20;" />
[...]

PR #343, opened 2016-10-31
finishing SVG cleaning

rettinghaus:
removes more unnecessary initial values:
* `@fill="black"`
* `@fill-rule="nonzero"`

for `<rect>` `@rx` now gets only set if it’s non zero:
> If neither ‘rx’ nor ‘ry’ are properly specified, then set both rx and
ry to 0.
(https://www.w3.org/TR/SVG/shapes.html)

And finally every path now gets a proper `stroke`. This seems to blow up
the SVG again, but setting black in stroke as a default value gives every sub
element an outline, which is bad SVG. And that way we would basically have to set
stroke-width to zero for every non-path element. Rendering text gets so
cleaner and coloring text gets easier as we only have to set `fill`. 
With that the last `@style` vanished. 

If #278 is addressed `stroke` and `stroke-width` may get moved up in the surrounding `g`.

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -198,10 +198,7 @@ void SvgDeviceContext::StartTextGraphic(Object *object, std::string gClass, std:
     if (object->HasAttClass(ATT_COLOR)) {
         AttColor *att = dynamic_cast<AttColor *>(object);
         assert(att);
-        if (att->HasColor()) {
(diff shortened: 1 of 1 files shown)


lpugin:
This is a great improvement for text. Thanks! There is one issue with slurs, ties, and braces, which changed. They should have a stroke in order to have a more roundish end. Could you look at it?

rettinghaus:
Found it and fixed it. Sorry I didn't noticed that in the first place.

PR #347, opened 2016-11-04
adding support for rest positioning in MusicXML importer

rettinghaus:
This brings the capability of reading and converting positions of rests and measure rests to the MusicXML importer. 

Also this fixes the annoying
> Unsupported key signature '0s'

### DIFF
--- a/src/mrest.cpp
+++ b/src/mrest.cpp
@@ -13,7 +13,7 @@ namespace vrv {
 // MRest
 //----------------------------------------------------------------------------
 
-MRest::MRest() : LayerElement("mrest-"), AttVisibility(), AttFermatapresent(), PositionInterface()
+MRest::MRest() : LayerElement("mrest-"), PositionInterface(), AttVisibility(), AttFermatapresent()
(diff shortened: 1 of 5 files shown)


src/mrest.cpp
```diff
@@ -13,8 +13,9 @@ namespace vrv {
 // MRest
 //----------------------------------------------------------------------------
 
-MRest::MRest() : LayerElement("mrest-"), AttVisibility(), AttFermatapresent()
+MRest::MRest() : LayerElement("mrest-"), AttVisibility(), AttFermatapresent(), PositionInterface()
```
lpugin: This needs to be in the same order as in the header (some compilers will complain about it)

rettinghaus:
Sorry I missed that.

rettinghaus:
I added support for hidden notes (`@visibility`) and rests (`<space>`)

rettinghaus:
xml:ids weren't referenced correct in the MEI output. That is fixed now.

PR #371, opened 2016-11-26
improved fermata

rettinghaus:
I added support for `@shape` on fermata. Without `att.extsym` we can't access the "veryshort" and "verylong" fermatas, so I commented them out for now.


lpugin:
you can leave smufl.h uncommented because this file is generated and will be overwritten.

rettinghaus:
so I commented them out in `supported.xsl`

lpugin:
OK, I am actually leaving them since their are in the Leipzig font

PR #376, opened 2016-12-05
Adds support for transpositions and hairpins in MusicXML importer

rettinghaus:
Hairpins are using ids for end points, so MusicXML wedges starting and ending on the same note will not get rendered.


lpugin:
> Hairpins are using ids for end points, so MusicXML wedges starting and ending on the same note will not get rendered.

Do we have a warning for this? If not can you add one?

rettinghaus:
Should that be a warning only for the importer or more global on the svg renderer?

lpugin:
In the renderer, I think.

PR #378, opened 2016-12-05
Adding iomidi

rettinghaus:
This adds iomidi.h/iomidi.cpp. Although they won't do much for now, it's a start. 
I did some renaming in the toolkit and main.cpp to bring MIDI and MEI output into a consistent naming scheme. 
Also this adds simple support for `@tie` on notes.


include/vrv/toolkit.h
```diff
@@ -62,7 +62,7 @@ class Toolkit {
     /**
      * Save an MEI file.
      */
-    bool SaveFile(const std::string &filename);
+    bool SaveMeiFile(const std::string &filename);
```
lpugin: There is only one native format for Verovio, namely MEI. So it should stay Save

include/vrv/iomidi.h
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * The main method for exporting the file to MEI.
+     */
+    virtual bool ExportFile();
+
+    /**
+     * Export the document to a MIDI file.
+     * Run trough all the layer and fill the midi file content.
+     */
+    void ExportMIDI(MidiFile *midiFile);
```
lpugin: I guess the idea is to move the content of Doc::ExportMIDI here. This is good. However, was we should do it to have the MidiFile class inside the MidiOutput class (and not the Toolkit). We could make it aligned with the MeiOutput class, namely to have an ExportFile() method and a GetOutput() method. The second one returns the output as a std::string, which has to be 64based encoded for MIDI. This is already happening in Toolkit::RenderToMidi

include/vrv/toolkit.h
rettinghaus: Changed it back

include/vrv/toolkit.h
```diff
@@ -111,7 +111,7 @@ class Toolkit {
      * Creates a midi file, opens it, and writes to it.
      * currently generates a dummy midi file.
      */
-    bool RenderToMidiFile(const std::string &filename);
```
rettinghaus: Is it really a rendering method? Not more of an export?

lpugin: I see it more as a rendering than a saving. Why would you save as MIDI ;-)? It does not really matter, but since we have renderToMidi it is introducing even more confusion to change it to SaveMidiFile.

PR #390, opened 2016-12-20
Adding language support

rettinghaus:
This adds support for `@xml:lang` on `<dir>`, `<rend>`, `<syl>`, `<tempo>`, and `<verse>` in MEI and MusicXML import.  It gets written to the corresponding g node in the SVG rendering and thus can be styled with CSS.

### DIFF
--- a/include/vrv/dir.h
+++ b/include/vrv/dir.h
@@ -20,7 +20,11 @@ class TextElement;
 // Dir
 //----------------------------------------------------------------------------
 
-class Dir : public ControlElement, public TextListInterface, public TextDirInterface, public TimeSpanningInterface {
+class Dir : public ControlElement,
+            public TextListInterface,
+            public TextDirInterface,
+            public TimeSpanningInterface,
+            public AttLang {
 public:
     /**
      * @name Constructors, destructors, and other standard methods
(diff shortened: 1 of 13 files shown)


lpugin:
Thanks for this. Do you think we should have some documentation in the gh-pages for it?

rettinghaus:
I can add this if you tell me where it should be found. In the features section along with *Tempi and directives* or *Lyrics*? Or a short notice on both? 

Speaking of: perhaps there should be an overview of supported elements and attributes as a simple table in the wiki.

lpugin:
Maybe in _Lyrics_ and then refer to it from other places?

Supported elements: there was something in place but it is outdated. Namely a yaml file in the gh-pages branch (./supported.yml) from which this page was generated http://www.verovio.org/supported-mei.xhtml

The idea is to list all the attribute classes used, with possible comments for each attribute, and then the list of element classes and specify what attribute classes they implement. This is maybe over the top and too much work to maintain. Using the wiki is an idea. Alternatively we can point people to this http://www.verovio.org/doxygen/html/hierarchy.html ?

rettinghaus:
@lpugin I will have a look at the yml file. This is a good and user friendly approach. For the technically advanced audience we should then add a link to http://www.verovio.org/doxygen/html/hierarchy.html 

@craigsapp Do you want to add `@xml:lang` support now to your humdrum importer?

PR #412, opened 2017-01-04
 improvements for tuplets and grace notes

rettinghaus:


craigsapp:
Looks good.  Note that this is anticipation of `@grace="acc"` displaying without a slash by default and `@grace="unacc"` without a slash by default (so currently will render backwards in verovio).

Issue #422, opened 2017-01-12
MEI staves to MIDI tracks


craigsapp:
The MIDI converter in verovio creates tracks for each staff in a MIDI file, but all data is stored in the first track.  Instead, each staff's notes should be placed in a separate track.

Here is an example score:

<img width="497" alt="screen shot 2017-01-12 at 04 02 31" src="https://cloud.githubusercontent.com/assets/3487289/21889106/1e108046-d87c-11e6-8a0c-d50c393ecce2.png">

MEI data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2017-01-12 04:04:22</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.13-dev-6997da1-dirty</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-000000156782000" midi.bpm="60">
                        <staffGrp xml:id="m-000000109657593" symbol="brace" barthru="true">
                            <staffDef xml:id="staffdef-000000044704737" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="1" lines="5" />
                            <staffDef xml:id="staffdef-000000007869779" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="2" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-000000086163972">
                        <measure xml:id="measure-L4" n="1" right="end">
                            <staff xml:id="staff-L4F2" n="1">
                                <layer xml:id="layer-L4F2" n="1">
                                    <note xml:id="note-L5F2" dur="1" oct="5" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
[...]

PR #425, opened 2017-01-13
MEI staves to MIDI tracks

rettinghaus:
This addresses #422 and adds MEI labels as midi track names.


lpugin:
Could you make the PR to develop directly? (develop-midi is somehow unstable with some timemap stuff)

PR #440, opened 2017-01-18
small fix for XML excerpts

rettinghaus:


lpugin:
Thanks, but I just fixed these.

PR #451, opened 2017-01-23
Metadata conversion

rettinghaus:
This brings many improvements to the metadata conversion from humdrum to MEI.


src/iohumdrum.cpp
```diff
@@ -631,7 +716,7 @@ string HumdrumInput::getDateString(void)
 {
     time_t t = time(0); // get time now
     struct tm *now = localtime(&t);
-    string dateStr = StringFormat("%d-%02d-%02d %02d:%02d:%02d", now->tm_year + 1900, now->tm_mon + 1, now->tm_mday,
+    string dateStr = StringFormat("%d-%02d-%02dT%02d:%02d:%02d", now->tm_year + 1900, now->tm_mon + 1, now->tm_mday,
```
craigsapp: Is putting a "T" between the date and time standard?  I think I copied this from elsewhere in the verovio code, so it would be good to have them match one way or the other.

src/iohumdrum.cpp
```diff
@@ -671,22 +756,22 @@ void HumdrumInput::getRespPeople(vector<vector<string> > &respPeople, vector<Hum
 {
 
     // precalculate a reference map here to make more O(N) rather than O(N^2)
-    addPerson(respPeople, references, "COM", "Composer"); // cmp
+    addPerson(respPeople, references, "COM", "composer"); // cmp
```
craigsapp: Lower case is fine, but remember that I was using upper case to indicate these are authority labels, and the lower-case ones were labels that I defined since I couldn't find a match in the LOC list.

src/iohumdrum.cpp
```diff
@@ -796,14 +881,31 @@ void HumdrumInput::insertTitle(pugi::xml_node &titleStmt, const vector<HumdrumLi
...
+          }
+        }
+        else if (key.compare(0, 3, "OTA") == 0) {
+          title.append_attribute("type") = "alternative";
+          if (lang) {
+              title.append_attribute("xml:lang") = language.c_str();
+          }
+        }
+        else if (key.compare(0, 3, "OTP") == 0) {
+          title.append_attribute("type") = "uniform";
```
craigsapp: `OTP` is the "popular title", are you sure you want to call that the "uniform" title?

http://www.music-cog.ohio-state.edu/Humdrum/guide.append1.html

src/iohumdrum.cpp
rettinghaus: This makes it an isodate, so MEI validates. (Check Guidelines)

src/iohumdrum.cpp
rettinghaus: On title only some types are allowed, so "uniform" fits best here.

src/iohumdrum.cpp
```diff
@@ -756,7 +841,7 @@ void HumdrumInput::insertTitle(pugi::xml_node &titleStmt, const vector<HumdrumLi
         plang = false;
         lang = false;
         key = references[i]->getReferenceKey();
-        if (key.compare(0, 3, "OTL") != 0) {
+        if (key.compare(0, 2, "OT") && key.compare(0, 1, "X")) {
```
craigsapp: This condition is cryptic, but studying it for a while it makes sense.  It means: don't process keys starting with neither "OT" or "X" (compare returns false if they match which was the confusing part).  Keys starting with "OT" or "X" are all title-related ("X" ones are in an older style which I replaced with the `@`/`@@` language modifiers which affect any reference key).

craigsapp:
src/iohumdrum.cpp changes look good.

src/iohumdrum.cpp
craigsapp: I am too primitive to validate :-)

pe-ro:
NO! The uniform title is a standard title assigned by an authority. A so-called popular title is nothing like that.

craigsapp:
Is there a place that the "popular title" can go?  I.e., this is a title not given by the composer -- typically the publisher so that it will help with sales.

"Tempest" sonata by Beethoven:

```
!!!COM: Beethoven, Ludwig van
!!!CDT: 1770///-1827///
!!!OTL: Piano Sonata no. 17 in D minor
!!!OTP: tempest
```

Here is the "uniform title" (in the USA):

```
   Sonatas, piano, no. 17, op. 31, no. 2, D minor
```
In Stanford library catalog "Tempest" and "Sturm" are listed at "title variation".

Example MARC21 fields for the work:

screen shot 2017-01-23 at 16 51 21

Field 246 "Varying Form of Title":

https://www.loc.gov/marc/bibliographic/bd246.html

src/iohumdrum.cpp
craigsapp: This is approximately the list I was using for roles:
https://www.loc.gov/marc/relators/relaterm.html

Here is a related discussion from @pe-ro to me in March 2013:

Moving on, just for fun, you can combine `<composer>` and `<persName>` like so --
```xml
<composer type="attributed">
  <persName authority="LCNAF" authURI="http://lccn.loc.gov/n80005115">Josquin</persName>
</composer>
```
This asserts that Josquin is the attributed composer and that the authorized form of "Josquin" (in the Library of Congress Name Authority File) is stored at http://... (which is "Josquin, des Prez, d. 1521", by the way).

 
Currently, there's no place to record the authorized form of the name directly in MEI.  The quick and dirty way to do this is to use an attribute (something like `@reg`) to store the authorized name, but that's "overloading" an attribute.  On the other hand, providing for multiple options within `<composer>` complicates the markup --

```xml
<composer type="attributed">
  <choice>
    <orig>Josquin</orig>
    <reg>
      <persName authority="LCNAF" authURI="http://lccn.loc.gov/n80005115"/>
    </reg>
  </choice>
</composer>
```

but this isn't allowed yet.  I'm not sure `<choice>` is a good idea for use in the header.

pe-ro:
`title/@type="alternative"` covers any titles by which a work is known other than those provided for in other options; that is, main, subordinate, abbreviated, or translated. So 
```
<title type="alternative">Sturm</title>
<title type="alternative">Tempest</title>
```
is the way to go for popular titles.

`@label` can be used to record a "helpful hint" for the title, e.g.,
```
<title type="alternative" label="Popular Title">Sturm</title>
<title type="alternative" label="Popular Title">Tempest</title>
```
Adding `@analog` is the way to equate this kind of title; that is, an alternative, with the 246 field in MARC, e.g.,
```
<title type="alternative" label="Popular Title" analog="marc:246">Sturm</title>
<title type="alternative" label="Popular Title" analog="marc:246">Tempest</title>
```
There's no simple way to indicate that one of the alternative titles is a translation of the other, but adding a language attribute
```
<title type="alternative" label="Popular Title" analog="marc:246" xml:lang="de">Sturm</title>
<title type="alternative" label="Popular Title" analog="marc:246" xml:lang="en">Tempest</title>
```
should be sufficient in most cases.

pe-ro:
> On title only some types are allowed, so "uniform" fits best here.

The list of values for `title/@type` is not closed.  In other words, other values are permitted in addition to the suggested ones.  The suggested values are very well-considered ones that should fit in the vast majority of cases, but suggestions nonetheless.

> Lower case is fine, but remember that I was using upper case to indicate these are authority labels, and the lower-case ones were labels that I defined since I couldn't find a match in the LOC list.

What's being communicated here with the change in case is actually a change in authority. But
```
<persName resp="CMP" authority="CraigsRoleList">Perry Roland</persName>
```
won't work because `@authority` applies to the content of the element; that is, "Perry Roland", not the resp attribute.

So, what's necessary is to separate the name and the role --
```
<persName role="CMP">Perry Roland</persName>
<resp codedval="CMP" authority="CraigsRoleList">AttributedComposer</resp>
```
which says the name "Perry Roland" has been assigned the role of "CMP" and that there's a role of "AttributedComposer" in Craig's List o' Roles that has been assigned the code of "CMP".  The value "CMP" in `@role` and `@codedval` establishes the link between my name and "AttributedComposer".

Read the paragraph above VERY carefully and you'll see that I didn't assert that there's a link between **me** and the role of "composer", only between **my name** and the role.  MEI does not yet encode people, only their names.  If you want to do prosopography, then use TEI. :grinning:

lpugin:
@craigsapp I leave you merge it

rettinghaus:
What about using `@analog` for adding the Humdrum Reference Record `<title type="alternative" label="popular" analog="humdrum:OTP">tempest</title>`?

craigsapp:
> What about using @analog for adding the Humdrum Reference Record `<title type="alternative" label="popular" analog="humdrum:OTP">tempest</title>`?

That sounds good, but perhaps overkill.  But it would be useful when converting back into Humdrum again...

craigsapp:
Your composition:

```
**kern
*clefG2
*M4/4
1r
```

Should more properly be:

```
**kern
*clefG2
*M4/4
1r
*-
```

In other words, adding `*-` to terminate the data started by `**kern`.  (The parser I am using in verovio doesn't care about it, but it is needed in order to validate a Humdrum file :-)

ahankinson:
Slightly off topic, but do the `humdrum:` and `marc:` attributes need some sort of namespace? They look a bit strange without an URI to define them, but I don't know if attribute values can have a namespace....

pe-ro:
The attributes themselves aren't in the humdrum or marc namespace.  Actually, the values aren't either -- prefacing the value with, for lack of a better term, a "pseudo-namespace" is a quick-and-dirty way of constraining the value space of `@analog`.  As far as I know, there's no real method of doing this.

PR #453, opened 2017-01-24
changed string to std::string

rettinghaus:
This PR changes `string` to `std::string` adopting normal Verovio coding style.

### DIFF
--- a/include/vrv/iohumdrum.h
+++ b/include/vrv/iohumdrum.h
@@ -62,15 +62,15 @@ namespace humaux {
         ~HumdrumTie();
         HumdrumTie &operator=(const HumdrumTie &anothertie);
         void insertTieIntoDom(void);
-        void setStart(const string &id, Measure *starting, int layer, const string &token, int pitch,
+        void setStart(const std::string &id, Measure *starting, int layer, const std::string &token, int pitch,
             hum::HumNum starttime, hum::HumNum endtime);
-        void setEnd(const string &id, Measure *ending, const string &token);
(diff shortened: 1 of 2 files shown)


craigsapp:
Adding the prefixes to include/vrv/iohumdrum.h is proper.  They were missing but there is no compiler error, probably because there is a `using namespace std;` embedded in include/hum/humlib.h (that I should remove).

For the `std::` prefixes in src/iohumdrum.cpp I am less keen due to decreased readability and increase in time to type them when adding new code, etc.  But I will let you add them, and I can remove them again later if I don't like them.  There will probably be other things that will need a `std::` prefix, and there will be a complication in identifying them since humlib.h will also interact to hide their need.

Here are the current `using` statements in iohumdrum.cpp:

```
using namespace hum; // humlib  namespace
using namespace vrv; // verovio namespace
using namespace pugi; // pugixml namespace
using namespace std;
```

The `pugi` namespace is not necessary since I already prefixed pugi library variables with the namespace to keep track of where they come from (so that `using` line can be removed.  And then the `using namespace std;` line should be removed as well (but there will still be a hidden one in humlib.h).   I will do that because I have to merge conflicts since we are working on the same file at the same time.

craigsapp:
There actually were no conflicts, so no problems so far (I thought github was saying there were conflicts).

craigsapp:
This commit removes the using statements:

https://github.com/rism-ch/verovio/commit/e3bc1f20d63334115ed4b9fa2bb9c04a1a9dea54

I will check what you just pushed...

Issue #458, opened 2017-01-26
Proposed head.shape implementation


billmoser:
Hello -

I have a proposal for implementing the head.shape attribute for noteheads.
I've implemented this in my copy of master.  Attached are sample mei and pdf
illustrating the I/O.  Details of the changes are also attached (with my comments
interspersed).  I'll summarize them here:
  1)  several glyphs have been uncommented in fonts/supported.xsl, requiring
    generate_all.sh to be rerun.  This of course has the side effect that 
    several other files in data and include have been (re)regenerated
  2) "Boilerplate" has been added to include/vrv/note.h, src/iomei.cpp, and
    src/note.cpp in order to allow the head.shape attribute to be recognized.
  3) src/view_element.cpp has two additions:
        a) a function which translates MEI head.shape strings into SMuFL codes
        b) a 3-line conditional in View::DrawNote which replaces the 
           fontNo for the notehead if the head.shape attriute exists

I have a few questions:
1) Do you want these changes? If so, how should I do the pull request, 
   from a fork or a branch, and a fork or branch of what?
2) There are many more SMuFL glyphs than there are MEI head.shapes. 
   Is sticking with the MEI subset necessary, or is it possible for me to go
   with the full set of SMuFL codes/strings.  This would then differ from MEI's
   data.HEADSHAPE.list
3) is view_element.cpp the appropriate place for the string/code converter?
4) my notehead replacement is simple and context-free.  That works for me, but 
   does it work for the project overall?

Regards,
Bill

Examples:

[head-shape-mei.txt](https://github.com/rism-ch/verovio/files/733675/head-shape-mei.txt)
[head-shape.pdf](https://github.com/rism-ch/verovio/files/733669/head-shape.pdf)

Changes:

[Changes.txt](https://github.com/rism-ch/verovio/files/733683/Changes.txt)

ahankinson:
Hi Bill,

This is more a comment on the mechanics of this contribution, rather than its implementation. I will point you to the contribution guidelines for the MEI repo, since they are largely the same as the practices for Verovio:

https://github.com/music-encoding/music-encoding/blob/develop/CONTRIBUTING.md#how-to-contribute

[...]

PR #459, opened 2017-01-27
The commit implements the head.shape attribute for notes.

billmoser:
Here is an overview of the changes:
 1) several glyphs have been uncommented in fonts/supported.xsl, requiring
    generate_all.sh to be rerun.  This of course has the side effect that
    several other files in data and include have been (re)regenerated
 2) "Boilerplate" has been added to include/vrv/note.h, src/iomei.cpp, and
    src/note.cpp in order to allow the head.shape attribute to be recognized.
 3) src/view_element.cpp has two additions:
        a) a function which translates MEI head.shape strings into SMuFL codes
        b) a 3-line conditional in View::DrawNote which replaces the
           fontNo for the notehead if the head.shape attribute exists

I have a few questions/concerns:
1) There are many more SMuFL glyphs than there are MEI head.shapes.
   Is sticking with the MEI subset necessary, or is it possible for me to go
   with the full set of SMuFL codes/strings.  This would then differ from MEI's
   data.HEADSHAPE.list
2) Is view_element.cpp the appropriate place for the string/code converter?
3) This notehead replacement is simple and context-free.  As @craigsapp notes
   in issue #458, the note stems are not attaced to the heads for the new
   glyphs.  Its ugly -- not sure how to fix it.


lpugin:
@billmoser thanks for this. Could you make the pull request to the rism-ch/develop branch (and not master)? Also, I have a few questions, so could you contact me be email at laurent.pugin@rism-ch.org ?

pe-ro:
A few observations --

1. Use of `@altsym` or `@extsym` on `<note>` (or any other element, in fact) implies that there's a one-to-one mapping between the element and a SMuFL codepoint.  Currently, you can't use one of these attributes to replace just the notehead or just the stem.
2. The values provided in `data.HEADSHAPE.list` (used by `@head.shape`) are merely suggestions.  In fact, any NMTOKEN value can be used.  So, theoretically, one can use a SMuFL name (e.g., "noteShapeArrowheadLeftWhite"), a SMuFL code point (e.g., "U+E1C8"), or some other value like "fooBar".  Ultimately, however, it's up to the consuming software to determine which values it can/cannot support.
3. Relying on SMuFL too much can lead to privileging the visual over the semantic.
3. I believe adding `<noteHead>` and `<stem>` children to `<note>` is preferable to extending `@altsym` and `@extsym` to permit multiple (presumably compose-able) values.  Treating note heads and stems as objects in their right (as we've already done with accidentals and articulations) allows these things to have their own attributes.  This approach, however, is not without its down side.  We'll have to approach this carefully.

billmoser:
FYI, I have closed this pull request and opened a new one per @lpugin's comment above.

ahankinson:
@pe-ro what about adding `@head.sym` and `@stem.sym` on note?

ahankinson:
Note that it should be possible to change the base branch of a pull request without closing and opening a new one.

rettinghaus:
@ahankinson How to? I couldn't figure it out neither.

pe-ro:
@ahankinson, adding `@head.sym` and `@stem.sym` is a possibility as well.  Presumably, `@head.sym` and `@stem.sym`, like `@altsym` would point to symbols defined inside the MEI file.  This means we'd also need to add `@head.extsym` and `@stem.extsym` to accommodate externally-defined (i.e., SMuFL) symbols.  This swells the number of attributes.  More annoying, unless head and stem are promoted to element status there's no place to attach `@authority` and `@authURI` attributes to capture the fact that the external symbols I'm using **aren't** SMuFL or to indicate where the canonical definition of the SMuFL symbol is located.

ahankinson:
Can we put `@authority` and `@authURI` on the symbol, and not on the actual object? This would make more sense. So `@head.sym` points to a symbol IDREF that defines the internal or external symbol. I think this would make more sense, since the authority and authURI  refers to the symbol, not the stem or head. (In other words, `@authority` on `<noteHead />` wouldn't be as precise as `@authority` on `<symbol />`.) 

Something like this in pseudo-MEI:

```xml
<symbol xml:id="s123" authority="http://smufl/" data="u1234" /> <!-- SMuFL codepoint -->
<symbol xml:id="s456">(svg symbol)</symbol>  <!-- embedded SVG -->
...
<note head.sym="#s123" stem.sym=#s456 />
```

pe-ro:
The indirection used in your example is exactly what happens already in the case of user-defined symbols, for example --
```
<symbolTable xmlns="http://www.music-encoding.org/ns/mei" xmlns:svg="http://www.w3.org/2000/svg">
  <symbolDef xml:id="mySymbol1">
    <curve startid="#mySymbol1" endid="#mySymbol1" lform="dotted" lwidth="0.2mm" curvedir="above" bulge="10 50%"/>
    <graphic startid="#mySymbol1" target="www.mySite.org/cuteGraphic.png" width="10px" height="11px"/>
    <graphic target="www.mySite.org/cuteGraphic.png" width="10px" height="11px" ulx="1" uly="1"/>
    <line startid="#mySymbol1" x2="1" y2="2" ho="0" vo="0" endho="0" endvo="0"/>
    <symbol startid="#mySymbol1" glyphnum="#x000" authority="notSMuFL"/>
    <symbol startid="#mySymbol1" glyphnum="U+E770" authority="SMuFL"/>
  </symbolDef>
  <symbolDef xml:id="mySymbol2">
    <svg:svg/>
  </symbolDef>
</symbolTable>
```
A symbol definition may contain either MEI "graphic primitives" (line, curve, graphic, and symbol elements) or an `svg:svg` element, but not both.  `<symbol>` can point an external symbol defined by SMuFL or some other mechanism.

Is this what you had in mind?  If so, it's a little convoluted; that is, using `@head.sym` to point to an internally-defined symbol that in turn references an external symbol.

ahankinson:
Yes, it's what I had in mind. I suspect it's no less convoluted than having separate noteHead and stem elements, each with their own authority URI entry pointing to an external code point. As I see it, the advantage of 'routing' it all through symbol is that we get to use existing mechanisms, rather than needing to define new ones -- we only have to define and specify the behaviour of `@*.sym` as pointing to an internally-defined symbol, rather than needing to expand the constellation of elements available for note definitions.

src/view_element.cpp
```diff
@@ -47,6 +47,21 @@
 #include "verse.h"
 #include "vrv.h"
 
+/**
+  * Translates a MEI head.shape attribute value to a SMuFL glph code
+  * (see http://music-encoding.org/documentation/3.0.0/data.HEADSHAPE.list)
+  * Note that only a few of the strings have been mapped thus far.
+ **/
+wchar_t getFontNumber(std::string s)
```
lpugin: Could you move the function to a Note::GetHeadShapeGlyph method?

src/view_element.cpp
```diff
@@ -47,6 +47,21 @@
...
+{
+    if (s == "filldiamond") return vrv::SMUFL_E0DB_noteheadDiamondBlack;
```
lpugin: Instead of a string s you will then you this->GetHeadShape() which will be a data_HEADSHAPE_list type.  Also, I normally do not vrv:: prefix SMuFL codes.

pe-ro:
Using `<symbolTable>/<symbolDef>` means that either 
1. you know _a priori_ all the symbols/characters the file uses or will use; or
2. you're willing to edit two places in the file: the symbolTable to add the symbol and the element's pointer to the symbol you just added.

Also, using a large number of SMuFL code points means the symbolTable would end up duplicating SMuFL.  

For these reasons, it's better to allow each element to point directly to its SMuFL analog.

ahankinson:
On the other hand, using the symbol table will allow you to use more than just SMuFL for the element symbols with the same mechanism. I think this is an important consideration -- that we don't have to introduce a lot of new functionality within MEI to support this.

Editing in two places is already part of 'extended' MEI techniques, like the facs / zone, or recording sections of the file. Having a symbol table separate from the 'logical' flow of the music is an established pattern.

I don't think we would be in danger of replicating SMuFL within a single file, since you would only ever use a subset of SMuFL in any given encoding.

ahankinson:
And if you're concerned about duplicating information, consider that for each `head` or `stem` child element we would also need to record the authority and the symbol pointer, and you would need a full scan of the file to determine which SMuFL glyphs were in use. With the symbolTable functionality this gets reduced down to a single IDREF, making it a much more succinct and centralized method of addressing external symbols.

You would also only need to create a single symbol entry for any given *type* of symbol, after which you can simply address it. If you have a special note head that you want to use from SMuFL, you only have to edit the file in two places once, after which you can just use the IDREF.

kepper:
> 
> I don't think we would be in danger of replicating SMuFL within a single file, since you would only ever use a subset of SMuFL in any given encoding.

or, since MEI supports anyURI, you could have a project-specific symbol table in a separate file, where each individual encoding just points to those things it actually needs…

I think the real decision is about who this feature is targeting at – novice users might be confused by having to add a symbol definition elsewhere, while advanced users might appreciate the additional flexibility they gain.

The question is if we can find a single solution that fits both, or if we have to come up with two solutions – one "quick and dirty" in-place solution with attributes, and one "bloated" solution with external symbol tables. Of course this will inevitably raise the question why MEI again has two solutions for one problem… (which I would be fine with – we can strip this down to one for "MEI Go!" if necessary…)

ahankinson:
I guess ultimately you should be able to give SMuFL codepoints as a URI, which would wrap both the authority and the actual symbol into a single address. `@head.sym="http://smufl/U+E770"` would obviate the need for a separate symbol definition table.

pe-ro:
Both approaches are supported now --  
1. For those wishing to point from an element directly to its SMuFL analog, `@extsym` is available.
2. `@altsym` can be used by those who want to point to an internal symbol definition (which may actually reference an externally-defined symbol).

Either or both approaches can be ignored by applications, preferably via a customization.  Conversion from one to the other shouldn't be too difficult, either.

This, however, doesn't address the initial question of how to handle note head shapes because it doesn't address whether / how to "de-compose" the note.  `<note altsym="wherever"/>` would seem to indicate that there's an alternative symbol **for the entire note**, while 
```
<note>
  <noteHead altsym="whatever"/>
  <stem/>
</note>
```
implies that the alternative symbol applies **only to the note head**.

By providing codepoints for both note heads and stems, SMuFL offers more detail than MEI currently does.  But, the central question is, Do we need this level of detail?  One can argue either way -- we need it for completeness' sake or, since stems are usually drawn with graphic primitives (i.e., lines) anyway, why bother?  If we choose to implement the detail then I believe the best way is to create sub-elements --
```
<note altsym="whatever">
  <noteHead altsym="whatever"/>
  <stem altsym="whatever"/>
</note>
```
where the `@altsym` attribute applies to the thing to which it's attached.

pe-ro:
> I guess ultimately you should be able to give SMuFL codepoints as a URI, which would wrap both the authority and the actual symbol into a single address. `@head.sym="http://smufl/U+E770"` would obviate the need for a separate symbol definition table.

But it also bypasses the advantages of `<symbolTable>` --
1. the possibility to "compose" symbols from multiple sources in the symbol table; and
2. transmit the symbol table along with the rest of the encoding.

This is why `@altsym` (an internal reference) and `@extsym` (an external reference) exist.

kepper:
> Am 01.02.2017 um 17:36 schrieb pe-ro <notifications@github.com>:
> 
>  If we choose to implement the detail then I believe the best way is to create sub-elements --
> 
> <note altsym="whatever">
>   <noteHead altsym="whatever"/>
>   <stem altsym="whatever"/>
> </note>
> 
> where the @altsym attribute applies to the thing to which it's attached.
> 

I remember this question popping up every now and then, and especially the mensural folks seemed to be very interested in separate noteHead elements, for all kinds of weird things they've found. Since this is only required for very specific cases, and we surely wouldn't require everyone to provide individual noteHead and stem elements for every single note, I'm fine with this addition.

I wonder if we need to have <flag/> then as well. And, we need to make it absolutely clear that the duration still lives in note/@dur…

pe-ro:
Adding `<noteHead>` is not without peril, i.e., it opens the question of whether to add `<stem>` and `<flag>`.  I'd prefer to head down that path fairly slowly, adding `<noteHead>`, `<stem>`, and `<flag>` incrementally.

ahankinson:
If we're giving our preferences, I'd prefer to not add them at all... I think we can accomplish 95% of what we could do with separate elements with `@head.*` and `@stem.*` attributes.

kepper:
> 
> If we're giving our preferences, I'd prefer to not add them at all... I think we can accomplish 95% of what we could do with separate elements with @head.* and @stem.* attributes.

Just for the record – Beethovens Werkstatt wouldn't use individual elements (hence my question about others that are asking for this feature…)

pe-ro:
> I think we can accomplish 95% of what we could do with separate elements with @head.* and @stem.* attributes.

Granted, but that means adding 4 attributes to `<note>` -- `@head.altsym`, `@head.extsym`, `@stem.altsym`, and `@stem.extsym` -- which is somewhat kludgey.  In my opinion, the attribute names themselves provide a clue to the more idiomatic XML solution of treating head (and possibly, but not necessarily stem) as full-fledged objects.

pe-ro:
Treating the note head and stem(s?) as objects could result in something like the following --
```
<note dur="4" dur.ges="16p">
  <noteHead shape="foobar" mod="centerdot brack" color="aqua" fill="solid" fillcolor="blue"
    rotation="none" altsym="#whatev"/>
  <stem dir="ne" len="7" mod="none" pos="center" visible="true" with="above" x="0" y="0" altsym="#whatev"/>
  <accid accid="1qf" accid.ges="f" altsym="#whatev"/>
  <artic artic="dot" artic.ges="dnbow" altsym="#whatev"/>
</note>
```
where `<note>` has all of the logical attributes (as well as the visual ones that apply to the entire note) and `<noteHead>` and `<stem>` have all the visual attributes.  Alternatively, we could allow `<note>` to continue to have _all_ of its current attributes (to maintain backward compatibility for a while), but disallow (or warn about) data repeated there _and_ on the new elements via Schematron.

This may "discombobulate" CMN folks for a bit, but will ultimately be better because it achieves a level of markup detail that works for both CMN and Mensural notation.

ahankinson:
I don't think a head or stem *is* a full-fledged object, they're just "attributes" of a note. 

I'm still not convinced that this isn't privileging the visual over the logical. An accidental is broken into its own element because it defines a separate logical function; it seems the only reason to break out stems and heads is to define visual characteristics which, until now, have been adjusted through `att.*.vis` attribute classes.

I think promoting visual attributes of an object to a first-class object in itself is a way in which lies madness. 

```
<clef>
    <shape />
    <line />
</clef>
```

?

lpugin:
I would recommend to move the discussion to music-encoding or MEI-L

PR #460, opened 2017-01-28
Implementing basic support for the head.shape attribute for notes

billmoser:
Here is an overview of the changes:
     1) several glyphs have been uncommented in fonts/supported.xsl, requiring
        generate_all.sh to be rerun.  This of course has the side effect that
        several other files in data and include have been (re)regenerated
     2) "Boilerplate" has been added to libmei/attconverter.cpp, libmei/attypes.h
        include/vrv/note.h, src/iomei.cpp, and src/note.cpp in order to allow the
        head.shape attribute to be recognized as a note attribute.
     3) src/view_element.cpp has two additions:
            a) a function which translates MEI data.HEADSHAPE.list items into SMuFL codes
            b) a 3-line conditional in View::DrawNote which replaces the
               fontNo for the notehead if the head.shape attribute exists

    I have a few questions/concerns:
    1) There are many more SMuFL glyphs than there are MEI head.shapes.
       Is sticking with the MEI subset necessary, or is it possible for me to go
       with the full set of SMuFL codes/strings? This would then differ from MEI's
       data.HEADSHAPE.list.  I have implemented it this way for now (see the
       data.HEADSHAPE.list enum in atttypes.h, and the Headshape-related converters
       in attconverter.cpp).
    2) Is view_element.cpp the appropriate place for the data.HEADSHAPE.list/SMuFL
       code converter?
    3) This notehead replacement is simple and context-free.  As @craigsapp notes
       in issue #458, the note stems are not attaced to the heads for the new
       glyphs.  Its ugly -- not sure how to fix it.


libmei/atttypes.h
```diff
@@ -318,6 +318,12 @@ enum data_HEADSHAPE_list {
     HEADSHAPE_list_slash,
     HEADSHAPE_list_square,
     HEADSHAPE_list_x,
+    
+    /* The following SMuFL glyphs are not part of MEI's data.HEADSHAPE.list */
+    HEADSHAPE_list_circlex,
```
lpugin: This is problematic because the fill is auto generated. The proper way to fix it is to have the MEI data.HEADSHAPE.list extended accordingly. The issue should be raised on the music-encoding repository

libmei/attconverter.cpp
```diff
@@ -702,6 +706,12 @@ data_HEADSHAPE_list AttConverter::StrToHeadshapeList(std::string value) const
     if (value == "slash") return HEADSHAPE_list_slash;
     if (value == "square") return HEADSHAPE_list_square;
     if (value == "x") return HEADSHAPE_list_x;
+
+    /* The following SMuFL glyphs are not part of MEI's data.HEADSHAPE.list */
+    if (value == "circlex") return HEADSHAPE_list_circlex;
```
lpugin: The fill is auto generated. See the comment on atttypes.h

src/view_element.cpp
```diff
@@ -47,6 +47,51 @@
...
+    if (s == "filldiamond") return vrv::SMUFL_E0DB_noteheadDiamondBlack;
+    if (s == "diamond") return vrv::SMUFL_E0DD_noteheadDiamondWhite;
+    if (s == "fillisotriangle") return vrv::SMUFL_E0BE_noteheadTriangleUpBlack;
+    if (s == "isotriangle") return vrv::SMUFL_E0BD_noteheadTriangleUpWhite;
+    if (s == "x") return vrv::SMUFL_E0A9_noteheadXBlack;
+    if (s == "circlex") return vrv::SMUFL_E0B3_noteheadCircleX;
+    return 0;
+}
+*/
+wchar_t getFontNumber(vrv::data_HEADSHAPE_list s)
```
lpugin: As commented in the previous PR we can move this function to a Note::GetHeadShapeGlyph method with no parameter. vrv:: prefixes should be removed

billmoser: Yes this is easy to do.  I will implement on my fork.

libmei/atttypes.h
billmoser: I saw that you raised the issue over at at music-encoding (https://github.com/music-encoding/music-encoding/issues/400), thanks.

billmoser:
It appears that any changes are dependent on changes to the MEI schema, so I suppose we need to wait until a new version is available which addresses these issues.  Even with a new schema, there is still a rendering problem. Currently, verovio uses information derived from the glyph's bounding box to compute the stem attachment points.  This method will not work for many (most?) of the SMuFL head shapes.  The font metadata files (well, bravura_metadata.json, anyway) contain the necessary information (stemUpSE, stemDownNW) but using these to implement stem drawing is a non-trivial change to view_element.cpp

lpugin:
Indeed. However changes in the MEI schema should not be that much of a delay as soon as a decision is taken since we can very well have the verovio:develop branch being based on the develop version of MEI. It is currently not the case but I also see a need for this.

Concerning the use of the SMuFL metadata this is indeed a fairly complex question. Verovio uses its own simple metadata bounding box system. I did look at switching to the SMuFL json one but postponed it because it is not clear how to generate additional information, and in particular the stemUpSE or stemDownNW you are mentioning. Having these would make it worthwhile. There were some discussions on the SMuFL mailing list about it and there is this script https://gist.github.com/anonymous/91548ecd79effc11ae7f , but it does not generate this information. I am not sure about the status of this.

So I think for now we can still adjust Verovio. We can do this nicely by putting most of it in the Note class and have methods returning the note head positioning according to the SMuFL glyph. I assume that if we have left, centred, right for horizontal placement and top, centred and bottom for vertical placement we can already make it work reasonably well for quite some cases without, but I have not look precisely at the note head shapes being considered.

billmoser:
I think you're right about the placement options.  left-centered-right for horizontal, and top-centered-bottom for vertical placement would be fine for the glyphs which are symmetrical on these axes (which is most of them, I believe.   

Generating the additional info such as stemUpSE and stemDownNW is simple for the fonts which include the information.  For the Bravura font, the bravura_metadata.json file contains this information, so it is easy to write a node/js script to transform the data into whatever format is most easily ingested in verovio (c++). However, I don't know how to interpret the information in the original .json file. For example there is this information:
"4stringTabClef": {
            "bBoxNE": [
                1.088, 
                2.016
            ], 
I don't know what the units are here.  Even if the units are unknown, we may be able to infer an angle, I suppose (arctan(1.088/2.016) = .495rad), which may then be applied to approximate the stem start location based on the bounding box.  I don't know anything about font rendering (or any other type of rendering for that matter, I'm a back-end cs/math guy), so maybe I'm missing something which is basic to those in the know.

lpugin:
It would be great to read the smufl json file and use it. I haven't look at their interpretation but I assume this to be explained in the SMuFL specifications.

However, my main concern is still how to generate such a file for other fonts than Bravura. So I would rather start with a simple placement option as I describe. What do you think?

Eventually, we could define somewhere a list of glyph that need more refined placement information and have it extracted from the bravura_metadata.json for Bravura and from a file created by hand for other font. This would be acceptable as long as the number of glyphs that need this remain reasonably low.

billmoser:
I can look into it and see if the simpler scheme will work for a selection of glyphs.  We still need a place for the additional metadata though.

lpugin:
OK, I created a develop-noteheads branch with NoteHeadStemPosition data type. From there you can add a `NoteHeadStemPosition Note::GetStemPosition()` method that return the appropriate value looking at the Note::GetHeadShape and the m_drawingStemDir. The default should be NOTE_HEAD_EAST for STEMDIRECTION_up and NOTE_HEAD_WEST for STEMDIRECTION_down.  Does this make sense?

billmoser:
I've taken another look at this, and I don't think that this approach will work.  I've attached three files to indicate the issues.  For now, just consider the rendering of the diamond noteheads.  

The first file is just a baseline, showing how stems are rendered with no additional glyph-specific information.  
[no-glyph-specific-adjustment.pdf](https://github.com/rism-ch/verovio/files/769817/no-glyph-specific-adjustment.pdf)

The second shows how the stems are rendered using the glyph's bbox to compute a glyph-specific radius. The placement on the x-axis is better, but not perfect.  The y-axis placement is unchanged.  Note also that the beamed notehead placement is no better than before (perhaps this rendering is done differently, elsewhere in the code).
[glyph-specific-box-adjustment.pdf](https://github.com/rism-ch/verovio/files/769818/glyph-specific-box-adjustment.pdf)

For the last rendering, there is a variable in `View::DrawStem` of view_element.cpp used to shorten the stem in the y direction:
`int shortening = 0.25 * m_doc->GetDrawingUnit(staffSize);`
The last file shows what happens in the rendering when I hard-code this to 0.0.  Now, the y-axis location is correct (for the non-beamed diamonds), but the x-axis is clearly off (by a stem-width?).
[glyph-specific-adjustment-shortening-eq-0.pdf](https://github.com/rism-ch/verovio/files/769819/glyph-specific-adjustment-shortening-eq-0.pdf)

It appears that, in the current context, glyph-specific "fudge-factors" such as the `shortening` variable, are required for correct rendering of these sorts of notehead shapes.  I think it makes more sense, then, to consider approaches such as SMuFL's, which use more sophisticated notions for notehead/stem attachment.  This is substantially beyond the scope of what I had contemplated when issuing the initial pull request.  I may be able to explore this more in depth, but I'm not sure what the timeframe will be

PR #471, opened 2017-02-05
adding support for lower accidentals on ornaments

rettinghaus:
This PR adds support for `@accidlower` and `@accidupper` on `<mordent>`, `<trill>`, and `<turn>`. 
Reading/writing on MEI files is enabled. 

`@accidlower` gets rendered: 

ornaments_001

```xml
<mei xml:id="rootBeer" xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
  <meiHead xml:id="threeHeadedMonkey" type="geilerKopf" suppe="klar">
    <fileDesc>
      <titleStmt>
        <title xml:id="ohneTitel" />
      </titleStmt>
      <pubStmt>
      </pubStmt>
    </fileDesc>
    <encodingDesc xml:id="Haschisch">
      <appInfo xml:id="Gras">
        <application isodate="2017-02-03T08:54:16" version="0.9.14-dev-666e967-dirty">
          <name>Verovio</name>
          <p>Transcoded from MusicXML</p>
        </application>
      </appInfo>
    </encodingDesc>
  </meiHead>
  <music>
    <body xml:id="Mustopf">
      <mdiv>
        <score>
          <scoreDef xml:id="scoredef-000000179000538">
            <staffGrp xml:id="m-000000128416136">
              <staffDef xml:id="staffdef-000000207365861" clef.shape="F" clef.line="4" key.sig="1f" meter.count="3" meter.unit="4" n="1" lines="5" />
            </staffGrp>
          </scoreDef>
          <section xml:id="section-000000085245121">
            <measure xml:id="measure-000000153551412" n="1">
              <staff xml:id="staff-000000196791433" n="1">
                <layer xml:id="layer-000000134056409" n="1">
                  <note xml:id="note-000000160973676" dur="4" oct="3" pname="d" stem.dir="down" />
                  <note xml:id="note-000000193949307" dur="4" oct="3" pname="c" stem.dir="up" />
                  <note xml:id="note-000000072160996" dur="4" oct="3" pname="e" stem.dir="down" />
                </layer>
              </staff>
              <turn xml:id="turn-000000084674041" staff="1" startid="#note-000000160973676" form="inv" />
              <turn xml:id="turn-000000040585051" staff="1" startid="#note-000000193949307" form="norm" accidlower="f" />
              <mordent xml:id="mordent-000000125845991" staff="1" startid="#note-000000072160996" accidlower="s" form="inv" />
              <trill xml:id="trill-000000036940252" staff="1" startid="#note-000000072160996" accidlower="n" accidupper="s"/>
              <fermata staff="1" startid="#note-000000072160996"/>
            </measure>
            <measure xml:id="measure-000000017298058" n="2" right="end">
              <staff xml:id="staff-000000185461773" n="1">
                <layer xml:id="layer-000000198255235" n="1">
                  <note xml:id="note-000000040123086" dur="4" oct="2" pname="b" stem.dir="up">
                    <accid xml:id="accid-000000038842924" accid.ges="f" />
                  </note>
                  <note xml:id="note-000000000106427" dur="4" oct="3" pname="f" stem.dir="down" />
                  <note xml:id="note-000000081491865" dur="4" oct="2" pname="e" stem.dir="up" />
                </layer>
              </staff>
              <mordent xml:id="mordent-000000212755109" staff="1" startid="#note-000000040123086" form="norm" accidlower="n" />
              <mordent xml:id="mordent-000000070740074" staff="1" startid="#note-000000000106427" form="inv" long="true" />
              <fermata xml:id="fermata-000000183455043" staff="1" startid="#note-000000081491865" form="norm" place="above" />
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

@lpugin I added `GetOrnamentaccidCode` to `view_control.cpp`. Perhaps it would be better to make it more globally available, as it is the same as used in `accid.cpp`.


craigsapp:
What were you doing when you created that example? 😃

src/view_control.cpp
```diff
@@ -1929,6 +1959,41 @@ void View::DrawTurn(DeviceContext *dc, Turn *turn, Measure *measure, System *sys
     dc->EndGraphic(turn, this);
 }
 
+int View::GetOrnamentaccidCode(data_ACCIDENTAL_EXPLICIT ornamentaccid)
```
lpugin: It think the best would be to make is a static method of Accid and then call Accid:: GetOrnamentaccidCode.  I would rename it GetOrnamentAccidGlyph to be inline with other similar methods

rettinghaus: I renamed the function. The rest I'll leave up to you.

lpugin:
Could you also make it a static method of Accid? Thanks!

rettinghaus:
I'd feel better if you could do that.

craigsapp:
I am about to do that... (so warn me if someone is doing it in parallel :-)

Also, why is "accid" lower case in `GetOrnamentaccidGlyph`?  Maybe I will figure out what an `ornamentaccid` is in the process...

I don't see how to re-push a push that I am not owner of, so it might end up being merged directly into `develop`, unless you stop me before I am done.

craigsapp:
Notes:

* `ornamentaccid` is an MEI attribute group/class: http://music-encoding.org/documentation/3.0.0/att.ornamentaccid
But why are attribute groups not camelCased?  And why are two ornaments needed for a mordent?

* `Accid::GetOrnamentaccidGlyph` cannot return an empty accidental and will return a natural accidental by default.  Will that ever cause problems?

* Might need to be generalized further for allowing editorial brackets or parentheses around the ornament accidental (i.e, something along the lines of `Accid::GetSymbolStr()`).  Ideally the accidentals would be children of the ornaments?

```xml
<mordent form="lower">
     <accid accid="s" enclose="brack"/>
</mordent>
```

screen shot 2017-02-06 at 14 43 49

Here is the test data rendering with the move of the ornamentaccid function from View to Accid:

screen shot 2017-02-06 at 14 46 30

[looks the same]

rettinghaus:
@craigsapp Thanks! But just moving the code to `accid.cpp` is something I could have done easily by myself. I thought Laurent would have wanted to combine the functions so there is no duplicated code.

craigsapp:
I was following literally what @lpugin said :-)  Merging is probably good, but I don't see an easy way of doing that since there are complications such as brackets to consider.  The main `Accid` SMuFL glyph function adds enclosures on the accidental, but a static functions cannot have access to member variables of `Accid`, and there is no input parameter for passing such things to the `GetOrnamentalaccidGlyph()` function).

lpugin:
Indeed. The simplest would be to have a `static wchat_t Accid::GetAccidGlyph(data_ACCIDENTAL_EXPLICIT)` method the is called from both View::DrawMordent or from Accid::GetAccidStr.

PR #472, opened 2017-02-07
removing duplicate code on accid.cpp

rettinghaus:
Everyone ok with this?


craigsapp:
That looks great to me, and I probably won't get in trouble for approving it.

PR #484, opened 2017-02-22
Adding expansion elements

rettinghaus:
This adds support for `<expansion>` in MEI. 
(Mainly for @craigsapp to add repeats in `iohumdrum` and MIDI output :-) )

Also this brings reading/writing for `att.distances` on `<staffDef>`.


src/expansion.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace vrv {
+
+//----------------------------------------------------------------------------
+// Expansion
+//----------------------------------------------------------------------------
+
+Expansion::Expansion() : SystemElement("expansion-"), AttCommon(), AttPlist()
+{
+    RegisterAttClass(ATT_COMMON);
+    RegisterAttClass(ATT_PLIST);
```
lpugin: You forgot att.commonPart

src/expansion.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+Expansion::~Expansion()
+{
+}
+
+void Expansion::Reset()
+{
+    SystemElement::Reset();
+    ResetCommon();
```
lpugin: Same here

rettinghaus: sloppy me!

lpugin:
This looks good. I think we still need to add the functor for casting off and for converting from to the page-based MEI, but I will have a look at them.

craigsapp:
When refined line-breaking and pagination is implemented, `<section>` grouping used for `<expansion>` lists segmentations could be given preference for choosing line/page segmentation points.  For example if the segment falls one measure away from a line break:

screen shot 2017-02-22 at 12 57 52

An attempt could be made to force the line break to move one measure forwards/backwards to end the section at a line/page break:

screen shot 2017-02-22 at 12 58 01

PR #502, opened 2017-03-05
Added missing include to cstdlib

ChWick:
This include is required for the android gcc 4.9 compiler. Functions like atoi are defined in <cstdlib>.


craigsapp:
This file is actually from another repository, and I have been naughty and not paying attention to the submitted issues recently, where there is a similar report from 11 days ago:

https://github.com/craigsapp/midifile/issues/38

So I will fix the problem there and update the code in verovio from there as well.

PR #504, opened 2017-03-05
Removed unused warnings by adding a UNUSED define

ChWick:
Using a UNUSED define to 'hide' warnings of unused variables. I usually like to compile with -Wall and these unused variables generate > 700 warnings.

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -413,7 +413,7 @@ public:
     /**
      * Retrieve the time spanning layer elements between two points
      */
-    virtual int FindTimeSpanningLayerElements(FunctorParams *functorParams) { return FUNCTOR_CONTINUE; }
+    virtual int FindTimeSpanningLayerElements(FunctorParams *functorParams) { UNUSED(functorParams); return FUNCTOR_CONTINUE; }
 
     ///@}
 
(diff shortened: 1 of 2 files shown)


craigsapp:
```cpp
virtual int PrepareRpt(FunctorParams *functorParams) { 
    UNUSED(functorParams); 
    return FUNCTOR_CONTINUE; 
}
```

Is there an advantage to creating a define rather than doing it directly?

```cpp
virtual int PrepareRpt(FunctorParams *functorParams) {
    (void)functorParams; 
    return FUNCTOR_CONTINUE;
}
```

The main advantage that I can think of is that it make the explicit unused variables easy to find and easier to figure out what is happening.  The main disadvantage is that some other code included after `vrvdef.h` could re-define the UNUSED define, or the definition in `vrvdef.h` could overwrite an existing define for UNUSED.

in `include/vrv/vrvdef.h`:
```cpp
    #define UNUSED(x) {(void)x;}
```

Also this should work

```cpp
virtual int PrepareRpt(FunctorParams *) {
    return FUNCTOR_CONTINUE;
}
```

craigsapp:
In any case, that will be useful to add to the code.  

I just noticed that you submitted the pull request to the `master` branch. @lpugin will be complaining about that and wanting you to change it to the `develop` branch.

lpugin:
Thanks for this. PR should indeed be made to the develop branch. See instructions here https://github.com/rism-ch/verovio/blob/master/doc/contributing.md

Also, I removed the name of the parameter for the methods that do not use it (39104f127d9389a19d87d16e389641e4ca1c2a80). If it works this would be a better fix than an additional instruction with static cast to void, also because some compilers might still detect it as an unused variable.

Issue #436, opened 2017-01-17
loading fonts when outputting MIDI files


craigsapp:
When using `-t midi`, verovio still tries to load font information:

``` bash
$ ./verovio test1.krn -t midi
[Warning] Glyph and bounding box units-per-em for code '57442' miss-match (bounding box: 1000)
Output written to test1.mid.
```

In case layout is being done before MIDI is output (which should not be done), this is also true if `--no-layout` is given:

```bash
$ ./verovio test1.krn -t midi --no-layout
[Warning] Glyph and bounding box units-per-em for code '57442' miss-match (bounding box: 1000)
Output written to test1.mid.
```

(warnings generated because I purposefully removed one of the glyphs).

It would probably be good to not load font information when MIDI files are created.

lpugin:
>  this is also true if --no-layout is given

Not really, because --no-layout means do not cast of the layout and output all the file on one single system. The font is needed for this.

craigsapp:
>> this is also true if --no-layout is given

> Not really, because --no-layout means do not cast of the layout and output all the file on one single system. The font is needed for this.

If I do this:

```bash
$ ./verovio file.krn -t mei --all-pages --no-layout
```

is there any layout happening with these options? In this case my intent is for the raw output from `src/iohumdrum.cpp`, with no post-processing. 

If I do this:

```verovio
$ rm -rf /usr/local/share/verovio
$ ./verovio test.krn -t mei --all-pages --no-layout
```

I get the error:
```
The resources path /usr/local/share/verovio could not be found; please use -r option.
```

which is a directory only used to store the fonts, so verovio is trying to sneak a peek at them when I am not expecting it to.  I haven't look carefully to see what verovio is doing in this case yet -- it could just be checking to see if the directory is there, but it seems to also be trying to load the fonts in this case:

```bash
$ ./verovio test.krn -t mei --all-pages --no-layout -r .
```

```
Error] Font directory './Bravura' cannot be read
[Error] Bravura font could not be loaded.
[Error] Font directory './Leipzig' cannot be read
[Error] Leipzig font could not be loaded.
[Error] Expected 209 default SMUFL glyphs but could load only 0.
The music font could not be loaded; please check the contents of the resource directory.
```

This also happens when converting from MEI to MIDI, so perhaps verovio loads the fonts whenever it is run rather than whenever they are needed.

PR #512, opened 2017-03-08
Loading fonts only for SVG

rettinghaus:
This is a fix for #436
Just moved some lines, we can rethink this in due course.


rettinghaus:
added missing support for colors on chords

PR #514, opened 2017-03-11
Changing font styling in SVG

rettinghaus:
This PR adds ATT_TYPOGRAPHY to `g` elements in SVG output and removes styling via FontInfo. 
Three main reasons: 
* better MEI attribute representation in SVG 
* fixes nested `<rend>` elements
* better access through CSS

Additionally `@lyric.*` attributes from `<staffDef>` are now put into a single `@style` attribute on `<g class="staff">` elements. This will effect all lyrics, as they are the only textual elements within staff. And after all those attributes define a "style". Again, this brings better access through CSS for `<verse>` and `<syl>`.


src/svgdevicecontext.cpp
```diff
@@ -159,6 +160,25 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         m_currentNode.append_attribute("id") = gId.c_str();
     }
 
+    // this sets staffDef styles for lyrics
+    if (object->Is(STAFF)) {
+        Staff *staff = dynamic_cast<Staff *>(object);
+        std::string styleStr;
+        if (staff->m_drawingStaffDef->HasLyricFam()) {
```
lpugin: We need an assert(staff->m_drawingStaffDef);

src/svgdevicecontext.cpp
```diff
@@ -159,6 +160,25 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         m_currentNode.append_attribute("id") = gId.c_str();
     }
 
+    // this sets staffDef styles for lyrics
+    if (object->Is(STAFF)) {
+        Staff *staff = dynamic_cast<Staff *>(object);
+        std::string styleStr;
```
lpugin: Always `assert(staff);` after a `dynamic_cast`

src/view_text.cpp
```diff
@@ -53,12 +53,9 @@ void View::DrawRend(DeviceContext *dc, Rend *rend, int x, int y, bool &setX, boo
 
     FontInfo rendFont;
     bool customFont = false;
-    if (rend->HasFontname() || rend->HasFontsize() || rend->HasFontstyle() || rend->HasFontweight()) {
+    if (rend->HasFontsize()) {
         customFont = true;
-        if (rend->HasFontname()) rendFont.SetFaceName(rend->GetFontname().c_str());
         if (rend->HasFontsize()) rendFont.SetPointSize(rend->GetFontsize());
```
lpugin: Why checking again?

rettinghaus: ouch!

lpugin:
The plan is to have classes for note components (stem, notehead, flags, etc.). We can then preserve them in the SVG even though they are not MEI classes. To be discussed.

Issue #526, opened 2017-03-20
Editing a note on a page other than the last-rendered page


ahwitz:
I've been dealing with a weird failure/exception in Vida for a bit where if you drag-edited a note on one page it would sometimes fail the first time and then work every time after. I think I've finally narrowed it down to a block in Toolkit::Drag:
```
1046: Object *element = m_doc.GetDrawingPage()->FindChildByUuid(elementId);
1047: if (element->Is(NOTE)) {
    // ... do things
    return true;
}
return false;
```
https://github.com/rism-ch/verovio/blob/develop/src/toolkit.cpp#L1046

If I understand what's happening correctly, it's assuming that the last page rendered is where the requested edit will occur. In the case where two pages are rendered with Verovio and an edit is requested on the first, it doesn't seem to enter the `if` statement above and thus nothing happens. Because of how Emscripten prints errors, I'm getting a message along the lines of `abort(15)` with little useful debug information, and re-compiling Verovio with `-g3` in the Emscripten line doesn't really help much.

Regardless, replacing `m_doc.GetDrawingPage()->FindChildByUuid` with `m_doc.FindChildByUuid` eliminates the failure, but I have a feeling that this might lead to a bit of performance loss if I submitted a PR with that direct change. Would any of the following solutions be okay/preferred to submit as a PR, or is there a better alternative?

1. If line 1046 above can't find the element, default back to m_doc.FindChildByUUID.
2. If line 1046 above can't find the element, use Toolkit::GetPageWithElement to determine what "the current drawing page" should be.
3. Add an optional "page" property to the `action='drag'` JSON object which, if provided, is passed to `m_view.SetPage` before `Toolkit::Drag` is called.

I can submit a PR with any/all of these if desired, as I know what needs to happen to implement them.

lpugin:
Option 1. looks like a good fix.

PR #541, opened 2017-04-12
Fix for #526

ahwitz:
There seems to be a bug in the current develop branch where running Toolkit::Drag and then calling toolkit.renderPage returns the same SVG (without reflecting the pitch change). Running a git bisect right now to see if I can get you any more information.

I've nonetheless verified the first commit by testing it with toolkit.getMEI() - the PName/Oct change is reflected correctly and there's no console error, so it gets past the assertion in 1049 (old line number) fine.

### DIFF
--- a/src/toolkit.cpp
+++ b/src/toolkit.cpp
@@ -1043,7 +1043,14 @@ const char *Toolkit::GetHumdrumBuffer()
 bool Toolkit::Drag(std::string elementId, int x, int y)
 {
     if (!m_doc.GetDrawingPage()) return false;
+    
+    // Try to get the element on the current drawing page
     Object *element = m_doc.GetDrawingPage()->FindChildByUuid(elementId);
+
+    // If it wasn't there, go back up to the whole doc
+    if (!element) {
+        element = m_doc.FindChildByUuid(elementId);
+    }
     if (element->Is(NOTE)) {
         Note *note = dynamic_cast<Note *>(element);
         assert(note);


ahwitz:
git bisect tells me that the breaking change happened somewhere in one of the following three commits:
https://github.com/rism-ch/verovio/commit/2d77c382f63cf2253f4745ccdd20712fb31afc55
https://github.com/rism-ch/verovio/commit/b08e45338dd6b9e63086ea6d81faa493769fbdc8
https://github.com/rism-ch/verovio/commit/128648181b1db76a66fe790afcc2e38688cab9f8

These are three of the dynamic drawing commits; two are marked WIP and none seem to compile, so I can't narrow down the breaking change further. I don't know the bounding box code well enough to instinctually have an idea of where to look either. If you can't reproduce this in your editor and it seems to only be a Vida problem, let me know and I'll dig around more on my own.

lpugin:
This issue is due to the fact that quite some code calculating the layout was moved from the drawing part to specific layout calculation methods / functors. This was necessary for making it possible to fine tuning the layout and is a significant improvement. However, just redrawing the page after dragging does not work as before.

I fixed this here 2abc20e3c8a4f549279f66d8febb89c162890f2e by adding a new toolkit method that recalculate the pitch positions (but only this) for the current drawing page. We can think what exactly we want to happen here and what would be a good trade of between speed and re-adjustment, but for now calling redoPagePitchPosLayout before redrawing the page when dragging should fix the issue.

ahwitz:
Commenting on this just because the discussion's here: would it be straightforward for the SetAlignmentPitchPosParams and CalcStemParams functors to return a set of SVG IDs and their changed position values? If so, imagining Toolkit::Edit as something that takes in a set of MEI IDs and their corresponding changes and returns a set of SVG IDs and their corresponding changes would solve a good handful of speed problems. Would require a lot of JavaScript refactoring for anyone who uses Edit, though, but that might already be required with the new function.

lpugin:
Not sure what you mean, but when the Functor is applied to the entire dataset (as it happens when calculating the layout) then this will means thousands of IDs so you will raise other memory issues. SetAlignmentPitchPosParams just set the vertical position according to the clef and the pitch, so we can maybe optimise it the other way around. That is passing to SetAlignmentPitchPosParams an optional (?) list of IDs of the element to re-adjust. But I would suggest to look at this with a broader picture in mind on how editing / refreshing should work.

ahwitz:
The idea was to only return the IDs of elements whose positions have changed, so that the editing library would be repositioning elements instead of asking for an entire new copy of each affected system. I'd considered that changes would propagate out, but yeah, the size of that object is far greater than I'd thought it would be. Will brainstorm on the subway and see if I can throw any other ideas your way.

lpugin:
One option would be to add a parameter to the SetAlignmentPitchPosParams (in addition to an ID) specifying what parent container should be updated, going from updating the element itself only, the parent chord, measure, or system, etc. So you would do `vrvTootlkit.setAlignmentPitchPosParams("noteID", "chord")` to recalculate all the positions in the parent measure of the "noteID", or `vrvTootlkit.setAlignmentPitchPosParams("noteID", "system")` if you want the entire system to be recalculated.

ahwitz:
Aren't there cases where dragging a note would cause changes through the entire system, though? Would it be possible to track, throughout the functor call, what the highest element affected was?

ahwitz:
Actually, that might work, as long as you leave the page-level function you created above exposed. I can add a "refresh system" button as a fallback in case it gets cramped.

PR #548, opened 2017-04-16
Various fixes related to #541

ahwitz:
- Reinstated changes removed in d0da6fd68 to allow notes within a chord to not be in sequential pitch order
- Fixed typo in page.cpp
- Added ledger line calculations to Page::LayOutPitchPos to allow notes to be dragged farther down

There's still a few bugs with this that I couldn't fix:
- Moving a note with ledger lines up does not delete ledger lines below it
- Inserting a CalcChordNoteHeads functor call to determine whether any noteheads should be flipped/unflipped doesn't work. Is this because the noteheads haven't been re-X-aligned within Page::LayOutPitchPos?

### DIFF
--- a/include/vrv/bboxdevicecontext.h
+++ b/include/vrv/bboxdevicecontext.h
@@ -75,7 +75,7 @@ public:
     virtual void DrawEllipse(int x, int y, int width, int height);
     virtual void DrawEllipticArc(int x, int y, int width, int height, double start, double end);
     virtual void DrawLine(int x1, int y1, int x2, int y2);
-    virtual void DrawPolygon(int n, Point points[], int xoffset, int yoffset, int fill_style = AxODDEVEN_RULE);
+    virtual void DrawPolygon(int n, Point points[], int xOffset, int yOffset, int fillStyle = AxODDEVEN_RULE);
     virtual void DrawRectangle(int x, int y, int width, int height);
     virtual void DrawRotatedText(const std::string &text, int x, int y, double angle);
     virtual void DrawRoundedRectangle(int x, int y, int width, int height, double radius);
--- a/include/vrv/svgdevicecontext.h
+++ b/include/vrv/svgdevicecontext.h
@@ -184,7 +184,7 @@ private:
 
     /**
      * Flush the data to the internal buffer.
-     * Adds the xml tag if necessary and the <defs> from m_smufl_glyphs
+     * Adds the xml tag if necessary and the <defs> from m_smuflGlyphs
      */
     void Commit(bool xml_declaration);
 
@@ -215,7 +215,7 @@ private:
 
     // holds the list of glyphs from the smufl font used so far
     // they will be added at the end of the file as <defs>
-    std::vector<std::string> m_smufl_glyphs;
+    std::vector<std::string> m_smuflGlyphs;
 
     // pugixml data
     pugi::xml_document m_svgDoc;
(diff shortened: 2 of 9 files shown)


src/chord.cpp
```diff
@@ -188,13 +188,32 @@ int Chord::PositionInChord(Note *note)
...
+    int y1;
     ListOfObjects *childList = this->GetList(this); // make sure it's initialized
     assert(childList->size() > 0);
 
-    // The first note is the bottom
-    yMin = childList->front()->GetDrawingY();
-    // The last note is the top
-    yMax = childList->back()->GetDrawingY();
+    // Iterate through list of child Note objects to find lowest- and highest-pitched; not dependably in order
+    for (ListOfObjects::iterator it = childList->begin(); it != childList->end(); it++) {
```
lpugin: I don't think we need this because the list is orderer by pitch (see Chord::FilterList)

ahwitz: If you have a [C4, E4, G4] triad and drag the middle E down an octave in an editing interface so it's now [C4, E3, G4], is there a way to trigger FilterList from the functor calls in Page::LayOutPitchPos()?

lpugin: I guess the proper way to fix this would be to call `note->Modify(true)` in `Toolkit::Drag`. Can you try this?

ahwitz:
Updated this branch with a test commit - if I just call note->Modify in Toolkit::Drag, it doesn't propagate up to the chord and thus the stem doesn't get redrawn. If I call both note-Modify and note->GetFirstParent(chord)->Modify, everything works fine. I've included some debug logs in this commit to help determine what's going on; should the note Modify propagate up to the chord?

lpugin:
We can reasonably change it to this e8517b26edd6b4201e445aabf1f3c10c525ef0ca 

Can you try again?

ahwitz:
Functional.

It also turns out that the dots and ledger lines do not reset on drag, and because these appear to be arranged by staff (chord.cpp:420 for dots, staff.cpp:74 for ledger lines), I'm not sure how to safely reset those. Would it be a check for Staff::IsModified() somewhere?

lpugin:
Dots are reset in Dots::ResetHorizontalAlignment and ledger lines in Staff::ResetVerticalAlignment. This is indeed quite different and is not triggered by ::IsModified. The IsModified method is meant to reset cached flattened lists (e.g., the list of notes in the chord) and not for triggering re-alignment.

Now since many alignment sub-tasks are happening outside the drawing loop (i.e., dots or leger lines) for better positioning you cannot just call some the the alignment tasks as it is now. The reason if that the reseting is all happening in ::ResetHorizontalAlignment and ::ResetVerticalAlignment. If we want reseting for partial tasks, we need partial reseting methods.

So if you want dots or ledger lines to be updated when dragging (and not only at the end) you need for now to recalculate the full layout of the current page until we can clearly identify the set of actions to be performed for partial layout recalculation. Page::LayOutPitchPos is a bit of a temporary fix, and it would be good to have a clear overview of what we want to achieve here and in similar cases before we start implementing these partial layout recalculation methods. Any ideas how to proceed?

ahwitz:
> So if you want dots or ledger lines to be updated when dragging (and not only at the end)

One of the primary differences between your code and mine is that I actually do only recalculate alignment at the end of a drag, not during. I'll fool around with ResetHoriz/VerticalAlignment functors and see what I can do.

All that said, I do know a group of people who have ideas for what they want editing to achieve; I'll be talking with them later this afternoon and will report back in an email later.

Issue #427, opened 2017-01-16
dotted-rhythm forward/space conversion problem in MusicXML-to-MEI importer


craigsapp:
This example measure has problems converting into a multi-layer staff in MEI with the MusicXML-to-MEI importer:

Here is what the music looks like initially in MuseScore:

<img width="368" alt="screen shot 2017-01-15 at 16 08 25" src="https://cloud.githubusercontent.com/assets/3487289/21967474/e2212892-db3c-11e6-9b2b-b805bcae733e.png">

MusicXML data:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.0">
  <movement-title></movement-title>
  <identification>
    <encoding>
      <software>Finale 2014.5 for Mac</software>
      <software>Dolet Light for Finale 2014.5</software>
      <encoding-date>2017-01-13</encoding-date>
      <supports attribute="new-system" element="print" type="yes" value="yes"/>
      <supports attribute="new-page" element="print" type="yes" value="yes"/>
      <supports element="accidental" type="yes"/>
      <supports element="beam" type="yes"/>
      <supports element="stem" type="yes"/>
    </encoding>
  </identification>
  <part-list>
    <score-part id="P1">
      <part-name>Oboe</part-name>
      <score-instrument id="P1-I1">
        <instrument-name>Oboe</instrument-name>
      </score-instrument>
    </score-part>
  </part-list>
  <part id="P1">
    <measure number="1" width="285">
      <attributes>
        <divisions>8</divisions>
        <key>
          <fifths>3</fifths>
          <mode>major</mode>
        </key>
        <time>
          <beats>4</beats>
          <beat-type>4</beat-type>
        </time>
        <clef>
          <sign>G</sign>
          <line>2</line>
        </clef>
      </attributes>
      <sound tempo="114"/>
      <note default-x="17">
        <pitch>
          <step>F</step>
          <alter>1</alter>
          <octave>4</octave>
        </pitch>
        <duration>8</duration>
        <tie type="stop"/>
        <voice>1</voice>
        <type>quarter</type>
        <stem default-y="1">up</stem>
        <notations>
          <tied type="stop"/>
        </notations>
      </note>
      <note default-x="69">
        <rest>
          <display-step>B</display-step>
[...]

Issue #417, opened 2017-01-07
space element ignored in tuplet


craigsapp:
Verovio does not like space elements within tuplets:

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2017-01-07 06:47:37</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.13-dev-78fe7f7-dirty</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-000000075403378">
                        <staffGrp xml:id="m-000000008255812">
                            <staffDef xml:id="staffdef-000000028000452" clef.shape="G" clef.line="2" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-000000158477552">
                        <measure xml:id="measure-L2" n="1" right="end">
                            <staff xml:id="staff-L2F1" n="1">
                                <layer xml:id="layer-L2F1" n="1">
                                    <tuplet xml:id="tuplet-000000123897701" num="3" numbase="2" bracket.visible="true" num.format="count">
                                        <note xml:id="note-L3F1" dur="8" oct="4" pname="c" accid.ges="n" />
                                        <space xml:id="note-L4F1" dur="8" />
                                        <note xml:id="note-L5F1" dur="8" oct="4" pname="d" accid.ges="n" />
                                    </tuplet>
                                </layer>
                            </staff>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

This data gets rendered with no space:

[...]

PR #567, opened 2017-04-28
improved MusicXML import

rettinghaus:
several fixes and improvements
most importantly support for forward and backward elements


rettinghaus:
This should fix #427. 
But in some cases it will break while trying to add a space to a beam or a tuplet (similar to #417).

PR #578, opened 2017-05-03
Fix an error in the ls call for java build.sh

bcornec:


java/build.sh
```diff
@@ -9,7 +9,7 @@ mkdir -p target/classes/META-INF/lib
 
 swig -c++ -java -package org.rismch.verovio -outdir src/main/java/org/rismch/verovio verovio.i
 
-SRCFILES=`LS ../src/*.cpp`
+SRCFILES=`ls ../src/*.cpp`
```
lpugin: Unfortunately this will not work when `ls` is aliased and shows more that  just the file name. E.g,

```bash
rantanplan:verovio_develop laurent$ ls ./src/*.cpp
-rw-r--r--+ 1 laurent  ldap    7012 May  2 12:31 ./src/accid.cpp
-rw-r--r--+ 1 laurent  ldap    1587 Mar 14 10:23 ./src/anchoredtext.cpp
-rw-r--r--+ 1 laurent  ldap   16167 May  2 12:31 ./src/artic.cpp
-rw-r--r--+ 1 laurent  ldap   16243 Mar 14 10:23 ./src/att.cpp
-rw-r--r--+ 1 laurent  ldap    1577 May  2 12:31 ./src/barline.cpp
-rw-r--r--+ 1 laurent  ldap    9649 Apr 23 09:27 ./src/bboxdevicecontext.cpp
-rw-r--r--+ 1 laurent  ldap   21609 May  2 12:31 ./src/beam.cpp
-rw-r--r--+ 1 laurent  ldap    5675 Apr 12 13:21 ./src/boundary.cpp
-rw-r--r--+ 1 laurent  ldap   29865 Apr 23 09:27 ./src/boundingbox
...
```

I am not sure how to solve this.

bcornec: Using \ls avoids using aliases
Another way is to call /usr/bin/ls and in that case the shell won't replace it either.

Let me know which one is the most suitable and I'll propose and updated patch.
The problem is that currently LS is just not working at all :-(

bcornec: FYI, I'm packaging verovio for the Mageia distribution, and I may submit other comments due to what I find doing this.

lpugin: `\ls` looks like a good solution, thanks

rettinghaus: +1 for `\ls`
It should be used in the other shell scripts too.

bcornec:
I also updated the other script

lpugin:
I just noticed that this was a PR to master. Please next time make it to the develop branch.

bcornec:
Do you want me to also propose the same patch for devel ?

lpugin:
Yes please. I tried to cherry-pick it but go a missing -m error.

PR #579, opened 2017-05-03
Adding basic Figured Bass support

xhero:
It also parses the <harm> elements so accidentals are shown with the VerovioText font.


xhero:
Sorry not in master...

PR #580, opened 2017-05-03
Adding basic Figured Bass support

xhero:
On the correct branch this time.
It also parses the <harm> elements so accidentals are shown with the VerovioText font.


rettinghaus:
looks nice. But `f` an `fb` elements won't get rendered in containers of their correspondig class in the SVG.

lpugin:
It must be a Start/EndGraphic missing

PR #581, opened 2017-05-03
enhancing figured bass

rettinghaus:
This adds the missing `g.class="fb"` container to the SVG output. Also it sets the default font weight for figured bass to bold.


craigsapp:
Can you post a sample test of figured bass (here or wherever)?  Or is there already one on the verovio website?

rettinghaus:
Currently figures are aligned like simple `<harm>` and placed above by default. 

figures_001

```xml
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title/>
            </titleStmt>
            <pubStmt/>
        </fileDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef meter.count="4" meter.unit="4" meter.sym="common">
                        <staffGrp symbol="brace" barthru="true">
                            <staffDef n="1" clef.shape="G" clef.line="2" lines="5"/>
                            <staffDef n="2" clef.shape="F" clef.line="4" lines="5"/>
                        </staffGrp>
                    </scoreDef>
                    <section>
                        <measure n="1" right="end">
                            <staff n="1">
                                <layer n="1">
                                    <multiRest num="1"/>
                                </layer>
                            </staff>
                            <staff n="2">
                                <layer n="5">
                                    <note dur="4" oct="3" pname="c" stem.dir="up"/>
                                    <note dur="4" oct="3" pname="e" stem.dir="down"/>
                                    <note dur="2" oct="3" pname="g" stem.dir="down"/>
                                </layer>
                            </staff>
                            <harm tstamp="1" place="above" staff="2">
                                <fb>
                                    <f>♭</f>
                                </fb>
                            </harm>
                            <harm tstamp="2" place="below" staff="2">
                                <fb>
                                    <f>6♭</f>
                                    <f>4</f>
                                </fb>
                            </harm>
                            <harm tstamp="3" place="below" staff="2">
                                <fb>
                                    <f>7</f>
                                    <f>5</f>
                                    <f>4</f>
                                </fb>
                            </harm>
                            <harm tstamp="4" place="below" staff="2">
                                <fb>
                                    <f>_</f>
                                    <f>_</f>
                                    <f>♯</f>
                                </fb>
                            </harm>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

lpugin:
I am not convinced by the bold font. I would rather not have it by default

lpugin:
Also, you can `<f>—</f>` or `<f>–</f>` for extenders instead of underscore

rettinghaus:
Making them bold seems to be more traditional. I would propose to make it so for now as long we don't use the *correct* SMUFL glyph: 
https://w3c.github.io/smufl/gitbook/tables/figured-bass.html

MuseScore:
figures

lpugin:
@xhero ? What do you think?

lpugin:
We can still change it

craigsapp:
I would say this particular bold font does not look good:

screen shot 2017-05-05 at 8 28 23 am

and probably looks better in regular weight.

The MuseScore font looks nice:

screen shot 2017-05-05 at 8 29 47 am

There are also slash figures which will probably require dedicated symbols for the F text.  Example from MuseData renderer:

screen shot 2017-05-05 at 8 39 11 am


Also notice that the current verovio figures are left-justified to the center of notes.  They should either be left-justified to the left side of notes, or center-justified (as a stack group excluding the accidental qualifiers) on the center of the noteheads.  If they are left justified, be careful of cases where the accidental will precede the number -- the left-justification would still be for the number and not the accidental).

craigsapp:
Here is the style used for the C.P.E. Bach: Complete Works edition:

screen shot 2017-05-05 at 8 45 47 am

(click to view larger size) 

They are unbold.

rettinghaus:
I am wondering why notes are centered on time points unlike everything else. When a hairpin, a bass figure and a note *start* on `tstamp="1"` shouldn't they be all aligned to the same vertical position?

craigsapp:
Also notice that figured bass are not typically placed in an exactly level horizontal line such as for vereses and `<harm>`:

screen shot 2017-05-05 at 8 49 52 am

rettinghaus:
ok, we'll un-bold the fb font again. But what should be the default? SMUFL glyphs (to be added to Verovio text), or Times?

craigsapp:
> I am wondering why notes are centered on time points unlike everything else. When a hairpin, a bass figure and a note start on tstamp="1" shouldn't they be all aligned to the same vertical position?

SCORE aligns timepoints to the left side of noteheads.

lpugin:
> SCORE aligns timepoints to the left side of noteheads.

What happens when you have a whole note and a quarter note, of flipped note heads in chords?

craigsapp:
> ok, we'll un-bold the fb font again. But what should be the default? SMUFL glyphs (to be added to Verovio text), or Times?

Eventually you will probably need to use  SMUFL for things like this:

screen shot 2017-05-05 at 9 06 43 am
http://www.smufl.org/version/latest/range/figuredBass/

So you can do it now or later :-)

lpugin:
> I am wondering why notes are centered on time points unlike everything else. When a hairpin, a bass figure and a note start on tstamp="1" shouldn't they be all aligned to the same vertical position?

Yes, but their alignment differers. You would expect a tempo indication to be left-aligned, others will be centred. But this is a different topic to be discussed somewhere else.

Regarding Figured-bass we can maybe have two encoding / rendering option. The current one with a text font, and alternatively a more semantic approach where SMuFL font would be used. @xhero will discuss this during the conference in Tours so we can at least wait until then ;-)

xhero:
Speaking only of aesthetics, I prefer the "bold", i.e. traditional font for figured bass. But I know many many performers that prefer a normal looking (sometimes even sans-serif) font, and many modern editions do it this way. That's why I did not use from start the more traditional look.
Vertical alignment is again not standard. I personally dislike figures that "float" unevenly over or under notes, and I generally like to find an alignment that works over or under the highest or lower note, with the exception of those cases when only a couple figures would raise (or lower) everything unreasonably.

craigsapp:
> What happens when you have a whole note and a quarter note, of flipped note heads in chords?

Here is the behavior of SCORE for these cases:

screen shot 2017-05-05 at 9 12 51 am

The noteheads are left justified to the timepoints, with the whole note being wider and extending more to the right.

Here is an example where the stem is on the right side of the note:

screen shot 2017-05-05 at 9 16 52 am

(everything is still left-justified to the timepoint, even the wider whole notes).

The whole-notes have horizontal overlap in seconds:

screen shot 2017-05-05 at 9 24 51 am

craigsapp:
> Speaking only of aesthetics, I prefer the "bold", i.e. traditional font for figured bass. 

That is fine, but I would prefer "semi-bold":

screen shot 2017-05-05 at 9 29 08 am

Not all fonts include a full range of font weights however.  Ultimately it would depend on the purpose of the notation (performance/non-performance) and medium the music notation is displayed on (print/screen).  

This:

screen shot 2017-05-05 at 9 32 24 am

looks traditional in the sense of a 1980's Apple LaserWriter :-)


> But I know many many performers that prefer a normal looking (sometimes even sans-serif) font, and many modern editions do it this way. That's why I did not use from start the more traditional look.

That is something to consider.  Although at the moment verovio is not performance-quality notation for complicated cases such as keyboard notation, and it does not have an easy output path to paper.

> Vertical alignment is again not standard. I personally dislike figures that "float" unevenly over or under notes, and I generally like to find an alignment that works over or under the highest or lower note, with the exception of those cases when only a couple figures would raise (or lower) everything unreasonably.

More horizontal layout of figured bass is probably the style that will be adopted in the computer age, and I too would probably prefer flat layout. 

See the MuseData edition of Corelli music:

http://corelli.ccarh.org

Such as:

http://scores.ccarh.org/corelli/op3/op3n08.pdf

screen shot 2017-05-05 at 10 04 04 am

It is basically horizontal, but with occasional veritcal avoidances of the music as you mention.

 Note there will probably be complications related to how to vertically align a stack of two figures compared to one figure (such as "4 3" going to "6", does the "4" and "6" align horizontally, or does the "3" and "6" align?  And similarly for more figures: "9 6 4" followed by "6 5" for example.


 Non-horizontal layout is useful for generating tight vertical spacing of the music notation, which publishers like in order to save materials cost, and performers tend to like in terms of having less page turns.

I even prefer a horizontal layout for finger numbers, which is not standard:

screen shot 2017-05-05 at 9 49 14 am

Compare to the CPEB:CW layout of fingering in such cases where there is a number for each note:

screen shot 2017-05-05 at 9 52 48 am

This was a study sonata that C.P.E. Bach wrote for his students, and he added fingering numbers for each note.  I find that difficult to read since I can figure out the fingering numbers myself for simple music like this, so I added a hide-fingering feature:

screen shot 2017-05-05 at 9 54 16 am

lpugin:
I am looking at this and I wonder what to do with the alignment of the control events. For example, for slurs we will want to have the starting / ending point on the centre of the note head. This means a different position depending on the duration. This is feasible, but what about other control events, such as hairpins? And what about time stamps? To which position do they refer? The left one, the centre of a quarter note, or the center of a whole note?

image

craigsapp:
> I am looking at this and I wonder what to do with the alignment of the control events. For example, for slurs we will want to have the starting / ending point on the centre of the note head. This means a different position depending on the duration. This is feasible, but what about other control events, such as hairpins? And what about time stamps? To which position do they refer? The left one, the centre of a quarter note, or the center of a whole note?

In SCORE, timestamps are always defined as the left edge of notes.  There is an escape mechanism of an extra horizontal displacement for notes to be offset for chords with seconds or similar cases, where they need to be pushed off of their default timestamp position.

All events placed at the same timestamp as the note will be left justified to the left edge of the notes if they are not placed in relation to the notehead themselves, and there is a separate horizontal offset parameter to displace the item from that position if needed.  So for example, hairpins will by default be left-justified to noteheads if they start/stop at the same timestamp.  Likewise dynamic markings will be left justified to the left edge of noteheads.  And lyric text works the same way: it is left justified to the left edge of notehead by default.  Exceptions are done with a separate horizontal offset parameter.


For slurs, there will need to be the capability to align the slur to the left, right and center of the noteheads (both as an automatic decision and later as some sort of attribute).  The slur needs to be attached to notes rather than timestamps (in SCORE, notes are attached to timestamps, but slurs attach instead to the notehead.  Consider this example:

screen shot 2017-07-18 at 11 43 35 am

```
**kern
*M4/4
=1-
(8cL
8d
8e
8fJ)
=
(8cL<
8d
8e
8fJ)
=
(>8cL
8d
8e
8fJ)
=
(<8cL<
8d
8e
8fJ)
==
!!!RDF**kern: > = above
!!!RDF**kern: < = below
```

The slurs in the first two measures are correctly placed, where they are centered on the noteheads. 

However slurs attached to the beam side of notes in this example not typographically correct:  the slur needs to be centered on the beam, or shifted 1/2 note to the right in measure 3 so that the start and end of the slur is right-justified to the notes which it is attached to, and 1/2 note to the left in measure 4.

Here is what measure 3 and 4 should look like:

screen shot 2017-07-18 at 11 56 29 am


Parameter 8 for slurs in SCORE is used to control the justification of slur ends on notes:

http://wiki.ccarh.org/wiki/SCORE_slur_parameters#P8:_Slur.2Ftie_offset

screen shot 2017-07-18 at 12 07 24 pm

0 = Left to Left sides of notes
-1 = Center to Center
-4 = Right to Right
-5 = Center to Right
-6 = Right to Center
-7 = Center to Left
-8 = Left to Center
-9 = Left to Right

For ties, there are two separate cases:
-2 = a little to the right of the right edge of the start note to a little to the left of the left edge of the ending note.  This is typically used for ties.
-3 = a little more to the right of the right edge of the start note to a little to the left of the left edge of the ending note.  This is typically used for ties where the starting note posses an augmentation dot.

lpugin:
work-in-progess in the develop-xalign branch. Probably quite a few elements are now badly aligned (dynamics, figured-bass, etc.)

Issue #596, opened 2017-05-17
Stems should be rects


rettinghaus:
Stems are drawn now as SVG `path`, but should be a `rect` (as it was done in earlier versions).

lpugin:
Indeed. Should be changed here https://github.com/rism-ch/verovio/blob/develop/src/view_element.cpp#L1230-L1231

Issue #556, opened 2017-04-22
Implement currentColor in SVG


craigsapp:
It would be useful to stroke elements in SVG images with `currentColor` rather than `#000000`.

Info about currentColor:

https://developer.mozilla.org/en/docs/Web/CSS/color_value#currentcolor_keyword
https://css-tricks.com/cascading-svg-fill-color

Example CSS:

```css
<style>
        svg g.note {
                color: red;
                stroke: deeppink !important;
                fill: blue;
        }
</style>
```

Current SVG output from verovio uses a stroke of `#000000` for stems:

```xml
      <g class="note" id="note-L2F1">
        <use xlink:href="#E0A4" x="1061" y="1940" height="720px" width="720px" />
        <g class="stem" id="stem-1">
         <path d="M1277 1918 L1277 1310" stroke="#000000" stroke-width="20" />
         <g class="flag" id="flag-1">
          <use xlink:href="#E240" x="1267" y="1310" height="720px" width="720px" />
         </g>
        </g>
```

This will be rendered as:

<img width="1349" alt="screen shot 2017-04-22 at 8 19 41 am" src="https://cloud.githubusercontent.com/assets/3487289/25305661/7664fdb0-2734-11e7-9597-5f2a286d596d.png">

The stem remains black, even if `!important` is added to try to override the black color:
https://css-tricks.com/when-using-important-is-the-right-choice

However, if the stroke is set to `currentColor`, CSS styling can be applied:

```xml
      <g class="note" id="note-L2F1">
        <use xlink:href="#E0A4" x="1061" y="1940" height="720px" width="720px" />
        <g class="stem" id="stem-1">
         <path d="M1277 1918 L1277 1310" stroke="currentColor" stroke-width="20" />
         <g class="flag" id="flag-1">
          <use xlink:href="#E240" x="1267" y="1310" height="720px" width="720px" />
         </g>
        </g>
```

which is rendered as:

<img width="1335" alt="screen shot 2017-04-22 at 8 21 54 am" src="https://cloud.githubusercontent.com/assets/3487289/25305683/c458d140-2734-11e7-98f6-61e750f89c06.png">

In this case the stem is colored red from the CSS `color` parameter setting (not `stroke` since that is hard-coded to the soft variable `currentColor`).

Also if `fill="#000000` is ever used (it is not in my test image), then it should be replaced with `fill=currentColor` unless black is explicitly wanted such as from an MEI `color="#000000"` parameter.

[...]

PR #597, opened 2017-05-17
changed DrawStem and added color to beam

rettinghaus:
This changes stems from SVG `path` to `rect` to address #596 
Additionally it adds missing color support to beam and removes the no longer needed `stroke` from SVG.

drawstem_001

```xml
<?xml-model href="http://www.music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://www.music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title/>
      </titleStmt>
      <pubStmt/>
      <seriesStmt>
        <title>Brawing stems</title>
      </seriesStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef meter.count="3" meter.unit="8">
            <staffGrp>
              <staffDef n="1" clef.line="3" clef.shape="C" lines="5"/>
            </staffGrp>
          </scoreDef>
          <section>
            <measure n="1">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="c" color="blue"/>
                  <note dur="8" oct="4" pname="c" color="red"/>
                </layer>
              </staff>
            </measure>
            <measure n="2">
              <staff n="1">
                <layer n="1">
                  <beam color="lightgray">
                    <note dur="8" oct="4" pname="c" color="green"/>
                    <note dur="8" oct="4" pname="c"/>
                    <note dur="8" oct="4" pname="c"/>
                  </beam>
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

### DIFF
--- a/src/view_element.cpp
+++ b/src/view_element.cpp
@@ -307,7 +307,7 @@ void View::DrawArticPart(DeviceContext *dc, LayerElement *element, Layer *layer,
         xCorr = m_doc->GetGlyphWidth(code, staff->m_drawingStaffSize, drawingCueSize) / 2;
         // The position of the next glyph (and for correcting the baseline if necessary
         int glyphHeight = m_doc->GetGlyphHeight(code, staff->m_drawingStaffSize, drawingCueSize);
-        
+
         // Center the glyh if necessary
         if (Artic::IsCentered(*articIter)) {
             y += (articPart->GetPlace() == STAFFREL_above) ? -(glyphHeight / 2) : (glyphHeight / 2);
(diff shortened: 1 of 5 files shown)


src/view_element.cpp
```diff
@@ -1227,8 +1227,9 @@ void View::DrawStem(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
 
     dc->StartGraphic(element, "", element->GetUuid());
 
-    DrawVerticalLine(dc, stem->GetDrawingY(), stem->GetDrawingY() - stem->GetDrawingStemLen(), stem->GetDrawingX(),
-        m_doc->GetDrawingStemWidth(staff->m_drawingStaffSize));
+    DrawFilledRectangle(dc, stem->GetDrawingX() + m_doc->GetDrawingStemWidth(staff->m_drawingStaffSize) / 2,
```
lpugin: we should have x1 < x2, so - the half for x1 and + the half for x2

rettinghaus: makes sense

rettinghaus: refers to #556.

Issue #603, opened 2017-05-26
Issue Importing on Ubuntu/Python3


misingnoglic:
Hello,

I'm trying to host a web-app using verovio, and I've created a droplet on DigitalOcean to try to host the app. The droplet runs Ubuntu 16.04.2

I've successfully installed verovio on python3.6 using the installation steps in the wiki (and tweaking the memory of the droplet). However, when I try to import verovio, I get this large error:

```
root@polyphony:~/verovio/python# python3.6
Python 3.6.1 (default, Mar 22 2017, 10:05:53)
[GCC 5.4.1 20170304] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import verovio
Traceback (most recent call last):
  File "/root/verovio/python/verovio.py", line 18, in swig_import_helper
    fp, pathname, description = imp.find_module('_verovio', [dirname(__file__)])
  File "/usr/lib/python3.6/imp.py", line 296, in find_module
    raise ImportError(_ERR_MSG.format(name), name=name)
ImportError: No module named '_verovio'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/verovio/python/verovio.py", line 28, in <module>
    _verovio = swig_import_helper()
  File "/root/verovio/python/verovio.py", line 20, in swig_import_helper
    import _verovio
ImportError: /usr/local/lib/python3.6/dist-packages/verovio-0.9.12-py3.6-linux-x86_64.egg/_verovio.cpython-36m-x86_64-linux-gnu.so: undefined symbol: _ZN3hum17Tool_musicxml2hum7convertERSoRN4pugi12xml_documentE
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 63, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 30, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 23, in <module>
    from apport.packaging_impl import impl as packaging
  File "/usr/lib/python3/dist-packages/apport/packaging_impl.py", line 23, in <module>
    import apt
[...]

PR #604, opened 2017-05-26
updated setup.py to compile all files in setum/hum/*.cpp

misingnoglic:
Change related to issues found in #603

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -22,7 +22,7 @@ else:
 
 verovio_module = Extension('_verovio',
     sources=
-        glob('../src/*.cpp') +
+        glob('../src/*.cpp') + glob('../src/hum/*.cpp') +
         [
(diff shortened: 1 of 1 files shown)


ahankinson:
Why did you merge this to master?

craigsapp:
Oops, because I was presuming that the commit branch was develop and didn't check.  Probably easiest to also add manually to the develop branch since it will probably be useful to keep in the master and it is a small change.

ahankinson:
I will revert these. The pull request will have to be re-submitted against `develop`.

ahankinson:
Ah, but you've already done that. So I'll just remove it from master.

craigsapp:
Yes that is good.  I don't want to remove from the master just in case I instead remove the master ...

ahankinson:
Done.

PR #606, opened 2017-06-01
Bringing labels to life

rettinghaus:
This brings a cool new feature to SVG rendering: 
As discussed in music-encoding/music-encoding#361 MEI's `@label` is like `@title` in HTML. With this addition label attributes will get translated to `<title>` elements within the SVG `<g>` container. Standard behavior in browsers is then, when hovering over a glyph a tooltip opens, showing the content of `@label`. 

Additionally I updated pugixml to latest version.


lpugin:
This looks interesting, thanks. Could you try to do it in develop-MEI-4.0.0? I think you will need to inherit from AttLabelled (and not AttCommonPart). Also, I would make it child of LayerElement so it is available for all child classes without having to do it explicitly for each class. OK?

rettinghaus:
Done. Should `AttCommonPart` for MEI3 be moved to `LayerElement` too?

lpugin:
If you do not mind I would prefer to have this only in the MEI 4.0 branch. If we make a version 1.0.1 of Verovio still with MEI 3.0 it would not support this feature. This will avoid us merging conflicts. But if you really need it we can see. In that case it should definitely be moved to LayerElement

rettinghaus:
I think it is a useful feature. And other recent improvements definitely would justify a version 1.1. 
Wouldn't the jump to MEI 4.0 go well with a step up to Verovio 2.0?

lpugin:
Having a version 2.0 for MEI 4.0 is the right way to go. We can also have a 1.1 before. I'll have a try with the merge and see what the situation is.

PR #608, opened 2017-06-02
Adding labels

rettinghaus:


src/iomei.cpp
```diff
@@ -794,6 +793,7 @@ void MeiOutput::WriteMeiBreath(pugi::xml_node currentNode, Breath *breath)
     WriteControlElement(currentNode, breath);
     WriteTimePointInterface(currentNode, breath);
     breath->WriteColor(currentNode);
+    breath->WriteLabelled(currentNode);
```
lpugin: Reading and writting AttLabel should now hapen in Read / WriteLayerElement

rettinghaus: fixed

Issue #345, opened 2016-11-04
mRests in different layers merge automatically


craigsapp:
`<mRest>`'s in multiple layers merge into a single rest:

<img width="315" alt="screen shot 2016-11-03 at 18 08 13" src="https://cloud.githubusercontent.com/assets/3487289/19991197/cad4eda2-a1f0-11e6-9c63-cafb9e7a98f5.png">

MEI data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2016-11-03 18:06:04</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.13-dev-7fea990</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-0000000731620907">
                        <staffGrp xml:id="m-0000000340070014">
                            <staffDef xml:id="staffdef-0000001102740631" clef.shape="G" clef.line="2" key.sig="0" meter.count="4" meter.unit="4" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-0000001859886478">
                        <measure xml:id="measure-L1" n="0">
                            <staff xml:id="staff-0000001848995428" n="1">
                                <layer xml:id="layer-L1F2" n="1">
                                    <mRest xml:id="mrest-0000001902556676" />
                                </layer>
                                <layer xml:id="layer-L8F3" n="2">
                                    <mRest xml:id="mrest-0000000350665698" />
                                </layer>
                            </staff>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

Compare to the behavior of verovio with regular rests:

[...]

Issue #327, opened 2016-10-06
MEI tupletSpan example


rettinghaus:
The tupletSpan.mei from the examples isn't encoded correctly as the startid end endid attributes in the tupleSpan elements don't carry real references but just the ids (the '#' is missing). I'll fix it. 
My problem is, that Verovio actually understands it this way. Should Verovio be so kind?

rettinghaus:
(Same in other docs.)

rettinghaus:
And `<tempo>` in tempo2.mei has neither `@startid` nor `@tstamp` but gets rendered.

rettinghaus:
Can this be closed now?

PR #609, opened 2017-06-03
Develop loc

rettinghaus:
* adds `@loc` to `<mRest>` and `<note>`
* better handling for notes without `@pname`
* fixes ledger lines on rests
* addresses #327 and #345

mrests

```xml
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title />
      </titleStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef>
            <staffGrp>
              <staffDef meter.count="3" meter.unit="4" n="1" lines="5" clef.shape="G" clef.line="2"/>
              <staffDef meter.count="3" meter.unit="4" n="2" lines="1" />
            </staffGrp>
          </scoreDef>
          <section>
            <measure >
              <staff n="1">
                <layer n="1">
                  <mRest />
                </layer>
                <layer n="2">
                  <mRest />
                </layer>
              </staff>
              <staff n="2">
                <layer n="1">
                  <mRest />
                </layer>
              </staff>
            </measure>
            <measure>
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="c" />
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="g" />
                </layer>
              </staff>
              <staff n="2">
                <layer n="1">
                  <note dur="4" />
                  <note dur="8" dots="1"/>
                  <note dur="16" />
                  <note dur="4" />
                </layer>
              </staff>
            </measure>
            <measure right="end">
              <staff n="1">
                <layer n="1">
                  <note dots="1" dur="2" fermata="above" oct="4" pname="f" accid.ges="n" />
                </layer>
              </staff>
              <staff n="2">
                <layer n="1">
                  <note dur="4" />
                  <rest dur="4" />
                  <rest dur="4" fermata="above" />
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

### DIFF
--- a/src/chord.cpp
+++ b/src/chord.cpp
@@ -359,9 +359,10 @@ int Chord::CalcStem(FunctorParams *functorParams)
     this->GetYExtremes(yMax, yMin);
     params->m_chordStemLength = yMin - yMax;
 
-    int staffY = staff->GetDrawingY();
     int staffSize = staff->m_drawingStaffSize;
-    params->m_verticalCenter = staffY - params->m_doc->GetDrawingDoubleUnit(staffSize) * 2;
+
+    params->m_verticalCenter
+        = staff->GetDrawingY() - params->m_doc->GetDrawingUnit(staffSize) * (staff->m_drawingLines - 1);
 
     data_STEMDIRECTION stemDir = STEMDIRECTION_NONE;
 
--- a/src/layerelement.cpp
+++ b/src/layerelement.cpp
@@ -691,8 +691,7 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
         Note *note = chord->GetTopNote();
         assert(note);
         int loc = PitchInterface::CalcLoc(note->GetPname(), note->GetOct(), layerY->GetClefLocOffset(layerElementY));
-        // Once we have AttLoc on Note
-        // if (note->HasLoc()) loc = note->GetLoc();
+        if (note->HasLoc()) loc = note->GetLoc();
         this->SetDrawingYRel(staffY->CalcPitchPosYRel(params->m_doc, loc));
     }
     else if (this->Is({ CUSTOS, DOT })) {
@@ -704,9 +703,12 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
         Note *note = dynamic_cast<Note *>(this);
         assert(note);
         Chord *chord = note->IsChordTone();
-        int loc = PitchInterface::CalcLoc(note->GetPname(), note->GetOct(), layerY->GetClefLocOffset(layerElementY));
-        // Once we have AttLoc on Note
-        // if (note->HasLoc()) loc = note->GetLoc();
+        int loc = 0;
+        if (note->HasPname()) {
+          loc = PitchInterface::CalcLoc(note->GetPname(), note->GetOct(), layerY->GetClefLocOffset(layerElementY));
+        }
+        // should this override pname/oct
+        if (note->HasLoc()) loc = note->GetLoc();
         int yRel = staffY->CalcPitchPosYRel(params->m_doc, loc);
         // Make it relative to the top note one (see above) but not for cross-staff notes in chords
         if (chord && !m_crossStaff) {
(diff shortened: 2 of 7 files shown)


src/view_element.cpp
```diff
@@ -1586,7 +1586,7 @@ void View::DrawRestWhole(DeviceContext *dc, int x, int y, int valeur, bool cueSi
     if (y > (int)staff->GetDrawingY()
         || y < staff->GetDrawingY()
                 - (staff->m_drawingLines - 1) * m_doc->GetDrawingDoubleUnit(staff->m_drawingStaffSize))
-        DrawHorizontalLine(dc, x1, x2, y1, m_doc->GetDrawingStaffLineWidth(staff->m_drawingStaffSize));
+        DrawHorizontalLine(dc, x1, x2, y1, m_doc->GetDrawingStaffLineWidth(staff->m_drawingStaffSize) * 1.75);
```
craigsapp: 1.75 would be good to prepare for styling.  Maybe a class to store style settings?  This variable could be accessed with:

`Style::GetLedgerLineToStaffLineRatio()`

And could be changed with:

`Style::SetLedgerLineToStaffLineRatio()`

and/or a function called:

`m_doc->GetDrawingLedgerLineWidth(staff->m_drawingStaffSize()`

could be added.

rettinghaus: It's hardcoded for now, see: View::DrawLedgerLines
It has to be reviewed later on, as mRests should use the same functions as notes.

src/layerelement.cpp
```diff
@@ -704,9 +703,12 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
         Note *note = dynamic_cast<Note *>(this);
         assert(note);
         Chord *chord = note->IsChordTone();
-        int loc = PitchInterface::CalcLoc(note->GetPname(), note->GetOct(), layerY->GetClefLocOffset(layerElementY));
-        // Once we have AttLoc on Note
-        // if (note->HasLoc()) loc = note->GetLoc();
+        int loc = 0;
+        if (note->HasPname()) {
```
lpugin: For readability, I would prefer to have both `if` (this one and the one below) with our without `{}` but both the same. Small detail...

rettinghaus: Makes sense, so let's embrace braces.

src/chord.cpp
```diff
@@ -359,9 +359,10 @@ int Chord::CalcStem(FunctorParams *functorParams)
     this->GetYExtremes(yMax, yMin);
     params->m_chordStemLength = yMin - yMax;
 
-    int staffY = staff->GetDrawingY();
     int staffSize = staff->m_drawingStaffSize;
-    params->m_verticalCenter = staffY - params->m_doc->GetDrawingDoubleUnit(staffSize) * 2;
+
+    params->m_verticalCenter
+        = staff->GetDrawingY() - params->m_doc->GetDrawingUnit(staffSize) * (staff->m_drawingLines - 1);
```
lpugin: What does this changes?

src/note.cpp
```diff
@@ -397,9 +397,9 @@ int Note::CalcStem(FunctorParams *functorParams)
     params->m_isGraceNote = this->IsGraceNote();
 
     int staffSize = staff->m_drawingStaffSize;
-    int staffY = staff->GetDrawingY();
 
-    params->m_verticalCenter = staffY - params->m_doc->GetDrawingDoubleUnit(staffSize) * 2;
+    params->m_verticalCenter
+        = staff->GetDrawingY() - params->m_doc->GetDrawingUnit(staffSize) * (staff->m_drawingLines - 1);
```
lpugin: Same here?

src/chord.cpp
rettinghaus: Instead of moving two lines down from top it really points to the vertical center of a staff (important for staves that have not 5 lines).  
Alternatively we could limit automatic stem direction to staves with five lines.

src/note.cpp
rettinghaus: Same here!

src/chord.cpp
lpugin: Makes sense

Issue #389, opened 2016-12-20
implement <annot> display


craigsapp:
Rendering of `<annot>` data would be useful to have in verovio.  This would function in a very similar manner to the current `<harm>` implementation, but both should add horizontal collision avoidance between adjacent `<annot>` and `<harm>` entries (similar to the current implementation of `<verse>`.

It would also be useful if `@n` were implemented for `<annot>` (and perhaps `<harm>`) which would stack annotations in a similar manner to `<verse>`; i.e., `@n` would handle vertical collision avoidance between separate streams of annotations.

Test data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2016-12-19 17:26:49</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.13-dev-1745ac2</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef>
                        <staffGrp>
                            <staffDef clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section>
                        <measure n="1">
                            <staff n="1">
                                <layer n="1">
                                    <note dur="4" oct="4" pname="c" accid.ges="n" />
                                    <note dur="4" oct="4" pname="d" accid.ges="n" />
                                    <note dur="4" oct="4" pname="e" accid.ges="n" />
                                    <note dur="4" oct="4" pname="f" accid.ges="n" />
                                </layer>
                            </staff>
                            <annot n="1" label="test"   staff="1" tstamp="1.000000">this</annot>
                            <annot n="2" label="qstamp" staff="1" tstamp="1.000000">0</annot>
[...]

PR #641, opened 2017-06-26
Adding annotation text to SVG

rettinghaus:
this PR adds passing `<annot>` text to the SVG output as an `<desc>`


donbyrd:
This reminds me: I've been wondering what plans are for adding text annotation to Verovio output! I remember quite a bit of discussion awhile back in some issue, I forget which, but I don't recall any conclusion was reached. In the meantime, I've been using <syl> despite its annoying (for my purposes) habit of adding space before the next note -- as it should, since it thinks it's doing lyrics. But a couple of months ago it stopped adding space! That's nice for me, but still hardly ideal; I'd much rather use <annot>. So is there agreement to do it the way @rettinghaus has implemented?

pe-ro:
@donbyrd, it looks like you forgot some escape codes in your comment.  I'm interested in reading it in its entirety.  Can you fix it please?

donbyrd:
Oops. Fixed.

rettinghaus:
There were discussions about `<annot>` in #389 and in music-encoding/music-encoding/issues/424. I tried to go through them once again, but now I'm totally confused.

lpugin:
I think this is this appropriate implementation for non visible annotation as discussed here (https://github.com/rism-ch/verovio/issues/389#issuecomment-268623780) and once the `@visible` is added (if this is not the case already).

src/view_page.cpp
```diff
@@ -1208,6 +1208,10 @@ void View::DrawSystemEditorialElement(DeviceContext *dc, EditorialElement *eleme
     // if (element->m_visibility == Visible) {
     //    DrawSystemChildren(dc, element, system);
     //}
+    if (element->Is(ANNOT)) {
+        Annot *annot = dynamic_cast<Annot *>(element);
```
lpugin: I think we want to have a DrawAnnot method to be called from all DrawXXXEditorialElement. Eventually dealing with visible / invisible annot will be done in that method

rettinghaus: Had that at first, but removed it again as it just called `dc->AddDescription`. But it makes sense, especially when having `@plist` and wanting to do some nifty CSS pop-ups.

src/view_page.cpp
```diff
@@ -1304,11 +1311,19 @@ void View::DrawFbEditorialElement(DeviceContext *dc, EditorialElement *element,
...
     dc->EndTextGraphic(element, this);
 }
 
+void View::DrawAnnot(DeviceContext *dc, EditorialElement *element)
+{
+    assert(element);
+
+    dc->StartGraphic(element, "", element->GetUuid());
+
+    Annot *annot = dynamic_cast<Annot *>(element);
```
lpugin: always assert dynamic casts

src/svgdevicecontext.cpp
```diff
@@ -163,6 +163,11 @@ void SvgDeviceContext::StartGraphic(Object *object, std::string gClass, std::str
         }
     }
 
+    // make sure we are in a g container
+    while (strcmp(m_currentNode.name(), "g")) {
+        m_currentNode = m_currentNode.parent();
+    }
```
rettinghaus: This assures correct SVG structure in general and is needed for annotations within text elements. It just pushes the `g.annot` up to the parent `g`. (Tested thoroughly!)

lpugin: I am not convinced this is the way to go. Won't this change the hierarchy in the SVG? I have to have a closer look but probably the only thing to do is to call Start/EndTextGraphic when calling from DrawTextEditorialElement and Start/EndGraphic otherwise. This can be determined with an additional parameter in DrawAnnot. Wouldn't that work?

rettinghaus:
I got stuck. I close this for now.

lpugin:
Could you have a check if this https://github.com/rism-ch/verovio/commit/9a7435ae8e6e1a884368772219887b7f377e30d9 works (develop-annot branch) ?

Issue #642, opened 2017-06-29
Plaine & Easie character confusion


vonvlaho:
In the JS-toolkit seems to be a confusion of Plaine & Easie characters for grace notes. The IAML [recommends](http://www.iaml.info/plaine-easie-code) the "g" for acciaccaturas and the "q" for appoggiaturas, while verovio is doing it the opposite way. Will this be fixed? Otherwise we would switch to verovios semantics.

<img width="117" alt="g" src="https://user-images.githubusercontent.com/23097270/27699182-bf7f1526-5cf9-11e7-9800-1a8f27502198.png">

<img width="114" alt="q" src="https://user-images.githubusercontent.com/23097270/27699180-bf7ec29c-5cf9-11e7-82dd-ed270b4f195f.png">

craigsapp:
Fixed in commit https://github.com/rism-ch/verovio/commit/53b041a6d7cc204f41b2467738532b26ec2537dc

Example:

```
@clef:G-2
@keysig:xF
@timesig:3/8
@data:q8B''C/
```
Now generates:

<img width="370" alt="screen shot 2017-06-29 at 8 50 49 pm" src="https://user-images.githubusercontent.com/3487289/27704970-a8d014f2-5d0c-11e7-8445-28b74104dedd.png">

and 

```
@clef:G-2
@keysig:xF
@timesig:3/8
@data:g8B''C/
```

now creates:

<img width="357" alt="screen shot 2017-06-29 at 8 51 07 pm" src="https://user-images.githubusercontent.com/3487289/27704993-ba4dda48-5d0c-11e7-8fc7-03c31192c25c.png">

The correction was implemented in the `develop` branch (not the `master`), so you will have to use that version to see the revision.

craigsapp:
Also fixed in master branch: https://github.com/rism-ch/verovio/pull/646

vonvlaho:
Thanks for the fast reply. Can you also compile the verovio-toolkit.js? This seems not to have been updated yet. We're using this version: http://www.verovio.org/javascript/latest/verovio-toolkit.js.

lpugin:
Could you double check you are using version 1.1.5?
https://github.com/rism-ch/verovio/releases/tag/version-1.1.5

vonvlaho:
[...]

PR #646, opened 2017-06-30
fix PAE grace notes

rettinghaus:
fix for #642

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -92,7 +92,7 @@ bool PaeInput::ImportFile()
     parsePlainAndEasy(infile);
     return true;
 #else
-    LogError("Plain and Easy import is not supported in the build.");
+    LogError("Plaine & Easie import is not supported in the build.");
     return false;
 #endif
 }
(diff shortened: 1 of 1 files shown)


src/iopae.cpp
```diff
@@ -92,7 +92,7 @@ bool PaeInput::ImportFile()
     parsePlainAndEasy(infile);
     return true;
 #else
-    LogError("Plain and Easy import is not supported in the build.");
+    LogError("Plaine & Easie import is not supported in the build.");
```
craigsapp: Change in develop branch as well?

src/iopae.cpp
```diff
@@ -104,7 +104,7 @@ bool PaeInput::ImportString(std::string const &pae)
     parsePlainAndEasy(in_stream);
     return true;
 #else
-    LogError("Plain and Easy import is not support in the build.");
+    LogError("Plaine & Easie import is not support in the build.");
```
craigsapp: Change in develop branch as well?

src/iopae.cpp
rettinghaus: yes

src/iopae.cpp
rettinghaus: yes

PR #650, opened 2017-07-01
fixed spelling errors

rettinghaus:


pae-examples.xhtml
```diff
@@ -137,13 +137,13 @@ version-footer: true
                 <div class="highlight"><pre><code id="code_pae4"></code></pre></div>
             </p>
 
-            <h4>Grace notes (accaciature)</h4>
+            <h4>Grace notes (accacciature)</h4>
```
craigsapp: Are you sure it is spelled right now?  This website says it is "acciaccatura" (with the "i" in a different place):
https://en.wiktionary.org/wiki/acciaccatura

"acciaccature" is the Italian plural, but in an English context, using "acciaccaturas" would probably be preferrable.

rettinghaus: thanks for checking through! I didn't want to change to english on my own, but you're probably right.

pae-examples.xhtml
```diff
@@ -137,13 +137,13 @@ version-footer: true
...
             <div id="svg_pae5" style="overflow:auto;"/>
             <p>
                 <div class="highlight"><pre><code id="code_pae5"></code></pre></div>
             </p>
 
-            <h4>Grace notes (appogiature)</h4>
+            <h4>Grace notes (appoggiature)</h4>
```
craigsapp: Similar to acciaccature/acciaccaturas, a more preferred plural in English is appoggiaturas:
     https://en.wiktionary.org/wiki/appoggiatura

craigsapp:
English speakers can't handle plurals other than "s" :-), other than continually debating if it should be "octopusses" or "octopi":
     https://en.wiktionary.org/wiki/octopus

PR #652, opened 2017-07-03
Adding pae accidges

rettinghaus:


rettinghaus:
This adds accid.ges to notes according to the actual key. (Is needed for correct midi playback of PAE input.)

PR #653, opened 2017-07-03
improved tuplet bracket drawing

rettinghaus:
try this to fix the drawing error in humdrum-tools/verovio-humdrum-viewer/issues/37


craigsapp:
Thanks, that was fast!

It looks much better:

screen shot 2017-07-03 at 11 04 55 pm

Since single-note brackets are so small, perhaps it would be better to use double the width of the note for the bracket width.

PR #657, opened 2017-07-14
adding durdefault

rettinghaus:


rettinghaus:
I'm closing this one, as I can't decide where to put the static to keep the value.

lpugin:
Didn't we talk about putting in the the DurationInterface?

rettinghaus:
Can't remember. But where is the function it should be filled? Was it `CollectScoreDefs`?

PR #666, opened 2017-08-08
fix for #665

rettinghaus:


lpugin:
Thanks, that was quick!

rettinghaus:
More than two dots in PAE are not documented and should probably be added:
http://www.iaml.info/plaine-easie-code

lpugin:
@jenniferward : could you take care of this?

PR #696, opened 2017-09-21
Allow <ending> as sub-element of <score>.

yeonoson:
Converting the MusicXML example [45b-RepeatWithAlternatives](https://github.com/music-encoding/encoding-tools/blob/master/musicxml2mei/Lilypond-MusicXML-TestSuite-0.1/partwise/45b-RepeatWithAlternatives.xml) to MEI with the [musicxml2mei-3.0.xsl](https://github.com/music-encoding/encoding-tools/blob/master/musicxml2mei/musicxml2mei-3.0.xsl) stylesheet results in the following schema valid file: [45b-RepeatWithAlternatives.mei.txt](https://github.com/rism-ch/verovio/files/1322135/45b-RepeatWithAlternatives.mei.txt)

Verovio renders this as follows:
ending-old
and outputs two warnings:

    [Warning] Element <ending> within <score> is not supported and will be ignored
    [Warning] Element <ending> within <score> is not supported and will be ignored

This pull request fixes this, which results in the following rendering:
ending-new

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -4058,6 +4058,9 @@ bool MeiInput::ReadScoreBasedMei(pugi::xml_node element, Score *parent)
             else if (elementName == "section") {
                 success = ReadMeiSection(parent, current);
             }
+            else if (elementName == "ending") {
+                success = ReadMeiEnding(parent, current);
+            }
             else {
                 LogWarning("Element <%s> within <score> is not supported and will be ignored ", elementName.c_str());
             }
--- a/src/score.cpp
+++ b/src/score.cpp
@@ -14,6 +14,7 @@
 //----------------------------------------------------------------------------
 
 #include "editorial.h"
+#include "ending.h"
 #include "scoredef.h"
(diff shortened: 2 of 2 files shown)


lpugin:
Thanks! Do I understand correctly that the musicxml2mei-3.0.xsl  produces something like:
```xml
<score>
  <section>
    <measure/>
  </section
  <ending>
    <measure/>
  </ending>
   ...
</score>
```
You are right that this is valid. But does it really makes sense to close the section before the ending? @pe-ro?

yeonoson:
Yes, that is the structure produces by musicxml2mei-3.0.xsl. As MEI is new to me I can't judge if that makes sense.

kepper:
> 
> But does it really makes sense to close the section before the ending?

It does make sense – an ending is just a special section which may not be recursively nested. The guidelines seem pretty clear in this case:

http://music-encoding.org/documentation/3.0.0/shared/#sharedMdivContent

Issue #697, opened 2017-09-27
DARMS import


rettinghaus:
What is the status of DARMS input? I tried it with some files I found, but it doesn't seem to work. Also it is not documented on the Website. 
If it is not really needed, perhaps we should introduce `NO_DARMS_SUPPORT` and switch it of by default?

lpugin:
> we should introduce NO_DARMS_SUPPORT and switch it of by default?

I am OK with this because it is indeed currently probably quite broken

rettinghaus:
Then I'll add those lines over the next days.

craigsapp:
It does not work on the DARMS files that I have, so it could be for a certain variant of DARMS.

PR #698, opened 2017-09-28
addressing #697

rettinghaus:


src/keysig.cpp
```diff
@@ -61,21 +61,21 @@ int KeySig::octave_map[2][9][7] = {
 // KeySig
 //----------------------------------------------------------------------------
 
-KeySig::KeySig() : LayerElement("ksig-"), AttAccidental(), AttPitch()
+KeySig::KeySig() : LayerElement("keysig-"), AttAccidental(), AttPitch()
 {
     Init();
 }
 
 KeySig::KeySig(int alterationNumber, data_ACCIDENTAL_WRITTEN alterationType)
```
rettinghaus: This doesn't work. And as far I can see, this is only used in DARMS import?

Issue #439, opened 2017-01-18
short dynamics hairpin angle limit


craigsapp:
Short hairpin dynamics in verovio currently have the same height as long hairpin dynamics.  It would be better if the height of short hairpins were limited to a maximum angle:

<img width="329" alt="screen shot 2017-01-17 at 20 21 06" src="https://cloud.githubusercontent.com/assets/3487289/22050801/d0be0c7a-dcf2-11e6-99c2-e341317fbccb.png">

In this case the short hairpin on the eighth notes opens too wide given the length of the hairpin.

MEI data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2017-01-17 20:20:28</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.14-dev-fe01685</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-000000201540999">
                        <staffGrp xml:id="m-000000150502018">
                            <staffDef xml:id="staffdef-000000183058674" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-000000015777043">
                        <measure xml:id="measure-L3" n="1">
                            <staff xml:id="staff-L3F1" n="1">
                                <layer xml:id="layer-L3F1" n="1">
                                    <note xml:id="note-L4F1" dur="1" oct="4" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
                            <hairpin xml:id="hairpin-000000110600477" staff="1" tstamp="1.000000" tstamp2="1m+1.0" form="cres" />
                        </measure>
                        <measure xml:id="measure-L5" n="2">
[...]

PR #702, opened 2017-10-02
Restricting hairpin angles

DavidGarfinkle:
Our work at the hackathon as a quick-fix to #439.

As per Craig's comments, it seems a more aesthetic solution would be to allow the hairpin angles vary by width. But this way doesn't look too bad either:
hairpins2


lpugin:
Merge here 739d47d9d2d04f87881b0eab7580f5316adcff12

Issue #649, opened 2017-07-01
tuplet bracket + chord interaction


craigsapp:
Tuplet brackets are not drawn correctly when starting with a chord:

<img width="476" alt="screen shot 2017-07-01 at 1 43 21 pm" src="https://user-images.githubusercontent.com/3487289/27761773-4820ce50-5e63-11e7-91f5-717d3c5651ed.png">

As reported in pull request https://github.com/rism-ch/verovio/pull/634 (I don't see it in issues, so adding it here).

The problem results in the left edge of the tuplet being drawn at a large negative x position:

```xml
<g class="tuplet" id="tuplet-L3F1-L5F1">
        <use xlink:href="#E883" x="1679" y="2896" height="720px" width="720px" />
        <path d="M-2147482211 2774 L1639 2812" stroke="#000000" stroke-width="20" />
        <path d="M1883 2812 L2304 2737" stroke="#000000" stroke-width="20" />
        <path d="M-2147482211 2774 L-2147482211 2684" stroke="#000000" stroke-width="20" />
        <path d="M2304 2737 L2304 2647" stroke="#000000" stroke-width="20" />
```

The value -2147482211 is close to the maxumum value for signed 32-bit numbers, and could be a bad interaction between signed/unsigned ints (probably an underflow of an unsigned int which is then interpreted as a signed int).  The value may be 214783647 - 2147482211 = 1436 or -1436 or so.

A significant number of these tuplets (about 6 or so) in an SVG drawn directly on an HTML page will cause the SVG to not be displayed (will be drawn when displayed as an `<img>`, with the negative lines ommitted).

Test data:

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2017-07-01T13:43:57" version="2.0.0-dev-8215b04-dirty">
                    <name>Verovio</name>
[...]

PR #734, opened 2018-01-21
Bugfix 649 - Tuplet Chord Interaction

CCInc:
Chord elements don't have a SelfBB since they are not actually rendered, so `GetSelfX1()` and `GetSelfX2()` were returning `VRV_UNSET`. Instead, we check if it's a chord by calling `HasSelfBB` or reverting to ContentBB if it is a chord.

Proposed fix for #649 

Before:
image

After:
image


craigsapp:
The original first edition did not use brackets for the tuplets in this case 😉 

screen shot 2018-01-20 at 10 00 21 pm

CCInc:
Ah, well that would make even more sense :) I'll have to update my source file

PR #755, opened 2018-02-09
Adapt qt-binding to recent changes to option-handling and DeviceContext interface

yeonoson:


include/vrv/toolkit.h
```diff
@@ -18,9 +18,12 @@
 //----------------------------------------------------------------------------
 
 #include "checked.h"
-#include "jsonxx.h"
```
lpugin: Why changing this? Are you sure the CLI toolkit still works?

yeonoson: Yes, the CLI toolkit still works. 
Basically I just moved `#include "jsonxx.h"` from toolkit.h to toolkit.cpp. This is possible because jsonxx is only used in the header if USE_EMSCRIPTEN is defined.

This allows other projects that depend on vrv::Toolkit (such as vrvQt::Toolkit) to use it without adding "include/json" as an include directory.

lpugin: Moving the include to the implementation file is OK, but it needs to be included in all cases because it is now also used outside the emscripten build.

yeonoson: It is always included in the implementation file (toolkit.cpp, line 23) and has a conditional include in the header file (toolkit.h, lines 23-25).

lpugin: It makes sense

PR #758, opened 2018-02-14
Adding basic r/w for dots within notes and rests

rettinghaus:
This PR adds simple support for `<dot>` in `<note>` and `<rest>`. 

Single dots on notes (even within chords) are already working, rests need to be addressed. 
 
dots

```xml
<mei xmlns="http://www.music-encoding.org/ns/mei" xmlns:xlink="http://www.w3.org/1999/xlink" meiversion="3.0.0">
    <meiHead>
      <fileDesc>
        <titleStmt>
          <title>Dots</title>
        </titleStmt>
        <pubStmt></pubStmt>
      </fileDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef meter.count="3" meter.unit="4">
                        <staffGrp n="1" >
                            <staffDef n="1" lines="5" clef.line="2" clef.shape="G"/>
                        </staffGrp>
                    </scoreDef>
                    <section>
                        <measure n="1">
                            <staff n="1">
                                <layer n="1">
                                    <rest dur="4"><dot/></rest>
                                    <note pname="g" oct="4" dur="4"><dot/></note>
                                    <note pname="g" oct="4" dur="8"><sic><dot/></sic></note>
                                </layer>
                            </staff>
                        </measure>
                        <measure n="2">
                            <staff n="1">
                                <layer n="1">
                                  <chord dur="4">
                                    <note pname="g" oct="4" dur="4"><sic><dot/></sic></note>
                                    <note pname="d" oct="4" dur="4"></note>
                                  </chord>
                                </layer>
                            </staff>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -31,7 +31,9 @@ Dot::Dot() : LayerElement("dot-"), PositionInterface(), AttColor(), AttDotLog()
     Reset();
 }
 
-Dot::~Dot() {}
+Dot::~Dot()
+{
+}
 
 void Dot::Reset()
 {
@@ -51,7 +53,6 @@ int Dot::PreparePointersByLayer(FunctorParams *functorParams)
 {
     PreparePointersByLayerParams *params = dynamic_cast<PreparePointersByLayerParams *>(functorParams);
     assert(params);
-
     m_drawingNote = params->m_currentNote;
 
     return FUNCTOR_CONTINUE;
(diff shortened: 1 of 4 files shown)


lpugin:
Thanks, but this is only the reading / building tree side. It is kind of working because it uses rendering of dots designed for mensural notation. Supporting this will open all sorts of additional corner cases. As I already pointed out, if we want to go down this road we do it once for all and convert all `@dots` to `<dot>`. I made a proposal for introducing `@dots.ges`, so let's discuss it first.

PR #759, opened 2018-02-14
throw warning for children of rest

rettinghaus:
at least one should be notified, that this isn't supported :-)


lpugin:
Indeed!

PR #760, opened 2018-02-16
Added Midi Pedal Sustain Output

CCInc:
Fairly self explanatory - added a `GenerateMIDI` functor to `pedal.cpp`, which adds a midi control change event in the case of a pedal "up" or "down" event.

Also, added some control change functions to the MidiFile library (@craigsapp, would any of these functions be useful to PR into your main repo?)

### DIFF
--- a/src/pedal.cpp
+++ b/src/pedal.cpp
@@ -62,6 +62,7 @@ int Pedal::GenerateMIDI(FunctorParams * functorParams)
     if (!HasDir()) return FUNCTOR_CONTINUE;
 
     int velocity;
+    int tickOffset = 0;
     switch (GetDir())
     {
     case pedalLog_DIR_down:
@@ -69,6 +70,11 @@ int Pedal::GenerateMIDI(FunctorParams * functorParams)
         break;
     case pedalLog_DIR_up:
         velocity = 0;
+
+        // if you have a pedal event that ends right as another begins (ex. one in the end of measure 1 and start of measure 2), 
+        // they will overlap because they occur at the same midi time
+        // so we just add a -1 offset to the ending time to remedy this until a better method is thought of
+        tickOffset = -1;
         break;
     default:
         return FUNCTOR_CONTINUE;
@@ -77,16 +83,10 @@ int Pedal::GenerateMIDI(FunctorParams * functorParams)
     // todo: check pedal @func to switch between sustain/soften/damper pedals?
 
     double pedalTime = GetStart()->GetAlignment()->GetTime() * DURATION_4 / DUR_MAX;
-
-    // Note: even though this time may be technically correct, 
-    // if you have a pedal event exactly at the start & end of a measure, 
-    // they will overlap because they occur at the same midi time
-    // 
-    // Maybe should add an offset to fix?
     double starttime = params->m_totalTime + pedalTime;
     int tpq = params->m_midiFile->getTPQ();
 
-    params->m_midiFile->addSustainPedal(params->m_midiTrack, starttime * tpq, params->m_midiChannel, velocity);
+    params->m_midiFile->addSustainPedal(params->m_midiTrack, (starttime * tpq) + tickOffset, params->m_midiChannel, velocity);
 
     return FUNCTOR_SIBLINGS;
 }
--- a/src/pedal.cpp
+++ b/src/pedal.cpp
@@ -16,6 +16,9 @@
 #include "functorparams.h"
 #include "verticalaligner.h"
 #include "vrv.h"
+#include "MidiFile.h"
+#include "layerelement.h"
+#include "horizontalaligner.h"
 
 namespace vrv {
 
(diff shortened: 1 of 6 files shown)


craigsapp:
> Also, added some control change functions to the MidiFile library (@craigsapp, would any of these functions be useful to PR into your main repo?)

Yes that would be great.   In general you should do PRs for the midifile library at https://github.com/craigsapp/midifile, and then make me copy it (or do a PR to copy it) into verovio.  Otherwise, I will occasionally update the midifile code in verovio, so any changes done directly in verovio would tend to get lost.

In this case, Laurent can do the PR into verovio if he lies, and then I can back port it into the stand-alone midifile repository (but the indication of your authorship for the code would get lost :-).  Or you can do the same PR to the midifile library, which would give the same results.

The midifile codes looks great.  The only change I can think of is to change the `velocity` variable name in the `addSustainPedal` function to something like `position` instead.  A possible additional pair of convenience functions could be `addSustainPedalOn` and `addSustainPedalOff` which would set positions to 127 and 0 respectively using `addSustainPedal`.

lpugin:
Please make a PR for the MidiFile files into @craigsapp's repo first, and the we can update the files with the latest version.

src/pedal.cpp
```diff
@@ -46,4 +50,41 @@ void Pedal::Reset()
...
+int Pedal::GenerateMIDI(FunctorParams * functorParams)
+{
+    GenerateMIDIParams *params = dynamic_cast<GenerateMIDIParams *>(functorParams);
+    assert(params);
+
+    if (!HasDir()) return FUNCTOR_CONTINUE;
+
+    int velocity;
+    int tickOffset = 0;
+    switch (GetDir())
```
lpugin: See the coding guidelines and the clang-format use for {} layout (https://github.com/rism-ch/verovio/blob/master/doc/guidelines.md)

src/pedal.cpp
```diff
@@ -14,6 +14,10 @@
 //----------------------------------------------------------------------------
 
 #include "verticalaligner.h"
+#include "functorparams.h"
+#include "MidiFile.h"
```
lpugin: Includes from external libraries are listed separately (see https://github.com/rism-ch/verovio/blob/master/doc/guidelines.md#includes-and-forward-declarations)

src/pedal.cpp
```diff
@@ -46,4 +50,41 @@ void Pedal::Reset()
 // Pedal functor methods
 //----------------------------------------------------------------------------
 
+int Pedal::GenerateMIDI(FunctorParams * functorParams)
+{
+    GenerateMIDIParams *params = dynamic_cast<GenerateMIDIParams *>(functorParams);
+    assert(params);
+
+    if (!HasDir()) return FUNCTOR_CONTINUE;
```
lpugin: We should return `FUNCTOR_SIBLINGS` or `FUNCTOR_CONTINUE` consistently even if will  make no difference

lpugin:
Any progress on this?

CCInc:
I've been crazy busy lately, I will try to make these changes this weekend
:) hasn't been forgotten!

craigsapp:
I have added sustain on/off convenience functions to the `MidiFile` class:

https://github.com/craigsapp/midifile/commit/47ee5b1ba0bd34eca560ef7fbff871ba19277a23

src/pedal.cpp
```diff
@@ -58,13 +58,19 @@ int Pedal::GenerateMIDI(FunctorParams * functorParams)
     if (!HasDir()) return FUNCTOR_CONTINUE;
 
     int velocity;
+    int tickOffset = 0;
     switch (GetDir())
     {
     case pedalLog_DIR_down:
         velocity = 126;
```
craigsapp: Why 126 instead of 127?  In any case any value in the range from 64 to 127 will work as sustain pedal on.

src/pedal.cpp
```diff
@@ -58,13 +58,19 @@ int Pedal::GenerateMIDI(FunctorParams * functorParams)
...
         break;
     case pedalLog_DIR_up:
         velocity = 0;
+
+        // if you have a pedal event that ends right as another begins (ex. one in the end of measure 1 and start of measure 2), 
+        // they will overlap because they occur at the same midi time
+        // so we just add a -1 offset to the ending time to remedy this until a better method is thought of
```
craigsapp: I think that I have fixed the problem of ordering pedal up/down messages in the midifile library.  The `MidiFile::sortTracks()` function should now sort pedal on/offs occurring at the same time so that the pedal off is placed before the pedal on: 

https://github.com/craigsapp/midifile/blob/bb1e75842625ca75010407753758e48c12b02a78/src-library/MidiEventList.cpp#L584-L599

An alternate method would be to use the `MidiEvent::seq` variable to force a specific ordering of the pedal messages.  A lower `MidiEvent::seq` value would force an event to be sorted before a higher valued `seq`:

https://github.com/craigsapp/midifile/blob/9d7e042be14eab9a70ef81c2b5267af88053b268/src-library/MidiEventList.cpp#L548-L555

But your solution should also work well.

CCInc:
@lpugin I isolated the changes specific to Verovio, if the formatting is fixed correctly it should be good to go!

PR #766, opened 2018-02-20
preserve id on mdiv

rettinghaus:
forgotten or intentional?


rettinghaus:
Should work with score, too.

lpugin:
Forgotten. Could you fix it for Score and Pages too?

rettinghaus:
Added.
NB: imported `xml:id` from `score` gets lost and is replaced with created id from `pages`.

lpugin:
You can fix this in Pages::ConvertFrom(Score) using Object::SwapUuid

rettinghaus:
Ah yes. Wasn't aware of that function.

PR #767, opened 2018-02-22
make att.labelled available for textelements

rettinghaus:
This PR makes `@label` available for all text elements and adds corresponding labels to the parts of `pgHead`. Additionally it adds an `@type="autogenerated"` to `pgHead` so it can be identified in SVG more easily.


src/doc.cpp
```diff
@@ -166,6 +166,7 @@ bool Doc::GenerateHeaderAndFooter()
     // We mark it as generated for not having it written in the output
     pgHead->IsGenerated(true);
     pgHead->GenerateFromMEIHeader(m_header);
+    pgHead->SetType("autogenerated");
```
lpugin: Shall we have this in the footer too?

src/num.cpp
```diff
@@ -23,7 +23,7 @@ namespace vrv {
 // Num
 //----------------------------------------------------------------------------
 
-Num::Num() : TextElement("num-"), AttLabelled()
```
lpugin: Removing label on num will break auto pagination. Why not suggesting to add this to MEI?

lpugin: Sorry, I did not realized it is there through TextElement.

src/doc.cpp
rettinghaus: makes sense

PR #768, opened 2018-02-22
set lang on generated pgHead

rettinghaus:
if it's there, why not use it?

### DIFF
--- a/src/pghead.cpp
+++ b/src/pghead.cpp
@@ -62,6 +62,7 @@ bool PgHead::GenerateFromMEIHeader(pugi::xml_document &header)
             }
             Text *text = new Text();
             text->SetText(UTF8to16(titleNode.node().text().as_string()));
+            rend->SetLang(titleNode.node().attribute("xml:lang").as_string());
             rend->AddChild(text);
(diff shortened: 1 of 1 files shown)


src/pghead.cpp
```diff
@@ -62,6 +62,7 @@ bool PgHead::GenerateFromMEIHeader(pugi::xml_document &header)
             }
             Text *text = new Text();
             text->SetText(UTF8to16(titleNode.node().text().as_string()));
+            rend->SetLang(titleNode.node().attribute("xml:lang").as_string());
```
lpugin: You need to check that `@xml:lang` is there, don't you?

rettinghaus: Actually it works fine this way. But if it'd make you feel better, I can addy it. 8-)

lpugin: I am surprised it is not crashing if the attribute is not there.

lpugin: You are right, for the doc of pugixml: "In case the attribute handle is null, both functions [as_string() and value()] return empty strings - they never return null pointers."

Issue #318, opened 2016-09-29
collision measure numbers and harm indications


sannedubois:
If chords are like Cmaj7b5/D# and they are above the first note of the staff there can occur a collision of the chordsymbol and the measurenumber. I looked in finale 2012 how they handle it, but it's customizable. Best solution seems to be the vertical distance?

lpugin:
Closing because not documented enough

Issue #177, opened 2016-03-09
Configuration of visibility of bar numbers.


nikobeer:
In some circumstances it would be nice to have the possibility to configure whether bar numbers are rendered or not (e.g. „no bar numbers“, „only on system/page breaks“, „every nth bar“, „every bar number“ …).

rettinghaus:
At first we should react to the `@mnum.visible` on `<scoreDef>` to easily switch off numbering.

rettinghaus:
Now, that measure numbers are sub elements of measure, we could add them to every measure in SVG output by default, making them invisible. Then making them visible would be very easy with CSS.

lpugin:
I would suggest to revised `<scoreDef>` attributes first and then we can implement this accordingly

PR #769, opened 2018-02-22
Develop mnum

rettinghaus:
This PR implements `mNum` element and control functionality with `@mnum.visible`. 
* standard behavior: measure number is printed an system start (taken from measure@n)
* `mNum` elements are drawn (instead of generated measure numbers)
* measure numbers can be switched of with `mnum.visible="false"`


src/view_page.cpp
```diff
@@ -730,12 +731,24 @@ void View::DrawMeasure(DeviceContext *dc, Measure *measure, System *system)
...
-        // Draw measure number if > 1
-        if ((measure->HasN()) && (measure->GetN() != "0") && (measure->GetN() != "1")) {
-            DrawMNum(dc, measure);
+    if (m_drawingScoreDef.GetMnumVisible() != BOOLEAN_false) {
+        // Check if the first measure of the system
+        Measure *systemStart = dynamic_cast<Measure *>(system->FindChildByType(MEASURE));
+        MNum *mnum = dynamic_cast<MNum *>(measure->FindChildByType(MNUM));
+        if (measure == systemStart) {
+            // Draw auto measure number if > 1
+            if ((measure->HasN()) && (measure->GetN() != "0") && (measure->GetN() != "1")) {
```
lpugin: I would move this to a method in MEIInput following the model of `MeiInput::UpgradeStaffDefTo_4_0_0`. Otherwise the MEI output will have some `<mNum>` but not others. You can remove `"autogenerated"`.

lpugin:
Looks excellent, thanks

src/view_page.cpp
rettinghaus: I don't understand this. It generates mNum for the SVG, but it will not be added to measure, so "autogenerated" mNum will not appear in MEI output. Only mNum elements from the input file will be written to MEI output.

rettinghaus: Example: 
input
```xml
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title />
      </titleStmt>
      <pubStmt />
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef meter.count="4" meter.unit="4" mnum.visible="true">
            <staffGrp>
              <staffDef clef.shape="G" clef.line="2" n="1" lines="5" />
            </staffGrp>
          </scoreDef>
          <section>
            <measure n="1">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" />
                </layer>
              </staff>
              <mNum color="blue">1</mNum>
            </measure>
            <measure n="2">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" />
                </layer>
              </staff>
              <mNum place="above">2b</mNum>
            </measure>
            <measure n="3">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" />
                </layer>
              </staff>
              <mNum place="below"><rend halign="left">56</rend></mNum>
            </measure>
            <measure n="4">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" />
                </layer>
              </staff>
            </measure>
            <measure n="123456">
              <staff n="1">
                <layer n="1">
                  <note dur="4" oct="4" pname="e" />
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" />
                </layer>
              </staff>
            </measure>
            <measure n="6">
              <staff n="1">
                <layer n="1">
                  <bTrem>
                    <note xml:id="trem" dur="4" oct="4" pname="e" stem.mod="2slash"/>
                  </bTrem>
                  <note dur="4" oct="4" pname="f" />
                  <note dur="4" oct="4" pname="g" />
                  <note dur="4" oct="4" pname="a" visible="false" />
                </layer>
              </staff>
              <mNum><rend halign="right">greenBar</rend></mNum>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

output
```xml
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
    </meiHead>
    <music>
        <body>
            <mdiv xml:id="mdiv-0000000133842079">
                <score xml:id="score-0000001068443344">
                    <scoreDef xml:id="scoredef-0000000470251826" mnum.visible="true" meter.count="4" meter.unit="4">
                        <staffGrp xml:id="staffgrp-0000000069026394">
                            <staffDef xml:id="staffdef-0000000485434578" clef.shape="G" clef.line="2" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-0000000140450107">
                        <measure xml:id="measure-0000000460420296" n="1">
                            <staff xml:id="staff-0000000530490964" n="1">
                                <layer xml:id="layer-0000001757013251" n="1">
                                    <note xml:id="note-0000000074079660" dur="4" oct="4" pname="e" />
                                    <note xml:id="note-0000001663814007" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000001337448062" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000000778244885" dur="4" oct="4" pname="a" />
                                </layer>
                            </staff>
                            <mNum xml:id="mnum-0000001786371965" color="blue">1</mNum>
                        </measure>
                        <measure xml:id="measure-0000000165488486" n="2">
                            <staff xml:id="staff-0000001604476327" n="1">
                                <layer xml:id="layer-0000000481472510" n="1">
                                    <note xml:id="note-0000000390093674" dur="4" oct="4" pname="e" />
                                    <note xml:id="note-0000000036804627" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000000100075653" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000000491804370" dur="4" oct="4" pname="a" />
                                </layer>
                            </staff>
                            <mNum xml:id="mnum-0000000091489287" place="above">2b</mNum>
                        </measure>
                        <measure xml:id="measure-0000000968032657" n="3">
                            <staff xml:id="staff-0000001358425123" n="1">
                                <layer xml:id="layer-0000001152391004" n="1">
                                    <note xml:id="note-0000000080591935" dur="4" oct="4" pname="e" />
                                    <note xml:id="note-0000001593953935" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000001872772867" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000000025761590" dur="4" oct="4" pname="a" />
                                </layer>
                            </staff>
                            <mNum xml:id="mnum-0000001330830083" place="below">
                                <rend xml:id="rend-0000001219021476" halign="left">56</rend>
                            </mNum>
                        </measure>
                        <measure xml:id="measure-0000001400283488" n="4">
                            <staff xml:id="staff-0000001182737044" n="1">
                                <layer xml:id="layer-0000001152861876" n="1">
                                    <note xml:id="note-0000001552086698" dur="4" oct="4" pname="e" />
                                    <note xml:id="note-0000000437273177" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000000561245805" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000001110067011" dur="4" oct="4" pname="a" />
                                </layer>
                            </staff>
                        </measure>
                        <measure xml:id="measure-0000001705812388" n="123456">
                            <staff xml:id="staff-0000000429429196" n="1">
                                <layer xml:id="layer-0000001871443252" n="1">
                                    <note xml:id="note-0000001301242402" dur="4" oct="4" pname="e" />
                                    <note xml:id="note-0000000007589366" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000000852939189" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000000895605798" dur="4" oct="4" pname="a" />
                                </layer>
                            </staff>
                        </measure>
                        <measure xml:id="measure-0000000733765163" n="6">
                            <staff xml:id="staff-0000002048733510" n="1">
                                <layer xml:id="layer-0000000311306572" n="1">
                                    <bTrem xml:id="btrem-0000000859391512">
                                        <note xml:id="trem" dur="4" oct="4" pname="e" stem.mod="2slash" />
                                    </bTrem>
                                    <note xml:id="note-0000001369002162" dur="4" oct="4" pname="f" />
                                    <note xml:id="note-0000000679542776" dur="4" oct="4" pname="g" />
                                    <note xml:id="note-0000000757401486" dur="4" oct="4" pname="a" visible="false" />
                                </layer>
                            </staff>
                            <mNum xml:id="mnum-0000001511199433">
                                <rend xml:id="rend-0000000439777362" halign="right">greenBar</rend>
                            </mNum>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

fs

lpugin: This is problematic because it breaks two rules. The first is that the rendering process (i.e., the view) should never modify the document tree (i.e., the model), which is happening here. By calling `AddChild`, you are flagging the `Doc` as being modified. It would be normal for an application to rely on this flag to trigger another rendering (refresh) process, and this would create a loop.

The second rule is the the MEI input should also remain the same in the output, unless there is an MEI upgrade (e.g., 3.0 to 4.0), or a modification is explicitly requested. (There is a new option --mensural-to-measure for transforming mensural notation in sections into measure-based MEI.) The preservation of the data structure is even true for `@accid`, that are rendered as `<accid>` but still outputted as `@accid` in the MEI output, or for generated `<pgHead>`s that are not given in the output. With the current code, if you do a `getMEI` from the console, some measures will have a `<mNum>`. 

Upgrading the data to MEI 4.0 with `<mNum>` everywhere will fix the two issues.

lpugin: I can see it is not added to the tree (my bad). The problem then is the memory leak introduced... But most importantly, we will not have the bounding boxes for improving the layout. So it does have to be changed.

lpugin: Sorry to insist, but I do not see the point for not upgrading the data and adding `<mNum>` to the tree. There was already issues added about collisions with measure numbers (see #318), so with the new `<nNum>` element (in the tree) we will eventually have a chance to fix it, but not this way. Why don't you want to add an UpgradeMeasureTo_4_0_0?

rettinghaus: It was intentional to not include bounding boxes for now. 
And I really don't see the point to force mNums to MEI output. It is no mandatory element and Verovio generates them from some information from somewhere else (like pgHead does, too). If one wants to have mNums everywhere a XSLT should be efficient. 

P.S.:Memory leak is fixed.

lpugin: I am not convinced. In text encoding, it is a bad practice to have text content in attributes. This is why `<staffDef label="violin"/>` is now `<staffDef><label>violin</label></staffDef>`. Furthermore, if you do not want them in the MEI, why should it be in SVG? The approach of Verovio is to produce SVG the reflects the MEI as much as possible.

rettinghaus: Verovio generates `mNum` from `measure@n` for the first measure of every system. Do you want to have `<mNum>` added only for those measures? What should happen with `@n` then? Should it be replaced by `<mNum>`? This would lead to measures with `@n` and no `mNum` and others with reversed case. 
mNum should only be used 
> when the `@n` attribute on `<measure>` does not adaquately capture the appearance or placement of the measure number/label.

see https://github.com/music-encoding/music-encoding/commit/d8d94b1ba62e76fe544ca57ee2fb093ec96b620a
I wonder what @pe-ro thinks about that.

lpugin: I would add `<nNum>` to all measures when `@n` is given and drop `@n`.

lpugin: Alternatively, if you really want not to have `<mNum>` in the MEI output and continue to support `@n`, then you can follow the example of `pgHead` and mark it as generated with an additional member. You can do this from `Doc::GenerateHeaderAndFooter` which we could rename to something else such as `Doc::GenerateInternalMarkupDoc`. Then you need an Object::GenerateInternalMarkup` functor to be overridden in `Measure` where the insertion will happen. You also need to to override `MNum::Save` and `MNum::SaveEnd` accordingly (See RunningElement::Save/SaveEnd). Sounds like a lot of work but then we get everything we need. Up to you.

rettinghaus: Another question: will libmei have the meta-attribute class `att.mNum.vis`? I added the nenbers directly.

```xml
  <classSpec ident="att.mNum.vis" module="MEI.cmn" type="atts">
    <desc>Visual domain attributes.</desc>
    <classes>
      <memberOf key="att.color"/>
      <memberOf key="att.placement"/>
      <memberOf key="att.typography"/>
      <memberOf key="att.visualOffset"/>
      <memberOf key="att.xy"/>
    </classes>
  </classSpec>```

lpugin: `<mNum>` have to be part of the tree for proper layout, so doing this in `View::DrawMeasure` is not an option. There is a clear need to be able to show measure numbers not only at the beginning of a system but sometimes also within a system. This is a standard practice for some publishers, and it also makes sense to have this in dynamic layout where systems can end up being very long (see this issue #177). Not having `<mNum>` in the tree will not make it possible to manage the layout. (We probably do not want to see a slur going across a measure number, do we?). So we will indeed have `<mNum>` in all measures internally, and then the drawing algorithm would simply decide to render it or not depending on the options.

rettinghaus:
Done! 
Verovio creates a `mNum` for every `measure@n` if none is present and then prints the measure numbers on the beginning of every system (like before) plus the `mNum` information from the source file.

lpugin:
Great!

PR #779, opened 2018-02-28
badges

rettinghaus:
* bringing build badges to correct branch
* adding license badge

Having build information for the develop branch on the master branch is a bit misleading. Perhaps master should be added to CI, too?


lpugin:
I agree it is misleading, but since master is for the releases there is not much activity and less of a need to have CI. But we can very well have two sets of badges? Is this what you mean? I quite like seeing the develop status on the default page. Another option would be to change the default branch to develop, but I am not sure this is a good idea.

rettinghaus:
I would argue, that people will prefer to download the releases from GitHub instead of clone and build the branch. So having the develop branch as default maybe is a good idea after all?

lpugin:
Other opinions on this?

lpugin:
OK, the default branch is now develop

PR #784, opened 2018-03-02
use prefix operator for iteration

rettinghaus:


craigsapp:
Why?

lpugin:
I think it is more efficient because there is not copy to be made. (I doubt it makes a noticeable difference, but there is no reason not to change this.)

craigsapp:
OK.  If `++` is being applied to a built-in type (integer), then it will not matter since the code will be optimized by the compiler (such as by using `-O3` option).  If you define `++` for your own class, then there will be a difference that probably cannot be optimized out by the compiler.  

If you want to be efficient, then don't use XML 😛

rettinghaus:
Indeed it is faster for non-primitive types. It doesn't make much sense for built-in types, but I switched it there, too, for consistent coding style.

craigsapp:
Are you going to try to make me change it in the `iohumdrum.cpp` file?   At least I won't complain if you do it :-)

PR #785, opened 2018-03-03
prefix operator for iteration

rettinghaus:
@craigsapp asked for it:
that should address all iterations.


craigsapp:
Looks good :-) . We will see if I can remember to do them that way.

PR #790, opened 2018-03-06
minor code improvements

rettinghaus:


src/harm.cpp
```diff
@@ -64,9 +64,6 @@ void Harm::AddChild(Object *child)
     else if (child->IsEditorialElement()) {
         assert(dynamic_cast<EditorialElement *>(child));
     }
-    else if (child->Is(FB)) {
```
lpugin: This does not look right

rettinghaus: Yes, it does! Look above, this statement was doubled.

lpugin: Ah, yes - I introduced it when I fixed the bug on figured bass https://github.com/rism-ch/verovio/commit/b1e0c3ce3221f3231f3ee8ed88e991c7652151ec#diff-0d50c4abb090d2bafeccd263bc7e4544

PR #828, opened 2018-04-19
PAE ties as elements

### LINKED ISSUE #820: Plaine & Easie: ties not working
PAE-ties (`+`) are only converted to attributes and thus not shown. They should be converted to `tie` elements instead.

rettinghaus:
fixes #820

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -486,7 +486,10 @@ void PaeInput::parsePlainAndEasy(std::istream &infile)
         m_doc->m_scoreDef.SetProportNumbase(scoreDefMensur->GetNumbase());
         delete scoreDefMensur;
     }
-    if (m_tie != NULL) delete m_tie;
+    if (m_tie != NULL) {
+      delete m_tie;
+      m_tie = NULL;
+    }
     staffGrp->AddChild(staffDef);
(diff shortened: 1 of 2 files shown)


src/iopae.cpp
```diff
@@ -1317,17 +1319,15 @@ void PaeInput::parseNote(pae::Note *note)
...
+            m_tie = NULL;
         }
 
         if (note->tie) {
-            if (mnote->GetTie() == TIE_t)
-                mnote->SetTie(TIE_m);
-            else
-                mnote->SetTie(TIE_i);
-            m_last_tied_note = mnote;
+            m_tie = new Tie();
```
lpugin: You need to check in the PaeInput destructor if `m_tie == NULL` and delete it if not, otherwise there will be a memory leak.

src/iopae.cpp
```diff
@@ -483,6 +486,7 @@ void PaeInput::parsePlainAndEasy(std::istream &infile)
         m_doc->m_scoreDef.SetProportNumbase(scoreDefMensur->GetNumbase());
         delete scoreDefMensur;
     }
+    if (m_tie != NULL) delete m_tie;
```
lpugin: Please set m_tie to NULL

PR #862, opened 2018-05-14
minor improvements to code style

rettinghaus:


src/view_beam.cpp
```diff
@@ -336,7 +329,7 @@ void View::DrawFTrem(DeviceContext *dc, LayerElement *element, Layer *layer, Sta
     int dur = (dynamic_cast<AttDurationLogical *>(firstElement->m_element))->GetDur();
 
     if (dur > DUR_1) {
-        for (i = 0; i < elementCount; ++i) {
+        for (int i = 0; i < elementCount; ++i) {
```
lpugin: This used to be not supported by some compilers. I hope it is not the case anymore

PR #866, opened 2018-05-15
Neume Functionality

ATranimal:
This pull request contains all the code done at DDMAL in regards to implementing Neume compatibility in Verovio with respect to the upcoming MEI 4.0.0 Schema as provided by Ichiro Fujinaga. The scope of the pull request includes:

- Inclusion of relevant font glyphs
- New CPP and Header files for relevant elements (NC, Neume, Syllable)
- Implementation of Custos for neume mei
- Pitch Interface helper functions
- Drawing functions contained within view_neume.cpp
- Modifications to loading of MEI within iomei.cpp and view_element
- Addition of comparators to interfaces
- Preliminary work on extending editor functionality in the toolkit


include/vrv/nc.h
```diff
@@ -0,0 +1,66 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        note.h
```
lpugin: change to nc.h

include/vrv/neume.h
```diff
@@ -0,0 +1,90 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        note.h
```
lpugin: change to neume.h

include/vrv/att.h
```diff
@@ -266,6 +266,25 @@ class AttComparison {
     ClassId m_classId;
 };
 
+//----------------------------------------------------------------------------
+// InterfaceComparison
+//----------------------------------------------------------------------------
+
+class InterfaceComparison {
```
lpugin: Make the class inherit from AttComparison (we should actually rename AttComparison just Comparison because it can be use for any type of matching. I will do that once the PR is merged.)

include/vrv/object.h
```diff
@@ -379,6 +379,13 @@ class Object : public BoundingBox {
     void FindAllChildByAttComparison(ArrayOfObjects *objects, AttComparison *attComparison,
         int deepness = UNLIMITED_DEPTH, bool direction = FORWARD, bool clear = true);
 
+    /**
```
lpugin: Same here: with FindAllChildByAttComparison will work with a class inheriting from AttComparison

src/iomei.cpp
```diff
@@ -507,6 +518,14 @@ bool MeiOutput::WriteObject(Object *object)
         m_currentNode = m_currentNode.append_child("syl");
         WriteSyl(m_currentNode, dynamic_cast<Syl *>(object));
     }
+    else if (object->Is(SYLLABLE)) {
```
lpugin: Remove duplicate

src/iomei.cpp
```diff
@@ -1453,6 +1473,28 @@ void MeiOutput::WriteMultiRpt(pugi::xml_node currentNode, MultiRpt *multiRpt)
...
+    WriteLayerElement(currentNode, nc);
+    nc->WriteColor(currentNode);;
+    nc->WriteIntervalMelodic(currentNode);;
+    
+}
+
+void MeiOutput::WriteNeume(pugi::xml_node currentNode, Neume *neume)
+{
+    assert(neume);
+    neume->GenerateChildMelodic();
```
lpugin: I am quite skeptical with this call here. Writing should not change the data, there is a flaw in the design here.

src/neume.cpp
```diff
@@ -0,0 +1,148 @@
...
+        LogError("Adding '%s' to a '%s'", child->GetClassName().c_str(), this->GetClassName().c_str());
+        assert(false);
+    }
+
+    child->SetParent(this);
+    m_children.push_back(child);
+    Modify();
+}
+
+void Neume::Reset()
```
lpugin: Place Neume::Reset() before Neume::AddChild

src/view_neume.cpp
```diff
@@ -0,0 +1,279 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        view_syllable.cpp
```
lpugin: change to view_neume.cpp

include/vrv/object.h
JRegimbal: Do you mean that FindAllChildByInterfaceComparison will work?

include/vrv/att.h
JRegimbal: ClassId and InterfaceId are separate enums with separate checking methods, though. As it is having InterfaceComparison inherit from AttComparison would just add functionality for searching by class to InterfaceComparison. Is this desirable?

lpugin: You are right, I missed the ClassId. We whould have the following hierarchy:
```
class Comparison
class AttComparison : public Comparison
class InterfactComparison : public Comparison
```
And the the method / functor / functorparams acting on Comparison. Then we need only one set of method / functor / functorparams. OK?

include/vrv/object.h
lpugin: The method should be:
```
void FindAllChildByComparison(ArrayOfObjects *objects, Comparison *comparison,
         int deepness = UNLIMITED_DEPTH, bool direction = FORWARD, bool clear = true);
```
And then we can call it with either an `AttComparison` or and `InterfaceComparison`.

include/vrv/att.h
JRegimbal: Ok. I'll make a commit implementing this.

src/view_neume.cpp
```diff
@@ -111,6 +111,9 @@ void View::DrawNeume(DeviceContext *dc, LayerElement *element, Layer *layer, Sta
     
     std::map<std::string, NeumeGroup> *neumeGroups = Neume::GetGroups();
     
+    // Generate intm attribute for all nc children except the first
+    // This is necessary to properly render the neume
+    neume->GenerateChildMelodic();
```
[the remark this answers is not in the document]
ATranimal: If referring to the drawing of neume groups:

It's out of the norm in verovio to not search through the tree to draw each element individually, but I think for Neumes specifically this is the most efficient way to do it. If we were to handle drawing of neumes through each nc component, then we would have to search through the siblings and determine which position in the neume each component is and it would be more messy in my opinion. We can discuss this further, perhaps we can find a more elegant solution.

Issue #873, opened 2018-05-25
Beam-Rest Vertical Collision


CCInc:
Didn't see an issue for this already.

When rests are contained within beams, the beam tail height does not account for the size of the rest, causing difficulty reading.

![image](https://user-images.githubusercontent.com/2824685/40519007-f62809d4-5f82-11e8-9a8d-6ac8e7ade4e6.png)

```
<music xml:id="m-9">
    <body xml:id="m-10">
        <mdiv xml:id="m-11">
            <score xml:id="m-12">
                <scoreDef xml:id="m-13" lyric.name="Plantin MT Std" meter.count="4" meter.unit="4" music.name="Opus Std" page.botmar="15mm" page.height="279.4mm" page.leftmar="15mm" page.rightmar="15mm" page.topmar="15mm" page.width="215.9mm" ppq="256" text.name="Plantin MT Std">
                    <staffGrp xml:id="m-14">
                        <staffDef xml:id="m-15" clef.line="4" clef.shape="F" key.mode="major" key.sig="0" lines="5" n="1">
                            <!-- keyboard.piano.grand.bright -->
                            <instrDef xml:id="m-17" midi.channel="1" midi.instrname="Unnamed (bass staff)" midi.pan="63" midi.volume="100"/>
                        </staffDef>
                    </staffGrp>
                </scoreDef>
                <section xml:id="m-18">
                    <sb xml:id="m-20"/>
                    <measure xml:id="m-19" label="1" n="1">
                        <staff xml:id="m-21" n="1">
                            <layer xml:id="m-22" n="1">
                                <beam xml:id="m-24">
                                    <note xml:id="m-23" dots="1" dur="16" dur.ges="96p" oct="2" pname="a" pnum="45" stem.dir="up"/>
                                    <rest xml:id="m-25" dur="32" dur.ges="32p"/>
                                    <note xml:id="m-26" dur="16" dur.ges="64p" oct="2" pname="f" pnum="41" stem.dir="up"/>
                                    <note xml:id="m-27" dur="16" dur.ges="64p" oct="1" pname="b" pnum="35" stem.dir="up"/>
                                </beam>
                                <beam xml:id="m-29">
                                    <note xml:id="m-28" dots="1" dur="16" dur.ges="96p" oct="2" pname="a" pnum="45" stem.dir="up"/>
                                    <rest xml:id="m-30" dur="32" dur.ges="32p"/>
[...]

PR #874, opened 2018-05-25
Automatic Rest Location Detection within Beams

CCInc:
Related to #873 

When rests are within a beam element and sandwiched between two notes, this commit will automatically determine where to place the rests based on the average of the sandwiching note's locations.

Not sure if this is too much guesswork to be consistent with your project's goals - maybe add along as an option if you'd like?

Before:
image

After:
image

### DIFF
--- a/src/layerelement.cpp
+++ b/src/layerelement.cpp
@@ -954,6 +954,14 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
                 if (abs(locAvg - loc) > 3) {
                     loc = locAvg;
                 }
+                // if loc is odd, we need to offset it to be even 
+                // so that the rest dots do not collide with the staff lines
+                if (loc % 2 != 0) {
+                    // if it's above the staff, offset downwards
+                    // if below the staff, offset upwards
+                    if (loc > 4) loc--;
+                    else loc++;
+                }
             }
             else if (hasMultipleLayer) {
                 Layer *firstLayer = dynamic_cast<Layer *>(staffY->FindChildByType(LAYER));
--- a/src/layerelement.cpp
+++ b/src/layerelement.cpp
@@ -842,7 +842,7 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
             }
 
             // add offset
-            else if (staff->m_drawingLines > 1)
+            else if (staff->m_drawingLines > 3)
                 loc += 2;
         }
 
@@ -883,6 +883,8 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
                 ListOfObjects::const_iterator it = beamList->begin();
                 std::advance(it, restIndex);
                 ListOfObjects::const_reverse_iterator rit(it);
+                // iterate through the elements from the rest to the beginning of the beam
+                // until we hit a note or chord, which we will use to determine where the rest should be placed
                 for (; rit != beamList->rend(); ++rit) {
                     if ((*rit)->Is(NOTE)) {
                         Note *leftNote = dynamic_cast<Note *>(*rit);
@@ -910,6 +912,7 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
                                                                      layerY->GetClefLocOffset(layerElementY));
                         }
 
+                        // if it's a rest, use the middle of the chord as the rest's location
                         leftLoc = (topChordLoc + bottomChordLoc) / 2;
                         break;
                     }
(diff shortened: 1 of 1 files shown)


src/layerelement.cpp
```diff
@@ -866,10 +866,96 @@ int LayerElement::SetAlignmentPitchPos(FunctorParams *functorParams)
...
+                        Note *bottomChordNote = rightChord->GetBottomNote();
+                        if (bottomChordNote->HasPname()) {
+                            bottomChordLoc = PitchInterface::CalcLoc(bottomChordNote->GetPname(),
+                                                                     bottomChordNote->GetOct(),
+                                                                     layerY->GetClefLocOffset(layerElementY));
+                        }
+
+                        rightLoc = (topChordLoc + bottomChordLoc) / 2;
+                        break;
+                    }
```
CCInc: Is there a good way to reduce code replication between L887-L915 and L922-L950?

craigsapp:
That is great! Although some improvement can be made on placement of the rests, marked in red:

screen shot 2018-05-24 at 10 12 07 pm

The dots of the rests should never be placed on staff lines.  If the rests are above/below the staff lines, then the positions do not have to be quantized to spaces, but when inside the staff, they do have to be quantized to avoid the staff lines and align with the spaces.

CCInc:
Good point, how's this look? 12f69bd

image

craigsapp:
That looks great, the only small complaint is this beam/rest group:

screen shot 2018-05-26 at 12 11 58 am

The rest should go one space lower.  See the following figure:

screen shot 2018-05-26 at 12 37 15 am

where the second one has the best visual quality.

And whenever the stem lengths are optimally set, the second one will also still be the best (and the third option would not be possible):

screen shot 2018-05-26 at 12 39 04 am

-----------

The 2nd and 3rd beats of the fourth measure:

screen shot 2018-05-26 at 12 42 38 am

are acceptable, but probably would look slightly better one more space towards the center of the staff.  But I would not complain if they were left as is.

CCInc:
Here is the test case without the changes compared to the test case with the changes, with those changes highlighted:

(before)
image

(after)
image

Let me know if you'd like any additional changes

craigsapp:
Looks good to me, and infinitely better than the current placements in verovio, although I will let @lpugin check and merge the code.   Below are some very minor points.

Here are the ones highlighted in the  adjusted output that look perfect:

screen shot 2018-05-28 at 6 42 30 pm

screen shot 2018-05-28 at 6 43 30 pm

screen shot 2018-05-28 at 6 44 56 pm

screen shot 2018-05-28 at 6 45 32 pm

I was thinking that placing the rest-head dots on ledger-line positions would be acceptable, but they seem somewhat funny to me.  At a minimum, I would say that the rest-head dots should not be placed on the first ledger line away from the staff:

screen shot 2018-05-28 at 6 51 48 pm

The above example looks funny to me because of the distance between the rest-head dot and the first staff line, so it would be preferable that the rests are placed one diatonic step lower.  This case also involves more than one rest as well as being above the staff.  In such cases the rests should probably be aligned more to the note closer to the staff.  This is why I am saying one diatonic step down rather than one diatonic step up.  One diatonic step up would be acceptable, but one diatonic step down looks slightly better for this particular case.

For this example:

screen shot 2018-05-28 at 6 55 26 pm

The first two rests also do not feel quite right, and should be one diatonic step higher.  Moving down one diatonic step would not be appropriate in this case (unlike the previous case).  One reason is that the following two rests are better aligned to these two rests if the first two are moved up a diatonic step.

 I like the second pair of rests, however, eventhough the sixteenth rest is higher in relation to the 32nd rest when they are in their default positions in the middle of the staff.  Probably an ideal typesetting would align all four rests in this example instead of aligning them in two groups.

The rule for avoiding staff lines may also extend to ledger line positions, as the two cases in the example are more distracting if this rule is not followed off of the staff as well as on it.

There is one notational violation:

screen shot 2018-05-28 at 6 59 29 pm

The first rest touches the beam.  However, this is not a problem with rest placement, but rather a problem in stem length calculations, which should be occurring after placement of the rests (so that one is not your fault :-)

CCInc:
@craigsapp  I'd agree with that, here's what it looks like without any ledgers:

image

What do you think? If we need I can make more changes and open another PR. I think that you're right, it definately looks better for the rests under the staff:

image

The top one does not bother me as much but I'm probably just not seeing it right now, since I've been looking at these rests for the last week when writing this, so I'll defer to your judgement on that :)

craigsapp:
Yes, they definitely look nicer for some reason.

PR #899, opened 2018-06-06
Disable js-specific optimizations. Resolves #897

### LINKED ISSUE #897: Different behaviour in javascript toolkits with and without optimizations
A bug appears in the javascript toolkit when built with the default `-O3` flag that does not appear with optimizations disabled as of commit 693939c. The `</layer>` and `</staff>` tags appear before the last of the layer elements are written. These elements appear after the `</staff>` and are invalid.

This is the script run in node that produced the attached MEIs.
```javascript
var fs = require('fs');
eval(fs.readFileSync('verovio-toolkit.js').toString());
var mei = fs.readFileSync('mei4.mei').toString();

var options = {
    breaks: "none",
    pageHeight: 400,
    pageWidth: 600,
    noFooter: 1,
    noHeader: 1
};

var vrvToolkit = new verovio.toolkit();
vrvToolkit.setOptions(options);
vrvToolkit.loadData(mei);
console.log(vrvToolkit.getMEI(0, true));
```
[no-optimizations.txt](https://github.com/rism-ch/verovio/files/2076805/no-optimizations.txt)
[optimization-level-3.txt](https://github.com/rism-ch/verovio/files/2076806/optimization-level-3.txt)

JRegimbal:
Javascript optimizations in emscripten cause issue #897. This is one way of resolving it while keeping the LLVM optimizations at level 3.

### DIFF
--- a/emscripten/buildToolkit
+++ b/emscripten/buildToolkit
@@ -34,7 +34,7 @@ my $VEROVIO_ROOT  = "..";
 # Memory is increased (TOTAL_STACK) for processing large files (tested up to 7MB)
 # Empirically, the memory amount required is roughly 5 times the file size.
 # This can be disabled in the light version, which uses the default memory settings.
-my $FLAGS = "--llvm-opts 3 --js-opts 0";
+my $FLAGS .= "-O2";
 $FLAGS .= " -DNDEBUG";
 $FLAGS .= " --memory-init-file 0";
 $FLAGS .= " -std=c++11";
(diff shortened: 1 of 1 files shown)


JRegimbal:
I think I can find a better solution.

lpugin:
What is the resulting file size?

JRegimbal:
With the latest changes it's 6.0 MB rather than 16 MB with the first solution I had.

JRegimbal:
Better solution found by @lpugin in commit 63fe0c8.

PR #906, opened 2018-06-18
"Notehead" class added to SVG output

ahwitz:
While working on an editing interface, I realized that there are `g[class='note']` and `g[class='stem']` selectors available, but there didn't seem to be a reliable means of identifying what SVG `use` elements could reliably be considered a notehead, since the `xlink:href` attribute could have a variety of values.

This commit resolves that by adding an optional `className` parameter to `View::DrawSmuflCode` and `devicecontext::DrawMusicText` that is used in `View::DrawNote` to append a `notehead` class to all SVG elements that represent noteheads.

I have left `View::DrawMaximaToBrevis` as it is because I know absolutely nothing about those notation styles and don't want to semantically suggest something that is not correct.

- Added optional className parameter for View::DrawSmuflCode
- Added optional className parameter for devicecontext::DrawMusicText
- Added pugixml append_attribute for SVG `class` attribute
- Added "notehead" class for all notehead elements

- Fixed typo in doc/guidelines.md

(if you want another commit re-adding the line break in view-element that I accidentally deleted, let me know)

### DIFF
--- a/doc/guidelines.md
+++ b/doc/guidelines.md
@@ -26,7 +26,7 @@ The simplest way to fullfil the Verovio coding style is to use a clang-format to
 
 ### Downloading clang-format for OS X
 
-An easy way to install clang-format on OS X computers is to use [Hombrew](http://brew.sh).  Type this command in the terminal to install:
+An easy way to install clang-format on OS X computers is to use [Homebrew](http://brew.sh).  Type this command in the terminal to install:
 
 ```
 brew install clang-format
--- a/include/vrv/bboxdevicecontext.h
+++ b/include/vrv/bboxdevicecontext.h
@@ -81,7 +81,7 @@ public:
     virtual void DrawRoundedRectangle(int x, int y, int width, int height, double radius);
     virtual void DrawText(
         const std::string &text, const std::wstring wtext = L"", int x = VRV_UNSET, int y = VRV_UNSET);
-    virtual void DrawMusicText(const std::wstring &text, int x, int y, bool setSmuflGlyph = false);
+    virtual void DrawMusicText(const std::wstring &text, int x, int y, bool setSmuflGlyph = false, std::string className = "");
     virtual void DrawSpline(int n, Point points[]);
     virtual void DrawSvgShape(int x, int y, int width, int height, pugi::xml_node svg);
     virtual void DrawBackgroundImage(int x = 0, int y = 0){};
--- a/include/vrv/devicecontext.h
+++ b/include/vrv/devicecontext.h
@@ -154,7 +154,7 @@ public:
     virtual void DrawRoundedRectangle(int x, int y, int width, int height, double radius) = 0;
     virtual void DrawText(const std::string &text, const std::wstring wtext = L"", int x = VRV_UNSET, int y = VRV_UNSET)
         = 0;
-    virtual void DrawMusicText(const std::wstring &text, int x, int y, bool setSmuflGlyph = false) = 0;
+    virtual void DrawMusicText(const std::wstring &text, int x, int y, bool setSmuflGlyph = false, std::string className = "") = 0;
     virtual void DrawSpline(int n, Point points[]) = 0;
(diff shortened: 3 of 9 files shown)


lpugin:
I am not against this but could you give me an example where looking at `g[class='note']/use` does not work?

ahwitz:
As of right now, I fully acknowledge there doesn't seem to be anything else and that that XPath should work. The following could happen, however:

- Other elements that make sense to be drawn as `g[class='note']/use` may someday be added to the schema/Verovio.
- There may also be cases where graphics whose positions are calculated like noteheads but are not actually noteheads (like fret markers(I think?) in [lute tablature](https://tm.web.ox.ac.uk/sites/default/files/tm/images/media/italian_tablature.png)) are included as direct SVG descendants of the note object. Even though it may make sense to process their position like noteheads, I don't know if they'd want to be semantically treated as noteheads in the scope of an application using the SVG output.
- In general, having the path in the code to add `className` for other SMUFL codes may be useful in the future for other XPaths where there is ambiguity. In the really simple example I'm using for testing, `g[class='meterSig']` has two direct `use` children, and the only way to determine which one is which is by the `y` attribute. I highly doubt there's a use for that, but still...

lpugin:
Makes sense. Then the proper way to do it is to use the SMUFL ranges from the https://github.com/rism-ch/verovio/blob/develop/fonts/supported.xsl file. 

The file ./fonts/generate-h.xsl should be modified to generate the appropriate code for mapping a smufl code with its class name, and then all `<use>` in the SVG will have the corresponding SMUFL class. OK?

ahwitz:
Completely understood, but two questions:

- What happens if we determine that desired class names don't align with that XSL? For example, a hypothetical "raises-pitch" class would apply to some (but not all elements) from multiple sets.
- I had to change the `/include/vrv/smufl.h` SMUFL_COUNT constant from 267 to 130 in my copy of the code to get Verovio to run, and I think I'm going to run into problems related to that while testing this - is there documentation for what I need to do to fix this?

lpugin:
> What happens if we determine that desired class names don't align with that XSL? For example, a hypothetical "raises-pitch" class would apply to some (but not all elements) from multiple sets.

We will need to think about it when necessary.

> I  had to change the /include/vrv/smufl.h SMUFL_COUNT constant from 267 to 130 in my copy of the code to get Verovio to run, and I think I'm going to run into problems related to that while testing this - is there documentation for what I need to do to fix this?

You quite likely did run `make install` at some point but now use another version using the previous install resources. You should either re-run `make install` or use the `-r` option to set the path of the local version you are using.

PS You might need to add a `smufl.cpp` implementation file if you go for a static `std:map<wchar_t, std::string>` for the mapping.

ahwitz:
> We will need to think about it when necessary.

That's what I like to hear.

> PS You might need to add a smufl.cpp implementation file...

That's less what I like to hear... Can I attach it to the `Resources` class instead, like the font itself?

Roger on the `make install`. I take XCode for granted sometimes...

lpugin:
> That's less what I like to hear... Can I attach it to the Resources

Not really because we want the file to be fully generated by the XSLT. So it has to be its own file. But that is not a problem. We can have a `std::map<std::string, std::vector<wchar_t> >` that maps a classname with all the corresponding codes. You can instanciated it with something like:
```
s_smuflClasses = {
    {"noteheads", {1, 2}}, 
    {"clefs", {3, 4}},
};
```
and then get the class for a `code` with
```
  auto it = std::find_if(s_smuflClasses.begin(), s_smuflClasses.end(), 
   [code](const std::pair<std::string, std::vector<int> > & t) { 
      return (std::find(t.second.begin(), t.second.end(), code) != t.second.end());
   }
  );
  /// the class name should be in it->first (you can check with it != m_smuflClasses.end())
```

ahwitz:
(closing and will reopen on separate branch when I have time because requisite work is rather unlike what's been done here)

PR #929, opened 2018-08-23
adding "curvedir" attribute for "tie" element 

fzalkow:
A minor addition: Just came across the `curvedir` attribute for the `tie` element. Verovio seems to handle this correctly, but it is missing in page of Supported MEI elements

This is the first time for contributing to Verovio. According to the contribution guidelines I should create the pull request to the branch `gh-pages-develop`. Hope, I am doing this right.

Btw, thanks a lot for Verovio! :+1:


lpugin:
Many thanks for the PR. However, a PR should only have as file changes the files that were actually changed. Your PR has a long list quite likely because gh-pages-develop was quite behind. I updated the branch, so you would need to re-submit a PR.

Unfortunately, the list of supported elements / attributes has not been maintained. If you want to do it the file to maintain is https://github.com/rism-ch/verovio/blob/gh-pages/supported-mei.yml. This should be the only file change.

fzalkow:
As requested, I re-submitted my PR (#931).

Issue #944, opened 2018-10-24
Get MIDI pitches next to Timemap entries


sonovice:
I would like to implement a function that returns a pianoroll for a loaded MEI file, similar to the result of `renderToTimemap()`, but also containing the midi pitch:
```json
[
  {
    "tstamp": 125,
    "on": [
      {"d414233e148": 68},
      {"d414233e38": 72}
    ],
    "off": []
  },
  {
    "tstamp": 250,
    "on": [
      {"d414233e260": 77}
    ],
    "off": [
      {"d414233e148": 68}
    ]
  }
]
```
Having a look at the code in `doc.cpp` I found the method `CalculateMidiTimemap()` which seems to be a good base to start with.

Unfortunately, I am having a hard time to understand the underlying data structure. Any hints on where I could start?

sonovice:
Maybe even better: Have a function that returns the MIDI pitch for a given note ID.

sonovice:
I think I've managed to implement it, see sonovice/verovio@6b8c5ca4397864cd836f014daf0e8788d0c33be1

Too bad my linux does not want to compile the JS version using emscripten, so I cannot test it in the browser...

sonovice:
For the sake of completeness, here is an example on how to use the new method to get the exact result that I was looking for:

```js
vrvToolkit.renderToMIDI();
var timemap = JSON.parse(vrvToolkit.renderToTimemap());
for (event of timemap) {
    for (type of ["on", "off"]) {
        if (type in event) {
            var pitches = [];
            for (id of event[type]) {
                var pitch = vrvToolkit.getMIDIPitchForElement(id);
                var dict = {};
                dict[id] = pitch;
                pitches.push(dict);
            }
            event[type] = pitches;
        }
    }
}
console.log(timemap);
```

lpugin:
Thanks for this. I suggest to have the method returning a JSON object with all the MIDI values - for now only pitch, but maybe later also others.

k-ljung:
@sonovice do you have a demo or sample where you implemented a pianoroll view?

sonovice:
@k-ljung No, I don't since I am using the data for something completely different. But it should be quite easy using d3.js - The library is very helpful in creating (interactive) SVGs.

PR #946, opened 2018-10-25
Added "getMIDIValuesForElement" method

sonovice:
Added a method to get the MIDI Pitch for a Note with the given ID after `renderToMIDI()` has been called.
See #944


emscripten/buildToolkit
```diff
@@ -177,6 +177,7 @@ $exports .= "'_vrvToolkit_getOptions',";
 $exports .= "'_vrvToolkit_getPageCount',";
 $exports .= "'_vrvToolkit_getPageWithElement',";
 $exports .= "'_vrvToolkit_getTimeForElement',";
+$exports .= "'_vrvToolkit_getMIDIPitchForElement',";
```
lpugin: Rename it to `getMIDIValuesForElement` and sort it alphabetically.

include/vrv/toolkit.h
```diff
@@ -195,6 +195,13 @@ class Toolkit {
      */
     int GetTimeForElement(const std::string &xmlId);
 
+    /**
+     * Return the MIDI pitch of the element with the ID (xml:id).
+     * RenderToMidi() must be called prior to using this method.
+     * Returns -1 if no element is found.
+     */
+    int GetMIDIPitchForElement(const std::string &xmlId);
```
lpugin: Change it to GetMIDIValuesForElement and return a string / json object with for now only the pitch. We can then use the same method if one day we want to add other MIDI values

lpugin: Sounds good

sonovice: Done, ready to pull.

PR #997, opened 2019-02-21
fixing java build script for linux

rettinghaus:
This PR adds paths to java on linux (tested on Ubuntu 18.10) and should work on all other systems as well without any warning. 
Updated the importer.mei to MEI4.


bindings/java/build.sh
```diff
@@ -37,5 +37,5 @@ FILES="$SRCFILES \
  ../../libmei/atts_visual.cpp"
 
 CXXOPTS="-g -fpic -std=c++11 -I../../include -I../../include/vrv -I../../include/json -I../../include/hum -I../../include/midi -I../../include/pugi -I../../include/utf8 -I../../libmei -I/opt/local/include/ -I/System/Library/Frameworks/JavaVM.framework/Headers/"
-g++ -shared -o target/libverovio.jnilib $CXXOPTS $FILES verovio_wrap.cxx
+g++ -shared -o -I$JAVA_HOME/include -I$JAVA_HOME/include/linux target/libverovio.jnilib $CXXOPTS $FILES verovio_wrap.cxx
```
lpugin: Have you tested this on another platform? Isn't there a more cross-platform way to include the files?

PR #1015, opened 2019-03-17
Added emscripten_main.cpp to CMake list when building as library.

adrq:
Allows library to be used by any language with C function interface.

Signed-off-by: Adrian Quiroga <adrq22@yahoo.com>

### DIFF
--- a/emscripten/c_wrapper.cpp
+++ b/emscripten/c_wrapper.cpp
@@ -20,7 +20,7 @@ extern "C" {
 /** declarations */
 
 void *vrvToolkit_constructor();
-void *vrvToolkit_constructorResroucePath(const char * resourcePath); //alternate constructor which allows user to set custom data path
+void *vrvToolkit_constructorResourcePath(const char * resourcePath); //alternate constructor which allows user to set custom data path
 void vrvToolkit_destructor(Toolkit *tk);
 bool vrvToolkit_edit(Toolkit *tk, const char *editorAction);
 const char *vrvToolkit_getAvailableOptions(Toolkit *tk);
--- a/emscripten/npm-dev/package.json
+++ b/emscripten/npm-dev/package.json
@@ -1,6 +1,6 @@
 {
   "name": "verovio-dev",
-  "version": "2.1.0-12",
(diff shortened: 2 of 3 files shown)


lpugin:
In what context do you need this? This makes building it as a library dependent on emscripten, doesn't it?

adrq:
This does NOT make building as library depend on emscripten. The file `emscripten_main.cpp` is essentially a plain C wrapper around the Toolkit. This is used by the JS toolkit, however, we can also use this when building as a library, to allow the library to be called from languages that cannot call C++ code directly. This commit takes the emscripten C wrapper and adds it to the CMake list. This makes the final compiled library have callable C symbols, which will work with any language that can make a standard C function call (basically any language).

We could create a separate plain C wrapper file, and add that to the CMake list, but it would contain basically the same content as `emscripten_main.cpp`, so I figured it would be easier to just use the wrapper that is already there. This would make it easier to maintain I think?

I'm trying to call use the library is a project involving plain C and also Rust, and it is extremely useful to be able to call the library using a plain C function interface.

lpugin:
You are correct. Then we should take the opportunity to rename it to `c_wrapper.cpp`. Could you do it? It also has to be changed in `./emscripten/buildToolkit.sh`.

emscripten/c_wrapper.cpp
```diff
@@ -20,6 +20,7 @@ extern "C" {
 /** declarations */
 
 void *vrvToolkit_constructor();
+void *vrvToolkit_constructorResroucePath(const char * resourcePath); //alternate constructor which allows user to set custom data path
```
lpugin: can't it be named `vrvToolkit_constructor(const char * resourcePath)` ?

adrq: AFAIK, when compiling for C we can't have overloaded functions because of the name mangling. I tried compiling with the same function name, and both the emscripten and regular builds complain about a duplicate function definition. I don't know if there is a way around this without getting into complex macros and other preprocessing. Also, just committed a fix to the function name typo (fixed spelling Resrouce -> Resource).

PR #976, opened 2019-01-05
Added parsing of breath marks in MusicXML files

sonovice:

PR #1025, opened 2019-03-26
Corrected time offset for breath marks import from musicxml

sonovice:
See #976 , thanks @craigsapp !


sonovice:
Sorry that it took so long, I simply forgot about it over time...

PR #1029, opened 2019-03-31
fix PAE duration

rettinghaus:
fixes duration error from #966

TUPLETS_001
```
@clef:G-2
@keysig:bBEA
@timesig:2/4
@data:8,B/4(8{GB'E};3)({FD,B})/8'G-8.-6,B/4({8'EGB};3)({BBB})/8B-8.-6B/4({8CAC};3)({bCAC})/8,B-8.-6'A/
```

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -233,7 +233,7 @@ void PaeInput::parsePlainAndEasy(std::istream &infile)
         }
 
         // rhythmic values
-        else if (isdigit(incipit[i]) != 0) {
+        else if ((isdigit(incipit[i]) != 0) && (incipit[i-1] != ';')) {
             i += getDurations(incipit, &current_measure, i);
         }
 
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -674,7 +674,7 @@ int PaeInput::getTupletFermata(const char *incipit, pae::Note *note, int index)
 
         // Triplets are in the form (4ABC)
         // index points to the '(', so we look back
-        // if the resut is a number or dot, it means we have the long format
+        // if the result is a number or dot, it means we have the long format
         // i.e. 4(6ABC;5) or 4.(6ABC;5)
         if ((index != 0) && (isdigit(incipit[index - 1]) || incipit[index - 1] == '.')) {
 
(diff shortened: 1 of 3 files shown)


src/iopae.cpp
```diff
@@ -233,7 +233,7 @@ void PaeInput::parsePlainAndEasy(std::istream &infile)
         }
 
         // rhythmic values
-        else if (isdigit(incipit[i]) != 0) {
+        else if ((isdigit(incipit[i]) != 0) && (incipit[i-1] != ';')) {
```
lpugin: Make sure i < 0

rettinghaus: C++ doesn't actually care.

lpugin: I do ;-) Sorry, I meant make sure `i > 0`

src/iopae.cpp
```diff
@@ -233,7 +233,7 @@ void PaeInput::parsePlainAndEasy(std::istream &infile)
         }
 
         // rhythmic values
-        else if ((isdigit(incipit[i]) != 0) && (incipit[i-1] != ';')) {
+        else if ((isdigit(incipit[i]) != 0) && ((i = 0) || incipit[i-1] != ';')) {
```
lpugin: Fix it with
`else if ((isdigit(incipit[i]) != 0) && ((i <= 0) || incipit[i-1] != ';')) {`

src/iopae.cpp
lpugin: Thanks. Just to make it clear: C++ does care.

src/iopae.cpp
rettinghaus: @lpugin I tried to force an index error without any luck, and then I found this answer: https://stackoverflow.com/questions/1239938/accessing-an-array-out-of-bounds-gives-no-error-why

lpugin: Yes, with your code it could happen that you would read outside the range of the array. Depending how it is allocated, this can go through. But then anything can be there...

Everywhere else in Verovio we use `std::vector` and `at()` method, which will raise an error when accessing outside the range. This makes it easier to find a problem.

PR #1036, opened 2019-04-14
set accidental gestural for note based on key

dchisolm:
this seems to be what is needed to make it so midi
output is not all naturals.


dchisolm:
Updated. Thanks for the tips!

PR #1041, opened 2019-04-21
add clef support to ABC import

rettinghaus:


lpugin:
Reviewing this I realized that there are some placed where char array access is not safe. See in `readMusicCode` and similar methods. Could you :
* change this to std::string 
* use `.at` accessor
* always make sure you are not out of bounds

rettinghaus:
Now it should be fine

include/vrv/ioabc.h
```diff
@@ -56,7 +56,7 @@ class AbcInput : public FileInputStream {
     void parseABC(std::istream &infile);
 
     // parsing functions
-    int SetBarLine(const char *musicCode, int index);
+    int SetBarLine(std::string musicCode, int index);
```
lpugin: Please change this to `std::string &musicCode` for optimization

include/vrv/ioabc.h
```diff
@@ -75,7 +75,7 @@ class AbcInput : public FileInputStream {
 
     // input functions
     void readInformationField(char dataKey, std::string dataValue);
-    void readMusicCode(const char *musicCode, Section *section);
+    void readMusicCode(std::string musicCode, Section *section);
```
lpugin: Same here

PR #1044, opened 2019-04-26
Update pugixml

rettinghaus:
This PR updates pugixml to latest version and followed specification changes
See https://pugixml.org/docs/manual.html#v1.9

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -102,7 +102,7 @@ bool MusicXmlInput::ImportString(std::string const &musicxml)
         m_doc->Reset();
         m_doc->SetType(Raw);
         pugi::xml_document xmlDoc;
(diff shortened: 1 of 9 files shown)


src/iomusxml.cpp
```diff
@@ -576,6 +576,7 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
         if (!measure) {
             LogWarning("MusicXML import: Element '%s' could not be added to measure '%s'",
                 iter->second->GetClassName().c_str(), iter->first.c_str());
+            m_controlElements.erase(iter);
```
lpugin: This screws up the iterator and does not fix the memory leak because the ControlElement is still not deleted.

rettinghaus: Ok, so we just clear all remaining elements.

Issue #1042, opened 2019-04-23
Slurs across voices/layers in musicXML are not converted to MEI


wergo:
Similar to #1039, slurs are only imported when belonging to the same staff, layer, and number.

Slurs might start and end in different layers and even staves, but are uniquely identified by their number (hopefully). 

In my previous example, the slur below the notes would not get imported as it starts in layer 1 and ends in layer 2 (for musicXML see #1039):
<img width="449" alt="Op35-Tie-Excerpt-MuseScore" src="https://user-images.githubusercontent.com/45632600/56585193-a449f980-65dd-11e9-9915-0846988fa567.png">

Here is a fictitious example with slurs crossing staves and voices/layers with the slur numbering indicated in the screen shot (red numbers). Currently, slur 1 and 3 are not converted to MEI. (In the musicXML file, the ending for slur 3 occurs _before_ its starting element which might cause difficulties for the importer.)

<img width="466" alt="Beethoven Mondschein-Sonate-excerpt-FINALE" src="https://user-images.githubusercontent.com/45632600/56586496-1de2e700-65e0-11e9-9e8f-4a1f49f7b85c.png">

This is the musicXML code for this example:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
  <movement-title>Sonata Op. 27, Nr. 2</movement-title>
  <identification>
    <creator type="composer">Beethoven</creator>
    <encoding>
      <software>Finale v26 for Mac</software>
      <encoding-date>2019-04-23</encoding-date>
      <supports attribute="new-system" element="print" type="yes" value="yes"/>
      <supports attribute="new-page" element="print" type="yes" value="yes"/>
      <supports element="accidental" type="yes"/>
      <supports element="beam" type="yes"/>
      <supports element="stem" type="yes"/>
    </encoding>
  </identification>
  <defaults>
    <scaling>
      <millimeters>6.35</millimeters>
      <tenths>40</tenths>
    </scaling>
    <page-layout>
      <page-height>1871</page-height>
      <page-width>1323</page-width>
      <page-margins type="both">
        <left-margin>99</left-margin>
        <right-margin>80</right-margin>
        <top-margin>117</top-margin>
        <bottom-margin>80</bottom-margin>
      </page-margins>
[...]

PR #1045, opened 2019-04-26
MusicXML import: slurs across layers and staves are imported from MusicXML

wergo:
Currently, slurs are only imported from MusicXML, when they are in the same voice (layer) and staff (please see discussion on issue #1042).  

In this branch, slurs starts and stops are matched only by their `@number` attribute, not `@layer` and `@staff` anymore. Since in a partwise MusicXML a `@stop` attribute of a slur may occur _before_ the `@start` attribute has been called, an additional `m_slurStopStack` is introduced that contains unmatched slur stops (in addition to the `m_slurStack` for opened slurs). Since these should be matched within the same measure, the routine in `MusicXmlInput::OpenSlur` for matching them checks for measure number and slur number. Left-over open slurs and unmatched slur stops are reported at the end by their `xml:id`.

Note: There is a restriction in MusicXML that there can not be more than 6 different slurs (or other spanners) in one measure (see http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#ST-MusicXML-number-level.htm). I can imagine @craigsapp generating examples easily exceeding this restriction. :-)

I tested this code successfully with the examples from #1042 and some others. I am eager for comments and more discussion.

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -63,15 +63,23 @@ namespace musicxml {
 
     class OpenSlur {
     public:
-        OpenSlur(int staffN, int layerN, int number)
+        OpenSlur(int number)
         {
-            m_staffN = staffN;
-            m_layerN = layerN;
             m_number = number;
         }
 
-        int m_staffN;
-        int m_layerN;
+        int m_number;
+    };
+
+    class CloseSlur {
+    public:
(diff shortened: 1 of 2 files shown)


lpugin:
This looks good to me. @rettinghaus ?

lpugin:
Thanks

PR #1049, opened 2019-04-29
* Bar Lines: Support rendering dotted barlines

earboxer:
Adds support for dotted barlines. Makes dashed barlines' dashes longer.

46a-Barlines.xml from the [Unofficial MusicXML test suite](http://lilypond.org/doc/v2.18/input/regression/musicxml/collated-files).

The end of measure 3 has the dotted barline.

file


earboxer:
On second thought, we should be using the [SMuFL characters](http://www.smufl.org/version/latest/range/barlines/)

lpugin:
> On second thought, we should be using the [SMuFL characters](http://www.smufl.org/version/latest/range/barlines/)

I am not convinced characters are very appropriate for drawing lines.

PR #1051, opened 2019-05-08
musicXMLimport supports endings

wergo:
Here is a way to get endings into Verovio from musicXML. It works for my examples supporting `@type="discontinue"`, multi-staff scores, and endings spanning over multiple measures. Please test it.

If the `<ending>` element text is missing in musicXML, the `@number` attribute is taken instead (several musicXML exporters do this).

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -295,6 +295,8 @@ private:
     std::vector<std::pair<Tie *, musicxml::OpenTie> > m_tieStack;
     /* The stack for hairpins */
     std::vector<std::pair<Hairpin *, musicxml::OpenHairpin> > m_hairpinStack;
+    /* The stack of endings to be inserted at the end of XML import */
+    std::vector<std::pair<Measure *, musicxml::EndingInfo> > m_endingStack;
     /* The stacks for ControlElements */
     std::vector<Dir *> m_dirStack;
     std::vector<Dynam *> m_dynamStack;
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -94,14 +94,16 @@ namespace musicxml {
 
     class EndingInfo {
     public:
-        EndingInfo(std::string endingNumber, std::string endingType)
+        EndingInfo(std::string endingNumber, std::string endingType, std::string endingText)
         {
             m_endingNumber = endingNumber;
             m_endingType = endingType;
+            m_endingText = endingText;
         }
 
         std::string m_endingNumber;
         std::string m_endingType;
+        std::string m_endingText;
     };
 
 } // namespace musicxml
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -91,7 +91,7 @@ namespace musicxml {
         int m_dirN;
         std::string m_endID;
     };
-    
+
     class EndingInfo {
     public:
         EndingInfo(std::string endingNumber, std::string endingType)
@@ -99,13 +99,13 @@ namespace musicxml {
             m_endingNumber = endingNumber;
             m_endingType = endingType;
         }
-        
+
         void SetEndingType(std::string endingType) { m_endingType = endingType; }
-        
+
         std::string m_endingNumber;
         std::string m_endingType;
     };
-   
+
 } // namespace musicxml
 
 //----------------------------------------------------------------------------
@@ -203,6 +203,13 @@ private:
     bool HasAttributeWithValue(pugi::xml_node node, std::string attribute, std::string value);
     bool IsElement(pugi::xml_node node, std::string name);
     bool HasContentWithValue(pugi::xml_node node, std::string value);
+    ///@}
+
+    /*
+     * @name Helper method to check whether a ending measure number is already present in m_endingStack.
+     */
+    ///@{
+    bool NotInEndingStack(std::string measureN);
     ///@}
 
     /*
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -91,6 +91,18 @@ namespace musicxml {
         int m_dirN;
         std::string m_endID;
     };
+    
+    class EndingInfo{
+    public:
+        EndingInfo(std:string endingNumber, std:string endingType)
+        {
+            m_endingNumber = endingNumber;
+            m_endingType = endingType;
+        }
+        
+        std::string m_endingNumber;
+        std::string m_endingType;
+    }
 
 } // namespace musicxml
 
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -94,7 +94,7 @@ namespace musicxml {
 
     class EndingInfo {
     public:
-        EndingInfo(std::string endingNumber, std::string endingType, std::string endingText)
+        EndingInfo(std::string const &endingNumber, std::string const &endingType, std::string const &endingText)
         {
             m_endingNumber = endingNumber;
             m_endingType = endingType;
@@ -209,7 +209,7 @@ private:
      * @name Helper method to check whether a ending measure number is already present in m_endingStack.
      */
     ///@{
-    bool NotInEndingStack(std::string measureN);
+    bool NotInEndingStack(std::string const &measureN);
     ///@}
 
     /*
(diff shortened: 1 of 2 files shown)


src/iomusxml.cpp
```diff
@@ -591,6 +600,42 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+                if ((*jter)->GetUuid() != measureList.front()->GetUuid()) {
+                    int idx = section->GetChildIndex(*jter);
+                    section->DetachChild(idx);
+                    ending->AddChild(*jter);
+                }
+                logString = logString + ((*jter == measureList.back()) ? ")." : ", ");
+            }
+            LogMessage(logString.c_str());
+        }
+        m_slurStack.clear();
```
lpugin: I think you wanted to clear m_endingStack

src/iomusxml.cpp
```diff
@@ -591,6 +600,42 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
         measure->AddChild(iter->second);
     }
 
+    // manage endings stack
+    if (!m_endingStack.empty()) {
+        std::vector<std::pair<std::vector<Measure *>, musicxml::EndingInfo> >::iterator iter = m_endingStack.begin();
+        Object *section = iter->first.front()->GetParent();
+        for (; iter != m_endingStack.end(); ++iter) {
```
lpugin: A really minor thing: I usually always initialize `iter` in the loop and not above. Could you change them?

src/iomusxml.cpp
```diff
@@ -591,6 +600,42 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+            } else {
+                ending->SetN(iter->second.m_endingText);
+            }
+            ending->SetLendsym(LINESTARTENDSYMBOL_angledown); // default, does not need to be written
+            if (iter->second.m_endingType == "discontinue") {
+                ending->SetLendsym(LINESTARTENDSYMBOL_none); // no ending symbol
+            }
+            section->ReplaceChild(measureList.front(), ending);
+            ending->AddChild(measureList.front());
+            for (; jter != measureList.end(); ++jter) {
```
lpugin: Same here (initialization of `jter` in the loop))

src/iomusxml.cpp
```diff
@@ -591,6 +600,42 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+                logString = logString + (*jter)->GetUuid().c_str();
+                if ((*jter)->GetUuid() != measureList.front()->GetUuid()) {
```
lpugin: Could you add a comment to explain what is happening here?

src/iomusxml.cpp
```diff
@@ -2165,4 +2228,20 @@ std::string MusicXmlInput::ConvertKindToSymbol(std::string value)
         return "";
 }
 
+bool MusicXmlInput::NotInEndingStack(std::string measureN_)
```
lpugin: Please use `const std::string &measureN` as parameter (and do not make a copy in the method if not necessary)

include/vrv/iomusxml.h
```diff
@@ -92,6 +92,20 @@ namespace musicxml {
         std::string m_endID;
     };
 
+    class EndingInfo {
+    public:
+        EndingInfo(std::string endingNumber, std::string endingType, std::string endingText)
```
lpugin: change parameters to `const std::string &`

src/iomusxml.cpp
wergo: Many thanks, @lpugin, for looking over the code! 
Yes, my mistake. I will make all changes in a combined commit.

src/iomusxml.cpp
wergo: You are right. Here, I wanted to get a pointer to `section`, but I just realize that this pointer is creater further above in that method.

src/iomusxml.cpp
wergo: I will change that.

src/iomusxml.cpp
wergo: I create an `ending` element and replace that with the first `measure` in `section` with `ReplaceChild()`. In the loop, the remaining measures in `measureList` are added to `ending` (if there are any). I see that I can write this in a simpler way. Will come soon.

src/iomusxml.cpp
wergo: Ok.

include/vrv/iomusxml.h
wergo: Ok. But should we do the same also for the neighboring classes (`OpenHairpin`, `OpenSlur`, `CloseSlur`, ...)?

include/vrv/iomusxml.h
```diff
@@ -94,7 +94,7 @@ namespace musicxml {
 
     class EndingInfo {
     public:
-        EndingInfo(std::string endingNumber, std::string endingType, std::string endingText)
+        EndingInfo(std::string const &endingNumber, std::string const &endingType, std::string const &endingText)
```
lpugin: Sorry to be picky, could you change to `const std::string &endingNumber`? No difference, but I like these things to be consistent. Thanks

include/vrv/iomusxml.h
lpugin: Yes, this would be good.

src/iomusxml.cpp
```diff
@@ -2165,4 +2228,20 @@ std::string MusicXmlInput::ConvertKindToSymbol(std::string value)
...
+{
+    std::string measureN = measureN_;
+    std::vector<std::pair<std::vector<Measure *>, musicxml::EndingInfo> >::iterator iter;
+    for (iter = m_endingStack.begin(); iter != m_endingStack.end(); ++iter) {
```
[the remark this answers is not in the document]
wergo: Oh, sure, much shorter.

include/vrv/iomusxml.h
wergo: Done.

include/vrv/iomusxml.h
wergo: Alright. I will do.

wergo: I have another question, that I came across: there are two versions of this enumartion in the verovio project:  `LINESTARTENDSYMBOL_none` and `LINESTARTENDSYMBOL_NONE`. Should all be upper case?

grafik

craigsapp: The one with the capital letters means an undefined end symbol, and the lower case one means that there is no end symbol.

In libmei/atttypes.h:

```
/**
 * MEI data.LINESTARTENDSYMBOL
 */
enum data_LINESTARTENDSYMBOL {
    LINESTARTENDSYMBOL_NONE = 0,
    LINESTARTENDSYMBOL_angledown,
    LINESTARTENDSYMBOL_angleup,
    LINESTARTENDSYMBOL_angleright,
    LINESTARTENDSYMBOL_angleleft,
    LINESTARTENDSYMBOL_arrow,
    LINESTARTENDSYMBOL_arrowopen,
    LINESTARTENDSYMBOL_arrowwhite,
    LINESTARTENDSYMBOL_harpoonleft,
    LINESTARTENDSYMBOL_harpoonright,
    LINESTARTENDSYMBOL_H,
    LINESTARTENDSYMBOL_N,
    LINESTARTENDSYMBOL_Th,
    LINESTARTENDSYMBOL_ThRetro,
    LINESTARTENDSYMBOL_ThRetroInv,
    LINESTARTENDSYMBOL_ThInv,
    LINESTARTENDSYMBOL_T,
    LINESTARTENDSYMBOL_TInv,
    LINESTARTENDSYMBOL_CH,
    LINESTARTENDSYMBOL_RH,
    LINESTARTENDSYMBOL_none,
    LINESTARTENDSYMBOL_MAX
};
```

wergo: Ah, ok. I did not know the difference before. Thanks!

lpugin: Yes. _NONE mean there will be no attribute in the MEI output. With _none there will be an attribute with `"none"`.

lpugin:
Many thanks!

wergo:
Thanks for helping improve the code (and merging)!

PR #1062, opened 2019-05-15
Fixes possible null pointer dereference.

valeriyvan:
Solves static analyser issue.

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -1538,16 +1538,19 @@ int Doc::PrepareLyricsEnd(FunctorParams *functorParams)
 {
     PrepareLyricsParams *params = dynamic_cast<PrepareLyricsParams *>(functorParams);
     assert(params);
-
-    if ((params->m_currentSyl && params->m_lastNote) && (params->m_currentSyl->GetStart() != params->m_lastNote)) {
-        params->m_currentSyl->SetEnd(params->m_lastNote);
+    auto params_m_currentSyl = params->m_currentSyl;
+    if (!params_m_currentSyl) {
+        return FUNCTOR_STOP; // early return
+    }
+    if (params->m_lastNote && params_m_currentSyl->GetStart() != params->m_lastNote) {
+        params_m_currentSyl->SetEnd(params->m_lastNote);
     }
     else if (m_options->m_openControlEvents.GetValue()) {
-        if ((params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_i)
-            || (params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_m)) {
+        auto wordpos = params_m_currentSyl->GetWordpos();
+        if (wordpos == sylLog_WORDPOS_i || wordpos == sylLog_WORDPOS_m) {
             Measure *lastMeasure = dynamic_cast<Measure *>(this->FindChildByType(MEASURE, UNLIMITED_DEPTH, BACKWARD));
             assert(lastMeasure);
-            params->m_currentSyl->SetEnd(lastMeasure->GetRightBarLine());
+            params_m_currentSyl->SetEnd(lastMeasure->GetRightBarLine());
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/doc.cpp
```diff
@@ -1538,16 +1538,19 @@ int Doc::PrepareLyricsEnd(FunctorParams *functorParams)
 {
     PrepareLyricsParams *params = dynamic_cast<PrepareLyricsParams *>(functorParams);
     assert(params);
-
-    if ((params->m_currentSyl && params->m_lastNote) && (params->m_currentSyl->GetStart() != params->m_lastNote)) {
-        params->m_currentSyl->SetEnd(params->m_lastNote);
+    auto params_m_currentSyl = params->m_currentSyl;
```
lpugin: This line is useless, isn't it?

src/doc.cpp
```diff
@@ -1538,16 +1538,19 @@ int Doc::PrepareLyricsEnd(FunctorParams *functorParams)
...
+    if (!params_m_currentSyl) {
+        return FUNCTOR_STOP; // early return
+    }
+    if (params->m_lastNote && params_m_currentSyl->GetStart() != params->m_lastNote) {
+        params_m_currentSyl->SetEnd(params->m_lastNote);
     }
     else if (m_options->m_openControlEvents.GetValue()) {
-        if ((params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_i)
-            || (params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_m)) {
+        auto wordpos = params_m_currentSyl->GetWordpos();
```
lpugin: Please change auto to the actual type

src/doc.cpp
lpugin: I meant: you do not need another local variable

valeriyvan: No, we don't need extra variable.
But. 
1) Pointer dereference isn't free. If it's used twice it's not guaranteed compiler will optimise this. But it's not the point.
2) Point is it simplifies code reading and understanding.

src/doc.cpp
valeriyvan: Hm. Ok, I will do that. Type inference is sweet feature!

src/doc.cpp
lpugin: I find it more confusing. Having a check and early return is OK but without the extra variable.

valeriyvan:
Look how it is now, please.
Is it ok?
Or would you like to get rid of intermediate variable `params_m_currentSyl` at all?

lpugin:
Please remove the params_m_currentSyl variable. Otherwise it looks good

valeriyvan:
done

src/doc.cpp
```diff
@@ -1538,13 +1538,15 @@ int Doc::PrepareLyricsEnd(FunctorParams *functorParams)
...
+        return FUNCTOR_STOP; // early return
+    }
+    if (params->m_lastNote && params->m_currentSyl->GetStart() != params->m_lastNote) {
         params->m_currentSyl->SetEnd(params->m_lastNote);
     }
     else if (m_options->m_openControlEvents.GetValue()) {
-        if ((params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_i)
-            || (params->m_currentSyl->GetWordpos() == sylLog_WORDPOS_m)) {
+        sylLog_WORDPOS wordpos = params->m_currentSyl->GetWordpos();
+        if (wordpos == sylLog_WORDPOS_i || wordpos == sylLog_WORDPOS_m) {
```
lpugin: Sorry, one last thing I am noticing: please put both tests in ()

valeriyvan: https://en.cppreference.com/w/c/language/operator_precedence
Why are these extra parentheses needed?

lpugin: Are you offering me a C++ course? ;-)
It is a habit and prefer to have the code consistent. Thanks

valeriyvan: Well, your game - your rules.
As for me, parentheses which don't change natural operations precedence just create visual garbage. They might be useful in complex conditions. But rare in one-liners.
Pushed change.

craigsapp: > As for me, parentheses which don't change natural operations precedence just create visual garbage.

Absolutely not.

valeriyvan: Ok. Different people - different opinions.
I have returned parentheses back.

PR #1064, opened 2019-05-15
Gives initial values to variables before passing them by reference

valeriyvan:
Gives initial values to variables before passing them by reference to functions to stop Xcode complaining on garbage variables.
Solves 4 static analyser issues.


lpugin:
You would need to make the PR to https://github.com/craigsapp/midifile/
We can then update the files here

valeriyvan:
Are you aware that these files in these repos don't match?

lpugin:
Yes, the version in Verovio is older

PR #1069, opened 2019-05-16
Breaks: Add "line" option to uses encoded system breaks, but automatically determine page breaks.

### LINKED ISSUE #1056: Toolkit: Add breaks option for only line breaks [enhancement]
Currently, the options for 'breaks' are

* none - shows the whole song as 1 line
* auto - automatically formats the song
* encoded - Takes the system breaks (line breaks) and page breaks from the file

What do you think about an additional option for breaks

* line - Take the system breaks (line breaks) from the file, automatically paginate.

(So if you have a file that has breaks encoded for A4, you can change it to US Letter or US Legal, and the only change would be the lines (systems) moving to different pages).

earboxer:
Close #1056


earboxer:
There are a few things I think I know about this.

1. `System::CastOffPages` isn't running as often as I would want (on an example document, it's running only when there are already four systems on the page).
2. in `System::CastOffPages` (with `--breaks=line`) `this->m_drawingYRel` and `this->GetHeight()` are not set to accurate values (they are the same values that `--breaks=auto` gives when `params->m_currentPage->GetChildCount()` is 0 on the first page), so when it does run, a new page isn't being created.

earboxer:
Figured it out! Such a simple change for an 11 month wait... (https://github.com/rism-ch/verovio/pull/1069/commits/b343a4d0765b3dc24b896430f3770f454072dae1)

lpugin:
Was this ever merged?

earboxer:
No. I think Github automatically closed this PR when I merged it into earboxer:develop

/open

lpugin:
No problem. Can you re-open it?

earboxer:
Ok.

lpugin:
Look good. Could you just run clang-format and add a line to the changelog? Thanks!

Issue #935, opened 2018-10-02
clef changes in different layers rendered twice


craigsapp:
Clef changes are supposed to be given for every layer on a staff within a measure.

But the following example causes problems rendering in verovio, since the implicit timestamps of the two clefs cannot be made the same:

<img width="384" alt="screen shot 2018-10-02 at 7 04 57 am" src="https://user-images.githubusercontent.com/3487289/46353560-5eadeb00-c611-11e8-94bd-d76f62cf086d.png">

The problem is that the clef change in the first layer is at beat 1.5, but in the second layer it is at beat 2, so verovio renders this example instead as:

<img width="406" alt="screen shot 2018-10-02 at 7 10 51 am" src="https://user-images.githubusercontent.com/3487289/46353927-2eb31780-c612-11e8-99e7-322cbfda28ef.png">

How should this be fixed?  Related to previous discussion on how to encode clef changes in MEI (perhaps in the music-encoding issues), the clef in the second layer could be given a `@sameas` attribute that points to the clef in the first layer, causing the second clef to not be displayed.

Alternatively, adding an "invisible" attribute to clefs might be useful (although this would not be as optimal a solution as the `@sameas` one).

Another possibility is to give the clef in the second layer a time offset such as with `@tstamp` so that it overlaps with the clef change in the first layer, although such a solution is probably not optimal as well.

Test data:

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2018-10-02T07:11:32" version="2.0.0-dev-eebabfa">
                    <name>Verovio</name>
                    <p>Transcoded from Humdrum</p>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title />
                </titleStmt>
            </work>
        </workDesc>
    </meiHead>
    <music>
        <body>
[...]

Issue #850, opened 2018-05-06
MusicXML import of mid-measure clef changes within multiple layers


PhilippeCuvillier:
When a MusicXML measure contains a mid-measure clef change, this clef should be copied in all layers of the converted MEI file. This is related to discussions in #400 and #540.

Here is a test MusicXML file:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.0">
  <defaults>
    <scaling>
      <millimeters>7.2319</millimeters>
      <tenths>40</tenths>
    </scaling>
    <page-layout>
      <page-height>1545</page-height>
      <page-width>1194</page-width>
      <page-margins type="both">
        <left-margin>70</left-margin>
        <right-margin>70</right-margin>
        <top-margin>88</top-margin>
        <bottom-margin>88</bottom-margin>
      </page-margins>
    </page-layout>
    <system-layout>
      <system-margins>
        <left-margin>0</left-margin>
        <right-margin>0</right-margin>
      </system-margins>
      <system-distance>121</system-distance>
      <top-system-distance>70</top-system-distance>
    </system-layout>
    <music-font font-family="Maestro,engraved" font-size="20.5"/>
    <word-font font-family="Times New Roman" font-size="10.25"/>
  </defaults>
  <part-list>
    <score-part id="P1">
      <part-name print-object="no">MusicXML Part</part-name>
      <score-instrument id="P1-I1">
        <instrument-name>MusicXML Default 1</instrument-name>
      </score-instrument>
      <midi-instrument id="P1-I1">
        <midi-channel>1</midi-channel>
        <midi-program>1</midi-program>
      </midi-instrument>
    </score-part>
  </part-list>
  <!--=========================================================-->
  <part id="P1">
    <measure number="1" width="983">
      <print>
        <system-layout>
          <system-margins>
            <left-margin>70</left-margin>
            <right-margin>0</right-margin>
          </system-margins>
          <top-system-distance>211</top-system-distance>
        </system-layout>
        <measure-numbering>system</measure-numbering>
      </print>
      <attributes>
        <divisions>2</divisions>
        <key>
          <fifths>0</fifths>
          <mode>major</mode>
        </key>
        <time>
          <beats>4</beats>
          <beat-type>4</beat-type>
        </time>
[...]

Issue #540, opened 2017-04-11
MusicXML: notes after clef change not displayed correctly


phboer:
Expected:
![image](https://cloud.githubusercontent.com/assets/25896800/24929161/52caf594-1f05-11e7-97ae-a3f37164ea52.png)

Shown by Verovio:
![image](https://cloud.githubusercontent.com/assets/25896800/24929184/6ca17d8a-1f05-11e7-9a48-f208e05d7dcd.png)

The notes after the clef change are not displayed correctly. 

MusicXML fragement of the measure that contains the clef change:
```
    <measure number="18" width="160.91">
      <attributes>
        <clef number="2">
          <sign>F</sign>
          <line>4</line>
          </clef>
        </attributes>
      <note default-x="12.00" default-y="-25.00">
        <pitch>
          <step>A</step>
          <octave>4</octave>
          </pitch>
        <duration>12</duration>
        <voice>1</voice>
        <type>eighth</type>
        <stem>up</stem>
        <staff>1</staff>
        </note>
      <note>
        <rest/>
        <duration>6</duration>
        <voice>1</voice>
        <type>16th</type>
        <staff>1</staff>
        </note>
      <note default-x="85.65" default-y="-50.00">
        <pitch>
          <step>C</step>
          <octave>4</octave>
          </pitch>
        <duration>6</duration>
        <voice>1</voice>
        <type>16th</type>
        <stem>up</stem>
        <staff>1</staff>
        <beam number="1">begin</beam>
        <beam number="2">begin</beam>
        </note>
      <note default-x="110.20" default-y="-40.00">
        <pitch>
          <step>E</step>
          <octave>4</octave>
          </pitch>
        <duration>6</duration>
        <voice>1</voice>
        <type>16th</type>
        <stem>up</stem>
        <staff>1</staff>
        <beam number="1">continue</beam>
        <beam number="2">continue</beam>
        </note>
      <note default-x="134.76" default-y="-25.00">
        <pitch>
          <step>A</step>
          <octave>4</octave>
          </pitch>
        <duration>6</duration>
        <voice>1</voice>
        <type>16th</type>
        <stem>up</stem>
        <staff>1</staff>
        <beam number="1">end</beam>
        <beam number="2">end</beam>
        </note>
      <backup>
        <duration>36</duration>
        </backup>
      <direction placement="below">
        <direction-type>
          <pedal type="start" line="yes"/>
          </direction-type>
        <staff>2</staff>
        </direction>
[...]

Issue #400, opened 2016-12-28
Pitch placement in measure with intermediate clef change


craigsapp:
A problem occurs in clef assignment for notes when there is a clef change half-way through a measure, and the measure contains more than one layer:

<img width="1120" alt="screen shot 2016-12-28 at 10 24 27" src="https://cloud.githubusercontent.com/assets/3487289/21528554/3f9b8e26-cce8-11e6-9ed4-82e87522ccd9.png">

The note which looks like an E6 in m3 is supposed to be a G4, because the G4 is being displayed in bass clef instead of treble clef.  This is caused by the clef change in layer 1 affecting notes at earlier times in layer 2 before the clef change of layer 1 should be occurring in layer 2.

This is related to #229, but for cases where the clef change is not at the end of the measure but rather in the middle.

Test MEI data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2016-12-28 10:28:08</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.13-dev-2f22b77</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-000000007953841">
                        <staffGrp xml:id="m-000000201294796">
                            <staffDef xml:id="staffdef-000000015908972" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="1" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-000000010732145">
                        <measure xml:id="measure-L4" n="1">
                            <staff xml:id="staff-L4F1" n="1">
                                <layer xml:id="layer-L4F1" n="1">
[...]

PR #1071, opened 2019-05-17
MusicXmlImport: Clef changes are imported to all layers, using `@sameas`

wergo:
This PR addresses a highly discussed issue (#400, #850, #540, #935, #18) that clef changes have to appear in all layers in order to affect the susequent notes accordingly. The solution proposed in #400 was to insert a full clef change with all necessary attributes in the first layer and to add clef changes with a `@sameas` reference to the first in all others. This is now added to MusicXmlImport.

There is a ClefChangeStack that collects all clef changes and inserts them to the all layers in the right measure, staff, and time stamp (before a note/rest/space is inserted or at the end of a measure).

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1047,16 +1047,17 @@ bool MusicXmlInput::ReadMusicXmlMeasure(
             assert(staff);
             for (auto layer : *staff->GetChildren()) {
                 assert(layer);
-                for (musicxml::ClefChange clefChange : m_ClefChangeStack) {
-                    if (clefChange.m_measureNum == measureNum && clefChange.m_staff == staff
-                        && clefChange.m_scoreOnset == m_durTotal) {
-                        if (clefChange.isFirst) { // add clef when first in staff
-                            layer->AddChild(clefChange.m_clef);
-                            clefChange.isFirst = false;
+                std::vector<musicxml::ClefChange>::iterator iter;
+                for (iter = m_ClefChangeStack.begin(); iter != m_ClefChangeStack.end(); iter++) {
+                    if (iter->m_measureNum == measureNum && iter->m_staff == staff
+                        && iter->m_scoreOnset == m_durTotal) {
+                        if (iter->isFirst) { // add clef when first in staff
+                            layer->AddChild(iter->m_clef);
+                            iter->isFirst = false;
                         }
                         else {
                             Clef *sameasClef = new Clef(); // add clef with @sameas referring to original clef
-                            sameasClef->SetSameas(clefChange.m_clef->GetUuid().c_str());
+                            sameasClef->SetSameas(iter->m_clef->GetUuid().c_str());
                             layer->AddChild(sameasClef);
                         }
                     }
@@ -1109,6 +1110,8 @@ void MusicXmlInput::ReadMusicXmlAttributes(
                     meiClef->SetDisPlace(STAFFREL_basic_above);
             }
             m_ClefChangeStack.push_back(musicxml::ClefChange(measureNum, staff, meiClef, m_durTotal));
+            LogMessage("Clef change added: measure %s, staff %d, tstamp %d.",
+                       measureNum.c_str(), staff->GetN(), m_durTotal);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/iomusxml.cpp
```diff
@@ -1035,6 +1041,36 @@ bool MusicXmlInput::ReadMusicXmlMeasure(
...
+                            layer->AddChild(sameasClef);
+                        }
+                    }
+                }
+            }
+        }
+        std::vector<musicxml::ClefChange>::iterator iter;
+        for (iter = m_ClefChangeStack.begin(); iter != m_ClefChangeStack.end(); iter++) {
+            if (!iter->isFirst) // remove from stack, if already inserted
+                m_ClefChangeStack.erase(iter--);
```
lpugin: Can you double check that you do not have a memory leak here? Erasing the ClefChange does not delete the vrv::Clef. If you want to delete it because it is not used and was not added to a Layer, then you have to do
```
delete->iter->m_clef;
```
I don't think you need to erase the iter, so then a C++11 range loop would work too.

wergo: Thanks for the quick review!

I removed that loop and report un-matched clef changes at the end, before I clear the entire stack. (Coming from Java: how do I check memory leaks in Xcode?)

src/iomusxml.cpp
```diff
@@ -641,9 +641,11 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
 
     if (!m_ClefChangeStack.empty()) {
         for (musicxml::ClefChange iter : m_ClefChangeStack) {
-            LogWarning("MusicXML import: Clef change at measure %s, staff %d, time %d not inserted.",
-                iter.m_measureNum.c_str(), iter.m_staff->GetN(), iter.m_scoreOnset);
+            if (iter.isFirst)
```
lpugin: Sorry, it is difficult for me to have an overview, but don't you have to delete the clef? That is, something like:
```
delete iter.m_clef;
```

wergo: The ClefChange object just has pointers inside. And, in all my testing files, there was no single case of an uninserted clef change, so I cannot test that situation.

I checked with Xcode analyze and found just another potential leak in existing code: 
grafik

wergo: The memory leak in iomusicxml.cpp is here: https://github.com/rism-ch/verovio/blob/c83ac38e0fc0414fda9437915ef18f086b7caa1e/src/iomusxml.cpp#L1148 (or line 1198 in my version)

lpugin: Correct. And it has been there for a while...
https://github.com/rism-ch/verovio/commit/a6662949d0881bf763d8bf2a221d712f4d1c278c

(I will fix the one in view_slur.cpp, thanks for running the analyser)

> The ClefChange object just has pointers inside. And, in all my testing files, there was no single case of an uninserted clef change, so I cannot test that situation.

Yes, this is the question here: was the problem added to the document tree (e.g., through `Layer::AddChild`) or not? If not, then the object needs to be deleted explicitly.

Verovio is pretty robust because the memory is managed through its tree structure. That is, once an object is added as a child within the (Doc) tree, the parent will be in charge of deleting it. But we still need to make sure all objects created on the heap with `new` are all added to the tree. These memory leaks will be difficult to spot. The analyzer seem to be a good job but will probably not find all of them. I sometimes run the Apple Instruments that would probably spot this one if you have the corresponding data to test it.

lpugin: I am now looking again and it is looking good. I think I misunderstood what was happening because of the additional loop where I thought you where trying to delete unused clefs.

wergo: Many thanks for merging! The wonderful code structure makes it easy to contribute.

wergo: > Yes, this is the question here: was the clef added to the document tree (e.g., through Layer::AddChild) or not? If not, then the object needs to be deleted explicitly.

I will look into that to understand where it is added.

wergo: So, as it seems, the layer there gets added by the stacked `SelectLayer()` method. This should be fine then.

PR #1084, opened 2019-05-30
Last System Justification: Use parameter to set minimum width for justification.

earboxer:
https://github.com/rism-ch/verovio/issues/198#issuecomment-232763469 allowed the last line to be justified if it was 80% or more* of the width.

This PR hopes to make it possible to pass this as a parameter.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -457,6 +457,7 @@ public:
     OptionBool m_landscape;
     OptionBool m_mensuralToMeasure;
     OptionDbl m_midiTempoAdjustment;
+    OptionDbl m_minLastJust;
     OptionBool m_mmOutput;
     OptionBool m_noFooter;
     OptionBool m_noHeader;
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -457,7 +457,7 @@ public:
     OptionBool m_landscape;
     OptionBool m_mensuralToMeasure;
     OptionDbl m_midiTempoAdjustment;
-    OptionDbl m_minLastJust;
+    OptionDbl m_minLastJustification;
     OptionBool m_mmOutput;
     OptionBool m_noFooter;
     OptionBool m_noHeader;
--- a/src/system.cpp
+++ b/src/system.cpp
@@ -596,9 +596,8 @@ int System::JustifyX(FunctorParams *functorParams)
     // do not justify it if the non-justified width is less than a specified percent.
     if ((parent->GetIdx() == parent->GetParent()->GetChildCount() - 1)
         && (this->GetIdx() == parent->GetChildCount() - 1)) {
-        // TODO: Get this from the option.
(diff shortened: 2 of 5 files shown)


earboxer:
Note: This is not ready yet: I'm not sure what's the right way to pass parameters into this code.

src/system.cpp
```diff
@@ -592,12 +592,13 @@ int System::JustifyX(FunctorParams *functorParams)
...
-    // Eventually we should make this a parameter
+    // Check if we are on the last page and on the last system:
+    // do not justify it if the non-justified width is less than a specified percent.
     if ((parent->GetIdx() == parent->GetParent()->GetChildCount() - 1)
         && (this->GetIdx() == parent->GetChildCount() - 1)) {
-        // HARDCODED
-        if (params->m_justifiableRatio > 1.25) {
+        // TODO: Get this from the option.
+        double minLastJust = 0.8;
+        if (minLastJust > 0 && params->m_justifiableRatio > (1/minLastJust)) {
```
lpugin: Please at () around each condition. What will happen with 0? No justification? Is this the desired behavior?

src/system.cpp
```diff
@@ -592,12 +592,13 @@ int System::JustifyX(FunctorParams *functorParams)
...
 
-    // Check if we are on the last page and on the last system - do no justify it if ratio > 1.25
-    // Eventually we should make this a parameter
+    // Check if we are on the last page and on the last system:
+    // do not justify it if the non-justified width is less than a specified percent.
     if ((parent->GetIdx() == parent->GetParent()->GetChildCount() - 1)
         && (this->GetIdx() == parent->GetChildCount() - 1)) {
-        // HARDCODED
-        if (params->m_justifiableRatio > 1.25) {
+        // TODO: Get this from the option.
```
lpugin: To get it from the options, you need to add a `*Doc` member to the JustifyXParams class, and the access the options from it. OK?

include/vrv/options.h
```diff
@@ -457,6 +457,7 @@ class Options {
     OptionBool m_landscape;
     OptionBool m_mensuralToMeasure;
     OptionDbl m_midiTempoAdjustment;
+    OptionDbl m_minLastJust;
```
lpugin: Maybe rename to `m_minLastJustification`? This will make it easier to understand what it is.

src/system.cpp
earboxer: With 0 it will always justify. If C++ had a concept of 1/0 == ∞, then the logic wouldn't need the 0 check: the justifiableRatio (which is the percentage that it would be scaled by to justify) could never be greater than infinity, so the `FUNCTOR_STOP` won't be returned, and the justification will happen.

src/system.cpp
earboxer: Thanks!

PR #1088, opened 2019-06-04
MusicXml Import: Support for arpeggios

wergo:
Support for arpeggio import from MusicXml added (including parsing of `@number`, `@direction`, and `@color`). Relevant noteIds are collected and added to the `@plist` attribute of `<arpeg>`.


lpugin:
Thanks!

PR #1095, opened 2019-06-11
Adding iOS framework to project

Noroxs:
I would like to include Verovio into my iOS project. To simplify this action I added a new target (Cocoa Touch Framework) to the .xcodeproj file.

How to use it:
1. Integrate Verovio via git submodule into your current iOS project
2. Include `Verovio.xcodeproj` file within your project (Drag & Drop or Add Files to ...)
3. Select the new target `VerovioFramework`
4. Build the selected `VerovioFramework` target
5. Select your project target again
6. Include the `VerovioFramework.framework` file as `Embedded Binaries` and `Linked Framworks and Libraries` to your project target
7. Build the target
8. Use Verovio within your iOS App

This pull request only applies changes to the project file and the framework target itself.
Carthage support could be added later.


VerovioFramework/VerovioFramework.h
```diff
@@ -0,0 +1,91 @@
...
+
+#import <UIKit/UIKit.h>
+
+//! Project version number for VerovioFramework.
+FOUNDATION_EXPORT double VerovioFrameworkVersionNumber;
+
+//! Project version string for VerovioFramework.
+FOUNDATION_EXPORT const unsigned char VerovioFrameworkVersionString[];
+
+// In this header, you should import all the public headers of your framework using statements like #import <VerovioFramework/PublicHeader.h>
```
lpugin: Can't you use wildcards for this? It is annoying to have to add each new header by hand. In other builds (Python, Java, etc.*) we do not have to do it.

Noroxs: I'm sorry but Objective-C doesn't offer any kind of wildcard imports.
It could be possible to ignore the imports within the umbrella file. If you import a file that is not added to the umbrella header will lead to a warning `Missing submodule '...'`.

You also have to include every new file within the new framework target. 

If you don't want to do this I could do the maintenance part with adding new files to the framework target + adjusting the umbrella header file.

Noroxs: It could be possible to add a script to add all new files to the framework target + adjust the umbrella header. But in this case all files, needed or not, will be added to the project as well.

lpugin: > It could be possible to add a script to add all new files to the framework target + adjust the umbrella header

This sounds good to me.

Please move the output VerovioFramework to a `./bindings/ios` directory?

Noroxs: It looks like the script approach is not possible. There is no documented way to add files via script or command-line tool to the project file. The project file can become corrupted with any new Xcode version.
I also googled for another approach but could not find one. It's always needed to manually add the file to the target and adjust the visibility level (private, project, public). Adding the import to the umbrella header just fixes the warning.

If you can't approve this pull request I will continue with the fork and adapt all changes there.

Noroxs: I moved the VerovioFramework files (VerovioFramework.h + Info.plist) to `./bindings/iOS`directory

Noroxs:
Do you think it's possible for you to accept this pull request or is there anything to improve?

lpugin:
I think it is possible to improve it. What we can try is to have the framework including only one header file (./bindings/iOS/all.h). This file should be excluded from the git repository and generated from a script when the framework is built. The we can easily have the script listing in this file all the header files as we have for all the bindings. Can you try this?

Noroxs:
Yes I can implement such script. 
But there is an additional issue. All files (header and implementation) need to be added to the xcodeproj target (VerovioFramework). I checked it and there is now way to do this via a script. At some point it has to be done manually (clicking a checkbox). 

I already did this for this pull request but all new files should be added like this as well.

tools/create_ios_framework_headers.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
```
lpugin: Please move the file to ./bindings/iOS

lpugin:
It shows missing files for me (still pointing to ./ ?) Can you double check this?

image

tools/create_ios_framework_headers.sh
Noroxs: I moved it to the desired location

Noroxs:
Could you please check the latest version and check if you can build the `VerovioFramework` target?

Noroxs:
Do you still have the missing files problem? Did you select the correct scheme/target within your Xcode? I tried to reproduce this issue by cloning the project several times at several locations but I couldn't reproduce it.

Issue #1109, opened 2019-06-27
Default velocity in MIDI is to low


peterkorgaard:
When rendering MIDI all events are given the same velocity. This is fine, but I think the default level is to low. In MIDI velocity goes from 0-127. The default velocity in rendered MIDI from Verovio is 50 which is below half. Is it possible to change the default level to e.g. 100?

lpugin:
See #1110

PR #1110, opened 2019-06-27
add support for midi velocity

rettinghaus:
This PR addresses #1109.
It adds Verovios default midi values to `options.h` and sets default velocity to default MIDI 1.0 value (which is 90). 
Additional support for r/w of `@vel` on `<note>` and MIDI playback.


include/vrv/options.h
```diff
@@ -35,6 +35,13 @@ class OptionGrp;
 #define MIN_SCALE 1
 #define MAX_SCALE 1000
 
+//----------------------------------------------------------------------------
```
lpugin: Please move these to vrvdef.h

rettinghaus: Didn't put them there, but ok.

lpugin: I meant the midi values

rettinghaus: As you wish …

PR #1126, opened 2019-07-24
Xcode 11 Beta support

Noroxs:
Disable the option for the target device `Mac`. This will requires macOS 10.15


Noroxs:
Hello @lpugin. Is there a possibility that this request will be considered?
It's a bug fix for the current Xcode 11 version. For some reason the `SUPPORTS_UIKITFORMAC` is automatically enabled but requires the beta version of macOS Catalina (10.15).

PR #1142, opened 2019-09-13
adding support for graceGrp

rettinghaus:
This PR adds support for `<graceGrp>`. 

For test cases I've taken the [feature example](https://github.com/rism-ch/verovio/files/3610542/gracenotes.zip) and updated it to MEI4. 

No magic happens here. AFAICS the attach attribute doesn't need any special treatment.


src/layerelement.cpp
```diff
@@ -16,7 +16,6 @@
 #include "barline.h"
 #include "beam.h"
 #include "beatrpt.h"
-#include "btrem.h"
```
lpugin: Do you know why btrem was removed?

rettinghaus: I think, because `BTREM` isn't used in layerelement.cpp. 
I've put it back in.

PR #1152, opened 2019-10-03
Render expansion element with --expand <expansion>

wergo:
This PR enables Verovio to expand the `@plist` references of an `expansion` element, given by a command line argument `--use-expansion`. It resolves all references and any recursive expansion elements upon MEI import and adds them to the structure. 

Usage: `verovio --expand <expansionXmlId> -f mei -o outfile.mei infile.mei`

When cloning the repeated/expanded elements, it checks for all referring attributes (`@startid`, `@endid`, `@plist`, `@sameas` etc.) so that they get updated correctly. It also adds a new `expansion` element to the target file with the `xml:id="expansion-notated-000000xxxxxxxxxx"` to be able to reconstruct the original file (in case the source MEI was overridden). 

The toolkit has two new methods:
`tk.getNotatedIdForElement("measure-000321923") `returns the xml:id string
of the original/notated element (the same if already the original or
when no expansion was done there).

`tk.getExpansionIdsForElement(std:string)` returns a json string of all
repeated occurrences of that element. The first is always the
notated/original element and others appear in the order of the MEI file.

Using the toolkit works like this:
```
tk.setOption("useExpansion","expansion-full")
tk.loadFile("Beethoven-Op35-Eroica-Variationen.mei")
tk.getExpansionIdsForElement("measure-000002319081")
...
```

Known issues: time pointing elements across the sections are not handled at the moment (see Op. 35, Var. XI).

Test MEI files:
https://iwk.mdw.ac.at/goebl/Beethoven-Op35-Eroica-Variationen.mei
https://iwk.mdw.ac.at/goebl/Beethoven_Op126Nr6.mei
https://iwk.mdw.ac.at/goebl/sonata01-3.mei


include/vrv/doc.h
```diff
@@ -120,6 +125,15 @@ class Doc : public Object {
      */
     bool GetMidiExportDone() const;
 
+    /**
+     * WG: render expansion recursion
+     */
+    xsdAnyURI_List useExpansion(xsdAnyURI_List expansionList, xsdAnyURI_List existingList, Object *prevSection);
```
lpugin: Always use `CamelCase` names (and not `camelCase` ones) for methods

include/vrv/doc.h
```diff
@@ -409,6 +423,11 @@ class Doc : public Object {
     /** Record notation type for document */
     data_NOTATIONTYPE m_notationType;
```
lpugin: Put all members declarations below and do not mix them with method declarations

include/vrv/doc.h
```diff
@@ -409,6 +423,11 @@ class Doc : public Object {
...
 
+    /** The expansion map indicates which xmlId has been repeated (expanded) elsewhere */
+    std::vector<std::vector<std::string> > m_expansionMap;
```
lpugin: I think we need to add a new ExpansionMap class to handle this. The we can move some of the methods in Doc there, which would make it clearer.

Also, I can imaging you could use a <std::map> to map each ids with the expanded ones, but maybe I am overlooking something.

include/vrv/doc.h
```diff
@@ -409,6 +423,11 @@ class Doc : public Object {
...
+
+    std::vector<std::string> getExpansionIdsForElement(const std::string &xmlId);
```
lpugin: Change to `CamelCase` (and elswhere too)

include/vrv/options.h
```diff
@@ -474,6 +474,7 @@ class Options {
     OptionInt m_pageMarginRight;
     OptionInt m_pageMarginTop;
     OptionInt m_pageWidth;
+    OptionString m_useExpansion;
```
lpugin: Could be changed to `instantiateExpansion`? or `realizeExpansion`? or `implementExpansion`?

src/doc.cpp
```diff
@@ -1596,4 +1598,172 @@ int Doc::PrepareTimestampsEnd(FunctorParams *functorParams)
...
+std::vector<std::string> *Doc::getUuidList(Object *object, std::vector<std::string> *idList)
+{
+    for (Object *o : *object->GetChildren()) {
+        idList->push_back(o->GetUuid());
+        idList = getUuidList(o, idList);
+    }
+    return idList;
+}
+
+bool Doc::updateIds(Object *object)
```
lpugin: This should be done through Functors. It is a bit cumbersome with interfaces but basically have one `Object::UpdateIds(FunctorParams *params)` functor methods, that would cast them (as you do here) but then call the `InterfaceUpdateIds(FunctorParams *functorParams, Object *object)` method to be defined for the InterfaceTimePoint, InterfaceTimeSpanning, etc. classes. (Look at ResetDrawing and InterfaceResetDrawing for an example).

(The loop inside the children is done by the functor mechanism)

src/doc.cpp
```diff
@@ -1596,4 +1598,172 @@ int Doc::PrepareTimestampsEnd(FunctorParams *functorParams)
...
+        ids.push_back(newXmlId.c_str());
+        m_expansionMap.push_back(ids);
+    }
+    return true;
+}
+
+std::vector<std::string> Doc::getExpansionIdsForElement(const std::string &xmlId)
+{
+    std::vector<std::vector<std::string> >::iterator iter;
+    for (iter = m_expansionMap.begin(); iter != m_expansionMap.end(); ++iter) {
```
lpugin: This lookup would not be necessary if you use a std::map instead

src/iomei.cpp
```diff
@@ -2718,6 +2718,34 @@ bool MeiInput::ReadDoc(pugi::xml_node root)
...
+        else {
+            xsdAnyURI_List expansionList = start->GetPlist();
+            // std::cout << "[useExpansion] xml:id=\"" << expansionId.c_str() << "\" plist={";
+            // for (std::string s : expansionList) std::cout << s.c_str() << ((s != expansionList.back()) ? " " :
+            // "}.\n");
+            xsdAnyURI_List existingList;
+            existingList = m_doc->useExpansion(expansionList, existingList, start);
+            // save original/notated expansion as element in expanded MEI
+            Expansion *originalExpansion = new Expansion();
+            originalExpansion->SetUuid("expansion-notated");
```
lpugin: I don't really like this id... Maybe we need to add a bool member to flags them as expansion? See m_isAttribute example

include/vrv/doc.h
wergo: Many thanks for all your comments. I'll change that certainly.

include/vrv/doc.h
wergo: Ok, I can create a new file called ExpansionMap (or inside doc.cpp?).

Re: using std::map for this: as the lookup is not always pairwise (when there are more than one repetition and we have often different numbers of repetitions for the elements). 

Do you think of this structure std::map<std::string, std::vector<std::string>> that has each element as key and a complete list of element Ids as value: 
"orig1"; <"orig1","rend2">
"rend2"; <"orig1","rend2">
When adding a new pair "orig1","rend3" would result in this:
"orig1"; <"orig1","rend2","rend3">
"rend2"; <"orig1","rend2","rend3">
"rend3"; <"orig1","rend2","rend3">

That is slower in construction, but faster in retrieving the content. Is that, what you have in mind?

include/vrv/options.h
wergo: We had a extensive discussion with David about this. We originally used "renderExpansion", which may be confused with the toolkit render methods. Thank you for your ideas! Instantiate sounds much like coding, realize could be confused with the FRBR expression, is there really something implemented when calling this argument?

What about "expandExpansion". This is probably naming it most precisely, but sounds funny.

src/iomei.cpp
wergo: Do I understand correctly that the `m_isAttribute` flag is used to output the MEI content that could be written both as elements or attributes is output as it was at input?

However, for an `isExpansion` flag, there is no valid MEI to reflect this in a MEI file. That's why I thought it would be helpful to have one expansion element written that remembers the notated section configuration.

This flag would be useful for displaying repetitions dynamically in an app, so it could be used in parallel with the added expansion element?

src/doc.cpp
wergo: That's true; see above.

include/vrv/doc.h
lpugin: > I can create a new file called ExpansionMap

Yes, create a new file.

> Do you think of this structure std::map<std::string, std::vectorstd::string> 

Yes, exactly.

include/vrv/options.h
lpugin: > What about "expandExpansion"

Or just "expand"?

src/iomei.cpp
lpugin: `m_isAttribute` is used to flag an Object when it is converted from an MEI attribute into an element by verovio. For example `note@accid` become `note/accid@accid` in verovio, so we can support both but implement only the most complex one. When we output the MEI, we can convert it back to an attribute. Potentially, we could ignore it in the output and use verovio to convert some attributes to elements.

It looks like `m_isExpansion` would be useful for similar reason. Once flags, the element could be easier not be written in the output, or written, and it that case verovio could be used to expand MEI files.

include/vrv/options.h
wergo: Yes, perfect, that's the solution. Thanks for that!

include/vrv/doc.h
wergo: Just one more question: do you see the `std::map<std::string, std::vectorstd::string>` to live in that `expansionMap` class, or could that stay within `doc` (which seems simpler to me)? If not, how would you link the expansionMap to the doc?

lpugin: In ExpansionMap. Doc would simply have one ExpansionMap member. It does not make any difference, it is mostly not to overload the Doc class, which is already quite dense. (In the end, it is mostly to have the methods in the ExpansionMap and not in Doc)

src/iomei.cpp
wergo: Dear Laurent, I have implemented all requested changes (std::map, --expand, Object::isExpansion, ExpansionMap as separate class), except the functor approach to the recursive UpdateIds(Object*) method. I wont find the time at the moment, but hope to do this soon. Best, Werner

PR #1159, opened 2019-10-16
set currentColor as default color

### LINKED ISSUE #556: Implement currentColor in SVG

rettinghaus:
fixes #556
(after two years all original problems are gone, so now it's time to finally address it)

### DIFF
--- a/Verovio.xcodeproj/project.pbxproj
+++ b/Verovio.xcodeproj/project.pbxproj
@@ -2637,10 +2637,9 @@
 			};
 			buildConfigurationList = 8F086EA3188534680037FD8E /* Build configuration list for PBXProject "Verovio" */;
 			compatibilityVersion = "Xcode 3.2";
-			developmentRegion = English;
+			developmentRegion = en;
 			hasScannedForEncodings = 0;
 			knownRegions = (
-				English,
 				en,
 			);
 			mainGroup = 8F086E9E188534680037FD8E;
--- a/include/vrv/devicecontextbase.h
+++ b/include/vrv/devicecontextbase.h
@@ -17,6 +17,7 @@
 
 namespace vrv {
 
+#define AxNONE -1
 #define AxWHITE 255 << 16 | 255 << 8 | 255
 #define AxBLACK 0
 #define AxRED 255 << 16
--- a/src/bboxdevicecontext.cpp
+++ b/src/bboxdevicecontext.cpp
@@ -35,8 +35,8 @@ BBoxDeviceContext::BBoxDeviceContext(View *view, int width, int height, unsigned
     m_drawingText = false;
     m_textAlignment = HORIZONTALALIGNMENT_left;
 
-    SetBrush(AxBLACK, AxSOLID);
-    SetPen(AxBLACK, 1, AxSOLID);
+    SetBrush(AxNONE, AxSOLID);
+    SetPen(AxNONE, 1, AxSOLID);
 
     m_update = update;
 
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -856,6 +856,7 @@ void SvgDeviceContext::AddDescription(const std::string &text)
 std::string SvgDeviceContext::GetColour(int colour)
 {
     std::ostringstream ss;
+    ss << "#";
     ss << std::hex;
 
     switch (colour) {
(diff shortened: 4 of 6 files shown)


Verovio.xcodeproj/project.pbxproj
```diff
@@ -2637,10 +2637,9 @@
 			};
 			buildConfigurationList = 8F086EA3188534680037FD8E /* Build configuration list for PBXProject "Verovio" */;
 			compatibilityVersion = "Xcode 3.2";
-			developmentRegion = English;
```
lpugin: Can you discard the changes on this file?

rettinghaus: X-Code wanted this change, which seems understandable to me. 
You may check this: 
https://stackoverflow.com/questions/7051120/why-doesnt-my-file-move-into-en-lproj-but-instead-into-a-new-english-lproj

But if you still want it to be reverted, I'll do so.

lpugin: > https://stackoverflow.com/questions/7051120/why-doesnt-my-file-move-into-en-lproj-but-instead-into-a-new-english-lproj

This is from 8 years ago... What version of XCode are you using? I am on 10.1

craigsapp: I don't know if it is identical in purpose, but there is a environmental variable called `$MACOSX_DEPLOYMENT_TARGET` that specifies the oldest MacOS operating system that the code should be compiled to work on (such as "10.4" to run on computers as old as MacOS 10.4).  If the variable is not set, then only the version of the MacOS on the compiling computer and later can be used to run the compiled program.

rettinghaus:
I‘m using the latest version of X-Code. 
And they recommended eight years ago already to use ISO codes.

PR #1165, opened 2019-10-23
Revert "set currentColor as default color"

lpugin:
Reverts rism-ch/verovio#1159
It breaks some projects and we need to investigate what is the problem

### DIFF
--- a/Verovio.xcodeproj/project.pbxproj
+++ b/Verovio.xcodeproj/project.pbxproj
@@ -2637,9 +2637,10 @@
 			};
 			buildConfigurationList = 8F086EA3188534680037FD8E /* Build configuration list for PBXProject "Verovio" */;
 			compatibilityVersion = "Xcode 3.2";
-			developmentRegion = en;
+			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
 				en,
 			);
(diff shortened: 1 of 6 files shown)


rettinghaus:
*some projects* is somewhat vage for further investigation

lpugin:
Yes, this was an emergency fix. One project that was broken is https://measuringpolyphony.org/
Source code available here https://github.com/MeasuringPolyphony/measuring_polyphony_jekyll
With the change to `currentColor` staff lines disappear.

rettinghaus:
`currentColor` doesn't "break" the project, it's this CSS line:
https://github.com/MeasuringPolyphony/measuring_polyphony_jekyll/blob/f371936e27e3db3682a0d1c435dfdd6ed19cc111/css/transcriptions.css#L376

So `currentColor` does exactly what it is supposed to do. Reverting accepted changes due to some questionable CSS is itself somewhat questionable.

What we could do to avoid unwanted effects from CSS is to add `color:black` as default in the `g.page-margin` style setting.

lpugin:
> Reverting accepted changes due to some questionable CSS is itself somewhat questionable.

Well, since now you fixed what broke the project, I am going to ask you to redo the PR. I hope you will not question this ;-)

PR #1180, opened 2019-11-02
enhance control on footers and headers

### LINKED ISSUE #1178: auto-generated headers
By default Verovio generates an `pgHead` out of the information in `title` (as far as no `pgHead` is encoded). 
The `--no-header` option supresses the `pgHead` in the output.
Problem is, there is no way to differentiate between encoded and generated headers. I have some files with encoded `pgHead`, others without, but only want the encoded ones to be shown. 

Perhaps the `--no-header` should be changed in a way breaks are handled: 
* `--header auto` the default behavior like now
* `--header encoded` uses only encoded `pgHead`
* `--header none` prints no header at all

rettinghaus:
This PR makes the `--no-footer` and `--no-header` options deprecated and introduces the options `--footer` and `--header` instead, which work equally to `--breaks` with the values `auto` (default), `encoded` and `none`.

fixes #1178


tools/main.cpp
```diff
@@ -224,6 +224,8 @@ int main(int argc, char **argv)
         // deprecated - some use undocumented short options to catch them as such
         { "border", required_argument, 0, 'b' },
         { "ignore-layout", no_argument, 0, 'i' },
+        { "no-footer", no_argument, 0, 'y' },
```
lpugin: We can group them under a 'd' shortcut, which we will keep for deprecating other options in the future. (You should get the long option name with `long_options[option_index].name`). With this we avoid to arbitrary shortcuts that did not exist before for deprecating options

PR #1204, opened 2019-11-26
Fixing calculation of shift of slur position

valeriyvan:
Not sure my fix is right. Hope somewhere close.

### DIFF
--- a/src/slur.cpp
+++ b/src/slur.cpp
@@ -364,13 +364,12 @@ void Slur::AdjustSlurPosition(Doc *doc, FloatingCurvePositioner *curve,
         }
         if (dist != 0) posXRatio = (float)posX / ((float)dist / 2.0);
 
-        shift = 0;
         // Keep the maximum shift on the left and right
         if (curveDir == curvature_CURVEDIR_above) {
             shift = intersection;
         }
         else {
-            shift = intersection;
+            shift = -intersection;
         }
         // shift += doc->GetDrawingUnit(100);
         if (shift > 0) {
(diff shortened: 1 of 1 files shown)


src/slur.cpp
```diff
@@ -364,18 +363,18 @@ void Slur::AdjustSlurPosition(Doc *doc, FloatingCurvePositioner *curve,
...
 
-        shift = 0;
         // Keep the maximum shift on the left and right
+        int shift;
         if (curveDir == curvature_CURVEDIR_above) {
             shift = intersection;
         }
         else {
-            shift = intersection;
+            shift = -intersection;
```
lpugin: Why are you changing this?

valeriyvan: I don't understand logic of calculation. 
But doing same in both branches of `if` suggests here is a bug. 
I thought it was meant something like that.
No? What should it be then?

lpugin: OK. I have to look at this because we need to make sure we fix it in the right direction.

rettinghaus: @lpugin what's with this PR?

valeriyvan: Nothing from my side. Current calculation is definitely wrong and has a bug. But I am unable to fix it.

lpugin: I can see that doing then same in a if/else makes no sense, but could provide us with an MEI snippet that illustrates the bug?

valeriyvan: No, I can't provide MEI. Having such MEI I would probably have hints how to fix.

valeriyvan:
Fixed with PR #1506

PR #1206, opened 2019-11-27
update readme

rettinghaus:


README.md
```diff
@@ -37,10 +37,3 @@ Contributing
 ------------
 
 If you are willing to contribute to Verovio, please follow the [contribution](doc/contributing.md) and [coding style](doc/guidelines.md) guidelines.
-
-Example output
---------------
-
-This page was generated with version 0.9.13-dev-b953d3b
-
-![Example page](https://raw.githubusercontent.com/rism-ch/verovio/gh-pages/images/verovio-exp-0.9.13-dev.png)
```
lpugin: Please leave the example. I am going to update it

README.md
```diff
@@ -8,7 +8,7 @@ See the JavaScript toolkit running in the [MEI Viewer](http://www.verovio.org/me
 
 ![Choice interaction](https://raw.githubusercontent.com/rism-ch/verovio/gh-pages/movies/reflow.gif)
 
-Verovio can also render Plain and Easy (PAE) code and DARMS code and has a basic MusicXML import function. It is designed as a tool usable as a one-step conversion tool and it is written in pure standard C++ with no dependencies on third-party frameworks and libraries.
```
lpugin: Looking at this: I would prefer to keep `with no dependencies on third-party frameworks and libraries.`

rettinghaus: But wouldn't this be incorrect? We use JSON++, pugixml, Midifile, etc. 
How about 
> with no dependencies on external libraries.

PR #1211, opened 2019-12-03
MusicXML conversion: convert line-type for slurs and ties

earboxer:
Will allow dotted, dashed, or wavy style curves.

These are currently not rendered, so this change is mostly under-the-hood.

I added lform to AttCurvature; inheriting it from AttCurveRend did not work (presumably because they both implement ReadCurvature and WriteCurvature).

### DIFF
--- a/include/vrv/slur.h
+++ b/include/vrv/slur.h
@@ -17,7 +17,7 @@ namespace vrv {
 // Slur
 //----------------------------------------------------------------------------
 
-class Slur : public ControlElement, public TimeSpanningInterface, public AttColor, public AttCurvature {
+class Slur : public ControlElement, public TimeSpanningInterface, public AttColor, public AttCurvature, public AttCurveRend {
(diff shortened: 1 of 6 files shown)


lpugin:
files in `./libmei` should not be edited because they are generated and modifications will be lost. See https://github.com/rism-ch/verovio/blob/master/doc/guidelines.md#libmei

> inheriting it from AttCurveRend did not work (presumably because they both implement ReadCurvature and WriteCurvature).

Not sure what you mean here. AttCurveRend does not implement ReadCurvature but ReadCurveRend. This is the way to do this. Could you be more specific about the problem encountered?

earboxer:
Ah, I see. It takes a little more effort to add CurveRend to ties and slurs, thanks for the pointers!

earboxer:
@lpugin I have fixed this to no longer change libmei

PR #1223, opened 2019-12-08
Add support for principal-voice

### LINKED ISSUE #1215: Support principle-voice brackets when importing from MusicXML
MusicXML has a `<principal-voice>` element which allows you to put a start bracket or end bracket above a note, to mark sections of a song. This is used for [Hauptstimme and Nebenstimme](https://en.wikipedia.org/wiki/Hauptstimme), but in my case I'd like to use these brackets for marking piano introductions, as is common for hymns in various hymnals (see https://github.com/w3c/musicxml/issues/275).

Here's the documentation for principal-voice:
https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-principal-voice.htm

Would it be possible for Verovio to import these elements when converting MusicXML to MEI? (For that matter, does MEI have an equivalent?)

Here's a MusicXML code example for the start bracket:
```
      <direction placement="above">
        <direction-type>
          <principal-voice type="start" symbol="plain"/>
        </direction-type>
      </direction>
```
And the end bracket:
```
      <direction placement="above">
        <direction-type>
          <principal-voice type="stop" symbol="plain" />
        </direction-type>
      </direction>
```

(In the MusicXML file, the `<direction>` element is placed immediately before the `<note>` where you want the bracket to appear.)

Finally, here's an example of what I'd expect the brackets to look like:
60059748-004b0000-96ab-11e9-86d5-f2552bb085e8

rettinghaus:
This PR adds support for `principal-voice` in MusicXML importer.
The `bracktSpan@func` get the generic value `analytical`, in addition `@type` is set to `principal-voice`.
Closes #1215

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1094,7 +1094,7 @@ bool MusicXmlInput::ReadMusicXmlMeasure(
             // and with earliest end note.
             if (iter->second->GetPname() == (*jter)->GetPname() && iter->second->GetOct() == (*jter)->GetOct()
                 && (iter->second->GetScoreTimeOnset() < (*jter)->GetScoreTimeOnset()
-                       && (*jter)->GetScoreTimeOnset() < lastScoreTimeOnset)) {
+                    && (*jter)->GetScoreTimeOnset() < lastScoreTimeOnset)) {
                 iter->first->SetEndid("#" + (*jter)->GetUuid());
                 lastScoreTimeOnset = (*jter)->GetScoreTimeOnset();
                 tieMatched = true;
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1524,7 +1524,7 @@ void MusicXmlInput::ReadMusicXmlDirection(
         hairpinNumber = (hairpinNumber < 1) ? 1 : hairpinNumber;
         if (HasAttributeWithValue(wedge.node(), "type", "stop")) {
             // match wedge type=stop to open hairpin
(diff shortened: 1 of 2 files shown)


src/iomusxml.cpp
```diff
@@ -1639,6 +1640,34 @@ void MusicXmlInput::ReadMusicXmlDirection(
...
+        int voiceNumber = wedge.node().attribute("number").as_int();
+        voiceNumber = (voiceNumber < 1) ? 1 : voiceNumber;
+        if (HasAttributeWithValue(lead.node(), "type", "stop")) {
+            std::vector<std::pair<BracketSpan *, musicxml::OpenSpanner> >::iterator iter;
+            for (iter = m_bracketStack.begin(); iter != m_bracketStack.end(); ++iter) {
+                    int measureDifference = m_measureCounts.at(measure) - iter->second.m_lastMeasureCount;
+                    iter->first->SetTstamp2(std::pair<int, double>(measureDifference, timeStamp));
+                    m_bracketStack.erase(iter);
+                    return;
+            }
```
lpugin: There is a problem with this loop. It makes no sense since it contains a `return` statement

rettinghaus: Yes, sorry. The original plan was to allow nested brackets. But those shouldn't turn up, so I took the shortcut. 
Want to do? Remove the loop entirely or add an only an if statement for later improvements?

lpugin: Please remove the loop. (Double check that this does not create a memory leak with objects remaining in the m_bracketStack)

PR #1225, opened 2019-12-09
* Note: Add UpdateFromTransPitch method to apply a pitch to a note.

earboxer:
Implements

> Then take the resulting transposition and place it back into the note attributes.

from @craigsapp's [comment](https://github.com/rism-ch/verovio/issues/1189#issuecomment-563405899)

I call it `Note::UpdateFromTransPitch`

Eventually, we'll want a loop like
```c++
transposer->setTransposition("+M2")
while (Transposable transposableItem = iterator.iterate()){
  TransPitch pitch = transposableItem::GetTransPitch();
  transposer->Transpose(pitch);
  transposableItem::UpdateFromTransPitch(pitch);
}
```

We would need to implement `Transposable`.

Then we could just implement these two methods for `Harm`, `StaffDef`, `ScoreDef`, `KeySig` (and any others relevant classes that we may have missed), to complete the system.

### DIFF
--- a/include/vrv/note.h
+++ b/include/vrv/note.h
@@ -190,6 +190,7 @@ public:
     int GetChromaticAlteration();
 
     TransPitch GetTransPitch();
+    void UpdateFromTransPitch(TransPitch tp);
 
     //----------//
     // Functors //
--- a/include/vrv/note.h
+++ b/include/vrv/note.h
@@ -190,8 +190,11 @@ public:
     int GetChromaticAlteration();
 
     TransPitch GetTransPitch();
+
+private:
     void UpdateFromTransPitch(TransPitch tp);
 
+public:
     //----------//
     // Functors //
     //----------//
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -459,6 +459,43 @@ TransPitch Note::GetTransPitch()
     return TransPitch(pname, this->GetChromaticAlteration(), this->GetOct());
 }
 
+void Note::UpdateFromTransPitch(TransPitch tp)
+{
+    data_PITCHNAME pname = static_cast<data_PITCHNAME>(tp.m_pname + PITCHNAME_c);
+    this->SetPname(pname);
+
+    Accid *accid = this->GetDrawingAccid();
+    if (accid->HasAccidGes()) {
+        data_ACCIDENTAL_GESTURAL gestural = ACCIDENTAL_GESTURAL_NONE;
+        switch (tp.m_accid) {
+            case -2: gestural = ACCIDENTAL_GESTURAL_ff; break;
+            case -1: gestural = ACCIDENTAL_GESTURAL_f; break;
+            case 0: gestural = ACCIDENTAL_GESTURAL_NONE; break;
+            case 1: gestural = ACCIDENTAL_GESTURAL_s; break;
+            case 2: gestural = ACCIDENTAL_GESTURAL_ss; break;
+            default: break; // TODO: Return an error here.
+        }
+        accid->SetAccidGes(gestural);
+    }
+    else {
+        data_ACCIDENTAL_WRITTEN written = ACCIDENTAL_WRITTEN_NONE;
+        switch (tp.m_accid) {
+            case -3: written = ACCIDENTAL_WRITTEN_tf; break;
+            case -2: written = ACCIDENTAL_WRITTEN_ff; break;
+            case -1: written = ACCIDENTAL_WRITTEN_f; break;
+            case 0: written = ACCIDENTAL_WRITTEN_NONE; break;
+            case 1: written = ACCIDENTAL_WRITTEN_s; break;
+            case 2: written = ACCIDENTAL_WRITTEN_x; break;
+            case 3: written = ACCIDENTAL_WRITTEN_xs; break;
+            default: break; // TODO: Return an error here.
+        }
+        accid->SetAccid(written);
+    }
+
+    // Since we didn't read the oct.ges when getting TransPitch, we don't need to account for that.
+    this->SetOct(tp.m_oct);
+}
+
 //----------------------------------------------------------------------------
 // Functors methods
 //----------------------------------------------------------------------------
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -492,8 +492,12 @@ void Note::UpdateFromTransPitch(TransPitch tp)
         accid->SetAccid(written);
     }
 
-    // Since we didn't read the oct.ges when getting TransPitch, we don't need to account for that.
-    this->SetOct(tp.m_oct);
+    if (this->GetOct() != tp.m_oct) {
+        if(this->HasOctGes()) {
+            this->SetOctGes(this->GetOctGes() + tp.m_oct - this->GetOct());
+        }
+        this->SetOct(tp.m_oct);
+    }
 }
 
 //----------------------------------------------------------------------------
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -465,25 +465,38 @@ void Note::UpdateFromTransPitch(TransPitch tp)
     this->SetPname(pname);
 
     Accid *accid = this->GetDrawingAccid();
+    bool transposeGesturalAccid = false;
+    bool transposeWrittenAccid = false;
     if (accid->HasAccidGes()) {
+        transposeGesturalAccid = true;
+    }
+    if (accid->HasAccid()) {
+        transposeWrittenAccid = true;
+    }
+    // TODO: Check the case of both existing but having unequal values.
+    if (!accid->HasAccidGes() && !accid->HasAccid()) {
+        transposeGesturalAccid = true;
+    }
+
+    if (transposeGesturalAccid) {
         data_ACCIDENTAL_GESTURAL gestural = ACCIDENTAL_GESTURAL_NONE;
         switch (tp.m_accid) {
             case -2: gestural = ACCIDENTAL_GESTURAL_ff; break;
             case -1: gestural = ACCIDENTAL_GESTURAL_f; break;
-            case 0: gestural = ACCIDENTAL_GESTURAL_NONE; break;
+            case 0: gestural = ACCIDENTAL_GESTURAL_n; break;
             case 1: gestural = ACCIDENTAL_GESTURAL_s; break;
             case 2: gestural = ACCIDENTAL_GESTURAL_ss; break;
             default: break; // TODO: Return an error here.
         }
         accid->SetAccidGes(gestural);
     }
-    else {
+    if (transposeWrittenAccid) {
         data_ACCIDENTAL_WRITTEN written = ACCIDENTAL_WRITTEN_NONE;
         switch (tp.m_accid) {
             case -3: written = ACCIDENTAL_WRITTEN_tf; break;
             case -2: written = ACCIDENTAL_WRITTEN_ff; break;
             case -1: written = ACCIDENTAL_WRITTEN_f; break;
-            case 0: written = ACCIDENTAL_WRITTEN_NONE; break;
+            case 0: written = ACCIDENTAL_WRITTEN_n; break;
             case 1: written = ACCIDENTAL_WRITTEN_s; break;
             case 2: written = ACCIDENTAL_WRITTEN_x; break;
             case 3: written = ACCIDENTAL_WRITTEN_xs; break;
@@ -493,7 +506,7 @@ void Note::UpdateFromTransPitch(TransPitch tp)
     }
 
     if (this->GetOct() != tp.m_oct) {
-        if(this->HasOctGes()) {
+        if (this->HasOctGes()) {
             this->SetOctGes(this->GetOctGes() + tp.m_oct - this->GetOct());
         }
         this->SetOct(tp.m_oct);


src/note.cpp
```diff
@@ -459,6 +459,43 @@ TransPitch Note::GetTransPitch()
...
+            case -2: gestural = ACCIDENTAL_GESTURAL_ff; break;
+            case -1: gestural = ACCIDENTAL_GESTURAL_f; break;
+            case 0: gestural = ACCIDENTAL_GESTURAL_NONE; break;
+            case 1: gestural = ACCIDENTAL_GESTURAL_s; break;
+            case 2: gestural = ACCIDENTAL_GESTURAL_ss; break;
+            default: break; // TODO: Return an error here.
+        }
+        accid->SetAccidGes(gestural);
+    }
+    else {
```
earboxer: >  For @accid/@accid.ges, the output accid data must be placed into the same location from where it was extracted (this is important).

I think we should be changing both of these, because sometimes, both of them are given: 
```
<accid accid="s" accid.ges="s" enclose="paren" />
<accid accid="x" accid.ges="ss" />
<accid accid="f" accid.ges="f" />
```

src/note.cpp
```diff
@@ -459,6 +459,47 @@ TransPitch Note::GetTransPitch()
...
+    data_PITCHNAME pname = static_cast<data_PITCHNAME>(tp.m_pname + PITCHNAME_c);
+    this->SetPname(pname);
+
+    Accid *accid = this->GetDrawingAccid();
+    if (accid->HasAccidGes()) {
+        data_ACCIDENTAL_GESTURAL gestural = ACCIDENTAL_GESTURAL_NONE;
+        switch (tp.m_accid) {
+            case -2: gestural = ACCIDENTAL_GESTURAL_ff; break;
+            case -1: gestural = ACCIDENTAL_GESTURAL_f; break;
+            case 0: gestural = ACCIDENTAL_GESTURAL_NONE; break;
```
earboxer: Should probably be ACCIDENTAL_GESTURAL_n

src/note.cpp
```diff
@@ -459,6 +459,47 @@ TransPitch Note::GetTransPitch()
...
+        }
+        accid->SetAccidGes(gestural);
+    }
+    else {
+        data_ACCIDENTAL_WRITTEN written = ACCIDENTAL_WRITTEN_NONE;
+        switch (tp.m_accid) {
+            case -3: written = ACCIDENTAL_WRITTEN_tf; break;
+            case -2: written = ACCIDENTAL_WRITTEN_ff; break;
+            case -1: written = ACCIDENTAL_WRITTEN_f; break;
+            case 0: written = ACCIDENTAL_WRITTEN_NONE; break;
```
earboxer: I think if a **written** accidental was specified (a sharp, for example), we would want to write a natural if the transposed pitch had no Chromatic Alteration.

src/note.cpp
craigsapp: (1) and (2) have already been discussed, and that is correct.

(3) has two subcases: (3a) both are given and both are the same.  In this case, both should be changed to the same values (as appropriate for their data type).  The official view is that if both are the same, then only `@accid` should be given -- this is probably noted by the MEI Schematron validation. 
 (3b):  Both are present, but they are not the same.  There can be many reasons for this, some of which could be anticipated and some not.  I think a warning should be printed when this case occurs, and use the  `@accid.ges` value (one warning regardless of whether or note there are more than one occurrences of the situation).  The problem is that the `@accid` will not be in a modern convention, so how to change it will not be certain.  The main legitimate case is for 17th-century music (and earlier) where a sharp sign will be visually displayed for a note that would otherwise be flat, and the sharp raises the flat to a natural.  I think in this case the `@accid` should remain a sharp and no changed, but there may be other conventions that would cause problems.

A warning message should also be given (once) if a microtonal accidental is encountered (we can think about microtonal transposition after the basics for the standard system has been implemented).

for (4) there are also two subcases: (4a) the note contains an implied `@accid.ges="n"`.  In which case that value should be used as input to the transposition system.  If the transposed note output has a non-natural accidental, then insert an explicit `@accid.ges` entry for that note with the updated accidental.  If the accidental from the transposition remains natural, then leave `@accid.ges` as implicit. (4b) there is no visual accidental, but there should be a non-natural `@accid.ges`.  The consensus is that this is poorly-formed data, and in any case cannot be transposed properly until the `@accid.ges` parameters are filled in correctly (the input is garbage, and the output will be garbage).  

I think for both sub-cases of (4), a warning message should be printed (one for the entire process).  @rettinghaus  will complain about that because he likes implicit `@accid.ges="n"`, so any data that he produces would generate the warning.  I don't like implicit `@accid.ges="n"` because then it is difficult to distinguish between cases (4a) which is good, and (4b) which is very bad.  In any case, I think transposing `@accid.ges` is the proper thing to do in this case.  When the key signature is transposed along with the notes, there should be no migration of the accidental between `@accid` and `@accid.ges`.  In other words, there should not be new accidentals being displayed visually on notes where there was no visual accidental.  So, in this case, there was no visual accidental in the input, so there should be no visual accidental in the output.

src/note.cpp
craigsapp: Yes, I am pretty sure that a visual accidental will need to remain visible provided the key signature is transposed correctly at the same time.  If the key signature is not transposed by the same amount as the notes, then the accidentals can migrate between `@accid` and `@accid.ges` (and that is much more complicated to sort out, particularly with cross-staff notes and cautionary/courtesy accidentals that are not labeled as such).

src/note.cpp
craigsapp: Yes, the meaning of "0" as a value for `tp.m_accid` is "natural".

craigsapp: You might also want to add `ACCID_GESTURAL_fff` and `ACCID_GESTURAL_sss` to match the full range available in MEI (code excerpt might not show those if they are already there).

lpugin:
> 
> Eventually, we'll want a loop like
> 
> ```c++
> transposer->setTransposition("+M2")
> while (Transposable transposableItem = iterator.iterate()){
>   TransPitch pitch = transposableItem::GetTransPitch();
>   transposer->Transpose(pitch);
>   transposableItem::UpdateFromTransPitch(pitch);
> }
> ```
> 
> We would need to implement `Transposable`.
> 
> Then we could just implement these two methods for `Harm`, `StaffDef`, `ScoreDef`, `KeySig` (and any others relevant classes that we may have missed), to complete the system.

Not sure what you mean with this but in Verovio this is achieved with `Functor` objects. This is already in place for `Note` and `KeySig`. See https://github.com/rism-ch/verovio/blob/06e62ae2ce14b5011bb542d2c35726c72aa0194c/src/note.cpp#L1000

We will need to add `Harm`, but not `StaffDef` and `ScoreDef` because Verovio turns their key signature attribute into a `KeySig`

include/vrv/note.h
```diff
@@ -190,6 +190,7 @@ class Note : public LayerElement,
     int GetChromaticAlteration();
 
     TransPitch GetTransPitch();
+    void UpdateFromTransPitch(TransPitch tp);
```
lpugin: Since they will be called from `Note::Transpose` they should be made `private`

src/note.cpp
earboxer: [data.ACCIDENTAL.GESTURAL](https://music-encoding.org/guidelines/v4/data-types/data.accidental.gestural.html) does not support triple-flat or triple-sharp.

craigsapp: I added a request to add triple sharp/flats to`@accid.ges`, so eventually it may be possible: https://github.com/music-encoding/music-encoding/issues/639

src/note.cpp
earboxer: 3b: I think detecting mismatching accid and accid.ges is out of scope for this pull request. Implementing "transpose both" will have the effect of making 17th century music (and any other different conventions) follow modern conventions.

4: Makes sense. Transposed music should **not** retain implicit `@accid.ges="n"`, but I don't think a warning is necessary (or if so, it should be in `GetChromaticAlteration`, since it would also affect MIDI generation).

(updated in 3168822c)

include/vrv/note.h
earboxer: Oh, cool! (I guess I didn't notice that earlier).

Updated in 52cef660

src/note.cpp
```diff
@@ -493,7 +506,7 @@ void Note::UpdateFromTransPitch(TransPitch tp)
     }
 
     if (this->GetOct() != tp.m_oct) {
-        if(this->HasOctGes()) {
+        if (this->HasOctGes()) {
```
craigsapp: Just checking: do you have and/or know about clang-format, which will fix these sorts of problems automatically: https://electronjs.org/docs/development/clang-format

Usage:

```bash
clang-format -i file.cpp
```

where the `-i` option will edit the file in place.

earboxer: Yeah, I have a git pre-commit hook set up to run this for me over files that I'm committing. (but I believe clang-format doesn't give a different return value when it succeeds, so my pre-commit hook doesn't remind me to stage changes that it made).

earboxer: (I guess I should be using `git-clang-format`)

src/note.cpp
craigsapp: That is good for (4).

For (3b), I would transpose as requested but print a warning that both `@accid` and `@accid.ges` are used at the same time and are different values.  In such cases `@accid.ges` should be used for the transposition system.  The result will be correct sounding data, but the visual notation may be incorrect.  In the future, transposition can be tested on cases where the two attributes are present and not equal to see how `@accid` should be adjusted based on the transposition of `@accid.ges`.  In 17th-century music, the accidentals are usually treated as relative signs ("sharp" meaning raise by a semitone, rather than the modern absolute meaning of make the chromatic inflection of the diatonic value a sharp), so leaving a `@accid` as sharp when `@accid` goes from natural to sharp is the best thing to do in such cases.  But it gets complicated with how naturals are used, so correct transposition of the notation of 16th century and earlier would need more study of how the accidental signs behave when transposed.

lpugin:
Great!

lpugin:
@earboxer do you need guidance to implement the calls to the methods from the functors?

earboxer:
@lpugin Yeah, I've tried

```diff
diff --git a/src/doc.cpp b/src/doc.cpp
index 8b13e664..84bf34d0 100644
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -1158,7 +1158,7 @@ void Doc::ConvertAnalyticalMarkupDoc(bool permanent)
 void Doc::TransposeDoc()
 {
     Transposer transposer;
-    // here pass the value of this->m_options.m_transpose.GetValue() to the transposer
+    transposer.SetTransposition(this->m_options->m_transpose.GetValue());

     Functor transpose(&Object::Transpose);
     TransposeParams transposeParams(this, &transposer);
diff --git a/src/note.cpp b/src/note.cpp
index cf76a1e8..e197f7bf 100644
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -1053,6 +1053,10 @@ int Note::Transpose(FunctorParams *functorParams)

     LogDebug("Transposing note");

+    TransPitch pitch = this->GetTransPitch();
+    params->m_transposer->Transpose(pitch);
+    this->UpdateFromTransPitch(pitch);
+
     return FUNCTOR_SIBLINGS;
 }
```

But it always has a segmentation fault when I use the transpose flag (`--transpose M2`) (but works fine when I don't pass the transpose flag.

lpugin:
I cannot see anything wrong in your code. Make a PR and I'll check.

craigsapp:
> But it always has a segmentation fault when I use the transpose flag (--transpose M2) (but works fine when I don't pass the transpose flag.

Do you mean that transposition works when hard-coding the transposition (i.e., not loading the transposition from the options), or do you mean that verovio does not crash when there is no transposition being done?  Presumably the first case, as the second case would not be too surprising.

If it is the first case, then maybe print the input option string before setting the transposition (does accessing that command-line option cause the problem?) . If the command-line option string is good, then you could run `Transpose::getIntervalClass(string)` and "M2" should return "6" in the base-40 setting (I should add a `Transpose::getIntervalClass(void)` function to return the current transposition state of the object).

Otherwise, I would suspect that the params variable is not initialized.

lpugin:
It crashing in `note.cpp` line 470. You do not check if `accid` is not NULL

earboxer:
So we should add the accid if it doesn't exist.

earboxer:
I should probably figure out how to use a step-through debugger for verovio...

@lpugin What did you use to find that out so quickly? (maybe we can add instruction to the wiki https://github.com/rism-ch/verovio/wiki/Building-instructions page)

lpugin:
> I should probably figure out how to use a step-through debugger for verovio...
> 
> @lpugin What did you use to find that out so quickly? (maybe we can add instruction to the wiki https://github.com/rism-ch/verovio/wiki/Building-instructions page)

I simply use xCode. (But you can also use Visual Studio if you are on Windows, or GDB.)

craigsapp:
On MacOS, segmentation faults are reported in the directory `~/Library/Logs/DiagnosticReports`.  There will be a file with the name of the program, and a timestamp and the computer name, such as:

```
verovio_2019-11-21-095243_oznin.crash
```

Then there will be a section in that report which traces backwards in the function stack, listing at the top which function the segmentation fault occured in, and what were the recursive list of functions that called that function, such as:


```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   verovio                             0x00000001034285b0 hum::HumdrumFileContent::getBaselines(std::__1::vector<std::__1::vector<int, std::__1:
1   verovio                             0x0000000103428180 hum::HumdrumFileContent::analyzeKernStemLengths() + 80 (vector:650)
2   verovio                             0x000000010323d130 vrv::HumdrumInput::convertHumdrum() + 608 (iohumdrum.cpp:547)
3   verovio                             0x000000010323e346 vrv::HumdrumInput::ImportString(std::__1::basic_string<char, std::__1::char_traits<cha
4   verovio                             0x000000010338d64e vrv::Toolkit::LoadData(std::__1::basic_string<char, std::__1::char_traits<char>, std::
5   verovio                             0x000000010338cc14 vrv::Toolkit::LoadFile(std::__1::basic_string<char, std::__1::char_traits<char>, std::
6   verovio                             0x00000001031ad37b main + 6139 (main.cpp:451)
7   libdyld.dylib                       0x00007fff5f7293d5 start + 1
```

For this report, the segmentation fault occurred in the `hum::HumdrumFileContent::getBaselines()` function.  And this function was called from `hum::HumdrumFileContent::analyzeKernStemLengths()` which was in turn called from `vrv::HumdrumInput::convertHumdrum()`, etc.


In linux, gdb can be used to extract the list of function calls to the function that caused the problem. (`backtrace` is the main command that needs to be run after the program crashes in gdb).

PR #1227, opened 2019-12-11
WIP: * Note::Transpose: Implement this

earboxer:
This is WIP because I get a segmentation fault when trying to run this.

picking up discussion from: https://github.com/rism-ch/verovio/pull/1225#issuecomment-564596165

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -1158,7 +1158,7 @@ void Doc::ConvertAnalyticalMarkupDoc(bool permanent)
 void Doc::TransposeDoc()
 {
     Transposer transposer;
-    // here pass the value of this->m_options.m_transpose.GetValue() to the transposer
+    transposer.SetTransposition(this->m_options->m_transpose.GetValue());
 
     Functor transpose(&Object::Transpose);
     TransposeParams transposeParams(this, &transposer);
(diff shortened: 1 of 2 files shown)


earboxer:
@craigsapp 
> Do you mean that transposition works when hard-coding the transposition (i.e., not loading the transposition from the options), or do you mean that verovio does not crash when there is no transposition being done? Presumably the first case, as the second case would not be too surprising.

verovio does not crash when there is no transposition being done.

> If it is the first case, then maybe print the input option string before setting the transposition (does accessing that command-line option cause the problem?) . If the command-line option string is good, then you could run Transpose::getIntervalClass(string) and "M2" should return "6" in the base-40 setting (I should add a Transpose::getIntervalClass(void) function to return the current transposition state of the object).

I don't think it's accessing the input option string, as that seems to be parsed correctly by the Transposer.

```
$ ./verovio cml.mei
Output written to cml.svg.
$ ./verovio --transpose invalidtransposingstring cml.mei
Interval requires a diatonic interval number: invalidtransposingstring
Segmentation fault: 11
$ ./verovio --transpose M2 cml.mei
Segmentation fault: 11
```


> Otherwise, I would suspect that the params variable is not initialized.

How do we check this? It looks like it should be right...

PR #1228, opened 2019-12-11
* Note: Create Accidental if it doesn't exist when transposing

earboxer:
Fixes the segfault as noted in https://github.com/rism-ch/verovio/pull/1225#issuecomment-564598449.

(So we *start* to have a working transposition system).

### DIFF
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -467,6 +467,10 @@ void Note::UpdateFromTransPitch(TransPitch tp)
     Accid *accid = this->GetDrawingAccid();
     bool transposeGesturalAccid = false;
     bool transposeWrittenAccid = false;
+    if (!accid) {
+        accid = new Accid();
+        this->AddChild(accid);
+    }
     if (accid->HasAccidGes()) {
         transposeGesturalAccid = true;
     }


src/note.cpp
```diff
@@ -467,6 +467,10 @@ void Note::UpdateFromTransPitch(TransPitch tp)
     Accid *accid = this->GetDrawingAccid();
     bool transposeGesturalAccid = false;
     bool transposeWrittenAccid = false;
+    if (!accid) {
```
lpugin: We probably want to add it only if necessary, don't we? So maybe first see boolean values only if `accid` exists, and then add it if necessary below. Would this work?

earboxer: From https://github.com/rism-ch/verovio/pull/1225#discussion_r355787424, we talked about what to do if there was no accidental given, and decided to always include the `@accid.ges="n"` in transposed scores. (So, we're always going to need `accid`).

earboxer: (With the current implementation, the only cases where this will add an empty `<accid>` object is with microtones and invalid transpositions)

craigsapp: Yes, the rule should be (1) never create `note@accid` (or equivalent in a child accid element), (2) never create `note@accid.ges` if `note@accid` is present, and  (3) only create `note@accid.ges` when there is an implicit `note@accid.ges="n"` and the implicit `n` changes to some other accidental (in which case the `note@accid.ges` value needs to become explicit).

lpugin: Ready to merge, then?

PR #1235, opened 2019-12-13
* Transposition: Transpose simple chord symbols in Harm elements

earboxer:
This finishes the basic functionality for transposition by interval.

test with a file like [71d-ChordsFrets-Multistaff.xml](https://github.com/cuthbertLab/musicxmlTestSuite/blob/master/xmlFiles/71d-ChordsFrets-Multistaff.xml)

Screen Shot 2019-12-13 at 17 37 31

transposed up a Major 3rd:

Screen Shot 2019-12-13 at 17 38 47

### DIFF
--- a/include/vrv/harm.h
+++ b/include/vrv/harm.h
@@ -72,6 +72,11 @@ public:
      */
     virtual int AdjustHarmGrpsSpacing(FunctorParams *functorParams);
 
+    /**
+     * See Object::Transpose
+     */
+    virtual int Transpose(FunctorParams *functorParams);
+
 protected:
     //
 private:
--- a/include/vrv/harm.h
+++ b/include/vrv/harm.h
@@ -11,6 +11,7 @@
 #include "controlelement.h"
 #include "textdirinterface.h"
 #include "timeinterface.h"
+#include "transposition.h"
 
 namespace vrv {
 
@@ -57,6 +58,12 @@ public:
      * Only supported elements will be actually added to the child list.
      */
     virtual void AddChild(Object *object);
+
+    /**
+     * Transposition related. The int tracks where we have iterated through the string.
+     */
+    TransPitch GetRootPitch(int &pos);
+    void SetRootPitch(TransPitch pitch, int endPos);
 
     //----------//
     // Functors //
--- a/include/vrv/harm.h
+++ b/include/vrv/harm.h
@@ -62,10 +62,10 @@ public:
     /**
      * Transposition related. The int tracks where we have iterated through the string.
      */
-    TransPitch GetRootPitch(unsigned int &pos);
-    void SetRootPitch(TransPitch pitch, unsigned int endPos);
-    TransPitch GetBassPitch();
-    void SetBassPitch(TransPitch pitch);
+    bool GetRootPitch(TransPitch &pitch, unsigned int &pos);
+    void SetRootPitch(const TransPitch &pitch, unsigned int endPos);
+    bool GetBassPitch(TransPitch &pitch);
+    void SetBassPitch(const TransPitch &pitch);
 
     //----------//
     // Functors //
--- a/include/vrv/harm.h
+++ b/include/vrv/harm.h
@@ -62,8 +62,10 @@ public:
     /**
      * Transposition related. The int tracks where we have iterated through the string.
      */
-    TransPitch GetRootPitch(int &pos);
-    void SetRootPitch(TransPitch pitch, int endPos);
+    TransPitch GetRootPitch(unsigned int &pos);
+    void SetRootPitch(TransPitch pitch, unsigned int endPos);
+    TransPitch GetBassPitch();
+    void SetBassPitch(TransPitch pitch);
 
     //----------//
     // Functors //
--- a/src/harm.cpp
+++ b/src/harm.cpp
@@ -79,7 +79,7 @@ void Harm::AddChild(Object *child)
     Modify();
 }
 
-TransPitch Harm::GetRootPitch(unsigned int &pos)
+bool Harm::GetRootPitch(TransPitch &pitch, unsigned int &pos)
 {
     Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
     assert(textObject);
@@ -100,14 +100,14 @@ TransPitch Harm::GetRootPitch(unsigned int &pos)
             else
                 break;
         }
-        return TransPitch(pname, accid, 4);
+        pitch = TransPitch(pname, accid, 4);
+        return true;
     }
     LogWarning("Failed to extract a pitch.");
-    // If we return octave -1, it knows it is an error.
-    return TransPitch(0, 0, -1);
+    return false;
 }
 
-void Harm::SetRootPitch(TransPitch pitch, unsigned int endPos)
+void Harm::SetRootPitch(const TransPitch &pitch, unsigned int endPos)
 {
     Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
     assert(textObject);
@@ -116,7 +116,7 @@ void Harm::SetRootPitch(TransPitch pitch, unsigned int endPos)
     textObject->SetText(pitch.GetPitchString() + &text[endPos]);
 }
 
-TransPitch Harm::GetBassPitch()
+bool Harm::GetBassPitch(TransPitch &pitch)
 {
     Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
     assert(textObject);
@@ -124,13 +124,13 @@ TransPitch Harm::GetBassPitch()
     for (unsigned int pos = 0; pos < text.length(); pos++) {
         if (text.at(pos) == L'/') {
             pos++;
-            return GetRootPitch(pos);
+            return GetRootPitch(pitch, pos);
         }
     }
-    return TransPitch(0, 0, -1);
+    return false;
 }
 
-void Harm::SetBassPitch(TransPitch pitch)
+void Harm::SetBassPitch(const TransPitch &pitch)
 {
     Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
     assert(textObject);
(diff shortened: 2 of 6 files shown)


craigsapp:
I am wondering about the meaning of `Cm7n11` and its transposition up a major third to `Em7n11`.  The question is about the natural sign for the 11th.  For `Cm7n11` that means the pitches `C E G Bb F`, and transposed, that means `E G# B D A` so `Em7n11` is definitely correct.  But I am wondering what is the chord symbol if transposed up a minor third.  That would be the pitches `Eb G Bb Db Ab` so the question is: is `Ab` represented by `n11` or by `b11`: `Ebm7n11` or `Ebm7b11`?

src/harm.cpp
```diff
@@ -209,4 +210,47 @@ int Harm::AdjustHarmGrpsSpacing(FunctorParams *functorParams)
...
 
+int Harm::Transpose(FunctorParams *functorParams)
+{
+    TransposeParams *params = dynamic_cast<TransposeParams *>(functorParams);
+    assert(params);
+
+    LogDebug("Transposing harm");
+
+    Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
+    std::wstring text = textObject->GetText();
```
lpugin: Always `assert` the result of a dynamic _cast when something should be return. In cases where it is not sure (such as this one), handle it properly.

src/harm.cpp
```diff
@@ -209,4 +210,47 @@ int Harm::AdjustHarmGrpsSpacing(FunctorParams *functorParams)
...
+
+    // Transpose chord names: "C".
+    if (text[0] >= 'A' && text[0] <= 'G') {
```
lpugin: Always use `.at` operators for vector element (or string character) access

src/harm.cpp
```diff
@@ -209,4 +210,47 @@ int Harm::AdjustHarmGrpsSpacing(FunctorParams *functorParams)
...
+        int i;
+        for (i = 1;; i++) {
+            switch (text[i]) {
+                case L'𝄫': accid -= 2; break;
+                case 'b':
+                case L'♭': accid--; break;
+                case '#':
+                case L'♯': accid++; break;
+                case L'𝄪': accid += 2; break;
+                default: goto break2;
```
lpugin: Please make sure there is no use of `goto`. Also, it is not a good idea to loop over the characters of the `text` variable and to change it within the loop. This needs to be refactored.

src/harm.cpp
```diff
@@ -209,4 +210,47 @@ int Harm::AdjustHarmGrpsSpacing(FunctorParams *functorParams)
...
+    LogDebug("Transposing harm");
+
+    Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
+    std::wstring text = textObject->GetText();
+
+    // Transpose chord names: "C".
+    if (text[0] >= 'A' && text[0] <= 'G') {
+        int accid = 0;
+        int i;
+        for (i = 1;; i++) {
```
lpugin: I don 't like seeing a loop without an end condition. Please refactor it.

src/harm.cpp
earboxer: I wanted to use `break 2;`, but C++ doesn't support that. https://stackoverflow.com/questions/1257744/can-i-use-break-to-exit-multiple-nested-for-loops recommends goto. (Also the text variable is not being changed within the loop).

Since this was confusing, I agree that it should be refactored.

earboxer:
@craigsapp

From the musicXML spec:

> If the degree-type is
    add, the degree-alter is relative to a dominant chord
    (major and perfect intervals except for a minor
    seventh).

So, we don't want to change the degree-alter, since it is relative to the interval.

Music21's midi conversion also works this way:

Ebm7add11 (Musescore did not understand Ebm7natural11) Music21 produces midi Eb Gb Bb Db Ab
```
      <harmony print-frame="no">
        <root>
          <root-step>E</root-step>
          <root-alter>-1</root-alter>
          </root>
        <kind text="m7">minor-seventh</kind>
        <degree>
          <degree-value>11</degree-value>
          <degree-alter>0</degree-alter>
          <degree-type>add</degree-type>
          </degree>
        </harmony>
```

We may want to consider changing musicXML parsing to call chord Cm7add11 instead of Cm7n11.
Nothing special needs to be done for these in the transposition system.

craigsapp:
> So, we don't want to change the degree-alter, since it is relative to the interval.

That is good, and makes it easier to transpose...

src/harm.cpp
earboxer: Does anything more need to be done besides `assert(textObject)`?

src/transposition.cpp
```diff
@@ -161,6 +161,19 @@ data_PITCHNAME TransPitch::GetPitchName()
...
+std::wstring TransPitch::GetPitchString()
+{
+    char pitchLetter = (m_pname + ('C' - 'A')) % 7 + 'A';
+    switch (m_accid) {
+        case -2: return std::wstring({ pitchLetter, L'𝄫' });
+        case -1: return std::wstring({ pitchLetter, L'♭' });
+        default: LogError("Transposition: Could not get Accidental for %i", m_accid);
+        case 0: return std::wstring({ pitchLetter });
+        case 1: return std::wstring({ pitchLetter, L'♯' });
+        case 2: return std::wstring({ pitchLetter, L'♯', L'♯' });
```
earboxer: When I tried to use the `𝄪` character, it would come out as `턪`.

craigsapp: That most likely means that it would need to be added to the verovio text font in order to be viewed as a double sharp.

craigsapp: But you are printing as text to standard output (or Javascript console), in which case it will more depend on the character encoding of the device, which is often now UTF-8 but not always.

Here are the encodings of double-sharp:
https://www.fileformat.info/info/unicode/char/1d12a/index.htm
Maybe using one of these might work.  Otherwise, I would just print "double-sharp" instead of the symbol since it is likely to get messed up regardless 😛

earboxer: I removed the code that printed this out when there was an error. (It wasn't able to print the wstring, and `.c_str()` got rid of the wide characters, so it wasn't useful).

src/harm.cpp
lpugin: Sorry, I was not clear. In this case, you need to exit if it is NULL. This should not happen, but it can. If `<harm>` is empty, for whatever reason, there will be no text child. (`assert` should be used to check that nothing went wrong but you expect this never to happen.)

src/harm.cpp
```diff
@@ -78,6 +79,73 @@ void Harm::AddChild(Object *child)
...
+    return TransPitch(0, 0, -1);
+}
+
+void Harm::SetRootPitch(TransPitch pitch, unsigned int endPos)
+{
+    Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
+    assert(textObject);
+    std::wstring text = textObject->GetText();
+
+    textObject->SetText(pitch.GetPitchString() + &text[endPos]);
```
lpugin: Access to character in `text` is not safe. Please change to `.at()` operator and make sure it is not out of range.

src/harm.cpp
```diff
@@ -78,6 +79,73 @@ void Harm::AddChild(Object *child)
     Modify();
 }
 
+TransPitch Harm::GetRootPitch(unsigned int &pos)
+{
+    Text *textObject = dynamic_cast<Text *>(this->GetChild(0, TEXT));
```
lpugin: Please change method to retrieve the child to `this->FindDescendantByType(TEXT, 1)` for consistency.

include/vrv/harm.h
```diff
@@ -58,6 +59,14 @@ class Harm : public ControlElement,
      */
     virtual void AddChild(Object *object);
 
+    /**
+     * Transposition related. The int tracks where we have iterated through the string.
+     */
+    TransPitch GetRootPitch(unsigned int &pos);
```
lpugin: For this method and `GetBassPitch` it would be clearer to pass TransPitch as parameter (`const TransPitch& pitch`) and to return a boolean indicating success and failure.

include/vrv/harm.h
```diff
@@ -58,6 +59,14 @@ class Harm : public ControlElement,
...
+    void SetRootPitch(TransPitch pitch, unsigned int endPos);
```
lpugin: Change `TransPitch pitch` to `const TransPitch& pitch`

src/harm.cpp
```diff
@@ -78,6 +79,73 @@ void Harm::AddChild(Object *child)
...
+    assert(textObject);
+    std::wstring text = textObject->GetText();
+
+    if (text.at(pos) >= 'A' && text.at(pos) <= 'G') {
```
lpugin: This will crash if `text` is an empty. Also, since you are retrieving the Text child in each method, it would be better to simply get it in `Harm::Transpose`. Then check that 1) it is not NULL, and 2) not empty. Then you can pass the `text` variable as parameter to the four methods, without actually setting it back in them but only at the end of `Harm::Transpose`. This means without calling `SetText` each time. This would make the four methods static-like methods, but I think it would be fine to leave them as they are. Alternatively, they could be moved to the Text class. This would be more correct in terms of classes encapsulation. However, having the transpose related methods in Harm makes it somehow clearer. I am fine either way.

src/harm.cpp
earboxer: Okay, I think I've fixed this. Though it's crashing elsewhere (without transposing) when given an empty harm:

gdb says

```
Thread 2 received signal SIGSEGV, Segmentation fault.
0x0000000100074bd1 in vrv::FloatingCurvePositioner::FloatingCurvePositioner (
    this=<optimized out>, object=<optimized out>, alignment=<optimized out>,
    spanningType=<optimized out>)
    at verovio/src/floatingobject.cpp:379
379	}
```

(But I don't really see what the problem is there...)

src/harm.cpp
earboxer: I changed it to check text.length() before any accesses. I did not refactor the textObject retrieval.

PR #1238, opened 2019-12-16
Added macOS support for CocoaPods

vitalii-unicsoft:
This small change allows Verovio framework to be built on macOS with CocoaPods.


lpugin:
@Noroxs does it work for you?

valeriyvan:
> @Noroxs does it work for you?

I could answer instead of @Noroxs. Yes, it does work.

Noroxs:
Thank you @valeriyvan. I have currently no MacBook available so I couldn't test.

PR #1242, opened 2019-12-18
Transpose based on tonic note

earboxer:
Implements transposition by tonic based on https://github.com/rism-ch/verovio/issues/1189#issuecomment-563405899 with the following caveat:

* When you transpose by tonic, it will still output a message to stderr about the failure to parse the interval string.


src/transposition.cpp
```diff
@@ -1090,6 +1169,17 @@ std::string Transposer::CircleOfFifthsToIntervalName(int fifths)
...
 
+//////////////////////////////
+//
+// Transposer::CircleOfFifthsToPitch -- Assuming the mode is major, guess the pitch
+// from the circle of fifths position.
+//
+TransPitch Transposer::CircleOfFifthsToPitch(int fifths)
+{
+    int intervalClass = CircleOfFifthsToIntervalClass(fifths);
+    return IntegerToPitch(GetMaxAccid() + 1 + intervalClass);
```
earboxer: @craigsapp There's probably a better way of doing this. (I think this is needed because `0` is equivalent to Cbbb instead of C).

earboxer:
This uses `TransPitch` to represent the transposition by tonic:

|tonic transposition string|TransPitch|
|---|---|
|`-C` | **C-1**|
|`C` | **C0**|
|`+C` | **C1**|
|`++C` | **C2**|


We will need to change the octave so that it takes into account where you're coming from:

E.g. If you're coming from **D0**, then both `-C`  and `C` should result in **C0**, `--C` being **C-1**.

src/doc.cpp
```diff
@@ -1158,7 +1158,23 @@ void Doc::ConvertAnalyticalMarkupDoc(bool permanent)
 void Doc::TransposeDoc()
 {
     Transposer transposer;
-    transposer.SetTransposition(this->m_options->m_transpose.GetValue());
+    bool setTransp = transposer.SetTransposition(this->m_options->m_transpose.GetValue());
+    if (!setTransp) {
+        // Set transposition by key tonic.
+        // Detect the current key from the keysignature.
+        KeySig *keysig = dynamic_cast<KeySig *>(this->FindDescendantByType(KEYSIG, 9));
```
lpugin: This is the right approach, but you need to look for it in the Doc::m_scoreDef. That is, doing `this->m_scoreDef.FindDescendantByType(KEYSIG, 1)`. (I would look only in the first level, because we want to find the scoreDef@key.sig. (If there are different on staffDefs, this is a different issue, I guess)

earboxer: I changed it to `this->m_scoreDef.FindDescendantByType(KEYSIG, 3)`, as the musicXML import places the keySig inside a StaffDef. (I'm testing with [13b-KeySignatures-ChurchModes.xml](https://github.com/cuthbertLab/musicxmlTestSuite/blob/master/xmlFiles/13b-KeySignatures-ChurchModes.xml)).

```xml
<music>
<body>
  <mdiv xml:id="mdiv-0000002032342191">
    <score xml:id="score-0000001847798602">
      <scoreDef xml:id="scoredef-0000000735621517">
        <staffGrp xml:id="staffgrp-0000000088293271">
          <staffDef xml:id="staffdef-0000000842034478" n="1" lines="5" ppq="1" clef.shape="G" clef.line="2" meter.count="1" meter.unit="4">
            <keySig xml:id="keysig-0000000350207968" mode="major" sig="2s" />
```

If we guaranteed the keySig (or key.sig) was in the scoreDef, we could do

```c++
KeySig *keysig = this->m_scoreDef->GetKeySig();
```

src/transposition.cpp
craigsapp: You are correct: C is not the zero point in the octave cycle, but rather the flattest C possible in the given base.  This is to allow easier calculating of the octave as compared to setting C-natural to be the zero point (Cb4 is enharmonically equivalent to B3, Cbb4 is enharmonically equivalent to Bb3).  So as you notice, the interval class and the pitch class integers are shifted from each other (P1 = 0, but C = max_accid).

However, I do not think the `+ 1` is needed: In Base-40: Cbb=0, Cb=1, C=2.  And in base-40 the maximum accidental is 2, so the C pitch class in any base is the maximum accidental count, not the maximum accidental plus one.

craigsapp: Also, the assumption about the mode being major is not necessary (or used in your code): the transposition is calculated from the interval class between the pitch classes of the tonics of the two keys.   The tonic is independent of the mode (which instead describes the pitch classes of other notes in the scale).  In other words, in C major, there is a C-natural, and in C minor, there is a C-natural, and in C dorian, there is a C-natural, and in C create-your-own-mode, there is a C natural.  Only for non-tonic notes can the accidental state of the notes change: in C major, there is an E-natural, but in C minor, there is an E-flat.

earboxer: When I don't use the `+ 1`, the pitch comes out flat.

From printing out some debug output, it seems like `IntegerToPitch` (or `GetMaxAccid`) is not behaving as you explained: `GetMaxAccid()` is returning 2, but `IntegerToPitch(0)` is returning `Cbbb0` (triple-flat)

```c++
    Transposer transposer;
    bool setTransp = transposer.SetTransposition(this->m_options->m_transpose.GetValue());
    std::cerr << "transposer.GetMaxAccid(): ";
    std::cerr << transposer.GetMaxAccid() << "\n";
    std::cerr << "transposer.IntegerToPitch(0): ";
    std::cerr << transposer.IntegerToPitch(0) << "\n";
    std::cerr << "transposer.IntegerToPitch(1): ";
    std::cerr << transposer.IntegerToPitch(1) << "\n";
    std::cerr << "transposer.IntegerToPitch(2): ";
    std::cerr << transposer.IntegerToPitch(2) << "\n";
    std::cerr << "transposer.IntegerToPitch(3): ";
    std::cerr << transposer.IntegerToPitch(3) << "\n";
```

```
transposer.GetMaxAccid(): 2
transposer.IntegerToPitch(0): Cbbb0
transposer.IntegerToPitch(1): Cbb0
transposer.IntegerToPitch(2): Cb0
transposer.IntegerToPitch(3): C0
```

craigsapp: Hmmm, I am probably subtracting one incorrectly somewhere in the Transpose class: I will check it out.

earboxer: It looks like it needs to be shifted back one, since B## is not accessible in base-40 (The off-by-one issue also applies to base-600)

```
transposer.IntegerToPitch(0): Cbbb0
transposer.IntegerToPitch(1): Cbb0
transposer.IntegerToPitch(2): Cb0
transposer.IntegerToPitch(3): C0
transposer.IntegerToPitch(4): C#0
transposer.IntegerToPitch(5): C##0
transposer.IntegerToPitch(6): C###0
transposer.IntegerToPitch(7): Dbb0
transposer.IntegerToPitch(8): Db0
transposer.IntegerToPitch(9): D0
transposer.IntegerToPitch(10): D#0
transposer.IntegerToPitch(11): D##0
transposer.IntegerToPitch(12): D###0
transposer.IntegerToPitch(13): Ebb0
transposer.IntegerToPitch(14): Eb0
transposer.IntegerToPitch(15): E0
transposer.IntegerToPitch(16): E#0
transposer.IntegerToPitch(17): E##0
transposer.IntegerToPitch(18): Fbb0
transposer.IntegerToPitch(19): Fb0
transposer.IntegerToPitch(20): F0
transposer.IntegerToPitch(21): F#0
transposer.IntegerToPitch(22): F##0
transposer.IntegerToPitch(23): Gbbb0
transposer.IntegerToPitch(24): Gbb0
transposer.IntegerToPitch(25): Gb0
transposer.IntegerToPitch(26): G0
transposer.IntegerToPitch(27): G#0
transposer.IntegerToPitch(28): G##0
transposer.IntegerToPitch(29): Abbb0
transposer.IntegerToPitch(30): Abb0
transposer.IntegerToPitch(31): Ab0
transposer.IntegerToPitch(32): A0
transposer.IntegerToPitch(33): A#0
transposer.IntegerToPitch(34): A##0
transposer.IntegerToPitch(35): Bbbb0
transposer.IntegerToPitch(36): Bbb0
transposer.IntegerToPitch(37): Bb0
transposer.IntegerToPitch(38): B0
transposer.IntegerToPitch(39): B#0
transposer.IntegerToPitch(40): Cbbb1
transposer.IntegerToPitch(41): Cbb1
transposer.IntegerToPitch(42): Cb1
```

src/transposition.cpp
```diff
@@ -281,9 +281,21 @@ void Transposer::SetTransposition(int transVal)
...
 {
     m_transpose = GetIntervalClass(transString);
+    return m_transpose != INVALID_INTERVAL_CLASS;
+}
+
+bool Transposer::SetTransposition(const TransPitch &fromPitch, const std::string &toString)
+{
+    TransPitch toPitch;
+    if (GetKeyTonic(toString, toPitch)) {
+        // TODO: Determine proper octave offset.
```
earboxer: Thinking about the logic for correcting the octave,

assuming fromPitch is octave 0 (writing **F0** as *F*)

* Transposing from *F* to `G`: no change needed (**G0** is the closest G)
* Transposing from *F* to `+G`: subtract one octave (the first `+` determines the direction) (for all other `+++G`s, also subtract one octave)
* Transposing from *F* to `-G`: no change needed (**G-1** is the closest downward G)

* Transposing from *G* to `F`: no change needed (**F0** is the closest F)
* Transposing from *G* to `+F`: no change needed (**F1** is the closest upward F)
* Transposing from *G* to `-F`: add one octave (so it gives the same result as `F`)

* Transposing from *C* to `B`: Add one octave (**B1** is the closest B)
* Transposing from *C* to `+B`: Do nothing (already was giving **B1**)
* Transposing from *C* to `-B`: Add one octave (**B0** is the closest downward B)

* Transposing from *B* to `C`: Subtract one octave (**C-1** is the closest C)
* Transposing from *B* to `+C`: Subtract one octave (**C0** is the closest upward C)
* Transposing from *B* to `-C`: do nothing (**C-1** is already the closest downward C)

Probably would be good to think of a way to simplify the logic here without needing 12 cases

src/transposition.cpp
craigsapp: The problem should be fixed with commit https://github.com/rism-ch/verovio/pull/1242/commits/2a621aba9148db9236e791cd1e508e931e6a1504

When I was calculating the interval classes of the diatonic pitches, I started with C = maxAccid + 1, where it should have been C = maxAccid.  In other words, I did the same thing that I was complaining about in your code (so you had to add one to fix the problem).

I will fix the +1 that I was complaining about in another commit.

earboxer: Got it in 1d2c4cc4809dbd2d2497c06770be32294f1c3786

craigsapp: But you beat me to it.

craigsapp: I recommend changing the name of `Transposer::IntegerToPitch()` to `Transposer::PitchClassToTransPitch()`.  This will help with possible confusion between interval classes and pitch classes: both of these are integers.  The name `IntegerToPitch` might imply to someone that both interval class or pitch class integers can be given to the function, but the two are not interchangeable (particularly due to the offset added to pitch classes).

src/transposition.cpp
craigsapp: I don't understand this case:

> Transposing from F to +G: subtract one octave (the first + determines the direction) (for all other +++Gs, also subtract one octave)

Here is how I am interpreting the pluses and the output octave for each cases:

F to G is a major second, so F0 to G0.
F to +G is a major second, so F0 to G0.
F to ++G is a major second and an octave, so F0 to G1.
F to +++G is a major second and two octaves, to F0 to G2.

earboxer: By `subtract one octave`, I guess I mean compared to the offset that GetKeyTonic gives:

`+G` = **G1**, so we need to subtract an octave from this so it become **G0**, as we expect.

earboxer: Maybe a better way of saying it (that might help with the logic):

A transposition with **_n_** plus `+` or minus `-` signs should never be more than **_n_** octaves away.

earboxer: And a transposition with **_0_** plus or minus signs should never be more than 1/2 an octave away.

craigsapp: `+G` cannot be interpreted as an absolute case since it is relative to the starting key.  `+G` will always be up a particular interval which is an octave or less.

The algorithm that I am thinking when there are pluses present:

From the key of F to `+G` for example:

(1) Start with F0 and G0 (both pitches in the 0th octave) and convert to both to integer pitches.
(2) if G0 is a smaller number than F0, add m_base to G0 
       (this is not necessary for the example case, but in the general case this will be possible such as when transposing from G up to F).
(3) for each left-over `+` sign, also add m_base to the target key's pitch interval.
(4) calculate the interval by subtracting F0 from the octave-adjusted G0 pitch.

Here is the reverse upward case: from G to +F:
(1) start with G0 and F0.
(2) F0 is smaller than G0, so add an octave to move it to F1.
(3) There are no more plus signs, so F1 no longer needs to be adjusted.
(4) Calculate the interval class: F1 - G0, which is up a minor seventh.


For the forced down cases:

For example from G to `-F`:

(1) Start with G0 and F0
(2) F0 is  less than G0, so no octave adjustment is necessary
(3) there are no extra `-` signs, so no further octave adjustment
(4) F0 - G0 = down a major second

Transpose from F major down to G major:

(1) Start with F0 and G0
(2) G0 is greater than F0, so change to G-1
(3) There are not extra `-` signs to no extra octave adjustment
(4) G-1 - F0 = down a minor seventh

------------

For the case where there is neither `+` nor  `-`, then the algorithm has to check which direction has the smallest absolute interval:

from key X to key Y:

(1) calculate the interval pitches for X0 and Y0
(2) If (Y0 - X0) is positive then compare to abs(Y0 - X-1) and choose whichever of X0 and X-1 gives the smallest value.
(2) If (Y0 - X0) is negative, then compare to abs(Y0 - X1) and choose whichever of X0 and X1 gives the smallest value.


So there are three cases to implement: When there are pluses, when there are minuses, and when there are neither pluses or minuses.  The first two cases could be merged into the same code, but it is probably clearer to keep them separate.

earboxer: I implemented it in c1b7f8e745989c13e8ec7f20a0ec8c0a336f1f31

I'm using four cases (I've split the case with neither pluses nor minuses into two) It could be simplified into an `if` and an `else if`, but right now I think it's pretty understandable.

earboxer:
@craigsapp

If you want to [Rename Transposer::IntegerToPitch()](https://github.com/rism-ch/verovio/pull/1242#discussion_r359953601), [Amend the comment about mode assumption](https://github.com/rism-ch/verovio/pull/1242#discussion_r359905270), and add any clarifying comments, feel free to commit them. (I'm resolving the discussions where these were mentioned).

I think at this point, the code is working as I expect it to.

src/transposition.cpp
craigsapp: Actually changed `IntegerToPitch` to be `IntegerPitchToTransPitch`, since it is not a pitch class but rather a pitch (includes octave).

earboxer:
Changes look good!

PR #1250, opened 2020-01-06
Fixed: Ahle verse numbers

ahankinson:
The verse numbers were being unnecessarily repeated. 

Also, I'm not sure why I keep only editing this file once per year. I guess it's true what they say -- "It's Ahle or nothing" 😁


craigsapp:
Next year you can fix the hyphens, and the year after you can fix the numbers to encode like this:

```xml
        <note xml:id="d23378e177" pname="e" oct="4" dur="4" stem.dir="down">
                              <verse n="1">
                              <label>1.</label>
                                 <syl>Je-</syl>
                              </verse>
                              <verse n="2">
                                 <label>2.</label>
                                 <syl>Tau-</syl>
                              </verse>
                              <verse n="3">
                                 <label>3.</label>
                                 <syl>Wei-</syl>
                              </verse>
                           </note>
```

(See https://github.com/rism-ch/verovio/issues/913)

PR #1259, opened 2020-01-14
Develop expand option for expansion element (and making sections invisble).

wergo:
I have added the removing functionality by creating a list of strings `reductionList` (containing all relevant siblings of the initial expansion element) which gets reduced for every element in `expansionList`(or for its parent when itself an `<expansion>`). In the end, all remaining elements in `reductionList` will be set invisible (line to be inserted).

### DIFF
--- a/src/expansionmap.cpp
+++ b/src/expansionmap.cpp
@@ -5,6 +5,8 @@
 // Copyright (c) Authors and others. All rights reserved.
 /////////////////////////////////////////////////////////////////////////////
 
+#include <iostream>
+
 #include "expansionmap.h"
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/expansionmap.cpp
```diff
@@ -33,17 +33,29 @@ void ExpansionMap::Reset()
 void ExpansionMap::Expand(const xsdAnyURI_List &expansionList, xsdAnyURI_List &existingList, Object *prevSect)
 {
     assert(prevSect);
-    
+    // find all siblings of expansion element to know what in MEI file
+    std::vector<std::string> reductionList;
+    for (auto o : prevSect->GetParent()->GetChildren()) {
```
lpugin: assert(prevSect->GetParent())

wergo: I have added this.

PR #1293, opened 2020-02-01
minor improvements on MusicXML import

rettinghaus:


src/iomusxml.cpp
```diff
@@ -1983,7 +1985,11 @@ void MusicXmlInput::ReadMusicXmlNote(pugi::xml_node node, Measure *measure, std:
         if (HasAttributeWithValue(node, "print-object", "no")) {
             Space *space = new Space();
             element = space;
-            space->SetDur(ConvertTypeToDur(typeStr));
+            if (!typeStr.empty()) {
+                space->SetDur(ConvertTypeToDur(typeStr));
+            }
+            // this should be mSpace
+            else space->SetDur(DURATION_1);
```
craigsapp: What are the implications of this?  If there is no type, then the space should be converted to be four quarter notes in duration?  And how does that related to other time signatures such as 2/4 or 7/8?

Is it that SetDur will be the visual display of the rest and SetDurPpq will be the actual duration?

rettinghaus:
This should be mSpace, but that is not yet available in Verovio.

PR #1307, opened 2020-02-17
Added adjustPageWidth option

palgranum:
This is a suggested implementation of the adjust page width option:
https://github.com/rism-ch/verovio/issues/1276

This is being referred to as cropping, but in my opinion it is equally useful (like adjustPageHeight) for extending the width. I might be missing something fundamental, but I can't see any other reliable way of getting acceptable layout when breaks are encoded. Otherwise one have to do a magic guess for a suitable width that fits the content, which probably will lead to compressed content or blank trailing space.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -462,6 +462,7 @@ public:
     OptionGrp m_general;
 
     OptionBool m_adjustPageHeight;
+    OptionBool m_adjustPageWidth;
     OptionIntMap m_breaks;
     OptionBool m_condenseEncoded;
     OptionBool m_condenseFirstPage;
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -494,9 +494,13 @@ Options::Options()
     m_general.SetLabel("Input and page layout options", "1-general");
     m_grps.push_back(&m_general);
 
-    m_adjustPageHeight.SetInfo("Adjust page height", "Crop the page height to the height of the content");
+    m_adjustPageHeight.SetInfo("Adjust page height", "Adjust the page height to the height of the content");
     m_adjustPageHeight.Init(false);
     this->Register(&m_adjustPageHeight, "adjustPageHeight", &m_general);
+
+    m_adjustPageWidth.SetInfo("Adjust page width", "Adjust the page width to the width of the content");
+    m_adjustPageWidth.Init(false);
+    this->Register(&m_adjustPageWidth, "adjustPageWidth", &m_general);
 
     m_breaks.SetInfo("Breaks", "Define page and system breaks layout");
     m_breaks.Init(BREAKS_auto, &Option::s_breaks);
--- a/src/page.cpp
+++ b/src/page.cpp
@@ -476,6 +476,11 @@ void Page::JustifyHorizontally()
     // Make sure we have the correct page
     assert(this == doc->GetDrawingPage());
 
+    if ((doc->GetOptions()->m_adjustPageWidth.GetValue()))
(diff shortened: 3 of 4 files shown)


lpugin:
Looks good. Thanks

craigsapp:
Looks good.  Here is an example of loading an SVG image created with the option into Microsoft Word:

Screen Shot 2020-02-17 at 09 46 21

Comparison to not using `--adjust-page-width` is given in the second image.  In this case the full page width is in the VewBox of the SVG image so the bounding box of the image in Word is wide.

samuelbradshaw:
Awesome! Is there a way to remove the margins entirely (there's still a half-inch margin on all sides)?

craigsapp:
Here are all of the page margin options:

```
 --page-margin-bottom <i>          The page bottom margin (default: 50; min: 0; max: 500)
 --page-margin-left <i>            The page left margin (default: 50; min: 0; max: 500)
 --page-margin-right <i>           The page right margin (default: 50; min: 0; max: 500)
 --page-margin-top <i>             The page top margin (default: 50; min: 0; max: 500)
```

They are all set to `50` by default.  You can try dropping them all down to `0`, but i think that this will sometimes cause clipping of things that extentend beyond the staff.

palgranum:
I discovered a potential problem with this implementation:
In Page.cpp I based the adjustedWidth on GetContentWidth(), however I have some mei-files that render fine with a fixed width, but this function returns 0. So, this can lead to laying out with zero width which is not good.
Maybe @lpugin or @craigsapp have some idea how to resolve this, because I don't fully understand why contentWidth can be 0 and if that is to be expected. If there is some issue with the files I used (they all seem to be single page mei's, but I don't know if that is coincidental), or if we could solve the problem by simply checking if GetContentWidth() != 0.

I included one example file, it was rendered with:
Verovio ganz.mei --breaks encoded --adjust-page-height true --adjust-page-width true

ScreenShot

[ganz.mei.zip](https://github.com/rism-ch/verovio/files/4219292/ganz.mei.zip)

craigsapp:
Ignoring the encoded system breaks:

```bash
verovio ganz.mei --adjust-page-height true --adjust-page-width true
```

Will give better results:

Screen Shot 2020-02-23 at 10 53 51


With `--breaks encoded` gives, as you note, a very narrow page:

Screen Shot 2020-02-23 at 10 55 12
I don't know what is causing that, but it could be related to when (and where) the page width adjustment is being done.  It is being done before system justification is made, so the result is that using `--adjustPageWidth` is similar to `--no-justification`:

Screen Shot 2020-02-23 at 11 00 29

I think that `--adjustPageWidth` is working at the above layout state: finding the widest unjustified system, and setting the page to that width.  If so, then instead it would be better to set the page to the input pageWidth option, then let page Justification stretch staves to this value, and then check for the right page break position.  The current `--adjustPageWidth` option will almost always strink the width of the page as compared to not using the option.   This is because the width of the unstretched longest system will be used to set the page width.  Instead the adjustPageWidth calculations should be made after justification is used to stretch most systems to the desired page width.  In cases where there is a single system that is not justified, the adjustPageWidth will generate an SVG image with a different page width than the input width; however, when there is any full-width justification (such as at system-breaks), then adjustPageWidth should not alter the width of the page.

palgranum:
I already did a new pull request:
https://github.com/rism-ch/verovio/pull/1316
I am using page->GetContentWidth, but after looking into this function I noticed that it only checks the first system, so I proposed an update to check all systems on the page and use the widest width for the adjustedPageWidth property.
I will add you Craig to review this PR.

PR #1315, opened 2020-02-20
Improve JavaScript build and NPM version

WolfgangDrescher:
- [x] Refactor import of verovio toolkit and emscripten module
- [x] VerovioToolkitProxy: Wrapper class as a JS-Proxy with magic function that calls the methods on the toolkit in th worker
- [x] Use webpack to bundle all scripts into a JS file that can be run as a normal JS file, as a Node-Module or a Worker
- [ ]  Make `./buildToolkit -H -w` work (`BINARYEN_ROOT must be set up in .emscripten`)
- [ ] Test if built `dist/index.js` will work as a default toolkit but also as a worker
- [ ] Review and improve new commands in `./buildToolkit`
- [ ] Write docs on how to implement verovio as a worker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ emscripten/build/
 # npm build
 emscripten/npm/node_modules
 emscripten/npm/dist
+emscripten/npm/src/VerovioEmscriptenModule.js
 
 # python build
 bindings/python/verovio_wrap.cpp
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ emscripten/build/
 # npm build
 emscripten/npm/node_modules
 emscripten/npm/dist
+!emscripten/npm/src/index.js
 emscripten/npm/src/VerovioEmscriptenModule.js
 
 # python build
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ emscripten/npm-dev/node_modules
 main
 DerivedData
 verovio.wasm
+
+# PHPStorm files
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,10 @@ tools/libverovio.so
 emscripten/data/
 emscripten/build/
 
+# npm build
+emscripten/npm/node_modules
+emscripten/npm/dist
+
 # python build
 bindings/python/verovio_wrap.cpp
 bindings/python/verovio.pyc
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,6 @@ bindings/qt/build-demo
 # Output files
 output.svg
 output.mei
-index.js
 emscripten/npm-dev/node_modules
 
 # VS Code files
--- a/emscripten/buildToolkit
+++ b/emscripten/buildToolkit
@@ -211,6 +211,9 @@ if ($? == 0) {
 		print "Building $NPM package\n";
 		# the cat order is different for the npm package because we want to wrap the Module in the init function
 		`cat verovio-npm-start.js $BUILD_DIR/verovio.js verovio-npm-end.js > $NPM/src/VerovioEmscriptenModule.js`;
+
+        print "Installing node modules\n";
+        `(cd $NPM && npm install)`;
 	}
 
 	# create a gzip version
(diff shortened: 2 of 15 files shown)


lpugin:
Please make sure you are not trying to achieve to many things in one go. I am not sure what you need to change in buildToolkit in order to make it work since it seems to work properly. Make a separate PR for this.

Regarding the VerovioEmscriptenAdapter.js, we certainly do not want another file that wraps all the methods. Same for the VerovioToolkit.js one. They look pretty redundant with the verovio-proxy.js functions - expect there is no class, but I don't think this is really the issue. Is it? Duplicating all this will not be good for maintenance.

WolfgangDrescher:
I don't think another PR is necessary since the build process would not change a lot. I just added another layer with webpack after builtToolkit has created the Module to make a browser version out of it (that can also be used in a node environment). See me changes in buildToolkit: `(cd $NPM && npm install)`, `(cd $NPM && npm run build)`.
I'm currently just having issues building the toolkit with the `-w` option.

Are your travis pipelines also building the npm version? If so: do they also use buildToolkit?

My intention is not to add another file with VerovioEmscriptenAdapter.js and VerovioToolkit.js but to replace and relocate them to make them reusable (e.g. the VerovioToolkitProxy needs the toolkit so we just need to import it there). Because i did not want to change to much of the existing code i did not merge VerovioEmscriptenAdapter directly into the VerovioToolkit and let them as they are but split them into two files.

In my opinion the buildToolkit should not do much more than building a JS file with the `Module` and adding an export statement at the end. Currently verovio-npm-start.js, verovio-proxy.js, etc. are concatenated in buildToolkit. I think bundling a runnable JS version with all its dependant classes like the toolkit, proxy, worker and helpers should be done with webpack.

Currently buildToolkit is building two JS files if i understand it correctly. One for general JS usage and one for npm. As far as i can see the only significant difference is the `module.exports` at the end of the npm version. Using webpack to bundle everything together it will not be necessary to have two different files anymore, since webpack adds a build factory at the beginning it will only export the modules if this is possible, otherwise it will assign them to window.

Am i correct that the wohle npm directory is shipped as an npm package? My modifications will build the JS into `~/emscripten/npm/dist/index.js`.

lpugin:
> I'm currently just having issues building the toolkit with the -w option.

Do you have the emdk installed? If yes, what do you get with `emcc --version`

> Are your travis pipelines also building the npm version? If so: do they also use buildToolkit?

No, the npm package is build only for releases, where a bunch of scripts are run locally. This includes `npm publish` at some point. 

Not everybody uses node and webpack, so we do have to keep buildToolkit generating a toolkit that bundles the Module and the verovio-proxy.js. If you need to have the 'raw' wasm output to build a module, just use the `verovio.js` produced by buildToolkit. I think this is what you need.

From there you can make a separate script, which might be simpler to start with. Once everything runs we can see if we can to leave it like this or merge it into buildToolkit.

WolfgangDrescher:
> If yes, what do you get with emcc --version

```
emcc (Emscripten gcc/clang-like replacement) 1.38.44 ((unknown revision))
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I tried installing emsdk like in your travis files like @musicEnfanthen quoted out in https://github.com/rism-ch/verovio/issues/1253#issuecomment-582843842 but it seems like there are issues with a SSL certificate when i run `./emsdk install latest`.

> This includes npm publish at some point.

I see. Then i keep working directly in the npm directroy.

> Not everybody uses node and webpack

Thats exactly why i am using it. Webpack will build me a plain JavaScript version that can be used as a standalone version. At the same time it should be possible to use the same file if you want to import it in node. Like this we have the advantage that during development we can benefit from node imports. Webpack will only be used in the build process and is not needed for users of the verovio package.

> keep buildToolkit generating a toolkit that bundles the Module and the verovio-proxy.js.

That basically what i am achieving to do. But with webpack instead.

> From there you can make a separate script, which might be simpler to start with. Once everything runs we can see if we can to leave it like this or merge it into buildToolkit.

Sounds good. So i will revert my changes in buildToolkit and add a new script (e.g. `bundleToolkit`) for building the toolkit with webpack.

musicEnfanthen:
> there are issues with a SSL certificate when i run `./emsdk install latest`.

Maybe related? [emscripten-core/emscripten#6723: ./emsdk install latest : Installation failed](https://github.com/emscripten-core/emscripten/issues/6723)

Does it work with `python ./emsdk install latest` ?

lpugin:
Please open an updated PR when appropriate.

PR #1316, opened 2020-02-20
Modified Page::GetContentWidth to take max width from all systems

palgranum:
It seems that for the adjustPageWidth to work properly, it should take all systems into consideration, so I added a loop through all of them. Not sure if m_children could have entries other than System*, and if you would prefer assert(system) rather than if(system)

### DIFF
--- a/src/page.cpp
+++ b/src/page.cpp
@@ -592,14 +592,19 @@ int Page::GetContentWidth() const
     // Make sure we have the correct page
     assert(this == doc->GetDrawingPage());
 
-    System *first = dynamic_cast<System *>(m_children.front());
-    assert(first);
-
+    int maxWidth = 0;
+    for (auto & child : m_children) {
+        System *system = dynamic_cast<System *>(child);
+        if (system) {
+            // we include the left margin and the right margin
+            int systemWidth = system->m_drawingTotalWidth + system->m_systemLeftMar + system->m_systemRightMar;
+            maxWidth = std::max(systemWidth, maxWidth);
+        }
+    }
     // For avoiding unused variable warning in non debug mode
     doc = NULL;
 
-    // we include the left margin and the right margin
-    return first->m_drawingTotalWidth + first->m_systemLeftMar + first->m_systemRightMar;
+    return maxWidth;
 }
 
 int Page::CalcJustificationStepSize(bool systemsOnly) const


craigsapp:
With this version of the code, result of 

```bash
verovio ganz.mei
```

```xml
viewBox="0 0 21000 29700"
```

Screen Shot 2020-02-23 at 23 24 11


-------

Adding `--adjust-page-height`:

```bash
verovio ganz.mei --adjust-page-height true 
```

```xml
viewBox="0 0 21000 18360"
```
Notice the third number remains the same (width of the page) and the fourth number gets smaller (height of page).

Screen Shot 2020-02-23 at 23 25 32


------------------------


Adding `--adjust-page-width`

```bash
verovio ganz.mei --adjust-page-height true --adjust-page-width true
```

The viewBox gets wider (was 21000 before `--adjust-page-width` added):

```xml
viewBox="0 0 22730 18360"
```

Ideally the width should stay the same in this case, since the music has more than one systems.  One way to fix it would be to limit the width adjustment to be equal or less than when `--adjust-page-width` is not used.  I don't think that veorvio ever shrinks systems to fit on a page, so I am wondering where the larger value comes from.


Screen Shot 2020-02-23 at 23 28 42


--------------------------

And finally adding `--breaks encoded`:

```bash
verovio ganz.mei --adjust-page-height true --adjust-page-width true --breaks encoded
```

```xml
viewBox="0 0 18790 22370"
```

Note that the width and the height switch places, which is suspicious, with the height of 22370 being the same as the width in the previous example.  The width of 18790 is slightly more than the height in the previous example, but not exactly equal.

Screen Shot 2020-02-23 at 23 48 11

palgranum:
Thank you for looking into it. I think the height/width switch is just a coincidence (also 22730 != 22370, they just look similar). The bars are laid out in the same configuration, but I think for some reason it ends up with a different width calculation due to something being different when laying out with encoded breaks.

My angle on this feature was to have some way of automatically setting the page width without magically guessing it when the breaks are encoded. I see that others might want it in order to crop width when having just a single system. I'm thinking then why wouldn't one use the 'breaks off' feature if it is known that it is only a single line of music. 
In any case, it leaves the question of what is the desired behavior of adjust-page-width when breaks are auto. Should it be ignored, or only be effective when number of systems = 1, or adjust width to the content (which means that the pageWidth value is used to layout, then a tiny amount can be cropped without compressing content)

lpugin:
This looks good to me. @craigsapp ?

PR #1323, opened 2020-02-27
Add support for reh element

rettinghaus:
This adds (basic) support for rehearsal marks in MEI and MusicXML. 
Additionally MusicXML import now uses tstamp for tempo and sets correct lb elements.


wergo:
Thanks a lot for implementhing rehearsal marks import!!

The import of `<reh>` seems to work fine. I have tested it on a musicXML of Mahler IV, exported from Finale. The staff attribution is always correct as far as I could check. 

There is a default value for language set to italian (`xml:lang="it"`). This is not very crucial, but what is the rational for that? (In the Mahler score, there are only numbers inside `<reh>`).

(Now, we would need to render them to the score... :-) )

include/vrv/reh.h
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @name Getter to interfaces
+     */
+    ///@{
+    virtual TextDirInterface *GetTextDirInterface() { return dynamic_cast<TextDirInterface *>(this); }
+    virtual TimePointInterface *GetTimePointInterface() { return dynamic_cast<TimePointInterface *>(this); }
+    ///@}
+
+    /**
+     * Add an element (text, rend. etc.) to a dynam.
```
lpugin: reh, not dynam

src/iomei.cpp
```diff
@@ -4010,6 +4029,21 @@ bool MeiInput::ReadPedal(Object *parent, pugi::xml_node pedal)
...
 
+bool MeiInput::ReadReh(Object *parent, pugi::xml_node reh)
+{
+    Reh *vrvReh = new Reh();
+    ReadControlElement(reh, vrvReh);
+
+    ReadTextDirInterface(reh, vrvReh);
+    ReadTimePointInterface(reh, vrvReh);
+    vrvReh->ReadColor(reh);
+    vrvReh->ReadLang(reh);
```
lpugin: Reading `AttVerticalGroup` is missing

src/iomusxml.cpp
```diff
@@ -1590,10 +1593,10 @@ void MusicXmlInput::ReadMusicXmlDirection(
         text->SetText(UTF8to16(dynamStr));
         dynam->AddChild(text);
         dynam->SetTstamp(timeStamp);
-        int staffNum = 1;
         pugi::xpath_node staffNode = node.select_node("staff");
-        if (staffNode) staffNum = staffNode.node().text().as_int() + staffOffset;
-        dynam->SetStaff(dynam->AttStaffIdent::StrToXsdPositiveIntegerList(std::to_string(staffNum)));
+        if (staffNode)
```
lpugin: Add {}

rettinghaus:
@wergo Rehearsal marks are italian by default, because MusicXML says so: https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-rehearsal.htm

craigsapp:
> There is a default value for language set to italian (xml:lang="it"). This is not very crucial, but what is the rationale for that? (In the Mahler score, there are only numbers inside `<reh>`).

Yes, I wonder about that as well.  Probably it means that the letters are enumerated in the Italian alphabet which does not include "J", "K", "W", "X", "Y":

https://en.wikipedia.org/wiki/Italian_orthography#Alphabet

So this is not really a language specification but rather an alphabet one.  If you want to automatically insert or delete rehearsal marks, then knowing what the allowed alphabet characters are and their sequence would be necessary.

rettinghaus:
@craigsapp There are multiple default values in MusicXML worth discussing ;-) 
LilyPond skips the "I" by default: https://lilypond.org/doc/v2.19/Documentation/notation/bars.html#rehearsal-marks

craigsapp:
> LilyPond skips the "I" by default: https://lilypond.org/doc/v2.19/Documentation/notation/bars.html#rehearsal-marks

That is a good one to skip; otherwise `I` looks like `1`.  But then what "langauge" would that be?

> There are multiple default values in MusicXML worth discussing ;-)

Would that be something natural to discuss? 😜

Issue #1176, opened 2019-10-29
Long loading times


Noroxs:
When loading the file [Weber: Der Freyschütze](https://github.com/Freischuetz-Digital/data-music/blob/master/musicSources/freidi-musicSource_A.xml) into Verovio it takes 70 to 100 seconds to load the file/first movement.

Other big files I tested (~37000 lines) only take 2 seconds of loading time.

How is the length of a file related to the load time? 
How could the load time be improved?

kepper:
I wouldn't recommend to operate on that file directly – that's a ~70mb file with 3100 measures of the Freischütz (split across 17 movements), but no single pitch at all. I suggest to read the readme.md of that repo about the setup of these files, and then go for the files in the [ready-for-use](https://github.com/Freischuetz-Digital/data-music/tree/master/ready-for-use) folder…

Noroxs:
This would only help for this specific file.
But imagine a case where I get such a big file and can't use a different one. 

So my questions are still relevant for me.

> How is the length of a file related to the load time?
How could the load time be improved?

lpugin:
This the times I get for loading and rendering the complete content. This looks fairly linear to me. 

Hummel_Concerto_for_trumpet.mei | 6MB
real	0m3.826s
user	0m3.429s
sys	0m0.290s

Brahms_StringQuartet_Op51_No1.mei | 659 KB
real	0m0.468s
user	0m0.396s
sys	0m0.049s

Ahle_Jesu_meines_Herzens_Freud.mei | 60 KB
real	0m0.067s
user	0m0.039s
sys	0m0.022s

So the best way to improve the loading / rendering time is to pass to Verovio only the data you want to render.

lpugin:
@Noroxs any ideas what the problem is for you?

Noroxs:
It's not directly a problem. It's more of an inconvenience. I like to gather some information like. 
Example:
- Number of pages in the document
- Number of measures per page

My intention is to give the user feedback as fast as possible. 
That's why I asked for optimisation methods.

PR #1327, opened 2020-03-02
Fixed most critical bottlenecks. 

r0mko:
This fix improves performance of PrepareDrawing by about 30% by using more optimal containers and faster string comparison

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -221,7 +221,8 @@ public:
      */
     virtual void CloneReset();
 
-    std::string GetUuid() const { return m_uuid; }
+    const std::string &GetUuid() const { return m_uuid; }
+    size_t GetUuidHash() const { return m_uuidHash; }
     void SetUuid(std::string uuid);
     void SwapUuid(Object *other);
     void ResetUuid();
@@ -1159,6 +1160,7 @@ private:
      */
     ///@{
     std::string m_uuid;
+    size_t m_uuidHash;
     std::string m_classid;
     ///@}
 
--- a/include/vrv/timeinterface.h
+++ b/include/vrv/timeinterface.h
@@ -118,6 +118,7 @@ public:
 protected:
     LayerElement *m_start;
     std::string m_startUuid;
+    size_t m_startUuidHash;
 
 private:
 };
@@ -217,6 +218,7 @@ public:
 private:
     LayerElement *m_end;
     std::string m_endUuid;
+    size_t m_endUuidHash;
 };
 
 } // namespace vrv
--- a/include/vrv/vrvdef.h
+++ b/include/vrv/vrvdef.h
@@ -287,7 +287,7 @@ typedef std::vector<BeamElementCoord *> ArrayOfBeamElementCoords;
 
 typedef std::vector<std::pair<int, int> > ArrayOfIntPairs;
 
-typedef std::vector<std::pair<LinkingInterface *, std::string> > ArrayOfLinkingInterfaceUuidPairs;
+typedef std::list<std::pair<LinkingInterface *, std::string> > ArrayOfLinkingInterfaceUuidPairs;
 
 typedef std::vector<std::pair<PlistInterface *, std::string> > ArrayOfPlistInterfaceUuidPairs;
 
--- a/include/vrv/vrvdef.h
+++ b/include/vrv/vrvdef.h
@@ -287,7 +287,7 @@ typedef std::vector<BeamElementCoord *> ArrayOfBeamElementCoords;
 
 typedef std::vector<std::pair<int, int> > ArrayOfIntPairs;
 
-typedef std::map<std::string, LinkingInterface *> ArrayOfLinkingInterfaceUuidPairs;
+typedef std::multimap<std::string, LinkingInterface *> ArrayOfLinkingInterfaceUuidPairs;
 
 typedef std::vector<std::pair<PlistInterface *, std::string> > ArrayOfPlistInterfaceUuidPairs;
 
--- a/include/vrv/vrvdef.h
+++ b/include/vrv/vrvdef.h
@@ -295,9 +295,9 @@ typedef std::vector<CurveSpannedElement *> ArrayOfCurveSpannedElements;
 
 typedef std::vector<std::pair<Object *, data_MEASUREBEAT> > ArrayOfObjectBeatPairs;
 
-typedef std::vector<std::pair<TimePointInterface *, ClassId> > ArrayOfPointingInterClassIdPairs;
+typedef std::list<std::pair<TimePointInterface *, ClassId> > ArrayOfPointingInterClassIdPairs;
 
-typedef std::vector<std::pair<TimeSpanningInterface *, ClassId> > ArrayOfSpanningInterClassIdPairs;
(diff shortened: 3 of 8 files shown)


lpugin:
Could you be a bit more specific about your claim or 1200x speed up and give an example file?
A random test on the Hummel file also used here https://github.com/rism-ch/verovio/issues/1176#issuecomment-547434052 confirms that it will make no difference in most cases.

develop branch:
real	0m3.914s
**user	0m3.594s**
sys	0m0.244s

pull request
real	0m3.915s
**user	0m3.580s**
sys	0m0.258s

This is not surprising because linking is not a widely used feature in MEI. For example, searching for `sameas` in the sample encodings give only 11 files https://github.com/music-encoding/sample-encodings/search?q=sameas&unscoped_q=sameas For this fix to make a significant difference I would expect a cases where you have hundreds for links in the file. What use case is this?

r0mko:
> Changing this does not work if we have two (or more) elements pointing to the same one. Changing to a map allows for only one - even though as it is in Object::PrepareLinking only one link will be establish and this need to be fixed.

My initial assumption was based on the current implementation which searches only for one instance of matching UUID. Also I made an investigation to find out whether or not we can have several values with same key and found no such cases for the file [https://github.com/Freischuetz-Digital/data-music/blob/master/musicSources/freidi-musicSource_A.xml](URL) mentioned in issue https://github.com/rism-ch/verovio/issues/1176. 
I agree that linking is not widely used in MEI files, but this particular commit (75632571971c3bd774e462bb34db3de9e6b78cab) addresses very specific and rare problem that causes extremely long loading of a file which apparently heavily employs linking.
I've measured duration for PrepareDrawing() with and without this patch applied and got the following results:
Before:
`Elapsed time for PrepareDrawing 46735 ms`
After:
`Elapsed time for PrepareDrawing 164 ms`
Which is not exactly 1200x speedup which I observed some time ago, but still well over 200x faster. 
I would propose changing map to multimap to allow storing more that one element for a key and do full search of entries, not only the first one.

lpugin:
Why are you keeping using this file? I think @kepper made it pretty clear it is not a typical file. More than 60 MB but not one single note... ;-) https://github.com/rism-ch/verovio/issues/1176#issuecomment-547421132

What is annoying is that you are adding a size_t hash (8 bytes) to every single object in Verovio to fix address a very specific and rare problem.

r0mko:
> What is annoying is that you are adding a size_t hash (8 bytes) to every single object in Verovio to address a very specific and rare problem.

Adding hashes to compare string addresses much more common problem. String comparison is used really heavily during PrepareDrawing. Hash comparison instead of string comparison alone gives about 15% speedup on iOS.

r0mko:
> Why are you keeping using this file? I think @kepper made it pretty clear it is not a typical file. More than 60 MB but not one single note... ;-) [#1176 (comment)](https://github.com/rism-ch/verovio/issues/1176#issuecomment-547421132)

This file was provided by eNote as an example. I have no idea of its purpose. Let's consider it as a benchmark which highlights an issue.

lpugin:
> Adding hashes to compare string addresses much more common problem.

Verovio avoids string comparisons as much as possible. Nearly all MEI attributes are converted to enums. But I do not think it is a good idea to use hashes on `@xml:id`. In MEI, they are meant to be unique, and validation checks this. Using a hash loosen the constraint, and it seems to me that this can yield unpredictable behavior in the case of hash collision. A that a lot of people use generated UUIDs (128bits) but of course if you have shorter `@xml:id`, the string comparison will be faster.

> Let's consider it as a benchmark which highlights an issue.

Do not use this file to benchmark rendering, this is a bad idea.

> Hash comparison instead of string comparison alone gives about 15% speedup on iOS.

Could you share the file for which you gain 15% of speed? It would be interesting to see on other platforms, maybe we can find another solution.

Pryadka:
> Changing this does not work if we have two (or more) elements pointing to the same one. Changing to a map allows for only one - even though as it is in Object::PrepareLinking only one link will be establish and this need to be fixed.

So, that means more than one object can have the same uuid?
Probably I missed something, but I can't understand why uuid's are generated using random integers as a suffix. Why we can't just use static counter (it is already exist - s_objectCounter) as suffix? If the problem is in a collision between random numbers, it can help.

craigsapp:
> Why we can't just use static counter (it is already exist - s_objectCounter) as suffix? If the problem is in a collision between random numbers, it can help. 

One problem with this method is that the MEI file may already have an ID such as "note-1", so if you generate a new ID "note-1" using this method, you need to check for an existing ID that is the same and skip it.  Using random numbers allows for avoiding doing this check since it should be a rare problem to have a collision (in the past there has been a problem with the random number being truncated, see https://github.com/rism-ch/verovio/issues/331)

> Moreover, this solution allow to avoid using random generator and correspondingly improve performance.

Faster to generate, but slower to avoid collisions.

Pryadka:
> > Why we can't just use static counter (it is already exist - s_objectCounter) as suffix? If the problem is in a collision between random numbers, it can help.
> 
> One problem with this method is that the MEI file may already have an ID such as "note-1", so if you generate a new ID "note-1" using this method, you need to check for an existing ID that is the same and skip it. Using random numbers allows for avoiding doing this check since it should be a rare problem to have a collision (in the past there has been a problem with the random number being truncated, see #331)
> 
> > Moreover, this solution allow to avoid using random generator and correspondingly improve performance.
> 
> Faster to generate, but slower to avoid collisions.

Anyway usage random instead of sequential can minimize a collision but not eliminate it. 
As a solution we can consider a collecting all input id's in a set and check if newly created uuid is not repeated. The std::set has search with log complexity, it's longer than random generation of one number, but can solve both problems.

craigsapp:
> Anyway usage random instead of sequential can minimize a collision but not eliminate it.

The probability of a collision can be calculated, and it is quite low (meaning less than a million in one I would guess).  An external tool can check for duplicate IDs in the generated SVG image if necessary.  Otherwise, no one will be killed if there are collisions, unlike in the airline industry were a one-in-a-million chance of a mid-air collision is unacceptable.  Note that the current automatic IDs are guaranteed to be unique across different element types, since the name of the element is prefixed to the random number.

Verovio calculates an ID for every element automatically provided that an element does not already have one on import.  A more useful discussion would be to talk about the necessity of having an ID for every single element (@kepper may have mentioned that before in the verovio issues).  There could be an option to suppress automatic IDs in the MEI data, and as a result, the SVG elements would also not have automatically generated IDs (and be smaller).

> As a solution we can consider a collecting all input id's in a set and check if newly created uuid is not repeated. The std::set has search with log complexity, it's longer than random generation of one number, but can solve both problems.

Keep in mind that verovio is mostly used for generating notation within web browser pages.  You are obviously targeting a higher resource system (smartphone, tablet).  XML is a very verbose data format already, so adding efficiencies that are at the expense of expanded memory will degrade the web-brower experience of verovio.

> The std::set has search with log complexity, it's longer than random generation of one number, but can solve both problems.

Random number generation is O(1), and using a set is O(log(N)), so the larger the file, the more computationally costly the process of verifying the uniqueness of the IDs.  If you absolutely need the IDs to be unique, then you should provide them in the input data; otherwise, note that collisions will otherwise be rare, so you would be wasting time by adding a uniqueness check for IDs.

PR #1331, opened 2020-03-06
Container staffGrp brackets and braces

rettinghaus:
Brackets and braces from `<staffGrp>` are now contained within a `<g class="grpSym">` referencing the MEI element `<grpSym>` until it's supported natively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## [unreleased]
+* Container brackets and braces in grpSym (@rettinghaus)
+* Support for reh elements (@rettinghaus)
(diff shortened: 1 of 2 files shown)


lpugin:
Could you provide a description? Thanks

rettinghaus:
Brackets and braces from `<staffGrp>` are now contained within a `<g class="grpSym">` referencing the MEI element `<grpSym>` until it's supported natively.

lpugin:
Thanks. Out of curiosity: why do you need this before it is supported natively?

rettinghaus:
Just thinking it would be nice to make them addressable by CSS.

PR #1340, opened 2020-03-11
Adjust voice

Pryadka:
This change aligns voices in seconds as it requires here https://github.com/eNote-GmbH/notengrafik/blob/master/testset/voices-in-seconds_expected.svg . Probably 0-margin is not perfect and it needs to be adjusted.


craigsapp:
https://github.com/eNote-GmbH/notengrafik/blob/master/testset/voices-in-seconds_expected.svg

Perhaps that is in a private repository?, as I cannot see it.

rettinghaus:
It is indeed. 
@Pryadka you should add a MEI test file and the Verovio output.

include/vrv/object.h
```diff
@@ -1132,7 +1132,7 @@ class Object : public BoundingBox {
     /**
      * Initialisation method taking a uuid prefix argument.
      */
-    void Init(std::string);
+    void Init(const std::string&);
```
lpugin: Add the missing `classId` name

src/bboxdevicecontext.cpp
```diff
@@ -300,8 +300,9 @@ void BBoxDeviceContext::DrawText(const std::string &text, const std::wstring wte
         m_textX -= extend.m_width;
     }
     else if (m_textAlignment == HORIZONTALALIGNMENT_center) {
-        m_textX -= (extend.m_width / 2);
+        m_textX -= extend.m_width / 2;
```
lpugin: Leave the ()

Pryadka:
Input data file :
[voices-in-seconds.txt](https://github.com/rism-ch/verovio/files/4318384/voices-in-seconds.txt)
Current implementation:
current
Expected:
expected

include/vrv/object.h
Pryadka: Done

src/bboxdevicecontext.cpp
Pryadka: Done

src/layerelement.cpp
```diff
@@ -1153,6 +1153,11 @@ int LayerElement::AdjustLayers(FunctorParams *functorParams)
                 }
                 else if (abs(previousNote->GetDrawingLoc() - params->m_currentNote->GetDrawingLoc()) > 1)
                     continue;
+                // check if the next note is simultaneous and put in with adjustable space
```
lpugin: I am not sure how this works. A note will always have a staff ancestor. When do you expect this not to be true? What is the logic of the change?
Also, you do not need to do a dynamic_cast to compare with a NULL.  (Please change `nullptr` to `NULL` for consistency)

Pryadka: Sorry, I didn't know about staff ancestors. I removed this check.

PR #1342, opened 2020-03-11
Allow chords with lyrics in MusicXML

rettinghaus:
This PR
* puts MusicXML lyrics in chords on MEI chord level
* fixes an ordering problem within `<staffDef>` that resulted in invalid MEI files


src/iomusxml.cpp
```diff
@@ -2369,7 +2370,13 @@ void MusicXmlInput::ReadMusicXmlNote(
                     verse->AddChild(syl);
                 }
             }
-            note->AddChild(verse);
+            if (element->Is(CHORD) || element->Is(NOTE)) {
+                element->AddChild(verse);
+            }
+            else {
+                // this should not happen
```
craigsapp: I can happen if the element is a rest (so it is good to have this else case).

PR #1349, opened 2020-03-20
add brace to Bravura and Petaluma

Pryadka:
The gliph for Smufl character brace (E000 https://www.smufl.org/version/latest/range/staffBracketsAndDividers/ )   is added. It needs for further replacing hardcoded implementation with font dependent.


lpugin:
Using a glpyh for braces is not a good idea because the glyph would need to be hugely distorted. The result is usually extremely ugly, and this is why the SMuFL glpyh is not used. This PR does not solve any issues and does not add a new feature.

Pryadka:
> Using a glpyh for braces is not a good idea because the glyph would need to be hugely distorted. The result is usually extremely ugly, and this is why the SMuFL glpyh is not used. This PR does not solve any issues and does not add a new feature.

Ok. It can be ugly for one customer and necessary for another. We can wrap corresponding code with #ifdef , or use some another approach.

Pryadka:
ugly_brace
This is the result of test rendering. 'Ugly' means too wide?
We still want to use different fonts for 'brace' rendering. As a solution can we consider using different scales for height and width? Yes, it needs significant code reworking, but gives us possibility to use scale correction for another glyphs in the future and solve 'brace' issue.

lpugin:
> 'Ugly' means too wide?

Well, this is a matter of taste but I find it ugly like this. Using different scales for height and width will not completely solve the issue. As it is now, Verovio makes the brace look similar whatever system height. This will not be the case with a glyph and scaling.

Pryadka:
> As it is now, Verovio makes the brace look similar whatever system height. This will not be the case with a glyph and scaling.
Why not? We will calculate the width using the same calculation, and then adjust the horizontal scale to preserve the size of the frames, as is now.

Pryadka:
> As it is now, Verovio makes the brace look similar whatever system height. This will not be the case with a glyph and scaling.

Why not? We will calculate the width using the same calculation, and then adjust the horizontal scale to preserve the size of the frames, as is now.

lpugin:
Stretching a brace glyph yields different results. 
image

Pryadka:
To be honest I don't understand. Why 2nd and 3rd images are different. BoundingBox context creates identical boxes for both, doesn't it?

Pryadka:
Hello @lpugin .

Having braces corresponding current font is still actual for us. I can provide 2 solution for supporting brace glyph.

1. Create some .xml files for different brace heights with different names, for instance E0001-brace, E0002-brace …

2. Read data from corresponding glyph file (E000-brace) and extract path. Apply for every point of this pass a transformation with keep the same coordinates for points at the beginning, the end and in the middle, and stretch coordinates in between. The transformation can be non-linear function or some smooth function. The width of unstretchable intervals should be found in experimental way. Horizontal coordinates of the pass can be kept as is or stretched depending on a system height.

What do you think about it.

Issue #1158, opened 2019-10-15
Replace xlink:href by href?


pfefferniels:
The SVG that verovio generates seems to make frequent use `xlink:href` attributes in `<use>` elements. `xlink:href` is deprecated in SVG2 and I just ran into a strange bug where SVG.js failed to calculate bounding boxes because of that. Could it be replaced by just `href`?

lpugin:
There is no plan to move to SVG2 in the near future. What is SVG.js?

lpugin:
As far as I remember the `xlink:href` was chosen because some viewers would not work otherwise. I think it is (or was) the case for Safari.

pfefferniels:
I see. [This page](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href) was warning that some browser won't support it anymore in some future point. 
SVG.js It's a library for [manipulating SVG](https://svgjs.com/docs/2.7/). I use it server-side in combination with svgdom. It only supports `href`, not `xlink:href` ...

lpugin:
Since you use it server-side the best for now is simply to pipe the output through a simple XSLT that changes the attribute to `href`. Would this work for you?

I think it will be a long way until most browsers will support SVG 2.0, and until them it will be wiser to stick with the current version.

lpugin:
As you can see here, `href` still has lower support than `xlink:href`:
https://caniuse.com/#search=svg%20xlink%3Ahref
https://caniuse.com/#search=svg%20href

pfefferniels:
> Since you use it server-side the best for now is simply to pipe the output through a simple XSLT that changes the attribute to `href`. Would this work for you?
> 
> I think it will be a long way until most browsers will support SVG 2.0, and until them it will be wiser to stick with the current version.

Yes, that's what I am going to do. Thanks for the information :)

kepper:
…and we would increase filesizes considerably. I'm frequently running into issues with that, so I'm not in favor. Could that be an option in Verovio?

rettinghaus:
Just tested it. It increases file sizes by about 2% to 3%. 
@kepper compared to the output older versions of Verovio created that's still small. ;-)

rettinghaus:
@kepper Tried to update Beethovens Werkstatt?

kepper:
[...]

PR #1358, opened 2020-03-21
fix wrong svg attribute

rettinghaus:
This removes the non-valid `href` attribute in SVG 1.1 output. 
We discussed that in #1158, no idea where that came from.


lpugin:
It came here https://github.com/rism-ch/verovio/commit/c66cfa682c741860a57285a3bece12c80377b8a8

rettinghaus:
@lpugin Ok, sorry. Must have been remnants of other tests.

Issue #1352, opened 2020-03-20
MusicXML conversion: Convert <rights> into <pgFoot> and <pgFoot2>


earboxer:
Example File: [51c-MultipleRights.xml](https://raw.githubusercontent.com/cuthbertLab/musicxmlTestSuite/master/xmlFiles/51c-MultipleRights.xml) from the Unofficial MusicXML Test Suite.

earboxer:
I implemented this in https://github.com/earboxer/verovio/commit/816833b36f00acc89c07ea1196d9c4be2a5ed731, but I'm not completely happy with the result.

A score exported from MuseScore to MusicXML appears like

![Screen Shot 2020-03-20 at 16 51 07](https://user-images.githubusercontent.com/12452738/77205335-0b607e80-6acb-11ea-8958-251e5288b6f4.png)

Which is not ideal, since it repeats the Copyright twice (as it also appears as a `<credit-words>` in the exported musicXML)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
  <work>
    <work-title>Title</work-title>
    </work>
  <identification>
    <creator type="composer">Composer</creator>
    <creator type="lyricist">Lyricist</creator>
    <rights>Copyright</rights>
    <encoding>
      <software>MuseScore 3.4.2</software>
      <encoding-date>2020-03-20</encoding-date>
      <supports element="accidental" type="yes"/>
      <supports element="beam" type="yes"/>
      <supports element="print" attribute="new-page" type="no"/>
      <supports element="print" attribute="new-system" type="no"/>
      <supports element="stem" type="yes"/>
      </encoding>
    </identification>
  <defaults>
    <scaling>
      <millimeters>7.05556</millimeters>
      <tenths>40</tenths>
      </scaling>
    <page-layout>
      <page-height>1584</page-height>
      <page-width>1224</page-width>
      <page-margins type="even">
        <left-margin>56.6929</left-margin>
        <right-margin>56.6929</right-margin>
        <top-margin>56.6929</top-margin>
        <bottom-margin>113.386</bottom-margin>
        </page-margins>
      <page-margins type="odd">
        <left-margin>56.6929</left-margin>
        <right-margin>56.6929</right-margin>
        <top-margin>56.6929</top-margin>
[...]

PR #1360, opened 2020-03-21
only print upper credits

rettinghaus:
This PR addresses #1352 in trying to only take credits above the music into account.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -532,12 +532,17 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
         section->AddChild(pb);
     }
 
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -532,12 +532,17 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
         section->AddChild(pb);
     }
 
+    pugi::xpath_node layout = root.select_node("/score-partwise/defaults/page-layout");
+    float bottom = layout.node().select_node("page-margins/bottom-margin").node().text().as_float();
+    LogWarning("%f", bottom);
```
earboxer: We probably don't want to keep this warning.

PR #1376, opened 2020-03-25
change stem automatic shortening

rettinghaus:
This PR slightly improves the stem shortening outside a staff. It now follows the same rules as LilyPond and MuseScore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+* Improve automatic stem shertening (@rettinghaus)
 * Complete beam refactoring
 * Support for mSpace elements (@rettinghaus)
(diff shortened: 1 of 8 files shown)


lpugin:
Thanks. (Maybe we don't want a line in the change log for this? I think we should try to make it useful for the users. For people who want to see precisely who made what, people can look at the commits)

rettinghaus:
@lpugin agreed

lpugin:
It seems that this is not working as expect.

Before 
image

After
image

I am reverting it. Could you double check?

craigsapp:
> This PR slightly improves the stem shortening outside a staff. It now follows the same rules as LilyPond and MuseScore.

Hopefully they use these values for stem shortening (original octave lengths on the left):

screen shot 2016-12-10 at 07 21 23

PR #1379, opened 2020-03-26
Get rid of redundant coping for BACKWARD processing

Pryadka:
This commit guarantees the following:
- use a reverse iterator to reverse process data instead of creating a copy of the collection and reversing it;
- using a constant reference instead of copying strings that are constant values. I think that this improvement should be applied to the whole project (an appropriate task should be created)

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -61,7 +61,7 @@ Object::Object() : BoundingBox()
     }
 }
 
-Object::Object(std::string classid) : BoundingBox()
+Object::Object(const std::string& classid) : BoundingBox()
 {
     Init(classid);
     if (s_objectCounter++ == 0) {
@@ -161,7 +161,7 @@ Object::~Object()
     ClearChildren();
 }
 
-void Object::Init(std::string classid)
+void Object::Init(const std::string& classid)
 {
     m_classid = classid;
     m_parent = NULL;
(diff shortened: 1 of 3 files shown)


src/object.cpp
```diff
@@ -786,46 +786,43 @@ void Object::Process(Functor *functor, FunctorParams *functorParams, Functor *en
...
             if (filters && !filters->empty()) {
-                bool hasComparison = false;
                 // first we look if there is a comparison object for the object type (e.g., a Staff)
-                ArrayOfComparisons::iterator comparisonIter;
-                for (comparisonIter = filters->begin(); comparisonIter != filters->end(); ++comparisonIter) {
-                    // if yes, we will use it (*comparisonIter) for evaluating if the object matches
-                    // the attribute (see below)
-                    Object *o = *iter;
-                    ClassIdComparison *attComparison = dynamic_cast<ClassIdComparison *>(*comparisonIter);
+                auto classId = iter->GetClassId();
```
lpugin, line 792: Do not use `auto` when not necessary

lpugin:
> using a constant reference instead of copying strings that are constant values. I think that this improvement should be applied to the whole project (an appropriate task should be created)

I agree. I'll create a develop-const branch for this to be reviewd. Make a PR there

PR #1388, opened 2020-03-30
add support for head.visible

rettinghaus:


src/view_element.cpp
```diff
@@ -1256,28 +1256,30 @@ void View::DrawNote(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
...
-            fontNo = SMUFL_E0A4_noteheadBlack;
+        if (drawingDur < DUR_1) {
+            DrawMaximaToBrevis(dc, noteY, element, layer, staff);
+        }
+        // Whole notes
+        else if (drawingDur == DUR_1) {
+            if (note->GetColored() == BOOLEAN_true)
+                fontNo = SMUFL_E0FA_noteheadWholeFilled;
+            else
+                fontNo = SMUFL_E0A2_noteheadWhole;
```
lpugin: Could you take the opportunity to add `{}`? Thanks

src/view_element.cpp
```diff
@@ -1256,28 +1256,30 @@ void View::DrawNote(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
...
+
+            DrawSmuflCode(dc, noteX, noteY, fontNo, staff->m_drawingStaffSize, drawingCueSize, true);
+        }
+        // Other values
+        else {
+            if ((note->GetColored() == BOOLEAN_true) || drawingDur == DUR_2)
+                fontNo = SMUFL_E0A3_noteheadHalf;
+            else
+                fontNo = SMUFL_E0A4_noteheadBlack;
```
lpugin: Ditto

rettinghaus: done

PR #1394, opened 2020-03-30
add warning

rettinghaus:
this adds a `@type="warning"` to `<annot>` that it may be rendered in a special way (perhaps with markdown `:warning:`).


lpugin:
Thanks. Why removing the Schematron processing instruction?

rettinghaus:
dunno, automatism?

PR #1401, opened 2020-04-05
major trill rendering improvement

### LINKED ISSUE #807: Improvements to trill line extenders
In this example using trill wavy line extensions:

screen shot 2018-03-28 at 4 33 23 am

There are three things that could use improvement:

(1) A wavy line should be attached to its "tr" symbol (no space between them).

(2) When a `trill@tstamp2` indicates the end of a measure, the wavy line should not extend over/thru a barline.  In the above example this can be see in two locations:  at the end of the first measure on the top staff, the wavy line is crossing over the barline.  It does not look bad since the line does not visually cross the barline, but if the barline were extended upwards, the trill line would incorrectly cross it.  A second case occurs at the last trill of both staves where the trill marker is overlapping with the final barline.

In such cases, the best thing to do is to position the wavy line as currently, then stretch the layout so that the barline ends after the line (as opposed to truncating the wavy line one cycle earlier). (Good luck implementing that :-)

(3) The last trill of the top staff is not horizontally aligned with the other trills on the staff.  In this case all trills should be horizontally aligned.  In other words, the trill line ending can/should be tweaked so that this happens.  This is similar to how dynamic markings and hairpins now behave in verovio.

Test data:

```xml
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2018-03-28T04:35:23" version="2.0.0-dev-314ae07">
                    <name>Verovio</name>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title />
                </titleStmt>
            </work>
        </workDesc>
    </meiHead>
    <music>
        <body>
            <mdiv xml:id="mdiv-0000001333036987">
                <score xml:id="score-0000001803235005">
                    <scoreDef xml:id="scoredef-0000000201960921" midi.bpm="400">
                        <staffGrp xml:id="staffgrp-0000001932407777" symbol="brace" barthru="true">
                            <staffDef xml:id="staffdef-0000000128916277" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="1" lines="5">
                                <label xml:id="label-0000001620817514" />
                            </staffDef>
                            <staffDef xml:id="staffdef-0000001664869736" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" n="2" lines="5" />
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-0000001681502571">
                        <measure xml:id="measure-L4" n="1">
                            <staff xml:id="staff-L4F2N1" n="1">
                                <layer xml:id="layer-L4F2N1" n="1">
                                    <note xml:id="note-L5F2" dur="4" oct="5" pname="c" accid.ges="n" />
                                    <note xml:id="note-L6F2" dur="4" oct="5" pname="c" accid.ges="n" />
                                    <note xml:id="note-L7F2" dur="2" oct="5" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
                            <staff xml:id="staff-L4F1N1" n="2">
                                <layer xml:id="layer-L4F1N1" n="1">
                                    <note xml:id="note-L5F1" dur="2" oct="4" pname="c" accid.ges="n" />
                                    <note xml:id="note-L7F1" dur="2" oct="4" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
                            <trill xml:id="trill-L5F2" staff="1" tstamp="1.000000" />
                            <trill xml:id="trill-L6F2" staff="1" tstamp="2.000000" />
                            <trill xml:id="trill-L7F2" staff="1" tstamp="3.000000" tstamp2="0m+5.0" />
                            <trill xml:id="trill-L5F1" staff="2" tstamp="1.000000" tstamp2="1m+5.0" />
                        </measure>
                        <measure xml:id="measure-L8" right="end" n="2">
                            <staff xml:id="staff-L8F2N1" n="1">
                                <layer xml:id="layer-L8F2N1" n="1">
                                    <note xml:id="note-L9F2" dur="2" oct="5" pname="c" accid.ges="n" />
                                    <note xml:id="note-L10F2" dur="2" oct="5" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
                            <staff xml:id="staff-L8F1N1" n="2">
                                <layer xml:id="layer-L8F1N1" n="1">
                                    <note xml:id="note-L9F1" dur="2" oct="4" pname="c" accid.ges="n" />
                                    <note xml:id="note-L10F1" dur="2" oct="4" pname="c" accid.ges="n" />
                                </layer>
                            </staff>
                            <trill xml:id="trill-L9F2" staff="1" tstamp="1.000000" tstamp2="0m+3.0" />
                            <trill xml:id="trill-L10F2" staff="1" tstamp="3.000000" tstamp2="0m+5.0" />
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

rettinghaus:
This PR adds support for `@lstartsym="none"` on trills and improves overall engraving rules for trills. Examples using IDs on the first line, time stamps on the second:

## Before:
Trills-Leipzig_old

## After:
Trills-Leipzig

# Bravura
## Before:
Trills-Bravura_old

## After:
Trills-Bravura

# Gootville
## Before:
Trills-Gootville_old

## After:
Trills-Gootville

# Petaluma
## Before:
Trills-Petaluma_old

## After:
Trills-Petaluma

### Test file:
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0" ?>
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>Trills without trill glyph</title>
      </titleStmt>
      <pubStmt>
        <respStmt>
        </respStmt>
      </pubStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef system.leftline="true">
            <staffGrp bar.thru="false" symbol="none">
              <staffDef n="1" lines="1" clef.shape="perc" clef.line="1"></staffDef>
            </staffGrp>
          </scoreDef>
          <section label="using IDs">
            <pb />
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" xml:id="one" />
                </layer>
              </staff>
              <trill startid="#one" endid="#two" />
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" xml:id="two" />
                </layer>
              </staff>
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" xml:id="three" />
                </layer>
              </staff>
              <trill startid="#three" endid="#four" lstartsym="none" />
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" xml:id="four" />
                </layer>
              </staff>
            </measure>
          </section>
          <section label="using time stamps">
            <sb />
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" />
                </layer>
              </staff>
              <trill tstamp="1" tstamp2="1m+1" />
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" />
                </layer>
              </staff>
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" />
                </layer>
              </staff>
              <trill tstamp="1" tstamp2="1m+1" lstartsym="none" />
            </measure>
            <measure>
              <staff n="1">
                <layer>
                  <note dur="1" loc="0" />
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

closes #807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+* Improved trill support (@rettinghaus)
 * Support for `trill@extender` (@rettinghaus)
 * Support for `note@head.visible` (@rettinghaus)
 * Improved spacing with crossing voices (@rettinghaus)
--- a/include/vrv/trill.h
+++ b/include/vrv/trill.h
@@ -26,6 +26,7 @@ class Trill : public ControlElement,
               public TimeSpanningInterface,
               public AttColor,
               public AttExtender,
+              public AttLineRend,
               public AttOrnamentAccid,
               public AttPlacement {
 public:
--- a/src/elementpart.cpp
+++ b/src/elementpart.cpp
@@ -82,7 +82,7 @@ wchar_t Flag::GetSmuflCode(data_STEMDIRECTION stemDir)
             case 6: return SMUFL_E24A_flag256thUp;
             case 7: return SMUFL_E24C_flag512thUp;
             case 8: return SMUFL_E24E_flag1024thUp;
-            default: return 0; break;
+            default: return 0;
         }
     }
     else {
@@ -95,7 +95,7 @@ wchar_t Flag::GetSmuflCode(data_STEMDIRECTION stemDir)
             case 6: return SMUFL_E24B_flag256thDown;
             case 7: return SMUFL_E24D_flag512thDown;
             case 8: return SMUFL_E24F_flag1024thDown;
-            default: return 0; break;
+            default: return 0;
         }
     }
 }
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2561,6 +2561,10 @@ void MusicXmlInput::ReadMusicXmlNote(
         trill->SetColor(xmlTrill.node().attribute("color").as_string());
         // place
         trill->SetPlace(trill->AttPlacement::StrToStaffrel(xmlTrill.node().attribute("placement").as_string()));
+        if (notations.node().select_node("ornaments/wavy-line")) {
+            trill->SetTstamp2(
+                std::pair<int, double>(0, (double)(m_durTotal) * (double)m_meterUnit / (double)(4 * m_ppq) + 0.99));
+        }
     }
 
     // turns
--- a/src/trill.cpp
+++ b/src/trill.cpp
@@ -22,11 +22,18 @@ namespace vrv {
 //----------------------------------------------------------------------------
 
 Trill::Trill()
-    : ControlElement("trill-"), TimeSpanningInterface(), AttColor(), AttExtender(), AttOrnamentAccid(), AttPlacement()
+    : ControlElement("trill-")
+    , TimeSpanningInterface()
+    , AttColor()
+    , AttExtender()
+    , AttLineRend()
+    , AttOrnamentAccid()
+    , AttPlacement()
 {
     RegisterInterface(TimeSpanningInterface::GetAttClasses(), TimeSpanningInterface::IsInterface());
     RegisterAttClass(ATT_COLOR);
     RegisterAttClass(ATT_EXTENDER);
+    RegisterAttClass(ATT_LINEREND);
     RegisterAttClass(ATT_ORNAMENTACCID);
     RegisterAttClass(ATT_PLACEMENT);
 
@@ -41,6 +48,7 @@ void Trill::Reset()
     TimeSpanningInterface::Reset();
     ResetColor();
     ResetExtender();
+    ResetLineRend();
     ResetOrnamentAccid();
     ResetPlacement();
 }
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -917,22 +917,22 @@ void View::DrawTrillExtension(
     assert(trill);
     assert(staff);
 
-    int y = trill->GetDrawingY() + m_doc->GetDrawingUnit(staff->m_drawingStaffSize) / 2;
+    int y
+        = trill->GetDrawingY() + m_doc->GetGlyphHeight(SMUFL_E566_ornamentTrill, staff->m_drawingStaffSize, false) / 3;
 
     // Adjust the x1 for the tr symbol
-    if ((spanningType == SPANNING_START) || (spanningType == SPANNING_START_END)) {
-        x1 += m_doc->GetGlyphWidth(SMUFL_E566_ornamentTrill, staff->m_drawingStaffSize, false);
+    if (trill->GetLstartsym() == LINESTARTENDSYMBOL_none) {
+        x1 -= trill->GetStart()->GetDrawingRadius(m_doc);
+        y += m_doc->GetDrawingUnit(staff->m_drawingStaffSize);
+    }
+    else if ((spanningType == SPANNING_START) || (spanningType == SPANNING_START_END)) {
+        int offsetFactor = 2;
+        if (trill->GetStart()->Is(TIMESTAMP_ATTR)) offsetFactor = 1;
+        x1 += m_doc->GetGlyphWidth(SMUFL_E566_ornamentTrill, staff->m_drawingStaffSize, false) / offsetFactor;
     }
 
-    // Adjust the x2 for extensions with @endid
-    if ((spanningType == SPANNING_END) || (spanningType == SPANNING_START_END)) {
-        LayerElement *end = trill->GetEnd();
-        assert(end);
-        if (!end->Is(TIMESTAMP_ATTR)) {
-            x2 = end->GetContentLeft()
-                - m_doc->GetGlyphWidth(SMUFL_E59D_ornamentZigZagLineNoRightEnd, staff->m_drawingStaffSize, false) / 2;
-        }
-    }
+    // Adjust the x2 for endid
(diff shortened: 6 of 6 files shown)


craigsapp:
One of the IDs has `#` and the other does not:

```
<trill startid="#one" endid="two" />
<trill startid="#three" endid="four" lstartsym="none" />
```

The start of the wavy line looks great in the After view (with the removed gap after the `tr`), but the ending point should not go beyond the left side of the ending note, or the note at the ending timestamp.  Either extra space should be added to the music to fit in an integral number of waves, or the waves must stop before the left edge of the next note.


Screen Shot 2020-04-05 at 04 43 50

Screen Shot 2020-04-05 at 04 43 57

The parts of the trill lines marked in red are strange.  A little overlap of the tail of a wave with the next note looks good.

For the ending points the first two are by note ID and the second two are by timestamp.  Since the timestamp is the time of the ending notes, I am wondering why all four line endings do not look the same?

How would a timestamp of `1m+0` look?   Would that stop at the barline of the next measure?

Also how would a `@tstamp2="5"` look like?  I would expect this to end at the barline (and not go over).  I would prefer `tstamp2="1m+0"` to look the same as `@tstamp2="5"`, but that is not a requirement.  It would just be more convenient to give the barline as `1m+0` rather than `5` since that would make it safer to change the time signature without messing up the endpoint of the trill.

Also, how would/should grace notes before the `@tstamp2` affect the length of the trill?  In general the ending grace notes of a trill are not under the trill, so I would think that grace notes coming before `@tstamp2` would stop the wavy line before the grace notes.  Using `@endid` would require the ID of the first grace notes to look the same.

rettinghaus:
With `@tstamp2="1m+0"` or `@tsamp2="0m+5"` it ends at the barline. 
But using `@endid` *includes* the last note, so it is supposed to be longer. 
Compare it with `<beamSpan>`, `<bracketSpan>`, `<octave>` etc.

craigsapp:
> But using `@endid` includes the last note, so it is supposed to be longer.
Compare it with <beamSpan>, <bracketSpan>, <octave> etc.

Then the question is whether the note referenced in `@endid` is supposed to be trilled or not.  If it is not supposed to be trilled, then my comment above remains valid.  But then, as you note, the meaning of `@endid` would be different from `<octave>`.  However, the graphical syntax of trill lines are different from ottava lines.  Ottava lines clearly are **inclusive** of the note referenced in the `@endid`.  Wavy lines should be **exclusive** in order for `@endid` to make sense.  The areas marked in red in the example above are strictly not allowed to occur in music notation.  The inclusive/exclusive property should not be possessed by `@endid` but rather should be owned by the element it occurs within.  So what I am saying is that the note referenced in `@endid` should not be included in the trill, but rather the trill stops at that note.  For `<octave>`, the ending note is inclusive, so the bracket must be placed after the start of the note.

There can be two possibilities for the meaning of `@endid`:

(1) The note is not in the trill (exclusive)
or
(2) The note is in the trill (inclusive)

If the note is not in the trill, then the wavy line cannot cross the left edge of the note (the tail of the wavyline extending over this point is OK).  If the note is in the trill, then the wavy line must continue to the end of the duration of the note.  So in both cases, the areas marked in red are not possible.

If the trill were inclusive, then there will be complications in controlling the visual end of the wavy line: in less common notational styles, the ending notes of a trill would be under the wavy line, but usually the wavy line should end before the terminal notes of the trill (since those notes are not themselves being trilled).

rettinghaus:
@craigsapp basically we are on the same page, so I changed the endid behavior now. Otherwise you would have to check alway the duration of the ending note to extend the line, which may disagree with the duration value of the trill. And there may be other caveats. 
But still I'm not comfortable with the (possible) inconsistent meaning of `@endid`.

rettinghaus:
trill

craigsapp:
The update looks great.

> But still I'm not comfortable with the (possible) inconsistent meaning of `@endid`.

I can see your concern, but it is being overly cautious.  I claim that the meaning of this attribute cannot be intrinsic to the attribute itself, but must come from the element that possesses it.  If you agree with that, then you should not worry.

If there is a problem with leaving the refined interpretation to the element, then there must be a new attribute created called `@endxid` (or similar) which means attach to the specified end element, but attach with the endpoint being exclusive rather than inclusive.   However, that seems like overkill.   Trill lines, and any many other lines such as a gliss, or a melodic-continuation line across staves are exclusive.  Brackets are usually inclusive (all of the cases I can think of at least).

Hairpin lines for crescendos are a special case.  Sometimes it is better than they end at the left edge of a note, while in other cases the right edge is better.  This could alternately be controlled by a justification attribute that specifies which side of the note (or the middle of the note) the hairpin should start/end on.

PR #1410, opened 2020-04-11
add support for group-barline

rettinghaus:


wergo:
@rettinghaus Thanks for fixing extender and ftrem issues!

PR #1420, opened 2020-04-16
improve bTrem and pedal in MusicXML import

rettinghaus:
This PR improves `bTrem` recognition (for cases without tremolo-type or unmeasured tremolos) and adds support for sostenuto pedal from MusicXML 3.1

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2625,7 +2625,7 @@ void MusicXmlInput::ReadMusicXmlNote(
         for (iter = m_trillStack.begin(); iter != m_trillStack.end(); ++iter) {
             int measureDifference = m_measureCounts.at(measure) - iter->second.m_lastMeasureCount;
             if (atoi(((iter->first)->GetN()).c_str()) == extNumber) {
-                (iter->first)->SetTstamp2(std::pair<int, double>(measureDifference, m_durTotal + 0.9999));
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -2193,6 +2187,14 @@ void MusicXmlInput::ReadMusicXmlNote(
             fTrem->SetBeams(beamFloatNum + beamAttachedNum);
             fTrem->SetBeamsFloat(beamFloatNum);
         }
+        else if (!HasAttributeWithValue(tremolo.node(), "type", "stop")) {
```
wergo, line 2190: Shouldn't that be `"type", "single"`?

rettinghaus: No. 
1. type attribute is not required (backward compatibility to MusicXML 1.1) and 
2. MusicXML 3.1 brings `type="unmeasured"`.

So there are three possibilities to encode a MEI `bTrem` in MusicXML.
See https://www.musicxml.com/for-developers/version-history/
(and https://github.com/w3c/musicxml/pull/200/files)

wergo: Ah, I have not looked at the version history, just at the manual: http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-tremolo.htm, but apparently that's not the latest info.

... and I have missed the `!`sign; sorry for that. I understand now. Thanks for clarification!

PR #1435, opened 2020-04-30
add cross-staff example

rettinghaus:


_tests/cross-staff/cross-staff-008.mei
```diff
@@ -0,0 +1,94 @@
...
+                </layer>
+              </staff>
+            </measure>
+            <measure>
+              <staff n="1">
+                <layer n="1">
+                  <beam beam.with="below">
+                    <note dots="1" dur="8" oct="5" pname="c"/>
+                    <space dur="16"/>
+                  </beam>
```
lpugin, line 77: Does this really make sense? It is a 'high' pitched beamed with the staff below. Also, I would not use `<space>` in beams for this basic example. I am not sure it is supported and should be tested separately.

rettinghaus: removed this measure

PR #1465, opened 2020-05-21
add VSCode tasks

rettinghaus:
This PR adds a `tasks.json` to build Verovio directly from VSCode on hitting `Ctrl+Shift+B`. 
For now it simply invokes cmake and make, only needs gcc to be installed. We could add further tasks later to build the js or the java toolkit.


rettinghaus:
… tested on linux and macos.

PR #1466, opened 2020-05-21
read phrase as slur

rettinghaus:


lpugin:
Not sure about this. Why don't you want to simply implement `<phrase>`?

rettinghaus:
I thought about this. I started implementing phrase, but gave up, as I didn’t want to copy viewslur. Viewslur should be something more generic like viewcurve to be used for DrawSlur DrawTie and DrawPhrase. Also we should introduce a MEI meta element for those shared attributes. 
That seemed to be a lot, and I wouldn’t dare to do this alone. 
My approach here looked like a useful addition for now.

lpugin:
I see your point. Actually, my preference on this would be to see `<phrase>` remove from MEI and replace by an attribute on `@slur`, which I think would be sufficient. It could be something like `@func="phase"` or `@phrase="true"`.

rettinghaus:
@lpugin sounds ok. But even then, Verovio would need to update the MEI while loading. ;-)

kepper:
I understand the wish to get rid of phrase marks – it's yet another curvy line to deal with. I german, we have the wonderful world "Bogensetzung" (roughly: placement of curved lines / arcs), which includes ties, slurs, phrase marks, legato "arcs" – basically all kinds of Bogen / arc. Something like that could make sense in MEI – a `<slur/>` would then be syntactic sugar for an `<arc func="slur"/>` (or whatever names we pick – no preference yet on my end). I would not, however, use `<slur/>` for this generic thing, as a slur is something specific.

lpugin:
Yes, once it it clarified in MEI we can intregrate this for backward compatibility with MEI 4.0

PR #1479, opened 2020-06-02
Added font style support

Pryadka:
All changes were done according to discussion https://github.com/rism-ch/verovio/issues/1460

- read character size information for bold and italic fonts;
- apply corresponding sizes for characters;

### DIFF
--- a/include/vrv/devicecontextbase.h
+++ b/include/vrv/devicecontextbase.h
@@ -102,8 +102,8 @@ public:
     {
         m_pointSize = 0;
         m_family = 0; // was wxFONTFAMILY_DEFAULT;
-        m_style = FONTSTYLE_normal;
-        m_weight = FONTWEIGHT_normal;
+        m_style = FONTSTYLE_NONE;
+        m_weight = FONTWEIGHT_NONE;
         m_underlined = false;
         m_supSubScript = false;
         m_faceName.clear();
--- a/include/vrv/devicecontextbase.h
+++ b/include/vrv/devicecontextbase.h
@@ -102,8 +102,8 @@ public:
     {
         m_pointSize = 0;
         m_family = 0; // was wxFONTFAMILY_DEFAULT;
-        m_style = FONTSTYLE_NONE;
-        m_weight = FONTWEIGHT_NONE;
+        m_style = FONTSTYLE_normal;
+        m_weight = FONTWEIGHT_normal;
         m_underlined = false;
         m_supSubScript = false;
         m_faceName.clear();
@@ -113,8 +113,8 @@ public:
 
     // accessors and modifiers for the font elements
     int GetPointSize() { return m_pointSize; }
-    int GetStyle() { return m_style; }
-    int GetWeight() { return m_weight; }
+    data_FONTSTYLE GetStyle() { return m_style; }
+    data_FONTWEIGHT GetWeight() { return m_weight; }
     bool GetUnderlined() { return m_underlined; }
     bool GetSupSubScript() { return m_supSubScript; }
     std::string GetFaceName() { return m_faceName; }
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -138,8 +138,8 @@ public:
      */
     ///@{
     /** Resource path */
-    static std::string GetPath() { return m_path; }
-    static void SetPath(const std::string &path) { m_path = path; }
+    static std::string GetPath() { return s_path; }
+    static void SetPath(const std::string &path) { s_path = path; }
     /** Init the SMufL music and text fonts */
     static bool InitFonts();
     /** Init the text font (bounding boxes and ASCII only) */
@@ -159,12 +159,12 @@ private:
 
 private:
     /** The path to the resources directory (e.g., for the svg/ subdirectory with fonts as XML */
-    static std::string m_path;
+    static std::string s_path;
     /** The loaded SMuFL font */
-    static GlyphMap m_font;
+    static GlyphMap s_font;
     /** A text font used for bounding box calculations */
-    static GlyphTextMap m_textFont;
-    static StyleAttributes m_currentStyle;
+    static GlyphTextMap s_textFont;
+    static StyleAttributes s_currentStyle;
     static const StyleAttributes k_defaultStyle;
 };
 
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -21,6 +21,8 @@
 #include <time.h>
 #endif
 
+#include "atttypes.h"
+
 namespace vrv {
 
 class Glyph;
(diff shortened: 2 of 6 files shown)


src/verse.cpp
```diff
@@ -175,7 +176,11 @@ int Verse::AdjustSylSpacing(FunctorParams *functorParams)
 
     // Use the syl because the content bounding box of the verse might be invalid at this stage
     int overlap = params->m_lastSyl->GetContentRight() - (firstSyl->GetContentLeft() + xShift);
-    overlap += lastSyl->CalcConnectorSpacing(params->m_doc, params->m_staffSize);
+    overlap += params->m_lastSyl->CalcConnectorSpacing(params->m_doc, params->m_staffSize);
+
+    if (params->m_lastSyl->GetUuid() == "syl-0000001599481773") {
+        int d = params->m_doc->GetDrawingUnit(params->m_staffSize);
+    }
```
lpugin, line 183: ??

src/verse.cpp
```diff
@@ -138,6 +138,7 @@ int Verse::AdjustSylSpacing(FunctorParams *functorParams)
             Syl *syl = dynamic_cast<Syl *>(*iter);
             assert(syl);
             syl->SetDrawingXRel(previousSylShift);
+            auto t = syl->GetUuid();
```
lpugin, line 141: Don't use `auto` in such cases

include/vrv/devicecontextbase.h
```diff
@@ -102,8 +102,8 @@ class FontInfo {
     {
         m_pointSize = 0;
         m_family = 0; // was wxFONTFAMILY_DEFAULT;
-        m_style = FONTSTYLE_NONE;
-        m_weight = FONTWEIGHT_NONE;
+        m_style = FONTSTYLE_normal;
+        m_weight = FONTWEIGHT_normal;
```
lpugin, line 106: Doesn't this mean that `normal` font style / weight will be added to every single text element in the SVG? I think we want to avoid this. @rettinghaus ?

src/view_element.cpp
```diff
@@ -1399,6 +1399,9 @@ void View::DrawSyl(DeviceContext *dc, LayerElement *element, Layer *layer, Staff
         dc->SetFont(m_doc->GetDrawingLyricFont(staff->m_drawingStaffSize));
     }
 
+    dc->GetFont()->SetStyle(syl->HasFontstyle() ? syl->GetFontstyle() : data_FONTSTYLE::FONTSTYLE_normal);
+    dc->GetFont()->SetWeight(syl->HasFontweight() ? syl->GetFontweight() : data_FONTWEIGHT::FONTWEIGHT_normal);
```
lpugin, line 1403: No need to prefix `FONTSTYLE_normal` and `FONTWEIGHT_normal`
Again, do we always want / need the font style / weigh to be set to normal?

src/view_text.cpp
```diff
@@ -356,6 +356,10 @@ void View::DrawText(DeviceContext *dc, Text *text, TextDrawingParams &params)
 
     dc->StartTextGraphic(text, "", text->GetUuid());
 
+    if(dc->GetFont()->GetWeight() != FONTWEIGHT_NONE && dc->GetFont()->GetStyle() != FONTSTYLE_NONE) {
```
lpugin, line 359: Please use `HasWeight()` and `HasStyle()` instead

src/verse.cpp
Pryadka: Sorry, this is about my further work. I'll remove it.

src/view_text.cpp
Pryadka: `FontInfo `doesn't have such methods. Should I implement them?

include/vrv/devicecontextbase.h
rettinghaus: Definitely! We have Verovio's "house style" set in the svg style element. So `@fontstyle` should only be added, if it is used in the input file.

Pryadka: Ok. I will care about it.

src/view_element.cpp
Pryadka: Currently if style/weight is not specified in the input, we assume that it is normal. Normal means default. From rendering point of view a font type cant be undefined. And for sure we can omit these attributes in the output if they are _normal.

src/verse.cpp
Pryadka: The same. Will be removed.

src/view_text.cpp
lpugin: Ah, sorry, I thought it was checking against an attribute class. You can leave it like this

Pryadka:
@lpugin  Please review it again.

include/vrv/vrv.h
```diff
@@ -128,35 +130,42 @@ bool Check(Object *object);
...
     /** The path to the resources directory (e.g., for the svg/ subdirectory with fonts as XML */
     static std::string m_path;
     /** The loaded SMuFL font */
-    static std::map<wchar_t, Glyph> m_font;
+    static GlyphMap m_font;
     /** A text font used for bounding box calculations */
-    static std::map<wchar_t, Glyph> m_textFont;
+    static GlyphTextMap m_textFont;
+    static StyleAttributes m_currentStyle;
+    static const StyleAttributes k_defaultStyle;
```
lpugin, line 168: Sorry, this was my mistake, but static member as normally prefixed with `s_` and not `m_`. Could you change them? Thanks!

src/verse.cpp
```diff
@@ -175,7 +175,7 @@ int Verse::AdjustSylSpacing(FunctorParams *functorParams)
 
     // Use the syl because the content bounding box of the verse might be invalid at this stage
     int overlap = params->m_lastSyl->GetContentRight() - (firstSyl->GetContentLeft() + xShift);
-    overlap += lastSyl->CalcConnectorSpacing(params->m_doc, params->m_staffSize);
+    overlap += params->m_lastSyl->CalcConnectorSpacing(params->m_doc, params->m_staffSize);
```
lpugin, line 178: What is the reason for this change? Was it wrong?

src/vrv.cpp
```diff
@@ -100,13 +120,34 @@ Glyph *Resources::GetGlyph(wchar_t smuflCode)
...
+}
+
 Glyph *Resources::GetTextGlyph(wchar_t code)
 {
-    if (!m_textFont.count(code)) return NULL;
-    return &m_textFont[code];
+    GlyphMap *currentMap = &m_textFont[m_textFont.count(m_currentStyle) != 0 ? m_currentStyle : k_defaultStyle];
+
+    if (currentMap->count(code) == 0)
+            return NULL;
```
lpugin, line 145: On one line or with {}

src/verse.cpp
Pryadka: Logical. lastSyl is current last syllable. But in this line we calculate space between current syllable and previous syllable. This fix is not related to the current task.

src/vrv.cpp
Pryadka: Done.

src/verse.cpp
lpugin: Correct. Leave it

include/vrv/vrv.h
Pryadka: Done

PR #1483, opened 2020-06-04
Adjust measure

Pryadka:
As it was mentioned in this discussion https://github.com/rism-ch/verovio/issues/1460 the implementation of text-font support is not enough for fixing syllable overlapping issue. It turns out that DIR element affects measure layouting. If it's right border lays farther than the measure right border layouting inside it will be broken. In order to get rid of this problem corresponding method for adjusting X coordinates should be implemented.

### DIFF
--- a/include/vrv/devicecontextbase.h
+++ b/include/vrv/devicecontextbase.h
@@ -102,8 +102,8 @@ public:
     {
         m_pointSize = 0;
         m_family = 0; // was wxFONTFAMILY_DEFAULT;
-        m_style = FONTSTYLE_normal;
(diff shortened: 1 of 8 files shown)


lpugin:
Thanks. I think most of this was already merged https://github.com/rism-ch/verovio/pull/1479
Could you double check why the PR is duplicating what was already merged?

Pryadka:
> Thanks. I think most of this was already merged #1479
> Could you double check why the PR is duplicating what was already merged?

Sorry, I did not remove merged files from the commit. I will do another request in a couple of minutes.

Issue #1460, opened 2020-05-15
Support for bold text font in lyrics


Pryadka:
Sometimes rendering of text has different view and one of this representation looks like a bug. 
Wrong represantation:
![BigMei](https://user-images.githubusercontent.com/5373894/82040051-35d35000-96a6-11ea-8c9a-690d1113499c.png)
The same xml-snippet in another file
![SinipetFromBigMei](https://user-images.githubusercontent.com/5373894/82040125-5bf8f000-96a6-11ea-9b36-fd6370d4f98a.png)
It looks like a bug in `int Verse::AdjustSylSpacing(FunctorParams *functorParams)`

lpugin:
Please post an MEI snippet with one single measure / staff with the problem

Pryadka:
this is staff with a problem. **You can see wrong text in the middle of 9 page**
[scorecopy9.mei.txt](https://github.com/rism-ch/verovio/files/4634267/scorecopy9.mei.txt)

and this is corresponding snippet from this file put to empty .mei
[verse.mei.txt](https://github.com/rism-ch/verovio/files/4634271/verse.mei.txt)

Please remove .txt extension in order to use it.

lpugin:
Thanks. Next time post the MEI snippet inline like https://github.com/rism-ch/verovio/issues/1455

The main issue is the use of `@fontweight="bold"`.  `DeviceContext::GetTextExtent` does not take into account the fontweight. To fix this, you need to modify the Resources class so it can load and use different fontweights / styles. Which means
* Have additional maps (same as  `Resources::m_textFont`) for `Times-italic`, `Times-italic-bold`, and `Times-bold` (see `./data/text`)
* Load the fonts following the normal text font model (call `InitTextFont("Times-italic")`, etc.)
* Add a `Resources::SelectTextFont(data_FONTWEIGHT fontWeight, data_FONTSTYLE fontStyle) method that can be called before GetTextExtent is call in order to select the appropriate font.

There might be some spacing issue to fix too, but you need to implement this first.

Pryadka:
Hello @lpugin . I implemented text-font loading and using. It works but still doesn't fix the issue. Before creating pull request for this implementation I'd like to discuss one point. 
[...]

PR #1484, opened 2020-06-04
Adjust measure right border if it doesn't cover all it's children

Pryadka:
As it was mentioned in this discussion https://github.com/rism-ch/verovio/issues/1460 the implementation of text-font support is not enough for fixing syllable overlapping issue. It turns out that DIR element affects measure layouting. If it's right border lays farther than the measure right border layouting inside it will be broken. In order to get rid of this problem corresponding method for adjusting X coordinates should be implemented.

### DIFF
--- a/include/vrv/measure.h
+++ b/include/vrv/measure.h
@@ -297,6 +297,7 @@ public:
      * See Object::AdjustXOverflow
      */
     virtual int AdjustXOverflow(FunctorParams *functorParams);
+    virtual int AdjustXOverflowEnd(FunctorParams *functorParams);
 
     /**
      * See Object::AdjustXPos
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -864,6 +864,31 @@ int Measure::AdjustXOverflow(FunctorParams *functorParams)
     return FUNCTOR_CONTINUE;
 }
 
+int Measure::AdjustXOverflowEnd(FunctorParams *functorParams)
+{
+    AdjustXOverflowParams *params = dynamic_cast<AdjustXOverflowParams *>(functorParams);
+
+    if (!params->m_currentWidest) {
+        return FUNCTOR_CONTINUE;
+    }
+
+    int measureRightX = GetDrawingX() + GetRightBarLineLeft() - params->m_margin;
+    if (measureRightX < params->m_currentWidest->GetContentRight()) {
+        LayerElement *objectX = dynamic_cast<LayerElement *>(params->m_currentWidest->GetObjectX());
+        if (!objectX) {
+            return FUNCTOR_CONTINUE;
+        }
+        Alignment *left = objectX->GetAlignment();
+        ArrayOfAdjustmentTuples boundaries{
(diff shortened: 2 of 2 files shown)


lpugin:
Do you have a minimal test file? I am not sure I understand why this is now done at the end of every measure (and not only at the end of a system). In many cases, we want content (text) to overlap measures. I am overlooking something?

lpugin:
What do you get with this?
```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Tests in space</title>
         </titleStmt>
         <pubStmt>
            <date>2017-04-27</date>
         </pubStmt>
         <notesStmt>
            <annot>Verovio appropriately places multiple directives, spacing them as desired.</annot>
            <annot type="warning">This isn't working as supposed in Safari due to <ref target="https://bugs.webkit.org/show_bug.cgi?id=112032">this bug</ref>.</annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="unknown">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4" lines="5"/>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <note dur="1" oct="4" pname="c">
                              <accid accid.ges="n"/>
                           </note>
                        </layer>
                     </staff>
                     <dir place="above" staff="1" tstamp="1">Allegro <rend>molto</rend></dir>
                  </measure>
                  <measure n="2">
                     <staff n="1">
                        <layer n="1">
                           <note dur="1" oct="4" pname="d">
                              <accid accid.ges="n"/>
                           </note>
                        </layer>
                     </staff>
                  </measure>
                  <measure n="3">
                     <staff n="1">
                        <layer n="1">
                           <note dur="1" oct="4" pname="e">
                              <accid accid.ges="n"/>
                           </note>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

Pryadka:
Without the changes it looks like this
without-changes

and with them
with_changes

In this case, we stretched the measure. This example has no lyrics. If it has some kind of lyrical syllable, it will be broken. We can try another solution, just don't register DIR items as m_currentWidest. Such a solution may lead to another problem, and I can’t even guess which one.

lpugin:
Could you post a minimal test file for the issue you are trying to solve now that the font issue in #1460 is solved?

craigsapp:
This behavior it not good:

Screen Shot 2020-06-05 at 1 14 24 AM

Unless the text extends into the page margin at a system break.

Consider this example:

Screen Shot 2020-06-05 at 1 13 45 AM

"si deve..." text extends across 1/2 of the page, and it is not desirable for the text direction to be forced to avoid crossing the barline.

PR #1485, opened 2020-06-04
Improve tempo changes in MIDI output

rettinghaus:
Until now tempo changes were moved to the beginning of the measure in MIDI output. 
I don't know what to do to change timemap generation accordingly.

### DIFF
--- a/include/vrv/tempo.h
+++ b/include/vrv/tempo.h
@@ -62,6 +62,11 @@ public:
     // Functors //
     //----------//
 
+    /**
+     * See Object::GenerateMIDI
+     */
+    virtual int GenerateMIDI(FunctorParams *functorParams);
+
 private:
     //
 public:
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -1207,24 +1207,6 @@ int Measure::CalcMaxMeasureDuration(FunctorParams *functorParams)
     m_scoreTimeOffset.push_back(params->m_maxCurrentScoreTime);
     params->m_maxCurrentScoreTime += m_measureAligner.GetRightAlignment()->GetTime() * DURATION_4 / DUR_MAX;
 
-    // search for tempo marks in the measure
-    Tempo *tempo = dynamic_cast<Tempo *>(this->FindDescendantByType(TEMPO));
-    if (tempo && tempo->HasMidiBpm()) {
-        params->m_currentTempo = tempo->GetMidiBpm();
-    }
-    else if (tempo && tempo->HasMm()) {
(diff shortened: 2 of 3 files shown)


lpugin:
Yes, they where moved to the beginning of the measure because very often this is where the tempo changes occur. Now they are moved to the end, which is not really an improvement. If you want to improve it, I think the best is to revert these changes. Then, still at the beginning of the measure, look at the alignment of all tempo indications in the measure (we can have more than one) and store them. This means that `Measure::m_currentTempo` needs to become a strcuture that can store multiple tempo values with their time position. Then, when generating the timemap, they have to be used to change the tempo each time a time position is passed. 

We can anticipate cases where it will be quite complicated to make it work. For example:
image
```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title />
            <respStmt />
         </titleStmt>
         <pubStmt><availability />
         </pubStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application isodate="2020-06-03T22:02:52" version="3.0.0-dev-cb54f8c">
               <name>Verovio</name>
               <p>Transcoded from MusicXML</p>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffGrp symbol="bracket">
                        <staffDef n="1" lines="5" clef.shape="G" clef.line="2" key.sig="1f" meter.count="3" meter.unit="4">
                           <label>P1</label>
                        </staffDef>
                        <staffDef n="2" lines="5" clef.shape="G" clef.line="2" key.sig="1f" meter.count="3" meter.unit="4">
                           <label>P2</label>
                        </staffDef>
                     </staffGrp>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <note dur="4" oct="5" pname="c" />
                           <note dur="4" oct="5" pname="c" />
                           <note dur="4" oct="5" pname="c" />
                        </layer>
                     </staff>
                     <staff n="2">
                        <layer n="1">
                           <note dur="2" oct="5" pname="c" />
                           <note dur="4" oct="5" pname="c" />
                        </layer>
                     </staff>
                     <tempo staff="1" tstamp="2.0"><rend fontname="VerovioText">&#xE1D3;</rend> = 70</tempo>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>

```

rettinghaus:
Will rethink this …

PR #1493, opened 2020-06-08
refactor labelling

rettinghaus:
This PR brings a refactoring of the labelling of staffs and staff groups, in order to handle more complex situations.

## before
before

## after
after


lpugin:
Could you add it as a test file in the score section?

rettinghaus:
will do …

PR #1495, opened 2020-06-09
Brace font support to rism

Pryadka:
Added usage of font-dependent braces.


src/view_page.cpp
```diff
@@ -551,7 +553,26 @@ void View::DrawBracketsq(DeviceContext *dc, int x, int y1, int y2, int staffSize
...
 {
     assert(dc);
-
+    if (m_doc->GetOptions()->m_useBraceGlyph.GetValue()) {
+        Glyph *glyph = Resources::GetGlyph(SMUFL_E000_brace);
+        if (glyph) {
+            FontInfo* font = m_doc->GetDrawingSmuflFont(staffSize, false);
+            int dummy, g_w, g_h;
+            glyph->GetBoundingBox(dummy, dummy, g_w, g_h);
+            const float ratio = static_cast<float>(glyph->GetUnitsPerEm()) / font->GetPointSize();
```
lpugin, line 562: I think you can use `Doc::GetGlyphHeight`

src/view_page.cpp
```diff
@@ -48,6 +48,8 @@
 #include "text.h"
 #include "tuplet.h"
 #include "vrv.h"
+#include "glyph.h"
+#include "doc.h"
```
lpugin, line 52: Alywas order alphabetically

src/options.cpp
```diff
@@ -647,6 +647,10 @@ Options::Options()
     m_usePgHeaderForAll.Init(false);
     this->Register(&m_usePgHeaderForAll, "usePgHeaderForAll", &m_general);
 
+    m_useBraceGlyph.SetInfo("Use Brace Glyph", "Use brace glyph from current font");
+    m_useBraceGlyph.Init(false);
+    this->Register(&m_useBraceGlyph, "useBraceGlyph", &m_general);
```
lpugin, line 652: Order alphabetically

include/vrv/svgdevicecontext.h
```diff
@@ -14,6 +14,7 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <set>
```
lpugin, line 17: Order alphabetically

src/view_page.cpp
```diff
@@ -551,7 +553,26 @@ void View::DrawBracketsq(DeviceContext *dc, int x, int y1, int y2, int staffSize
 void View::DrawBrace(DeviceContext *dc, int x, int y1, int y2, int staffSize)
 {
     assert(dc);
-
+    if (m_doc->GetOptions()->m_useBraceGlyph.GetValue()) {
+        Glyph *glyph = Resources::GetGlyph(SMUFL_E000_brace);
+        if (glyph) {
```
lpugin, line 558: `assert(glyph)` would be enough since all the glyphs listed in smufl.h have to be available.

src/view_page.cpp
Pryadka: Done

include/vrv/svgdevicecontext.h
Pryadka: Done

src/view_page.cpp
Pryadka: Done

src/view_page.cpp
Pryadka: I guess it's not good idea. In this case we will call `Doc::GetGlyphHeight` and `Doc::GetGlyphWidth` . Both methods call `GetBoundingBox` and do another calculations which is not necessary.

src/options.cpp
Pryadka: Done

src/view_page.cpp
```diff
@@ -555,23 +554,20 @@ void View::DrawBrace(DeviceContext *dc, int x, int y1, int y2, int staffSize)
...
-            font->SetWidthToHeightRatio(static_cast<float>(braceWidth) / widthAfterScalling);
-            DrawSmuflCode(dc, x, y2, SMUFL_E000_brace, staffSize * ratio * scale, false);
-            font->SetWidthToHeightRatio(currentWidthToHeightRatio);
-            return;
-        } else {
-            LogError("Cant find brace glyph(SMUFL_E000_brace). Use the standart brace pattern.");
-        }
+        assert(glyph);
+        FontInfo* font = m_doc->GetDrawingSmuflFont(staffSize, false);
+        int dummy, g_w, g_h;
```
lpugin, line 559: Doc::GetGlyphHeight does not work for this?

Pryadka: No. It performs some calculations which are not needed.

lpugin: Could you be more specific? You should not have to call glyph->GetUnitsPerEm from the View. Also, since the SMuFL spec are explicit about the size of the brace glyph (i.e., the height of a single five-line stave) you should not have to retrieve it.

PR #1505, opened 2020-06-17
Introduced systemMaxPerPage option

alexgnt:
In case option is not set (0 by default) used standart logic for page
layout and system's margins calculation based on spacingSystem value.

When the option is set - space between systems speads equaly based
on the amount of systems which fits on page but not exceding options
value. I.e. first system alligned to the top of page, last one tailed to
the bottom of the page.


include/vrv/options.h
```diff
@@ -529,6 +529,7 @@ class Options {
     OptionDbl m_spacingNonLinear;
     OptionInt m_spacingStaff;
     OptionInt m_spacingSystem;
+    OptionInt m_systemMaxPerPage;
```
lpugin, line 532: Order alphabetically

src/doc.cpp
```diff
@@ -830,7 +830,8 @@ void Doc::CastOffDoc()
     Page *currentPage = new Page();
     CastOffPagesParams castOffPagesParams(contentPage, this, currentPage);
     CastOffRunningElements(&castOffPagesParams);
-    castOffPagesParams.m_pageHeight = this->m_drawingPageHeight - this->m_drawingPageMarginBot;
+    castOffPagesParams.m_pageHeight
+        = this->m_drawingPageHeight - this->m_drawingPageMarginTop - this->m_drawingPageMarginBot;
```
lpugin, line 834: Looks unrelated, doesn't it?

src/options.cpp
```diff
@@ -752,6 +752,10 @@ Options::Options()
     m_spacingSystem.Init(3, 0, 12);
     this->Register(&m_spacingSystem, "spacingSystem", &m_generalLayout);
 
+    m_systemMaxPerPage.SetInfo("Max. System per Page", "Maximun number of systems per page");
+    m_systemMaxPerPage.Init(0, 0, 24);
+    this->Register(&m_systemMaxPerPage, "systemMaxPerPage", &m_generalLayout);
+
```
lpugin, line 758: Alphabetical order

src/system.cpp
```diff
@@ -760,16 +760,31 @@ int System::CastOffPages(FunctorParams *functorParams)
...
+            assert(childSystem);
+
+            systemsTotalHeight += childSystem->GetHeight();
+        }
+
+        placeOnNewPage = systemsTotalHeight + this->GetHeight() > params->m_pageHeight - headerAndFooterHeight;
+    }
+
+    const auto currentShift = params->m_shift + headerAndFooterHeight;
+    if (placeOnNewPage || (this->m_drawingYRel - this->GetHeight() - currentShift < 0)) {
```
lpugin, line 787: Again, I don't think you need to change that much. If we need to fix footer and header spacing, please fix is separately.

src/doc.cpp
lpugin: The value passed is now `this->m_drawingPageContentHeight` and there is no need to deal with margins here anymore

include/vrv/options.h
alexgnt: Done.

src/options.cpp
alexgnt: Done.

src/doc.cpp
alexgnt: removed

src/system.cpp
alexgnt: That is not about footer and header. Since we need to place not more systems than systemMaxPerPage I have to calculate how match space left on the page to decide place it on new page or not.

lpugin: This is what I do not understand. The standard algorithm already does this. You only need to add one test in the condition, namely to check if the option is set and if the maximum number of system on the page is already reached.

alexgnt: Ok. I see the issue. Looks like castOfDoc first calls the LayOutVertically for the page which contains all systems of the document what is break logic of margin calculation and I have incorrect drawingYRel offset, as result systems overlapping and I had to calculate system height in place. 

It looks like I need somehow to distinguish that case and return 0 as a system margin in all cases when the maxPerPage is set and we do not do CastOffPages... In that case I will not need any additional calculation in that method.

Is there any simple way to distinguish that case? Could you please advice?

lpugin: Here: https://github.com/rism-ch/verovio/blob/df5c905d131b887730af9e7ec1afb1bf2543b8ec/src/system.cpp#L784
one additional test to check also if you have reached the maximum number of system on the current page. That should be it

lpugin: See this https://github.com/rism-ch/verovio/commit/b99f57b3adfdd413b3ed799535ebba5fb0f24b47 
With a value of 2 (as coded) here it produces this - but renders normally with 0 or with a value higher than what fits on the page (e.g. 10)
Please make your PR with the code for adding the option itself to this branch (develop-max-system)

image

alexgnt: Hi @lpugin , it looks ok when all systems can fit the page, but when it's not... the systems overlapping each other, since initial margin calculated incorrectly and drawingYRel are shifted. 

I see the following picture:
image

So, my idea is to set initial system margins to 0 during first LayOutVertically for the one page document, and them it should make proper calculation when layout on smaller pages. The problem for me now - is to distinguish this cases.

alexgnt: Perhaps, you do not really understand what I'm trying to do here... but the tricky part is margins, since first system should be alligned to the top of page, and the last one to the bottom. Margins should be equal in the middle, here is the sample (except last page):
image

src/page.cpp
```diff
@@ -73,6 +73,45 @@ bool Page::IsSupportedChild(Object *child)
     return true;
 }
 
+int Page::GetSystemMargin()
```
alexgnt, line 76: Here is how the margins calculated.

src/system.cpp
lpugin: How id you produce this example?? It this the develop-max-system branch? Could you post the file and the parameter you used? 
Sorry, I do not understand what you are trying to achieve. If the content does not fit on a page with --spacing-staff and --spacing-system set to 0, then I do not understand how you are going to make it fit on the page. This is why the example implementation I gave will not put the maximum number of system on a page if they do not fit. I don't understand how it yielded the overlap you a showing here.
Why can't you simply have --spacing-staff and --spacing-system to 0 and then justify the content of the page vertically with --justify-vertically (with our without --justify-systems-only)?

alexgnt: The mei file is this one: https://github.com/music-encoding/sample-encodings/blob/master/MEI_4.0/Music/Complete_examples/Grieg_op.43_butterfly.mei 
with the following parameters:
verovio -a --system-max-per-page 10 -r /work/verovio/data /work/Samples/Grieg_op.43_butterfly.mei

I'm using forked version as base, so I'll try cherry pick my commit on the top of rism develop and double check, I'll check your suggestion as well.

alexgnt:
Hi @lpugin, I've cleaned up everything and created new PR https://github.com/rism-ch/verovio/pull/1514 please have a look. Now to achieve required behavior combination of flags should be used as per your suggestion.

PR #1509, opened 2020-06-19
Features from Neon branch

JRegimbal:
This is Yet Another Neon PR™ containing work done as part of development of Neon. Most of the changes are restricted to the `EditorToolkitNeume` class, but a number also extend to classes across Verovio. Major changes are summarized here grouped by this scope.

#### Outside `EditorToolkitNeume`

*  `--use-rotate` option: When rendering systems, take into account the `@rotate` attribute. Only applicable in facsimile mode.
  * This results in adjustments to the y-value of rendered elements on a rotated system.
* `--create-default-syl` option: Create a default, empty `syl` element in cases where a `syllable` is read without a `syl`.
* `--create--default-syl-b-box` option: Create a default bounding box for `syl` elements without one based on the bounding boxes of the `syllable`. Only applies in facsimile mode.
* Write the `@follows` and `@precedes` attributes, if present, to the SVG. This is used to track syllables that span across line breaks.
* `Clef::PitchDistanceTo(Clef*)`, which gets the difference in steps between a note on the same line/space from the calling clef to the parameter clef.
* Addition of `PitchInterface::AdjustPitchForNewClef` which preserves the location of the pitched element (i.e. space or line number) by adjusting the pitch based on a change in clef.
* `FacsimileInterface::GetDrawingY` takes into account the value of `@rotate`.
* Addition of `FacsimileInterface::GetDrawingRotate` and `FacsimileInterface::GetHeight`
* Addition of `Object::FindNextChild`, `Object::FindPreviousChild`, `Object::GenerateBoundingBox`.
* Improvement of `Object::SetChildZones`, which associates `zone` elements with graphical elements.
* Addition of `Staff::GetDrawingRotate` and `Staff::AdjustDrawingStaffSize`
* Allow manual setting of height and width of box in `BBoxDeviceContext::DrawText`

#### Changes in `EditorToolkitNeume`
* Add `changeStaff`, `resizeRotate` actions.
* Reduce logic in `EditorToolkitNeume::Chain`.
* Add `AdjustPitchFromPosition` function.
  * Remove redundant logic in `EditorToolkitNeume::Drag`, `EditorToolkitNeume::Insert`.
* Improvements to `EditorToolkitNeume::Group`, `EditorToolkitNeume::Ungroup` to handle `syl` elements.
* Return info objects.
* Add stable clef reordering algorithm.


include/vrv/bboxdevicecontext.h
```diff
@@ -80,8 +80,8 @@ class BBoxDeviceContext : public DeviceContext {
     virtual void DrawRectangle(int x, int y, int width, int height);
     virtual void DrawRotatedText(const std::string &text, int x, int y, double angle);
     virtual void DrawRoundedRectangle(int x, int y, int width, int height, int radius);
-    virtual void DrawText(
```
caitlin-hutnyk, line 83: Is it worth mentioning this in the PR description? For me it's not clear why this function is being changed from the description

include/vrv/clef.h
```diff
@@ -57,6 +57,8 @@ class Clef : public LayerElement,
      */
     static int ClefId(data_CLEFSHAPE shape, char line, data_OCTAVE_DIS octaveDis, data_STAFFREL_basic place);
 
+    int PitchDistanceTo(Clef *clef);
```
caitlin-hutnyk, line 60: This might be totally my bad from last summer, but it seems like this function (and some of the other clef helper functions) might be mostly unused? If they're only called once or twice it might not be work having them

src/editortoolkit_neume.cpp
```diff
@@ -729,80 +894,98 @@ bool EditorToolkitNeume::Insert(std::string elementType, std::string staffId, in
...
     layer->ReorderByXPos();
+    m_infoObject.import("status", status);
+    m_infoObject.import("message", message);
     return true;
 }
 
 bool EditorToolkitNeume::Merge(std::vector<std::string> elementIds)
 {
-    m_editInfo = "";
+    m_infoObject.reset();
```
caitlin-hutnyk, line 933: If this is getting called by every edit action individually might it be worthwhile to have it be called by the wrapper functions or in some more built-in way?

include/vrv/clef.h
JRegimbal: Good point. It seems to only appear in `PitchInterface::AdjustPitchForNewClef`. I'm hesitant to remove it though just because it does seem like something that could be useful later.

include/vrv/bboxdevicecontext.h
JRegimbal: Ah yeah good point. It adds the option to set the height and width of the text manually rather than using `TextExtend`. If they're unset then `TextExtend` is used instead. This is for drawing the `syl` bounding boxes from the `zone` elements.

src/editortoolkit_neume.cpp
JRegimbal: Yeah this could probably be moved to the `ParseEditorAction` function and then failures in parsing could be included in the info object as well.

JRegimbal: Addressed this in 1c0cb06

src/object.cpp
```diff
@@ -1668,25 +1710,148 @@ int Object::ReorderByXPos(FunctorParams *functorParams)
...
+            *uly = std::min(*uly, zone->GetUly());
+            *lrx = std::max(*lrx, zone->GetLrx());
+            *lry = std::max(*lry, zone->GetLry());
+            result |= true;
+        }
+    }
+    return result;
+}
+
 int Object::SetChildZones(FunctorParams *functorParams)
```
lpugin, line 1740: This should definitely not have its main implementation in `Object` because it overloads it with something too specific. Maybe you can have a interface pseudo functor? See `TimeSpanningInterface::InterfaceResetDrawing` for example. Something like `FacsimileInterface::InterfaceSetChildZones` redirect from Object, or better, from LayerElement and Staff, which I think are the only two classes implementing FacsimilerInterface.

JRegimbal:
@lpugin I'll fix the issues you pointed out and then start trying to clean up the facsimile stuff in Object. Do you think it would be better suited in LayerElement?

lpugin:
> @lpugin I'll fix the issues you pointed out and then start trying to clean up the facsimile stuff in Object. Do you think it would be better suited in LayerElement?

OK, this is good. I will look again once it is merge since it will be easier for me to have an overview. The goal should be to move as much as possible into the FacsimileInterface class / files.

include/vrv/options.h
```diff
@@ -496,6 +496,9 @@ class Options {
     OptionBool m_svgHtml5;
     OptionInt m_unit;
     OptionBool m_useFacsimile;
+    OptionBool m_createDefaultSyl;
+    OptionBool m_createDefaultSylBBox;
+    OptionBool m_useRotate;
```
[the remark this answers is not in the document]
JRegimbal: The second one creates zones for syls that don't have one. Do you have an alternative in mind to the options? We only put them here so they were happening by default.

src/view_text.cpp
```diff
@@ -212,9 +220,18 @@ void View::DrawLyricString(DeviceContext *dc, std::wstring str, int staffSize)
...
+        }
+        else {
+            dc->DrawText(UTF16to8(str), str);
+        }
         dc->ResetFont();
     }
+
+    if (!wroteText && params) {
+        dc->DrawText("", L"", params->m_x, params->m_y, params->m_width, params->m_height);
+    }
```
lpugin, line 234: Could you comment on this?

include/vrv/options.h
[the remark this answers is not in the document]
lpugin: Are they cases where you turn them off? And cases where you turn only one for them on? Because both are very specific to neumes, so I am questioning the need for having options. Do we need them at all? If yes, do we need two? We can leave them but we would need better names.
For the first one, I don't like the fact the MEIInput changes the content of the tree while reading. In other cases, this is happening after the document is loaded. See methods such as Doc::ConvertMarkupDoc. It would be better to add something for syllables there (a bitfield MARKUP_SYLLABLE_CONTENT) and if possible fix both empty syllables and facsimile zones in one functor. Not sure it can work for facsimile zone, though. What do you think? It is a bit over the top to have an additional functor for this but it would be clearer.

src/view_text.cpp
JRegimbal: If there's no text but a zone is specified (providing the height and width) then draw a blank space.

lpugin: OK. This comment in the code would be useful - it has to be clear that this is only with facs/zone

include/vrv/options.h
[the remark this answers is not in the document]
JRegimbal: Not where we do with Neon, but if there was hypothetically a neume editor that didn't work with facsimiles then it would make sense to create default `syl` elements for editing but not create a `zone`. With `m_useRotate`, that just be wrapped into the facsimile option since we wrote this when `@rotate` was not standard.

I'd be fine with adding a functor to handle the syllables and zones outside of `MEIInput`. Do you think it makes sense to keep the two syl options separate? Or at all. It could just add a `syl` element since a syllable should have text and only add a zone if the facsimile option is passed.

src/view_text.cpp
JRegimbal: Got it!

include/vrv/options.h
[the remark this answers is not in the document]
lpugin: I think it is fine not to have the options at all if this works for you. That is, making sure there is a `<syl>` in `<syllable>` in all cases. You can add a `<zone>` only if the Doc is of type facs. This is acceptable. Comment the code as much as necessary to make all this clear.

(For the future: it would be best to avoid addition of content in Verovio, for various reasons. ConvertMarkup goal is primarily to convert to another representation - e.g., attribute into element. I would prefer it not to become a place where we "fix" the encoding")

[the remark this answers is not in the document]
JRegimbal: Got it! And sorry about including that kind of feature here.

JRegimbal:
Going to run some tests to make sure that what I've done so far hasn't broken anything. I'll add another comment if everything seems fine.

JRegimbal:
Things don't seem to be broken as far as I can tell.

JRegimbal:
> We are definitely getting there. I am only asking to move some code outside Doc::PrepareDrawing - if you really think it should go there. I am actually not too sure because depending on the use case, it can be called several times. Are this now going to add Zone elements every time it is called? It this what you need to happen?

It shouldn't add zone elements every time since it's checking if a zone is associated first, but this only needs to run once. Can you think of a better function that will only be called once after loading a file?

src/doc.cpp
```diff
@@ -727,11 +729,80 @@ void Doc::PrepareDrawing()
...
+                            (*it)->GetClassName().c_str());
+                        delete zone;
+                        continue;
+                    }
+                }
+                Object *surface = this->GetFacsimile()->FindDescendantByType(SURFACE);
+                assert(surface);
+                surface->AddChild(zone);
+                fi->SetZone(zone);
+            }
```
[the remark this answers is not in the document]
JRegimbal: I tried adding it to `PrepareFacsimile` but since the zone is created using zone information from other elements in the syllable in many cases, those zones will need to be associated first. Is there a way to have a functor handle sibling elements first before returning to a specific element?

src/syl.cpp
```diff
@@ -192,4 +193,48 @@ int Syl::ResetDrawing(FunctorParams *functorParams)
...
 
+bool Syl::CreateDefaultZone(Doc *doc)
+{
+    Zone *zone = new Zone();
+    const int offsetUly = 100;
+    const int offsetLrx = 100;
+    const int offsetLry = 200;
+
+    LayerElement *syllable = dynamic_cast<LayerElement *>(this->GetFirstAncestor(SYLLABLE));
+    assert(syllable);
```
lpugin, line 204: This is problematic because with non neume notation there will not be a Syllable parent. Just check and return instead of assert it.

src/doc.cpp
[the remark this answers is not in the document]
lpugin: Not quite. What you could add a PreperFacsimileEnd end functor and trigger the loop over m_zonelessSyls at the end. E.g., the end of the Doc, or the end of a System. But you can leave it like it is. This is just for the record.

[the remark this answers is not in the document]
JRegimbal: Ah okay thanks for letting me know!

lpugin:
Finger crossed!

PR #1514, opened 2020-06-23
System max per page

alexgnt:
Added new option which supposed to be used in combination with the
foolowing options (incl. sample values):
--spacing-system 0
--spacing-staff 0
--justify-vertically
--justify-systems-only
--justify-include-last-page
--system-max-per-page 5

### DIFF
--- a/src/system.cpp
+++ b/src/system.cpp
@@ -769,7 +769,10 @@ int System::CastOffPages(FunctorParams *functorParams)
         currentShift += params->m_pgHead2Height + params->m_pgFoot2Height;
     }
 
-    if ((params->m_currentPage->GetChildCount() > 0) && (this->m_drawingYRel - this->GetHeight() - currentShift < 0)) {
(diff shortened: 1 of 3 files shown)


src/system.cpp
```diff
@@ -769,7 +769,10 @@ int System::CastOffPages(FunctorParams *functorParams)
         currentShift += params->m_pgHead2Height + params->m_pgFoot2Height;
     }
 
-    if ((params->m_currentPage->GetChildCount() > 0) && (this->m_drawingYRel - this->GetHeight() - currentShift < 0)) {
+    const auto systemMaxPerPage = params->m_doc->GetOptions()->m_systemMaxPerPage.GetValue();
+    const auto childsCount = params->m_currentPage->GetChildCount();
```
lpugin, line 773: We don't use `auto` in such cases. Rename `childsCount` into `childCount`

alexgnt: Done.

PR #1523, opened 2020-06-29
Introduced --bottom-margin-header option

alexgnt:
Added header's bottom margin option for additional flexibility
and ability to distinguish staff spacing and header spacing.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -557,6 +557,7 @@ public:
     OptionDbl m_defaultTopMargin;
     //
     OptionDbl m_bottomMarginHarm;
+    OptionDbl m_bottomMarginPgHead;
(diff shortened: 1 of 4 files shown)


include/vrv/options.h
```diff
@@ -557,6 +557,7 @@ class Options {
     OptionDbl m_defaultTopMargin;
     //
     OptionDbl m_bottomMarginHarm;
+    OptionDbl m_bottomMarginHeader;
```
lpugin, line 560: Rename to `m_bottomMarginPgHead`

alexgnt: Done. Also I've moved usage of this margin from systems to more appropriate place page alignment.

PR #1534, opened 2020-07-07
Use pending measure x for new system start coordinate

Pryadka:
During the castOf process we start new system and set variable `params->m_shift` to start position of the last measure which can't be put in the current system line, although we have to set this variable to start x of first pending measure if it exists. After applying this change we expect that all systems (in the meaning of parts of a big system) will have width not wider than page width and we will be able to get rid of horizontal scaling of systems with width wider than page. This also fixes the issue about wrong spaces between syllable chunks.

### DIFF
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -935,6 +935,13 @@ int Measure::CastOffSystems(FunctorParams *functorParams)
             params->m_currentSystem = new System();
             params->m_page->AddChild(params->m_currentSystem);
             params->m_shift = this->m_drawingXRel;
+            for (Object *oneOfPendingObjects : params->m_pendingObjects) {
+                if (oneOfPendingObjects->Is(MEASURE)) {
+                    Measure *firstPendingMesure = dynamic_cast<Measure *>(oneOfPendingObjects);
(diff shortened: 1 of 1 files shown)


lpugin:
Could you provide a minimal example and show it before and after your fix? Thanks!

Pryadka:
> Could you provide a minimal example and show it before and after your fix? Thanks!

before the fix
before_fix

after
after_fix

The first measure of the second system on the page is 'pending' measure.

lpugin:
Thanks. Could you post the MEI? I do not understand why the page starts measure 29 before the fix and on page 28 after. How do you explain it?

Pryadka:
> Thanks. Could you post the MEI? I do not understand why the page starts measure 29 before the fix and on page 28 after. How do you explain it?

[scorecopy2.mei.txt](https://github.com/rism-ch/verovio/files/4884866/scorecopy2.mei.txt)

After this change the layouting of whole document changed. I can't say what exactly happens in previous pages, but this is the result of new distribution of measures between system chunks. I can only say that the measure with number 30 is in the pending list.

Issue #1558, opened 2020-07-14
Only shift augmentation dots when flags are in the way


th-we:
Augmentation dots should only be shifted if the flags are actually in the way. In this example, only the marked cases need collision avoidance:

![image](https://user-images.githubusercontent.com/1147152/87441905-2ce9f680-c5e3-11ea-87a5-ae4c7d226070.png)

All other cases should be changed to use the default dot position.

craigsapp:
Are you going to give an MEI data sample for the example?

This is the same as issue https://github.com/rism-ch/verovio/issues/595

craigsapp:
Also observe the funny situation related to flag avoidance of the dots in multiple layers related to issue https://github.com/rism-ch/verovio/issues/655:

<img width="283" alt="Screen Shot 2020-07-14 at 9 03 26 AM" src="https://user-images.githubusercontent.com/3487289/87448658-f72c1a00-c5b0-11ea-919a-c0d304b9ef84.png">

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2020-07-14T09:03:35" version="3.0.0-dev-79cf926-dirty">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000000426257882">
    <score xml:id="score-0000001585895961">
     <scoreDef xml:id="scoredef-0000000203742499">
      <staffGrp xml:id="staffgrp-0000001169226644">
       <staffDef xml:id="staffdef-0000000085753793" n="1" lines="5">
        <clef xml:id="clef-0000001054794104" shape="G" line="2" />
       </staffDef>
      </staffGrp>
[...]

PR #1560, opened 2020-07-14
improve dots positioning

rettinghaus:


lpugin:
Could you give an explanation, of better, an example before / after?

rettinghaus:
@lpugin please see #1558 for before / after.

PR #1565, opened 2020-07-17
allow multiple fermatas on barlines

rettinghaus:


lpugin:
Leave the dynamic casts for consistency

PR #1576, opened 2020-07-24
Adding some glyphs for Sprechgesang and harmonics

rettinghaus:


lpugin:
Can you make two separate PRs with one for the update of Bravura and one for the new glyph, and make sure there are now other commits mixed in? Thanks

rettinghaus:
Here are only the new glyphs now. Fonts have to be regenerated after all merges.

PR #1588, opened 2020-07-30
Update fonts with generated new glyphs (added in f47530b7)

Monceber:
- Added generated xmls and updated font xmls with noteheads that were added to font previously


lpugin:
Could you investigate why you have all these file changes and drop the useless ones?

Issue #1299, opened 2020-02-09
System dividers attached to the lower system


EvgenyM:
At the moment, system dividers are clipped to the lower system. Instead they are supposed to be positioned equally between two systems (in the middle).

Engraving file sample: attached.
[wollny-valse-p1.mei.zip](https://github.com/rism-ch/verovio/files/4176977/wollny-valse-p1.mei.zip)

![image](https://user-images.githubusercontent.com/718183/74103643-25af5000-4b4e-11ea-8319-76567a070adc.png)

craigsapp:
Desired rendering, where the marks are centered between systems rather than attached to bottom system:

![centered](https://user-images.githubusercontent.com/3487289/74105689-fdefc600-4b14-11ea-89f5-832e9179f543.png)

craigsapp:
MEI test data:

```xml
<?xml version="1.0" encoding="utf8" standalone="no"?>
<?xml-model href="../schema/4.0.1/mei-all.rng.xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="../schema/4.0.1/mei-all.rng.xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="m-1" meiversion="4.0.0">
  <meiHead xml:id="m-2">
    <fileDesc xml:id="m-3">
      <titleStmt>
        <title xml:id="m-10">Suite für Geige und Klavier</title>
        <composer xml:id="m-11">P. Wollny</composer>
      </titleStmt>
      <pubStmt>
        <availability xml:id="m-16">
          <useRestrict xml:id="m-17"/>
        </availability>
      </pubStmt>
    </fileDesc>
    <encodingDesc xml:id="m-18">
      <appInfo xml:id="m-19">
        <application xml:id="sibelius" isodate="2019-9-8T18:25:51Z" version="19100">
          <name xml:id="m-21" type="operating-system">Windows 7</name>
        </application>
        <application xml:id="sibmei" type="plugin" version="2.2.0">
          <name xml:id="m-23">Sibelius to MEI Exporter (2.2.0)</name>
        </application>
        <application version="v1.0_beta" xml:id="app_2019251182952907">
          <name>mei30To40.xsl</name>
        </application>
      </appInfo>
    </encodingDesc>
    <workList xml:id="m-5">
[...]

PR #1602, opened 2020-08-11
Fix for #1299. Center system divider between systems

Monceber:
Base system divider position on locations of top of current system staff and bottom of previous system staff.

### DIFF
--- a/src/view_page.cpp
+++ b/src/view_page.cpp
@@ -157,41 +157,7 @@ void View::DrawSystem(DeviceContext *dc, System *system)
 
     Measure *firstMeasure = dynamic_cast<Measure *>(system->FindDescendantByType(MEASURE, 1));
 
-    // Draw system divider (from the second one) if scoreDef is optimized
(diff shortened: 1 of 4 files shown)


src/view_page.cpp
```diff
@@ -1316,6 +1282,56 @@ void View::DrawLayerList(DeviceContext *dc, Layer *layer, Staff *staff, Measure
...
+void View::DrawSystemDivider(DeviceContext *dc, System *system, Measure *firstMeasure)
+{
+    assert(dc);
+    assert(system);
+    assert(firstMeasure);
+
+    // Draw system divider (from the second one) if scoreDef is optimized
+    if (!firstMeasure || (m_options->m_systemDivider.GetValue() == SYSTEMDIVIDER_none)) return;
+    // initialize to zero, first measure is not supposed to have system divider
+    static int previousSystemBottomMarginY = 0;
```
lpugin, line 1294: We do not use static variables. Please remove it

PR #1603, opened 2020-08-12
Do not return Uuid by value

Monceber:
- GetUuid() is called a lot of times in TimeSpanningInterface. This scales with amount of time spanning elements in the file and ends up being a big perfomance bottleneck. Returning by const reference allows to get rid of ctor and dtor calls for std::string and time related to that. Depending on the size of the file this can cut up to 25% of execution time.

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -221,7 +221,7 @@ public:
      */
     virtual void CloneReset();
 
-    std::string GetUuid() const { return m_uuid; }
+    const std::string& GetUuid() const { return m_uuid; }
     void SetUuid(std::string uuid);
     void SwapUuid(Object *other);
     void ResetUuid();


craigsapp:
Basically using a reference as a pointer :-)

I was wondering what happens when the element is deleted:

```cpp
#include <string>
#include <iostream>
using namespace std;

class OBJ {
	public:
		OBJ(void) { value = "CREATED"; }
		~OBJ() { value = "DELETED"; }
		string value;
		const string& getString(void) { return value; }
};

int main(void) {
	OBJ *obj;
	obj = new OBJ;
	const string& myvalue = obj->getString();
	obj->value = "CHANGED";
	delete obj;
	cerr << "myvalue is: " << myvalue << endl;
	return 0;
}
```

The result is 

```
myvalue is: DELETED
```

lpugin:
I think this is OK even thought in theory not 100% safe. Once the object is deleted the memory could be reallocated and yield unpredictable behavior.

Could you post the file where it cuts 25% of execution time?

lpugin:
@craigsapp if I do 
```cpp

int main(void) {
	OBJ *obj;
	obj = new OBJ;
	const string& myvalue = obj->getString();
	obj->value = "CHANGED";
	delete obj;
	obj = new OBJ;
	cerr << "myvalue is: " << myvalue << endl;
	return 0;
}
```
I got
```
myvalue is: CREATED
```

Monceber:
@lpugin @craigsapp 
That's not quite how you'd use reference, and that example is actually bad way of using references, especially if `obj = new OBJ;` were to be omitted.
Consider another example:

```
#include <string>
#include <iostream>
using namespace std;

class MyString : public std::string {
	public:
		MyString() { cout << "created" << endl; }
		MyString(const MyString& copy) { cout << "copy created" << endl; }
		~MyString() { cout << "destroyed" << endl; }
};

class OBJ {
	public:
		OBJ(void) { }
		~OBJ() { }
		MyString value;
		const MyString& getString(void) { return value; }
		MyString getStringByVal(void) { return value; }
};

int main(void) {
	cout << "Create Object" << endl;
	OBJ *obj;
	obj = new OBJ;

	cout << "Get String By Value" << endl;
	if (obj->getStringByVal() == "test") {
		// noop
	}

	cout << "Get String By Const Ref" << endl;
	if (obj->getString() == "test")
	{
		// noop
	}

	cout << "Create new element based on return" << endl;
	MyString test = obj->getString();

	cout << "Delete Object" << endl;
	delete obj;
	return 0;
}
```

And this is the output you'd get:
```
Create Object
created
Get String By Value
copy created
destroyed
Get String By Const Ref
Create new element based on return
copy created
Delete Object
destroyed
```

And that's what happens here:
https://github.com/rism-ch/verovio/blob/23cac12568b38a5d5cc3380e77825c13db972b5c/src/timeinterface.cpp#L175-L185
And if the first condition fails? Well, you repeat the process again for **else-if**.

I can't post the file with 25% improvement, but you can see results even with this file:
https://github.com/music-encoding/sample-encodings/blob/master/MEI_4.0/Music/Complete_examples/Hummel_Concerto_for_trumpet.mei
Albeit it's just 7-8% improvement here, it would increase with the size of the file and amout of TimeSpanning elements in the file.

To further reiterate the difference, here's disassembly of return by value (just first **if statement** from the snippet above):
```
00007FF6CEB60583  mov         rax,qword ptr [this]  
00007FF6CEB6058B  cmp         qword ptr [rax+90h],0  
00007FF6CEB60593  jne         vrv::TimeSpanningInterface::SetStartAndEnd+0DDh (07FF6CEB6061Dh)  
00007FF6CEB60599  mov         rax,qword ptr [this]  
00007FF6CEB605A1  add         rax,98h  
00007FF6CEB605A7  mov         rcx,rax  
00007FF6CEB605AA  call        std::basic_string<char,std::char_traits<char>,std::allocator<char> >::empty (07FF6CE71A56Fh)  
00007FF6CEB605AF  movzx       eax,al  
00007FF6CEB605B2  test        eax,eax  
00007FF6CEB605B4  jne         vrv::TimeSpanningInterface::SetStartAndEnd+0DDh (07FF6CEB6061Dh)  
00007FF6CEB605B6  mov         rax,qword ptr [this]  
00007FF6CEB605BE  add         rax,98h  
00007FF6CEB605C4  mov         qword ptr [rsp+80h],rax  
00007FF6CEB605CC  lea         rdx,[rsp+20h]  
00007FF6CEB605D1  mov         rcx,qword ptr [element]  
00007FF6CEB605D9  call        vrv::Object::GetUuid (07FF6CE7121ADh)  
00007FF6CEB605DE  mov         qword ptr [rsp+88h],rax  
00007FF6CEB605E6  mov         eax,dword ptr [rsp+7Ch]  
00007FF6CEB605EA  or          eax,1  
00007FF6CEB605ED  mov         dword ptr [rsp+7Ch],eax  
00007FF6CEB605F1  mov         rax,qword ptr [rsp+80h]  
00007FF6CEB605F9  mov         rdx,rax  
00007FF6CEB605FC  mov         rcx,qword ptr [rsp+88h]  
00007FF6CEB60604  call        std::operator==<char,std::char_traits<char>,std::allocator<char> > (07FF6CE7202EEh)  
00007FF6CEB60609  movzx       eax,al  
00007FF6CEB6060C  test        eax,eax  
00007FF6CEB6060E  je          vrv::TimeSpanningInterface::SetStartAndEnd+0DDh (07FF6CEB6061Dh)  
00007FF6CEB60610  mov         dword ptr [rsp+90h],1  
00007FF6CEB6061B  jmp         vrv::TimeSpanningInterface::SetStartAndEnd+0E8h (07FF6CEB60628h)  
00007FF6CEB6061D  mov         dword ptr [rsp+90h],0  
00007FF6CEB60628  movzx       eax,byte ptr [rsp+90h]  
00007FF6CEB60630  mov         byte ptr [rsp+78h],al  
00007FF6CEB60634  mov         eax,dword ptr [rsp+7Ch]  
00007FF6CEB60638  and         eax,1  
00007FF6CEB6063B  test        eax,eax  
00007FF6CEB6063D  je          vrv::TimeSpanningInterface::SetStartAndEnd+10Eh (07FF6CEB6064Eh)  
00007FF6CEB6063F  and         dword ptr [rsp+7Ch],0FFFFFFFEh  
00007FF6CEB60644  lea         rcx,[rsp+20h]  
00007FF6CEB60649  call        std::basic_string<char,std::char_traits<char>,std::allocator<char> >::~basic_string<char,std::char_traits<char>,std::allocator<char> > (07FF6CE72FED8h)  
00007FF6CEB6064E  movzx       eax,byte ptr [rsp+78h]  
00007FF6CEB60653  test        eax,eax  
00007FF6CEB60655  je          vrv::TimeSpanningInterface::SetStartAndEnd+131h (07FF6CEB60671h)  
```
And here is return by const reference:
```
00007FF71F34D843  mov         rax,qword ptr [this]  
00007FF71F34D848  cmp         qword ptr [rax+90h],0  
00007FF71F34D850  jne         vrv::TimeSpanningInterface::SetStartAndEnd+8Eh (07FF71F34D8AEh)  
00007FF71F34D852  mov         rax,qword ptr [this]  
00007FF71F34D857  add         rax,98h  
00007FF71F34D85D  mov         rcx,rax  
00007FF71F34D860  call        std::basic_string<char,std::char_traits<char>,std::allocator<char> >::empty (07FF71EF0A574h)  
00007FF71F34D865  movzx       eax,al  
00007FF71F34D868  test        eax,eax  
00007FF71F34D86A  jne         vrv::TimeSpanningInterface::SetStartAndEnd+8Eh (07FF71F34D8AEh)  
00007FF71F34D86C  mov         rax,qword ptr [this]  
00007FF71F34D871  add         rax,98h  
00007FF71F34D877  mov         qword ptr [rsp+20h],rax  
00007FF71F34D87C  mov         rcx,qword ptr [element]  
00007FF71F34D881  call        vrv::Object::GetUuid (07FF71EF1DD72h)  
00007FF71F34D886  mov         rcx,qword ptr [rsp+20h]  
00007FF71F34D88B  mov         rdx,rcx  
00007FF71F34D88E  mov         rcx,rax  
00007FF71F34D891  call        std::operator==<char,std::char_traits<char>,std::allocator<char> > (07FF71EF102DFh)  
00007FF71F34D896  movzx       eax,al  
00007FF71F34D899  test        eax,eax  
00007FF71F34D89B  je          vrv::TimeSpanningInterface::SetStartAndEnd+8Eh (07FF71F34D8AEh)  
```

lpugin:
For the record, I cannot measure 7-8% speed improvement on the Hummel example. Doing
```bash
time ./verovio --all-pages Hummel_Concerto_for_trumpet.mei 
```
gives me roughly the same execution time (sometimes higher and sometimes lower)

What tool do you use to measure this? The xCode Instruments?

Monceber:
Yes, that was probably a bad example. I've measured it and got the 7-8% difference, but that indeed seems to be an issue of different execution time.

Then, here's another file:
[berlioz-Requiem_05__QurensMe_fredipi.zip](https://github.com/rism-ch/verovio/files/5068598/berlioz-Requiem_05__QurensMe_fredipi.zip)

It originally has only 12 parts, but I've extended it (by copy-paste) to have 120 parts total.
For timing I'm using time command as well as <chrono> code snippet in the main.cpp.

**string GetUuid();**
xml -> mei
_time ./verovio -r ../Data -a Release/berlioz-Requiem_05__QurensMe_fredipi.xml -f xml -t mei_
```
chrono	7053ms
real	0m7.129s
user	0m6.484s
sys	0m0.531s
```
xml -> svg
_time ./verovio -r ../Data -a Release/berlioz-Requiem_05__QurensMe_fredipi.xml_
```
chrono	16466ms
real	0m16.555s
user	0m15.797s
sys	0m0.328s
```

**const string& GetUuid();**
xml -> mei
_time ./verovio -r ../Data -a Release/berlioz-Requiem_05__QurensMe_fredipi.xml -f xml -t mei_
```
chrono	5938ms
real	0m6.012s
user	0m5.484s
sys	0m0.438s
```
xml -> svg
_time ./verovio -r ../Data -a Release/berlioz-Requiem_05__QurensMe_fredipi.xml_
```
chrono	13651ms
real	0m13.728s
user	0m13.031s
sys	0m0.359s
```

craigsapp:
It is definitely better to use a return value of `const string&` rather than `string` for strings that are commonly used as temporary variables, and there will be no significant danger in doing it that way (unless you do the weird corner case from my test program).  The purpose of this is to do string comparisons on temporary non-copies of the UUID rather than on temporary copies of the UUID.   A copy will be generated automatically if the `const string&` output is saved to a regular `string` or `const string`, so this is why the `const string&` return value is fairly safe and hard to abuse.

@lpugin:  That is strange &mdash; I think you somehow dropped the `&` from the `const string&` variable, causing a copy to occur instead of a reference:

```cpp
#include <string>
#include <iostream>
using namespace std;

class OBJ {
	public:
		OBJ(void) { value = "CREATED"; }
		~OBJ() { value = "DELETED"; }
		string value;
		const string& getString(void) { return value; }
};

int main(void) {
	OBJ *obj;
	obj = new OBJ;
	const string myvalue = obj->getString();
	obj->value = "CHANGED";
	delete obj;
	cerr << "myvalue is: " << myvalue << endl;
	return 0;
}
```

Otherwise, I do not see how that would be valid behavior from the compiler, since the const string reference is getting direct access to memory inside of the class, and if you change it in the class, the const string reference will update accordingly.

The line:

```cpp
	const string myvalue = obj->getString();
```

makes a copy of `OBJ::value`, while the code:

```cpp
	const string& myvalue = obj->getString();
```

does not make a copy.  So in other words, if you are going to store the output of `obj->getString()` for later use, it is better to make a copy of it  by not using a reference. (and there is no real reason to use a const for the copy unless you want to enforce not changing that copy, and a non-const `string&` cannot be set to a `const string&` to make unauthorized changes to the contents of the protected members of the class).

-----------------------

> That's not quite how you'd use reference, and that example is actually bad way of using references, especially if obj = new OBJ; were to be omitted.

That is what I was testing, since I haven't used const reference return values (but I should start using them).  I created the test program to exploit the bug that can potentially be caused if it is used incorrectly 😉 .  So as long as the const& value is treated like a pointer (which can become invalid after deleting the pointer), then there will be no problems.

> I can't post the file with 25% improvement,

Then for Beethoven's 9th symphony? :-)

---------------------------

There is little chance of const reference values from being used incorrectly, and they are safer than pointers.

@Monceber's example code gives the standard use case:

```cpp
obj->getString() == "test"
```

On this line, the `const string&` output of `getString()` does not make a copy of the string before comparing it to `"test"`.  Instead the memory contents pointed to by the `const string&` is compared directly to `"test"` without any copying being done.  

The reason why it is safe, is that the proper way of saving to a variable would be to not use a reference:

```cpp
string value = obj->getString();
```

In this case there would be a copy happening, and if `obj` is deleted or goes out of scope, the contents of `value` is still valid.

Issue #1359, opened 2020-03-21
MusicXML import of dynamics mixed with text


th-we:
```
      <direction>
        <direction-type>
          <dynamics>
            <p/>
          </dynamics>
        </direction-type>
        <direction-type>
          <words> cantabile</words>
        </direction-type>
      </direction>
```
Currently essentially imports as:
```
<dynam>p</dynam>
```
Should import as something like
```
<dynam>p cantabile</dynam>
```
(or with additional `<rend>` inside to separate the two parts).

Sample MusicXML:

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise>
  <part-list>
    <score-part id="P1">
      <part-name>Flute</part-name>
    </score-part>
  </part-list>
  <part id="P1">
    <measure number="1">
      <attributes>
        <divisions>1</divisions>
        <time>
          <beats>3</beats>
          <beat-type>4</beat-type>
        </time>
        <clef>
          <sign>G</sign>
          <line>2</line>
        </clef>
      </attributes>
      <direction>
        <direction-type>
          <dynamics>
            <p/>
          </dynamics>
        </direction-type>
        <direction-type>
          <words> cantabile</words>
        </direction-type>
      </direction>
      <note>
        <rest measure="yes"/>
        <duration>3</duration>
        <voice>1</voice>
      </note>
    </measure>
  </part>
</score-partwise>
```

wergo:
The problem is that currently only one `<direction-type>` within a `<direction>` is imported. 

Importing any number of children within a `<direction>` element should be added. Combining dynamics with words is another special case...

th-we:
If the dynamic symbols such as “p”, “f”, “sfz” etc. are converted to `<rend fontfam="VerovioMusic">&#xE520;</rend>` with the respective SMuFL codepoints, then Verovio can handle this.

```xml
<dynam tstamp="1" staff="1"><rend fontname="VerovioText">&#xE520;</rend> cantabile</dynam>
```

rettinghaus:
closed by #1604

rettinghaus:
@lpugin is there a reason to keep this open?

PR #1604, opened 2020-08-13
Added support of mixed dynamics for MusicXML

th-we:
Import of dynamics mixed with text from MusicXML.

Also fixes a couple of potential crashes when working with iterators.


th-we:
Forgot to mention:  This addresses #1359.

PR #1606, opened 2020-08-14
Implement support for extended ornaments

Monceber:
Added support for the ExtSym to such elements as Fermata, Turn, Trill, Mordent. This will allow to change how element is drawn based on @glyph.num that is provided as attribute.


src/iomusxml.cpp
```diff
@@ -3640,4 +3670,18 @@ void MusicXmlInput::ShapeFermata(Fermata *fermata, pugi::xml_node node)
     }
 }
 
+std::string MusicXmlInput::GetOrnamentGlyphNumber(const std::string &attributes) const
+{
+    static std::map<std::string, std::string> precomposedNames = {
```
Monceber, line 3675: @lpugin I know you've mentioned that **static** should not be used. This can be changed either to non-static variable, but that would lead to memory allocation on each entry. Alternatively it could be changed to if-else statments, but it doesn't seem the best approach either.
Should I remove **static** regardless?

src/view_text.cpp
```diff
@@ -423,4 +424,44 @@ void View::DrawSvg(DeviceContext *dc, Svg *svg, TextDrawingParams &params)
...
+
+    // Set glyph, fontname and style needed to draw
+    Text *text = new Text();
+    text->SetText(glyph);
+    Rend *rend = new Rend();
+    rend->SetFontname(m_doc->GetOptions()->m_font.GetStrValue());
+    rend->SetFontstyle(FONTSTYLE_normal);
+    rend->SetHalign(HORIZONTALALIGNMENT_center);
+    rend->AddChild(text);
+    rend->SetParent(object);
```
lpugin, line 446: Could you post some example encodings and corresponding renderings? There is something conceptually wrong here because you cannot modify the Doc tree (the model) when drawing it. With this code, new objects will be added to the tree every time it is rendered.

Monceber: Here's the MEI file used:

```
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>Example</title>
      </titleStmt>
      <pubStmt/>
    </fileDesc>
    <encodingDesc xml:id="encodingdesc-01">
      <appInfo xml:id="appinfo-01">
        <application xml:id="application-01" isodate="2020-07-06T10:32:25" version="3.0.0-dev-5706f96-dirty">
          <name >Verovio</name>
          <p>...</p>
        </application>
      </appInfo>
    </encodingDesc>
  </meiHead>
  <music>
    <body>
      <mdiv xml:id="mdiv-01">
        <score xml:id="score-01">
          <scoreDef xml:id="scoredef-01">
            <staffGrp xml:id="staffgrp-01">
              <staffGrp xml:id="staffgrp-02">
                <staffDef xml:id="staffdef-01" n="1" lines="5" ppq="16" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4">
                  <label xml:id="label-01">Piano</label>
                  <clef xml:id="clef-01" line="2" shape="G" facs="#zone-02"/>
                  <meterSig count="4" unit="4" facs="#zone-03"/>
                </staffDef>
              </staffGrp>
            </staffGrp>
          </scoreDef>
          <section xml:id="section-01">
            <pb xml:id="pb-01" />
            <measure xml:id="measure-01" n="1">
              <staff xml:id="staff-01" n="1" facs="#zone-01">
                <layer xml:id="layer-01" n="1">
                  <note xml:id="note-01" dur.ppq="16" dur="8" oct="5" pname="c" stem.dir="up" >
                    <accid xml:id="accid-04" accid="s" accid.ges="s"/>
                  </note>
                  <note xml:id="note-02" dur.ppq="16" dur="4" oct="5" pname="d" stem.dir="down" >
                    <accid xml:id="accid-05" accid="s" accid.ges="s" />
                  </note>
                  <note xml:id="note-03" dur.ppq="16" dur="4" oct="5" pname="e" stem.dir="down" accid.ges="s" />
                  <note xml:id="note-04" dur.ppq="16" dur="8" oct="5" pname="f" stem.dir="down" accid.ges="s"/>
                </layer>
              </staff>
              <mordent staff="1" startid="#note-01" form="lower" glyph.num="#xE56C" />
              <trill staff="1" startid="#note-02" form="upper" glyph.num="#xE56C"/>
              <turn staff="1" startid="#note-03" form="lower" long="true" glyph.num="#xE56C"/>
              <fermata staff="1" startid="#note-04" glyph.num="#xE56C"/>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

And the output looks as following:
image
Now that I look at it, X position for some elements need adjustment, but that can be adjusted.

Regarding how this is handled, I don't think this would be possible to do this on the Doc step. I think neither of these elements (Mordent, Trill, etc.) are allowed to have text/rend children according to MEI specification, so addding them is not an option. The thing is, <rend> only has SetParent() called in this code, but it's never added as child, so it's not actually added to corresponding elements (and we don't have multiple copies of them).
There's a problem that I can see now with memory leaking that I'll need to address though (assuming that you're fine with this approach).

rettinghaus: The approach should be the same as with "normal" mordent / trill glyphs. So, the glyphs should be copied to the SVG and have to be made available by regenerating the fonts.

Monceber: As mentioned by @lpugin, this approach is wrong. 
Partially reverting ca775fc should do the trick of handling SMuFL codes properly. I'll retest code and create new pull request for it.

Issue #1526, opened 2020-07-03
Group spacing support


kseniaosipova:
There would be a good option for improving rendering quality, to have a parameter for setting the spacing between groups of instruments (e.g. --spacing-group):

![image](https://user-images.githubusercontent.com/60135178/86469963-bde6e500-bd3a-11ea-90d2-bb9e03217048.png)

Also, for chamber music, to have parameters to control:

- spacing within individual staff group of the main instrument:

![image](https://user-images.githubusercontent.com/60135178/86470104-03a3ad80-bd3b-11ea-8750-b6a17851de1e.png)

- spacing between staff groups of the main instruments:

![image](https://user-images.githubusercontent.com/60135178/86470223-46658580-bd3b-11ea-8377-7a5d4ec306eb.png)

- spacing between staff groups of the main instruments and other instruments:

![image](https://user-images.githubusercontent.com/60135178/86470257-5bdaaf80-bd3b-11ea-930c-56c876ec27f9.png)

craigsapp:
Yes, spacing control of groups would be good to have as an additional control to spacing options for the staff and system:

```
 --spacing-staff <i>               The staff minimal spacing in MEI units (default: 8; min: 0; max: 24)
 --spacing-system <i>              The system minimal spacing in MEI units (default: 3; min: 0; max: 12
 --spacing-group <i>               The group minimal spacing in MEI units (default: 3; min: 0; max: 12
```

With min/max/default to be determined.

Adding such control would be useful for creating dynamic vertical adjustments for orchestral scores, as well as soloist/accompanist scores and other grouped scores.

craigsapp:
The basic implementation could be identical to the current staff/system spacing option.

There are other things to think about in the meantime.

In the first example, there are three groups, but also additional complications:

<img width="936" alt="Screen Shot 2020-07-03 at 6 21 50 AM" src="https://user-images.githubusercontent.com/3487289/86473298-8570e800-bcf5-11ea-8784-3a9b5c45d6c0.png">

[...]

PR #1612, opened 2020-08-17
Vertical staves/systems spacing calculation and justification improvements

alexgnt:
The PR introduces are option --spacing-bracket-group and --spacing-brace-group to explicitly set spacing inside of bracket and brace groups. Improve spacing calculation in between staves and systems.

To improve user control over vertical justification introduced options:
--justification-staff
--justification-system
--justification-brace-group
--justification-bracket-group

The option --justify-systems-only is removed as obsolete. When --justification-* options set to 0 except for --justification-system same effect would be achieved.

Addressing issues described in #1526

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -457,7 +457,6 @@ public:
     OptionBool m_evenNoteSpacing;
     OptionBool m_humType;
     OptionBool m_justifyIncludeLastPage;
-    OptionBool m_justifySystemsOnly;
     OptionBool m_justifyVertically;
     OptionBool m_landscape;
     OptionBool m_mensuralToMeasure;
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -520,6 +520,8 @@ public:
     OptionDbl m_spacingNonLinear;
     OptionInt m_spacingStaff;
     OptionInt m_spacingSystem;
+    OptionInt m_spacingBracketGroup;
+    OptionInt m_spacingBraceGroup;
     OptionDbl m_staffLineWidth;
     OptionDbl m_stemWidth;
     OptionIntMap m_systemDivider;
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -522,6 +522,10 @@ public:
     OptionInt m_spacingSystem;
     OptionInt m_spacingBracketGroup;
     OptionInt m_spacingBraceGroup;
+    OptionDbl m_justificationStaff;
+    OptionDbl m_justificationSystem;
+    OptionDbl m_justificationBraceGroup;
+    OptionDbl m_justificationBracketGroup;
     OptionDbl m_staffLineWidth;
     OptionDbl m_stemWidth;
     OptionIntMap m_systemDivider;
--- a/src/system.cpp
+++ b/src/system.cpp
@@ -362,11 +362,7 @@ int System::AlignVerticallyEnd(FunctorParams *functorParams)
     AlignVerticallyParams *params = dynamic_cast<AlignVerticallyParams *>(functorParams);
     assert(params);
 
-    if (this->GetIdx() > 0) {
-        params->m_cumulatedShift
-            = params->m_doc->GetOptions()->m_spacingStaff.GetValue() * params->m_doc->GetDrawingUnit(100);
-    }
-
+    params->m_cumulatedShift = 0;
     params->m_staffIdx = 0;
 
     m_systemAligner.Process(params->m_functor, params);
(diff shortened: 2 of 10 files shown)


src/verticalaligner.cpp
```diff
@@ -151,6 +187,54 @@ void StaffAlignment::SetStaff(Staff *staff, Doc *doc)
...
+             }
+
+             if (m_spacingType == SpacingType::None) {
+                 staffChild = staffParent;
+                 staffParent = staffChild->GetParent();
+                 if (!staffParent || !staffParent->Is(STAFFGRP)) {
+                     m_spacingType = SpacingType::Staff;
+                 }
+             }
+        }
```
lpugin, line 237: I wonder if it is necessary to do this for each StaffAlignment. This does not look optimized to me. I think doing it only once for the Doc::m_scoreDef and keeping some sort of a map of the spacingType for each `staff@n` would do the job. What do you think?

include/vrv/options.h
```diff
@@ -520,6 +519,12 @@ class Options {
     OptionDbl m_spacingNonLinear;
     OptionInt m_spacingStaff;
     OptionInt m_spacingSystem;
+    OptionInt m_spacingBracketGroup;
+    OptionInt m_spacingBraceGroup;
+    OptionDbl m_justificationStaff;
+    OptionDbl m_justificationSystem;
+    OptionDbl m_justificationBraceGroup;
+    OptionDbl m_justificationBracketGroup;
```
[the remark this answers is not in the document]
alexgnt: Will do.

src/system.cpp
```diff
@@ -632,21 +627,20 @@ int System::JustifyY(FunctorParams *functorParams)
     JustifyYParams *params = dynamic_cast<JustifyYParams *>(functorParams);
     assert(params);
 
-    bool systemOnly = params->m_doc->GetOptions()->m_justifySystemsOnly.GetValue();
+    const double systemJustificationFactor = params->m_doc->GetOptions()->m_justificationSystem.GetValue();
+    const double shift = systemJustificationFactor / params->m_justificationSum * params->m_spaceToDistribute;
```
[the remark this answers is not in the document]
alexgnt: That would, a rear case. It means that all factors should be set to 0 and justification enabled.
In any case, I'll add a check in Page::JustifyVertically to avoid reaching here with zeroes.

src/verticalaligner.cpp
```diff
@@ -109,13 +110,46 @@ void SystemAligner::FindAllIntersectionPoints(
     SegmentedLine &line, BoundingBox &boundingBox, const std::vector<ClassId> &classIds, int margin)
 {
     StaffAlignment *alignment = NULL;
-    for (auto &child : m_children) {
+    for (const auto child : m_children) {
```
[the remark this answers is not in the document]
alexgnt: From performance and memory consumption point of view that's the same.
As for me that's misleading to have references to a pointers (especially non constant), that's why I have replaced them with pointers where it possible.

src/verticalaligner.cpp
alexgnt: From my point of view that shouldn't be critical since there is not so big amount of staves in the documents. The code designed to use dynamic_cast only when it's needed and staff groups shouldn't have a lot of childs. 
Also, that would be quite strange that scoreDef holds the map of spacing type above the staff@n.

What if I would store found type in static map inside of StaffAlignment and would have lazy initialization for it. I.e. it will search for type in groups tree only when it's not found in that map and adding it to it?

That would be similar to what you want and placed inside class where it's really used.

alexgnt:
> Could you provide a basic example (one page) that we can run with various options? Thanks

Please have a look at this one.
[VerticalSpacingTestCases.zip](https://github.com/rism-ch/verovio/files/5090021/VerticalSpacingTestCases.zip)

src/verticalaligner.cpp
lpugin: > From my point of view that shouldn't be critical since there is not so big amount of staves in the documents. The code designed to use dynamic_cast only when it's needed and staff groups shouldn't have a lot of childs.

Yes, but this is not a reason not to put it in a better place. It is also conceptually. 

> Also, that would be quite strange that scoreDef holds the map of spacing type above the staff@n.

I don't agree. We are looking at StaffGrp/StaffDefs. These belong to a ScoreDef, not to a StaffAlignment. It is good to have each StaffAlignment having a m_spacingType member. It is the their calculation that is wrongly placed. 

However, I think there is additional problem with the way it is done now, namely that it does not seem to take into account system optimization (hidden empty staves / staffGrps). So in the end, the best way / place to do it would be from the SystemAligner. This means moving the code to the parent in a method SystemAligner::SetSpacing. There you will be able to check if each staff is actually visible of not. I assume this can change some spacingType values. The call to SystemAligner::SetSpacing it to be done from System::OptimizeScoreDefEnd. This will also be much cleaner.

src/system.cpp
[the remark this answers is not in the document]
alexgnt: added check in Page::JustifyVertically

src/verticalaligner.cpp
alexgnt: I've submitted commit with changed logic that should work correct with system optimization.
Now I'm struggling with moving these calculations to ScoreDef, by now all my tries and experiments gives only screwed up offsets.

lpugin: My latest proposal was to move it to SytemAligner. Did you try adding a SystemAligner::SetSpacing method to be called on System::OptimizeScoreDefEnd? I cannot see it.

src/verticalaligner.cpp
```diff
@@ -15,16 +15,49 @@
...
+            return staffDef && staffDef->GetDrawingVisibility() != OPTIMIZATION_HIDDEN;
+        }
+
+        StaffGrp *staffGrp = dynamic_cast<StaffGrp *>(object);
+        return staffGrp && staffGrp->GetDrawingVisibility() != OPTIMIZATION_HIDDEN;
+    }
+
+protected:
+    const Object *m_objectToExclude;
+};
```
lpugin, line 59: Move this to `comparison.h`

src/verticalaligner.cpp
```diff
@@ -15,16 +15,49 @@
...
 #include "tie.h"
 #include "vrv.h"
 
 namespace vrv {
+namespace {
+//----------------------------------------------------------------------------
+// VisibleStaffDefOrGRPObject
+//----------------------------------------------------------------------------
+
+class VisibleStaffDefOrGRPObject : public ClassIdsComparison {
```
lpugin, line 36: Rename class to `VisibleStaffDefOrGrpObject`

src/verticalaligner.cpp
```diff
@@ -15,16 +15,49 @@
...
+
+public:
+    VisibleStaffDefOrGRPObject(): ClassIdsComparison({ STAFFDEF, STAFFGRP }) {}
+
+    void skip(const Object *objectToExclude) { m_objectToExclude = objectToExclude; }
```
lpugin, line 41: Rename to `Skip`

Issue #617, opened 2017-06-09
Automatic rest collision avoidance when two layers


craigsapp:
When there are two layers (more as well, but just dealing with two automatically should be sufficient), it would be very useful if rests automatically avoid notes/rests in the other layers, if there is no explicit vertical placement of the rests.

Example:

<img width="571" alt="screen shot 2017-06-09 at 3 28 44 pm" src="https://user-images.githubusercontent.com/3487289/26977483-5f23fed4-4d28-11e7-8297-d66c6a4fd7a9.png">

Humdrum test data:

```
**kern
*M4/4
=1-
*^
4e	4r
4f	4r
4g	4r
4a	4r
=	=
4b	4r
4cc	4r
4dd	4r
4ee	4r
*v	*v
=
*-
```

Here would be about the preferable automatic placement of the rests in this example:

<img width="643" alt="screen shot 2017-06-09 at 3 39 53 pm" src="https://user-images.githubusercontent.com/3487289/26977971-fd92b96a-4d29-11e7-94dc-6015036bb51c.png">

Hand edited MEI data for this example, using `ploc` and `oloc` for the vertical position of the rest:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2017-06-09T15:29:31" version="0.9.14-dev-7463144-dirty">
                    <name>Verovio</name>
                    <p>Transcoded from Humdrum</p>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title />
                </titleStmt>
            </work>
        </workDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-0000000393413851">
                        <staffGrp xml:id="staffgrp-0000001222731228">
[...]

PR #1624, opened 2020-08-21
Automatic rest collision avoidance for two layers

th-we:
Fixes most of #617. Only some situations where rests stand against rests in both layers might not be handled ideally yet, depending on the surrounding notes.


lpugin:
Thanks for this. I am not 100% convinced we need this as an option (yet). Could you give an example where you would want to change the settings?
Somehow related: I don't particularly like the idea to store the structure in JSON internally. As pointed in the issue #617 it does not have to be JSON. The string-based maps / lookups are not very Verovio-like. Could you think of a way to transform the JSON into something that does not use string-based maps? It would be good to have something generic that we can use for other options. Eventually, these might be integrating into MEI, which mean they would be loaded as XML and not as JSON.

lpugin:
I did a quick test with a file of 24 pages of empty measures of rests. 
Current develop branch:
* 0.76s
This PR:
* 58.47s
I suspect the JSON lookup is one reason, but maybe not the only one.

Monceber:
> Could you give an example where you would want to change the settings?

The easiest example that comes to mind would be adjustments for different fonts that may have higher/lower glyphs for rests or other relevant symbols, and with default offsets they might end up overlapping 

> Could you think of a way to transform the JSON into something that does not use string-based maps?

I'll take a look at what can it be substituted with. I had an idea of using something akin to bitflags for tracking options, but I decided against - it wasn't exactly a generic approach.

craigsapp:
>> Could you think of a way to transform the JSON into something that does not use string-based maps?
> I'll take a look at what can it be substituted with. I had an idea of using something akin to bitflags for tracking options, but I decided against - it wasn't exactly a generic approach.

Related to this and similar discussions recently, note how the libmei code handles such cases: a finite list of possible strings for an attribute are converted into an enum.  See:
https://github.com/rism-ch/verovio/blob/develop/libmei/atttypes.h

This enum can then be used for random access to vectors (although I do not think this is done in libmei), or can be used in switches to access lookup data, such as:
https://github.com/rism-ch/verovio/blob/f8bab404/libmei/attconverter.cpp#L68-L100

Switches can be optimized by the compile to `O(log N)` efficiency, similar to map lookups, but there is no overhead in creating the switch at runtime since it is prepared at compile time:
Related: https://stackoverflow.com/questions/2596320/how-does-switch-compile-in-visual-c-and-how-optimized-and-fast-is-it

Monceber:
I wouldn't call that generic approach, tho. Enums limit what values can be used with that option - each time another option is added a new enum would need to be created and populated. And with file types like json/xml, where information is nested, going with this would mean either a whole lot of values in single enum or multiple enums with same constant names. Switches would work, but with nested containers they would create a huge code bloat.

lpugin (commented):
I am fine making adding / changing options as easy as possible. However, this cannot be at the cost of making the code less efficient. Also, if you add values in your option structure, this means you are going to change the code that uses it anyway. So adding an enum it not much extra work.

Monceber:
The point regarding enums is that they need to be mapped to actual values somewhere, and MEI does it by having enourmous amount of xEnumToStr and vice versa. Yes, either implementation will require changes in the code that uses it. However, if there is a completely new code it will require new enum, new FromStr method for converting to numeric values and maybe some other changes. This by no means is generic solution if you need to add new code each time (i.e. code that is relevant to parsing configuration, not the code for the new feature).
Now, don't get me wrong - I'm not trying to argue regarding code efficiency here. I completely agree with you and I'll try to find better solution. The point was that I don't think that using enums is that 'better solution'.

There are also two another points.
1. I was thinking about using hashtable as structure to hold data, instead of Json. That way internal implementation would be quite fast (O(1) on average). This would also allow to use different input formats as options (json, xml, etc.) as long as parsing is implemented. Do you have any strong opinions against using such approach?
2. Looks like performance degradation is not related to the Json, but rather to this PR as a whole (json lookup has been replaced with return 3; and I still got the same time). From the looks of it dynamic_cast is offender here, so I'll look what I can improve here.

lpugin:
> There are also two another points.
> 
>     1. I was thinking about using hashtable as structure to hold data, instead of Json. That way internal implementation would be quite fast (O(1) on average). This would also allow to use different input formats as options (json, xml, etc.) as long as parsing is implemented. Do you have any strong opinions against using such approach?
> 

I think this is a good direction. I assume you would use `int` as keys and not strings. Correct? Could you elaborate a bit? Allowing the use of different input format (e.g, XML) is good. 

>     2. Looks like performance degradation is not related to the Json, but rather to this PR as a whole (json lookup has been replaced with return 3; and I still got the same time). From the looks of it dynamic_cast is offender here, so I'll look what I can improve here.

Yes, json parsing cannot cause such a degradation. The dynamic_cast can be quite expensive and can be quite slow. (This is the reason why we have this home-brewed typing in Verovio with Object::m_type and Object::Is(). Quite a significant overhead in terms of code writing and maintenance, but it is much faster than using the C++ RTTI to check types.)

Monceber:
>I think this is a good direction. I assume you would use int as keys and not strings. Correct? Could you elaborate a bit? Allowing the use of different input formats (e.g, XML) is good.

Correct. For instance, in following json input:
```
{
  "0": {
    "0": {
      "0": "1",
      "1": "1",
      "2": "1"
    }
  }
}
```
We'd get three strings "000", "001", "002". Then use hash for these strings as key to store corresponding values in the container. The hash would be of the integer type, so we'd be avoiding string lookup.

lpugin:
I am not sure I fully understand. Your
```
int GetRestOffsetFromOptions(const std::string &layer, const std::pair<int, std::string>& location, bool isTopLayer) const;
```
would become
```
int GetRestOffsetFromOptions(const std::string &layer, const std::pair<int, int>& location, bool isTopLayer) const;
```
Is this correct?
Currently the keys are hardcoded values such as `"restOnTopLayer"`. Assuming you do not want to use enums, what would they then be?

Monceber:
Not quite. `GetRestOffsetFromOptions` wouldn't change. It's just that we're getting value from the options, we would pass a string that corresponds for the path (e.g. "restOnTopLayerrestOnLinelong") and get value based on the hash of this string.

lpugin:
I am not convinced. This looks quite weird. I think we could have something much closer from you original proposal. Something like:
```
enum {
  noAccidental,
  restOnTopLayer,
  restOnBottomLayer,
  s
};

typedef std::map < int, std::map < int, int > >  rests;
rests defaultRests{
    {noAccidental, {
        { restOnTopLayer, 1},
        { restOnBottomLayer, 2}
    }},
    { s, {
        { restOnTopLayer, 5}
    }}
};
```
In any case, this is not the issue as this stage. First you need to make sure the layout approach works in terms of efficiency, and since the json lookup is not the issue we can forget the question for now. Once the layout speed issue is solved we can come back to this.

lpugin:
For the record:
```
enum RestLayer {
  RL_UNSET = -1,
  RL_sameLayer,
  RL_otherLayer
};

enum RestAccidental {
  RA_UNSET = -1,
  RA_none,
  RA_s,
  RA_f,
  RA_x,
  RA_n
};

enum RestLayerPlace {
  RLP_UNSET = -1,
  RLP_restOnTopLayer,
  RLP_restOnBottomLayer
};

enum RestNotePlace {
  RNP_UNSET = -1,
  RNP_noteInSpace,
  RNP_noteOnLine
};

typedef std::map < RestLayer, std::map < RestAccidental, std::map < RestLayerPlace, std::map < RestNotePlace, std::map < int, int > > > > > rests;

rests defaultRests{
    {RL_sameLayer, {
        { RA_none, {
          { RLP_restOnTopLayer, { 
            { RNP_noteInSpace, {
              { 1, 1}, {2, 3} /* ... */ }
            }
          }}
        }}
    }},
};
```
Then you can do
```
int pos = defaultRests.at(RL_sameLayer).at(RA_none).at(RLP_restOnTopLayer).at(RNP_noteInSpace).at(1);
```

Monceber:
Yes, I understand how it can be done and as mentioned earlier in this thread that's how MEI does it.
First of all, this replaces just configuration part of the problem, i.e. we change hardcoded json string with hardcoded map. If that was the only problem then maybe that would be a better approach. However, there are other issues.

First, reading options from commandline to provide custom offsets. I guess you might disagree since you don't find option necessary atm, but what would happen if it's actually needed? You'd need to add functions like these:
```
std::string StrToRestAccidental(const std::string& accid)
{
  std::map<std::string, RestAccidental> accidMap = {{"noAccidental", RA_none}, ...};
  return accidMap[accid];
}
```
for each enum. So that already adds 4 funcitons. If anywhere in the future we need to expand this list somehow (just a projection here), we would need another enum, another function and so on. And we still end up with maps with string keys (unless we go for if-elseif-elseif-else approach, which is even worse, imo).

Second, this is just a specific implementation for just this problem. If a need arises for another similar option for another feature? Should we add 4 other enums and 4 'parsing' functions for them? As I've mentioned before, this just bloats code that by no means is necessary.
With generic solution we are not tying ourselves with specific values for concrete example (i.e. restLayer, restAccidental, etc.). We create one approach that would work for the future instances that need similar behavior. True, if there are changes in values we would still need to change calling code, but parsing part will remain intact and unchanged.

lpugin:
I think was also want to avoid a configuration file bloat ;-). All the layout in Verovio is done without any of these. I think we need to think carefully before we go down the path of having this type of configuration files. How are you going to have users tweak these options?
```
{'otherLayer':{'noAccidental':{'restOnTopLayer':{'noteInSpace':{'1':3,'2':3,'4':5,'8':5,'16':7,'32':7,'64':9,
    '128':9,'long':5,'breve':5},'noteOnLine':{'1':2,'2':4,'4':6,'8':4,'16':6,'32':6,'64':8,'128':8,'long':6,'breve'
    :4}},'restOnBottomLayer':{'noteInSpace':{'1':-5,'2':-5,'4':-5,'8':-5,'16':-5,'32':-7,'64':-7,'128':-9,'long':-5
    ,'breve':-5},'noteOnLine':{'1':-6,'2':-6,'4':-6,'8':-4,'16':-4,'32':-6,'64':-6,'128':-8,'long':-6,'breve':-6}}}
    ,'s':{'restOnTopLayer':{'noteInSpace':{'1':3,'2':5,'4':7,'8':5,'16':7,'32':7,'64':9,'128':9,'long':5,'breve':5}
    ,'noteOnLine':{'1':2,'2':4,'4':6,'8':6,'16':8,'32':8,'64':10,'128':10,'long':6,'breve':4}},'restOnBottomLayer':
    {'noteInSpace':{'1':-5,'2':-5,'4':-5,'8':-5,'16':-5,'32':-7,'64':-7,'128':-9,'long':-5,'breve':-5},'noteOnLine'
    :{'1':-6,'2':-6,'4':-6,'8':-6,'16':-6,'32':-6,'64':-6,'128':-8,'long':-6,'breve':-6}}},'f':{'restOnTopLayer':{
    'noteInSpace':{'1':3,'2':5,'4':5,'8':5,'16':7,'32':7,'64':9,'128':9,'long':5,'breve':5},'noteOnLine':{'1':4,'2'
    :4,'4':6,'8':6,'16':8,'32':8,'64':10,'128':10,'long':6,'breve':4}},'restOnBottomLayer':{'noteInSpace':{'1':-5,
    '2':-5,'4':-5,'8':-5,'16':-5,'32':-7,'64':-7,'128':-9,'long':-5,'breve':-5},'noteOnLine':{'1':-6,'2':-6,'4':-6,
    '8':-4,'16':-4,'32':-6,'64':-6,'128':-8,'long':-6,'breve':-6}}},'x':{'restOnTopLayer':{'noteInSpace':{'1':3,'2'
    :3,'4':5,'8':5,'16':7,'32':7,'64':9,'128':9,'long':5,'breve':5},'noteOnLine':{'1':2,'2':4,'4':6,'8':6,'16':8,
    '32':8,'64':10,'128':10,'long':6,'breve':4}},'restOnBottomLayer':{'noteInSpace':{'1':-5,'2':-5,'4':-5,'8':-5,
    '16':-5,'32':-7,'64':-7,'128':-9,'long':-5,'breve':-5},'noteOnLine':{'1':-6,'2':-4,'4':-6,'8':-4,'16':-4,'32':
    -6,'64':-6,'128':-8,'long':-6,'breve':-6}}},'n':{'restOnTopLayer':{'noteInSpace':{'1':3,'2':3,'4':5,'8':5,'16':
    7,'32':7,'64':9,'128':9,'long':5,'breve':5},'noteOnLine':{'1':2,'2':6,'4':6,'8':6,'16':8,'32':8,'64':10,'128':10
    ,'long':6,'breve':4}},'restOnBottomLayer':{'noteInSpace':{'1':-7,'2':-5,'4':-7,'8':-5,'16':-5,'32':-7,'64':-7,
    '128':-9,'long':-5,'breve':-5},'noteOnLine':{'1':-6,'2':-6,'4':-6,'8':-6,'16':-6,'32':-6,'64':-6,'128':-8,'long'
    :-6,'breve':-6}}}},'sameLayer':{'restOnTopLayer':{'noteInSpace':{'1':-1,'2':1,'4':3,'8':1,'16':3,'32':3,'64':5,
    '128':5,'long':3,'breve':1},'noteOnLine':{'1':0,'2':0,'4':2,'8':2,'16':2,'32':2,'64':4,'128':4,'long':2,'breve':
    2}},'restOnBottomLayer':{'noteInSpace':{'1':-3,'2':-1,'4':-3,'8':-1,'16':-1,'32':-3,'64':-3,'128':-5,'long':-3,
    'breve':-3},'noteOnLine':{'1':-2,'2':-2,'4':-4,'8':-2,'16':-2,'32':-4,'64':-4,'128':-6,'long':-2,'breve':-2}}}}
```
So I am fine experimenting with this, but we shoud not put the cart before the horse. We need to have it working with a internal structure that is efficient and easy to read and understand. We should now shape the internal structure for possibly passing options. 

Of course with enum you need to map the with the strings. But this would not be very different from having `#defines`, which you would do anyway (we are not going to keep hard-corded string keys in the code). This can easily be done with a simple mapping class - we do not need to use exactly the same as the LibMEI code.

PR #1628, opened 2020-08-21
use SMuFL default for stem x position

rettinghaus:
This PR moves shifting stems to actual stem calculation, relying on (default) SMuFL anchor points for note heads.


lpugin:
Does this still work OK when changing the stem width parameter?

rettinghaus:
At least as far as I tested it.

PR #1629, opened 2020-08-21
Fixes memory leak of Zone

valeriyvan:


lpugin:
Thanks. (The whole method might need refactoring at some point.)

PR #1630, opened 2020-08-21
Noteheads and notehead modifiers

Monceber:
Added handling of the recently generated notehead glyphs. There's also placeholder for the noteheads that don't have generated SMuFLs but have support in MEI format. Also added handling of @head.mod, namely parenthesis (since it appears to be the only one modifier with corresponding symbols present).

### DIFF
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -52,6 +52,7 @@ Note::Note()
     , AttColor()
     , AttColoration()
     , AttCue()
+    , AttExtSym()
     , AttGraced()
     , AttMidiVelocity()
     , AttNoteAnlMensural()
@@ -67,6 +68,7 @@ Note::Note()
     RegisterAttClass(ATT_COLOR);
     RegisterAttClass(ATT_COLORATION);
     RegisterAttClass(ATT_CUE);
+    RegisterAttClass(ATT_EXTSYM);
     RegisterAttClass(ATT_GRACED);
     RegisterAttClass(ATT_NOTEANLMENSURAL);
     RegisterAttClass(ATT_NOTEHEADS);
@@ -91,6 +93,7 @@ void Note::Reset()
     ResetColor();
     ResetColoration();
     ResetCue();
+    ResetExtSym();
     ResetGraced();
     ResetNoteAnlMensural();
     ResetNoteHeads();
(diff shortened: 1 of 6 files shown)


src/note.cpp
```diff
@@ -398,6 +401,53 @@ wchar_t Note::GetMensuralSmuflNoteHead()
...
 
+wchar_t Note::GetNoteheadGlyph(const int duration) const 
+{
+    static std::map<std::string, wchar_t> additionalNoteheadSymbols
+        = { { "noteheadDiamondBlackWide", SMUFL_E0DC_noteheadDiamondBlackWide },
+            { "noteheadDiamondWhiteWide", SMUFL_E0DE_noteheadDiamondWhiteWide },             
+            { "noteheadNull", SMUFL_E0A5_noteheadNull } };
+
+    if (HasGlyphName()) {
+        auto glyph = GetGlyphName();
```
lpugin, line 412: Don't use `auto`

src/note.cpp
```diff
@@ -398,6 +401,53 @@ wchar_t Note::GetMensuralSmuflNoteHead()
     return code;
 }
 
+wchar_t Note::GetNoteheadGlyph(const int duration) const 
+{
+    static std::map<std::string, wchar_t> additionalNoteheadSymbols
+        = { { "noteheadDiamondBlackWide", SMUFL_E0DC_noteheadDiamondBlackWide },
+            { "noteheadDiamondWhiteWide", SMUFL_E0DE_noteheadDiamondWhiteWide },             
+            { "noteheadNull", SMUFL_E0A5_noteheadNull } };
```
lpugin, line 409: Are these missing in MEI? If yes, then it would be good to see if they should be added to MEI.

lpugin (changes requested):
Please add at least one test file to https://github.com/rism-ch/verovio.org/tree/gh-pages/_tests/note

Monceber:
Added PR for example file:
https://github.com/rism-ch/verovio.org/pull/9

src/note.cpp
Monceber: They are missing in MEI, yes. It would be good if noteheadNull were to be added, but other too feel more like alternatives to diamond shape that is already in the MEI. Albeit slightly wider, I'm not sure if that warrants a new change in MEI.

Monceber: But that feels more like enhancement that can be addressed separately when (or if) MEI adds them.

lpugin: OK, thanks.

petermcculloch:
The slash notehead shape -- should that be E100 noteheadSlashVerticalEnds instead of E101?  E101 will fit better with the stem and is inline with how Dorico does it:

https://steinberg.help/dorico/v2/en/dorico/topics/notation_reference/notation_reference_notes_notehead_designs_r.html

Monceber:
@petermcculloch Well, the problem is that there's only E101 currently drawn/generated for Leipzig/Bravura fonts. E100 will need to be added first before it can be used.

rettinghaus:
@petermcculloch E100 doesn't match E102 and E103 as well. 
LilyPond also uses slashes with horizontal ends: http://lilypond.org/doc/v2.19/Documentation/notation/note-head-styles.html

PR #1636, opened 2020-08-25
Develop support different fonts

Pryadka:


lpugin:
Why was this closed?

Pryadka:
> Why was this closed?

Sorry for this. I created pull request earlier than it should be done. It has not been tested yet on our side. I guess new request will be created soon.

PR #1640, opened 2020-08-25
Rewrite ungroup code

JRegimbal:
Essentially split apart ungrouping neumes and syllables and make it
possible to keep ligatures intact when ungrouping neumes unless the
entire neume is the ligature.


lpugin:
No memory leak? ;-) Please double check.

JRegimbal:
@lpugin I'm probably going to feel silly once I notice it, but I can't seem to spot the memory leak. I think that all the heap memory I'm allocating end up being objects that are added to the document which handles deallocation later, right?

lpugin:
> @lpugin I'm probably going to feel silly once I notice it, but I can't seem to spot the memory leak. I think that all the heap memory I'm allocating end up being objects that are added to the document which handles deallocation later, right?

no worries! I was just kidding. I know it can be tricky to see them.

Issue #1555, opened 2020-07-13
Support for SMuFL's engravingDefaults as Verovio options


rettinghaus:
Many of Verovios styling options resemble those from SMuFL's engravingDefaults. The main difference is that SMuFL specifies "all measurements expressed in staff spaces", while Verovio uses MEI's virtual units. 

Wouldn't it make sense to follow the SMuFL specifications (as long as the options do not correspond to MEI attributes)?

https://w3c.github.io/smufl/gitbook/specification/engravingdefaults.html

rettinghaus:
(Basically, we somewhat agreed on that already in #368.)

lpugin:
It would be good to support SMuFL options. At the same time we need to keep the door open for additional styling features in MEI that will be using MEI units. For this reason, and for ensuring consistency, it is preferable to keep Verovio options in MEI units. What about adding a single option (e.g., --smuflEngraving) that takes a JSON object (or file in the command line?) and that converts them to Verovio / MEI values?

rettinghaus:
Taking a JSON object sounds good. But I think it would be confusing to keep all the options around then. For full SMuFL compatibility we would have to add even more, as right now many values are hard coded, e.g. `beamSpacing`, `beamThickness`, `legerLineThickness`, and `thickBarlineThickness`. 
Perhaps we should keep direct options to a minimum instead and deprecate things like `--lyric-line-thickness`.

lpugin:
We can leave existing options for now. We do not want to force people to use SMuFL engraving options. Once we have a better styling structure in MEI we can clean them up.

lpugin:
The issue should have been closed already.

PR #1644, opened 2020-08-27
Implement support for engravingDefaults as option

Monceber:
Solution for the #1555 

Added an option that would work as following:
./verovio -a -r ../Data test.mei **--engraving-defaults config.json**

Existing options are retained in default configuration that is read when no option is provided. Existing options are left untouched and would override corresponding value when provided, e.g.
./verovio -a -r ../Data test.mei --engraving-defaults config.json **--stem-width 0.4**

Stem width option would be used even if there is stemThickness configuration in the config.json.


lpugin:
It is good to have this new option to load SMuFL options. However, internally, it should be the other way around. That is, simply copy the JSON / SMuFL option to each Verovio option as appropriate, and not the other way around. This will be simpler and more efficient.

lpugin (changes requested):
See comment

lpugin (commented):
For some reasons I can still see the old changes. Maybe the simplest is to redo a fresh PR with the latest solution? The only thing I see needs to be changed is `Options::sync` to be capitalized as `Options::Sync`

Monceber:
I'll create new PR if that the case.

PR #1652, opened 2020-08-31
Automatic rest collision avoidance for two layers Mk.2

Monceber:
Improved performance compared to the previous implementation. About 56 pages of rests are now rendered in ~2.4 seconds compared to previous 75 seconds. Same is rendered in ~1 second without changes, but I think it's the most amount of improvements one can get from this.


Monceber:
@lpugin 
Regarding our previous conversation about options - would you rather have none at all and have configurations hardcoded in the map of maps (as in yout example), or would you be fine if json options were to be parsed into such map?

lpugin:
This looks much better. Speed seems to be fine on the example I previously tested with.

I had a look at the `dyanmic_cast` vs `static_cast` question. Please have a look at this https://github.com/rism-ch/verovio/commit/6c9aab66d4a7e63f6c33b74172d6080af3a0eb9b#commitcomment-41935904 Comments are welcome.

Regarding the map values and the option, we should go for the map of maps following the example I gave. I have not looked at enough examples to understand what the implication it would have to have the values hard-coded and not changeable with an option. If feel it should be OK for now. If you think it is necessary to have the option now, then we can look at what is the best way to parse the JSON for such things with as less code as possible (see https://gist.github.com/lpugin/2b11de6faaf67affe3fcf4b12097b428).

On thing that came to my mind is that it could be good to have some sort of schema users can have in hand to make sure the values they pass with such options are valid. This can be quite helpful. This can open the door to have the code for reading (and writing?) them generated from the Schema. This would be very much MEI / Verovio like. Of course, it can be based on a JSON Schema for the options instead of XML. This might be over the top for one option, though. Opinions on this?

Monceber:
Went with hardcoded map of map for the time being. Using XML/Json schema to generate code is interesting idea, but I agree a decent amount of options will be needed to warrant adding this.

include/vrv/functorparams.h
```diff
@@ -1927,6 +1959,32 @@ class UnsetCurrentScoreDefParams : public FunctorParams {
...
+        m_initialElementId = elementId;
+        m_searchDirection = searchDirection;
+        m_isInNeighboringLayer = anotherLayer;
+    }
+    Object *m_relativeElement;
+    int m_initialElementId;
+    bool m_searchDirection;
+    bool m_isInNeighboringLayer;
+};
+
```
lpugin, line 1987: Order alphabetically within the list of FunctorParams subclasses

include/vrv/options.h
```diff
@@ -20,6 +20,7 @@
 
 #include "attalternates.h"
 #include "atttypes.h"
+#include "jsonxx.h"
```
lpugin, line 23: This can be removed for now

include/vrv/options.h
```diff
@@ -101,6 +104,7 @@ class Option {
 protected:
     std::string m_title;
     std::string m_description;
+    bool m_isSet;
```
lpugin, line 107: Can be removed too?

src/layerelement.cpp
```diff
@@ -1622,6 +1612,34 @@ int LayerElement::LayerCountInTimeSpan(FunctorParams *functorParams)
...
+
+    Layer *currentLayer = static_cast<Layer *>(GetFirstAncestor(LAYER));
+    if (!currentLayer || (currentLayer != params->m_layer) || IsScoreDefElement() || Is(MREST)) 
+        return FUNCTOR_SIBLINGS;
+    if (!GetDurationInterface() || Is(MSPACE) || Is(SPACE) || HasSameasLink())
+        return FUNCTOR_CONTINUE;
+
+    const double duration = !GetParent()->Is(CHORD)
+        ? GetAlignmentDuration(params->m_mensur, params->m_meterSig)
+        : static_cast<Chord *>(GetParent())->GetAlignmentDuration(params->m_mensur, params->m_meterSig);
```
lpugin, line 1628: This will need to be a `vrv_cast` once merged with the latest version of develop. Not sure what is the best way to resolve it.

src/layerelement.cpp
```diff
@@ -1622,6 +1612,34 @@ int LayerElement::LayerCountInTimeSpan(FunctorParams *functorParams)
     return (this->Is(CHORD)) ? FUNCTOR_SIBLINGS : FUNCTOR_CONTINUE;
 }
 
+int LayerElement::LayerElementsInTimeSpan(FunctorParams *functorParams)
+{
+    LayerElementsInTimeSpanParams *params = static_cast<LayerElementsInTimeSpanParams *>(functorParams);
```
lpugin, line 1617: Will now be a `vrv_params_cast`

src/layerelement.cpp
```diff
@@ -1622,6 +1612,34 @@ int LayerElement::LayerCountInTimeSpan(FunctorParams *functorParams)
...
+    assert(params);
+
+    Layer *currentLayer = static_cast<Layer *>(GetFirstAncestor(LAYER));
```
lpugin, line 1620: Will now be a `vrv_cast`

lpugin (changes requested):

PR #1655, opened 2020-09-03
Fix offsets for rests

Monceber:
- switched places for same and other layer in default rest offsets (correct order)
- removed invalid condition


lpugin:
Could you add some test files to verovio.org? Thanks

PR #1662, opened 2020-09-07
Add precomposed trills and mordents

rettinghaus:
This PR mainly adds baroque trills and mordent glyphs.


lpugin (changes requested):
Please add one or more examples

rettinghaus:
Range of precomposed ornaments in FontForge 
Bildschirmfoto 2020-09-08 um 09 08 03

PR #1666, opened 2020-09-09
Develop stem in beam

Pryadka:
Root cause of this issue is wrong note-head position calculation. For both notes in the beam, the first and the last, the width of the first note was used for note-head position calculation, although both width were calculated. This commit also includes:
- replace naked pointer with reference for obtaining some fields;
- replace method at() of std::vector with operator[], because the second approach is a little bit faster (at() does the same as [] does, but also performs index range checking and throws exception. We don't need this functionality and don't catch corresponding exception)
- some code refactoring aimed to make code more readable and nice.


lpugin:
Looks good. I am only not convince with the change from `.at` to `[]`. It was previously always `[]` and this was changed to `.at` since the impact on speed is not significant. It is true that we do not catch exceptions, but `.at` does make bug tracking much easier. There are many places were `[]` can yield undefined behavior, especially where we have an array holding index positions in another array.

Pryadka:
> Looks good. I am only not convinced with the change from `.at` to `[]`. It was previously always `[]` and this was changed to `.at` since the impact on speed is not significant. It is true that we do not catch exceptions, but `.at` does make bug tracking much easier. There are many places were `[]` can yield undefined behavior, especially where we have an array holding index positions in another array.

Yes, performance will grow not very significant, but nevertheless. Before replacing, I analyzed the use of each index and concluded that the wrong index could not appear. `beamElementCoords` is const and we have checks for cases when i-1 or i+1 can be out of range. I would prefer to use [], but if you prefer at(), I will turn it back. Please tell me you final decision.

lpugin:
> Yes, performance will grow not very significant, but nevertheless. Before replacing, I analyzed the use of each index and concluded that the wrong index could not appear. `beamElementCoords` is const and we have checks for cases when i-1 or i+1 can be out of range. I would prefer to use [], but if you prefer at(), I will turn it back. Please tell me you final decision.

You are right that as it stands wrong index cannot not appear. But the code can be modified and it reminds me too much of nasty bugs terribly difficult to track, so please change them back to `.at()` since I prefer to stay on the safe side. All the other changes look good and do make the code more readable. Thanks.

Pryadka:
@lpugin Done.

PR #1670, opened 2020-09-10
Take into account grace factor of the beam's notes

Pryadka:
When a note head width is calculated we have to take into account grace notes.


lpugin:
Thanks. I had just noticed the problem too

PR #1671, opened 2020-09-10
Improve vertical justification of last page

Monceber:
Changed logic so that last page would be justified regardless of the additional options, if `--justify-vertically` option has been provided.
Justification for last page is based on the spacing/proportions of the previous one if there are not enough of systems in it.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -456,7 +456,6 @@ public:
     OptionBool m_condenseTempoPages;
     OptionBool m_evenNoteSpacing;
     OptionBool m_humType;
-    OptionBool m_justifyIncludeLastPage;
     OptionBool m_justifyVertically;
     OptionBool m_landscape;
     OptionBool m_mensuralToMeasure;
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -527,10 +527,6 @@ Options::Options()
     m_humType.Init(false);
     this->Register(&m_humType, "humType", &m_general);
 
-    m_justifyIncludeLastPage.SetInfo("Justify including the last page", "Justify including the last page");
-    m_justifyIncludeLastPage.Init(false);
-    this->Register(&m_justifyIncludeLastPage, "justifyIncludeLastPage", &m_general);
-
     m_justifyVertically.SetInfo("Justify vertically", "Justify spacing vertically to fill the page");
     m_justifyVertically.Init(false);
     this->Register(&m_justifyVertically, "justifyVertically", &m_general);
--- a/src/page.cpp
+++ b/src/page.cpp
@@ -521,6 +521,7 @@ void Page::JustifyVertically()
     assert(pages);
     if (pages->GetLast() == this) {
         int idx = this->GetIdx();
+        const int childSystems = GetChildCount(SYSTEM);
         if (idx > 0) {
             Page *penultimatePage = dynamic_cast<Page *>(pages->GetPrevious(this));
             assert(penultimatePage);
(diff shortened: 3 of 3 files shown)


craigsapp:
> Justification for last page is based on the spacing/proportions of the previous one if there are not enough of systems in it.

That is great!

What happens if there is only one page of music and `--justify-vertically` is used?

Monceber:
If that's the case, behavior remains unchanged from what it is right now - justification will be based on calculated values (since there is no previous page).

craigsapp:
> If that's the case, behavior remains unchanged from what it is right now - justification will be based on calculated values (since there is no previous page).

OK.  But note this silly situation (which I probably reported in the issues but cannot find):

https://editor.verovio.org/?file=https://raw.githubusercontent.com/rism-ch/verovio.org/gh-pages/_tests/clef/clef-001.mei

Then select PDF download, and then view the resulting PDF:

Screen Shot 2020-09-10 at 7 56 14 AM

Monceber:
@craigsapp 
I think this is a separate issue (and I'm not sure it's a correct one on top of that). A check can be added that justification should be skipped if first page is the last page. However, we will kill some proper use cases with this approach.
For instance, we want to have proper justification in cases when first page has more than 10 staves (approximately). And running justification for such examples is more realistic than running it for the score with one page and two staves, imo.

lpugin:
The issue reported by @craigsapp was introduced in the refactoring of the justification. This is clearly an issue and it should be fixed. I would suggest to fix it before removing the include-last-page option.

rettinghaus:
@lpugin but why is it working in the browser and only the PDF is looking this strange?

lpugin:
I think is it because it has the option "justifyIncludeLastPage" set to true for the PDF.

craigsapp:
> For instance, we want to have proper justification in cases when first page has more than 10 staves (approximately). And running justification for such examples is more realistic than running it for the score with one page and two staves, imo.

Yes, if someone wants to print short examples like this, they should not turn on the vertical justification option.  Although for non-technical users, how to do this might not be apparent, since they would be using an interface such as the MEI viewer and printing from the menu without awareness of the options (and if it is possible to change them).

Perhaps a maximum vertical staff spacing option could be useful to autodetect such cases: if staves would vertically be adjusted beyond a certain distance (such as 100 diatonic steps, or about 10 staff heights), then vertical justification for that page would be disabled (not set to the maximum).  If someone wanted to override the default maximum vertical spacing, they could set the maximum to a larger value, such as 1000 diatonic steps, which would probably be more than the page height.

Of course, such an option would be outside of the scope of this pull request.

> @lpugin but why is it working in the browser and only the PDF is looking this strange?

I suspect that vertical justification is being added only in the PDF rendering, while in the viewer it is not. (as I see @lpugin has just stated).

Monceber:
For the time being I've made change so that we are basing such cases (when there is only one page) on the amount of the staves that are present. In case when we have only a handful of them, we won't justify to avoid situation like above (where two staves are stretched over the whole page). When there is more reasonable amount for us to justify page, then we do it.

This approach is probably is not the best out there and in the long run option for the maximum spacing as @craigsapp suggested is a way to go. However at this point it does solve this edge case. If there are objections to this approach, please let me know.

craigsapp:
That sounds good for now, and your solution should only fail for short pages with only a few staves on the system.  This might occur in on-screen situations, but all of the current online verovio interfaces that I know of do not use vertical justification for the screen display, only for creating PDFs.

Monceber:
@lpugin Are you ok with these changes?

PR #1672, opened 2020-09-10
Develop brackets

rettinghaus:
This PR 
* improves drawing of staffGrp brackets, so that the glyphs won't touch the staff
* adds support for repeatEndingLineThickness from engravingDefaults
* sorts engraving defaults in Sync(), following the list in https://w3c.github.io/smufl/gitbook/specification/engravingdefaults.html

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -533,6 +533,7 @@ public:
      */
     OptionGrp m_generalLayout;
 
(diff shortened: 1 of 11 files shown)


lpugin (changes requested):
Could you run clang-format? Thanks

PR #1675, opened 2020-09-10
Removes two assignment operators result of which were not used

valeriyvan:
May be this code is needed and its current state just indicates some other bug which I miss.


lpugin:
Please comment the two lines instead of removing them

PR #1678, opened 2020-09-11
Revert the change which leads to whong beam slope calculation

Pryadka:
This is hotfix for this issue https://github.com/rism-ch/verovio/issues/1677 
- revert improper change


rettinghaus (approved):

rettinghaus (approved):

lpugin:
@Pryadka I've notice some weird things with slurs and articulations with beams. For example here https://www.verovio.org/app.html. I am trying to nail it down to a minimal example so we can see if this is is related to the latest changes made to the beam calculation but could not yet. Just to let you know while I am still looking at this.

PR #1679, opened 2020-09-12
Fixes memory leak.

valeriyvan:
Also gets rid of unneeded cast of size_t to int.

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -11545,7 +11545,6 @@ template <class ELEMENT>
 void HumdrumInput::addTextElement(
     ELEMENT *element, const std::string &content, const std::string &fontstyle, bool addSpacer)
 {
-    Text *text = new Text;
     std::string data = content;
     if (element->GetClassName() == "Syl") {
         // Approximate centering of single-letter text on noteheads.
@@ -11564,7 +11563,11 @@ void HumdrumInput::addTextElement(
     std::vector<std::string> pieces;
     hre.split(pieces, data, "\\\\n");
 
-    for (int i = 0; i < (int)pieces.size(); i++) {
+    const size_t piecesSize = pieces.size();
+
+    Text *text = piecesSize > 0 ? new Text : NULL;
+
+    for (size_t i = 0; i < piecesSize; i++) {
         data = pieces[i];
         text->SetText(UTF8to16(data));
 
@@ -11591,7 +11594,7 @@ void HumdrumInput::addTextElement(
             }
         }
 
-        if (i < (int)pieces.size() - 1) {
+        if (i < piecesSize - 1) {
             // Need to add another text element, but add lb before it.
             Lb *lb = new Lb;
             element->AddChild(lb);


src/iohumdrum.cpp
```diff
@@ -11591,7 +11594,7 @@ void HumdrumInput::addTextElement(
             }
         }
 
-        if (i < (int)pieces.size() - 1) {
+        if (i < piecesSize - 1) {
```
craigsapp, line 11597: This line of code is precisely why I do not use unsigned integers for indexes.  If `piecesSize` is `0`, then `piecesSize-1` is `MAX_INT`, and that is highly likely to be larger than `i`.  It does not matter if `piecesSize` is guaranteed to never be `0` at this case, because in the future something above the code could change, and maybe `0` would be allowed.  Hell will freeze over before I use `size_t` for small indexes in code where I am the primary maintainer (which is `iohumdrum.cpp`).  So: use type `int` for `piecesSize`.

src/iohumdrum.cpp
```diff
@@ -11564,7 +11563,11 @@ void HumdrumInput::addTextElement(
     std::vector<std::string> pieces;
     hre.split(pieces, data, "\\\\n");
 
-    for (int i = 0; i < (int)pieces.size(); i++) {
+    const size_t piecesSize = pieces.size();
```
craigsapp, line 11566: By the logic I give for using `int` in the comment for line 11597, your fix is proper for long-term stability of the code.

However, note that I claim that `pieces` can never have a size less than one, so there is a guarantee that no memory leak can occur (which would happen if `pieces` had a size of 0).  In other words, once this function is entered, it is committed to inserting at least one text element, even if it were an empty string.

As an aside: the for-loop and the lines:

```
    hum::HumRegex hre;
    std::vector<std::string> pieces;
    hre.split(pieces, data, "\\\\n");
```

Are a fairly recent addition to the function to allow for line breaks in the text (mostly for `<dir>` elements).  A string such as `line1\nline2` will be converted into two text elements containing `line1` and `line2` with an `<lb>`  between them.  So `pieces` will have a size of 1 if there are not line breaks, and otherwise will be greater than one if there are any line breaks indicated by `\n` characters in the input string.

craigsapp (changes requested):

src/iohumdrum.cpp
valeriyvan: If `piecesSize` is 0 `for`-loop will not be executed. So we are safe.

src/iohumdrum.cpp
valeriyvan: How `hum::HumRegex::split` behaves in case of empty string? I suppose in this case returned vector will be empty. Or may contain empty string as the only element. Both implementations seem logical.

May be you are right and leak will never happen. To say that for sure one should know if `hum::HumRegex::split` ever returns empty vector.

Proposed change makes code work properly despite of logic of `hum::HumRegex::split`.

Think of possible case when logic of `hum::HumRegex::split` changes. Then here real leak might happen.

lpugin: @craigsapp what do you want to do with this?

lpugin:
Closing since the PR has now conflicts. Please redo if necessary

craigsapp:
Oops it got lost in the past.  I will check on it.

PR #1683, opened 2020-09-15
add missing Schleifer and short piano pedal glyphs

rettinghaus:
Added glyphs:
newglyphs


lpugin:
Could you post an image (e.g., screenshot) of the glyphs added? This is always useful to see. This means that people interested do not need to either download the new font file and open it with FontForge to see the glyphs, or go glyph by glyph and look in SMuFL. Thanks. (I added a comment to the contributing guidelines about it)

rettinghaus:
Sure. I always forget, because I just saw them.

PR #1687, opened 2020-09-17
Fix support for legerLineExtension and legerLineThickness

rettinghaus:
The given values for ledger lines weren't handled correctly. This PR fixes this issue and allows smaller values for `subBracketThickness` to fully support engraving defaults for Bravura. 

NB: Interestingly engravingDefaults for all fonts included in Verovio have "legerLineExtension" set to 0.4 which seems really large. E.g. MuseScore doesn't use this value (correctly?) and shortens the ledger lines, only Dorico works as expected. Verovio's default value was/is 0.27 (or 0.54vu). So I think, that the corresponding value in Leipzig's metadata.json should be adapted to this.

### DIFF
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -877,14 +877,13 @@ int Note::CalcLedgerLines(FunctorParams *functorParams)
 
     if (!this->HasLedgerLines(linesAbove, linesBelow, staff)) return FUNCTOR_CONTINUE;
 
-    // HARDCODED
     const int ledgerLineExtension
         = params->m_doc->GetOptions()->m_ledgerLineExtension.GetValue() * params->m_doc->GetDrawingUnit(staffSize);
-    int leftExtender = 2.5 * ledgerLineExtension;
-    int rightExtender = 2.5 * ledgerLineExtension;
+    int leftExtender = ledgerLineExtension;
+    int rightExtender = ledgerLineExtension;
     if (drawingCueSize || (this->GetDrawingDur() >= DUR_8)) {
-        leftExtender = 1.75 * ledgerLineExtension;
-        rightExtender = 1.25 * ledgerLineExtension;
+        leftExtender = 0.7 * ledgerLineExtension;
+        rightExtender = 0.5 * ledgerLineExtension;
     }
 
     if (drawingCueSize) {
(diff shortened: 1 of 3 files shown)


lpugin:
Could you illustrate this? Thanks

rettinghaus:
Current default rendering:
ledger_old

New default rendering:
ledger_new

New rendering with engraving defaults taken from Leipzig:
ledger_new_ed

New rendering with Bravura font  and engraving defaults taken from Bravura:
ledger_new_b

BTW: the generated `leipzig-5.2_metadata.json` isn't conformant with SMuFL metadata format, as the engravingDefaults values are all passed as string.

rettinghaus:
And the MEI source:
```xml
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>Leger lines test</title>
      </titleStmt>
      <pubStmt>
        <respStmt>
        </respStmt>
      </pubStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef meter.count="4" meter.unit="4">
            <staffGrp symbol="bracket" bar.thru="false">
              <staffDef n="1" clef.line="3" clef.shape="C" lines="5">
              </staffDef>
            </staffGrp>
          </scoreDef>
          <section>
            <measure>
              <staff n="1">
                <layer n="1">
                  <note pname="c" oct="5" dur="2"/>
                  <beam>
                    <note pname="c" oct="5" dur="8"/>
                    <note pname="d" oct="5" dur="8"/>
                    <note pname="c" oct="5" dur="8"/>
                    <note pname="b" oct="4" dur="8"/>
                  </beam>
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

lpugin:
@rettinghaus could you have a look at this https://github.com/rism-ch/verovio/pull/1689 ?

rettinghaus:
Rebased it on latest develop. Anything left to do, @lpugin?

craigsapp:
Screen Shot 2020-09-21 at 10 04 13 AM

How would you select between the two?  The bottom widths are very much preferred.  However, when there are adjacent notes with ledger lines that are spaced closely together, the top case is needed to provide a visual gap between the ledger lines.   

@lpugin left it small so that the same width was used in all situations, but a rendering refinement would automatically use the shorted ledger lines if otherwise no visual gap (maybe 0.5 vu) is present with the long forms.

rettinghaus:
@craigsapp you are wrong about "that the same width was used in all situations". To avoid collision in beams all notes shorter than a quarter get shortened ledger lines, this wasn't changed. 
You can change default ledger lines with the new options `ledgerLineThickness` and `legerLineExtension` (or by providing `engravingDefaults`).

PR #1689, opened 2020-09-17
Develop leipzig

lpugin:
Update script that generates SMuFL metadata
- changes strings to number
- chang value for legerLineExtension

Fonts need to be re-generated


rettinghaus (approved):
Looks great!

PR #1691, opened 2020-09-18
Develop support for plain text fonts

Monceber:
Added scripts for generating bbox XML output for the plain text fonts. They has similar behavior to **generate_all.sh** to extract data from font. The way to run is as following:
```
./generate_text_font.sh fontname.extension
```
It will generate *-bounding-boxes.svg file as well as .g2n mapping file (to map symbol names to Unicode numbers). Afterwards, XML will be generated based on the bboxes and map input.

Code has been adjusted accordingly to take option for text font to allow changing default font from Times to a custom one.


Monceber:
Doesn't look like Travis picked up correct code for the build:
```
/Users/travis/build/rism-ch/verovio/tools/main.cpp:384:26: error: no member named 'SetFont' in 'vrv::Resources'
if (!vrv::Resources::SetFont(options->m_font.GetValue())) {
```

This PR has no call to SetFont at line 384:
```
if (!vrv::Resources::SetMusicFont(options->m_font.GetValue())) {
```

Should I re-create PR again?

src/vrv.cpp
```diff
@@ -84,37 +84,45 @@ bool Resources::InitFonts()
...
+{
+    return LoadFont(fontName);
+}
+
+bool Resources::SetTextFont(const std::string &fontName, const bool usePostfixes)
+{
     struct TextFontInfo_type {
         const StyleAttributes m_style;
-        const std::string m_fileName;
+        const std::string m_postfix;
```
lpugin, line 106: What does `postfix` mean? Looks like a weird name change.

include/vrv/options.h
```diff
@@ -581,6 +581,7 @@ class Options {
     OptionDbl m_thickBarlineThickness;
     OptionDbl m_tieThickness;    
     OptionDbl m_tupletBracketThickness;
+    OptionString m_textFont;
```
lpugin, line 584: Fix alphabetical order

lpugin (changes requested):

src/vrv.cpp
Monceber: Previous since it was hardcoded it was referring to the filename (e.g. Times-italic). Now that we can change text font by option, this refers to the postfix for the weight/style combination that should be added to the text font name when doing InitTextFont(). We have postfixes like -bold, -italic, -bold-italic and when init font (for instance, Arial) we get file following file names: Arial-bold, Arial-italic, etc.

lpugin (approved):

lpugin:
@Monceber sorry for reverting but it needs to investigated why this breaks the JS toolkit.

PR #1692, opened 2020-09-18
fix and speed up tie import

rettinghaus:
I stumbled upon some bugs with tie import. 
So this PR fixes my issues, while reducing general memory usage and processing time.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2708,9 +2708,15 @@ void MusicXmlInput::ReadMusicXmlNote(
         pugi::xpath_node startTie = notations.node().select_node("tied[@type='start']");
         pugi::xpath_node endTie = notations.node().select_node("tied[@type='stop']");
         if (endTie) { // add to stack if (endTie) or if pitch/oct match to open tie on m_tieStack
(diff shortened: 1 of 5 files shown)


src/iomusxml.cpp
```diff
@@ -2708,9 +2708,15 @@ void MusicXmlInput::ReadMusicXmlNote(
         pugi::xpath_node startTie = notations.node().select_node("tied[@type='start']");
         pugi::xpath_node endTie = notations.node().select_node("tied[@type='stop']");
         if (endTie) { // add to stack if (endTie) or if pitch/oct match to open tie on m_tieStack
-            m_tieStopStack.push_back(note);
+            if (note->GetPname() == m_tieStack.back().second->GetPname() && note->GetOct() == m_tieStack.back().second->GetOct()) {
```
lpugin, line 2711: Just to make sure: don't you need to check that m_tieStack is not empty before accessing `.back()`?

rettinghaus: On Friday I don't need that, but on Monday I don't feel so sure about that anymore … ;-)

PR #1698, opened 2020-09-24
Develop beam

Pryadka:
This request is compiled from previously done commit which fixed stem positions for different note heads within the beam(request 1666) and the set of fixes for its errors.

### DIFF
--- a/src/beam.cpp
+++ b/src/beam.cpp
@@ -265,6 +265,9 @@ void BeamSegment::CalcBeamInit(
     }
     beamInterface->m_beamWidth = beamInterface->m_beamWidthBlack + beamInterface->m_beamWidthWhite;
 
+    if (!m_firstNoteOrChord) return;
+    if (!m_lastNoteOrChord) m_lastNoteOrChord = m_firstNoteOrChord;
+
     Note *firstNote = NULL, *lastNote = NULL;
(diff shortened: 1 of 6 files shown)


lpugin:
I will have a look. Could you respond to the couple of comments I made there https://github.com/rism-ch/verovio/commit/6e534fad88b0a196ab86c9e0d32540a976d788b8 Thanks!

src/beam.cpp
```diff
@@ -293,8 +293,9 @@ void BeamSegment::CalcBeamInit(
     }
 
     // Only if not only rests. (Will produce non-sense output anyway)
-    if (m_beamElementCoordRefs.size() != nbRests) {
-        this->m_avgY /= (m_beamElementCoordRefs.size() - nbRests);
+    const auto elementCount = static_cast<decltype(m_avgY)>(m_beamElementCoordRefs.size());
```
lpugin, line 281: Would a cast to `(int)` just work? I know this might be more correct, but I don't think we are going to ever have problems with `.size()` cast to `int`. This looks overly complicated, and I don't like seeing `auto` in such places because it makes it harder to understand what the type actually is.

Pryadka: Ok. done.

PR #1701, opened 2020-09-24
improve tuplet drawing

rettinghaus:
This PR takes glyph height of tuplet numbers into account, to get a better result with different fonts. 

_Examples from Verovio test suite rendered with Petaluma_
Current:
tuplet-017
New:
tuplet-017 new

Current:
tuplet-018
New:
tuplet-018 new


rettinghaus:
Is the adjustment of the length of the vertical lines wanted, or should this value be unchanged?

lpugin:
> Is the adjustment of the length of the vertical lines wanted, or should this value be unchanged?

I think is it good to adjust it.

PR #1703, opened 2020-09-28
Prevent slurs over breaks looking like ties

rettinghaus:
As Gould states on slurs over system breaks: 
> The whole slur should tilt in the direction of the pitches.

Current behaviour:
slursold

Proposed:
slursnew


lpugin:
image

I wonder why this one is not corrected. Any ideas? Because it is not "attached" to the note head?

rettinghaus:
How should it look? This PR just addresses starting / ending points for open slurs.

lpugin:
I thought you would want that one to do up. But I can merge it as it is if you want to.

rettinghaus:
@lpugin As I know, you prefer smaller changes for review. ;-) 
So please merge.

PR #1704, opened 2020-09-28
improve group symbol position

rettinghaus:
Positioning of brackets was still hard coded, although bracket thickness could be set. Changed that, to work in all cases. 
Also the default value for `subBracketThickness` was equal to `BracketThickness`, but should approximate staff line width.


rettinghaus:
before:
brackold
after:
bracknew

craigsapp:
It would be useful to provide the MEI data used to create the before / after (and add them to the test set if useful).

rettinghaus:
@craigsapp this is actually from the test set.

rettinghaus:
score-008.mei

lpugin:
@rettinghaus 

> this is actually from the test set.

This is good. Then (next time) a link to the file would be enough

PR #1710, opened 2020-09-30
Fix beam slope behavior

Monceber:
Fix for https://github.com/rism-ch/verovio/issues/1541


rettinghaus:
Looks great!

Before:
beamSlopeOld

After:
beamSlopeNew

Test file: https://github.com/rism-ch/verovio.org/blob/gh-pages/_tests/beam/beam-059.mei

PR #1722, opened 2020-10-05
Improve rendering of the situations when grace notes/groups collide with beams

Monceber:
Fixes two problems:
1. Beams containing grace notes/groups that end up overlapping with parent beam:
image

Fixes to following:
image

Coupled with #1721 it would look better, since it fixes related problem (grace groups within beams).

2. Clef change happening inside of the beam. Originally it would just overlap, now code tries to adjust beam vertical position to avoid colliding with clef:
image

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -805,6 +805,16 @@ public:
     virtual int CalcArtic(FunctorParams *) { return FUNCTOR_CONTINUE; }
 
     /**
+     * Calculate the vertical position adjustment for the beam if it overlaps with layer elements
+     */
+    virtual int AdjustBeams(FunctorParams *) { return FUNCTOR_CONTINUE; }
+
+    /**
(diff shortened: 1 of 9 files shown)


include/vrv/object.h
```diff
@@ -804,6 +804,16 @@ class Object : public BoundingBox {
...
+    /**
+     * Calculate the vertical position adjustment for the beam if it overlaps with layer elements
+     */
+    virtual int AdjustBeams(FunctorParams *) { return FUNCTOR_CONTINUE; }
+
+    /**
+     * Apply position adjustment that has been calculated previously
+     */
+    virtual int AdjustBeamsEnd(FunctorParams *) { return FUNCTOR_CONTINUE; }
+
```
lpugin, line 816: Usually I put them together (see AlignVertically above) so we straight away see there is an `*End` functor. But you did put a specific comment so it makes sense to have it like it.

PR #1724, opened 2020-10-05
Remove variable onStaffLine not read anymore after reсent code change

valeriyvan:


lpugin:
We should wait until #1725 is clarified

PR #1730, opened 2020-10-06
update font metadata

rettinghaus:
after updating the script, someone forgot to re-generate the metadata file ;-)


rettinghaus:
Should this go directly to develop branch?

PR #1751, opened 2020-10-16
fix(gh-actions): fix repo for deploy step (2)

musicEnfanthen:
After the fix in #1750 , the deploy step still failed due to permission issues. 

This PR tries to solve that by introducing a secret for a ssh deploy key, that needs to be set on the repos `rism-ch/verovio` and `rism-ch/verovio.org` (see below)

It also fixes the runner conditions in the c++ build, so that specified os and compiler are actually used and installed correctly. Additionally, it rearranges the dependencies between jobs: C++ build and JS build are now running in parallel (what reduces total compile time).

Ready to merge.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -19,16 +19,17 @@ env:
   EMSCRIPTEN_CACHE_FOLDER: emsdk-cache
 
   # gh-pages
-  GH_PAGES_REPO: rism-ch/verovio.org
+  GH_PAGES_REPO: ${{ github.repository_owner }}/verovio.org  # --> resolves to rism-ch/verovio.org
   GH_PAGES_BRANCH: gh-pages
-  GH_PAGES_FOLDER: gh-pages
 
   # build artifacts
   CLI_BUILD: cli-build
   TOOLKIT_BUILD: toolkit-build
 
-  # temp
-  TEMP_FOLDER: temp
+  # temporary directories
+  GH_PAGES_DIR: gh-pages-dir
+  TEMP_DIR: temp-dir
+
 
 jobs:
   ###############################################################################################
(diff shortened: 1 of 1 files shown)


musicEnfanthen:
It took quite some time to figure this out, because there seems to be no comprehensive explanation anywhere. So I am adding some hints for future reference.

Helpful pages with parts of the solution: [[1](https://github.community/t/automating-push-to-public-repo/17742/7)], [[2](https://github.community/t/push-built-files-to-repo-using-actions/18186)], [[3](https://dev.to/susomejias/use-token-for-github-actions-with-private-repositories-22m6)]
and finally: [[4](https://eriksamuelsson.com/access-private-submodules-in-github-actions/)]

**Summary:**
The gh-actions workflow has a built-in token (`GITHUB_TOKEN`) that only allows to git push to the same repo in which the workflow is defined (cf. here: [[5](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions)]).

But Verovio js builds need to be pushed from one repo (`rism-ch/verovio`) to another repo (`rism-ch/verovio.org`). To do so, a deploy key (ssh key) needs to be created on `rism-ch/verovio.org`, whose private can be used as a secret in `rism-ch/verovio`. (Instructions to create ssh-key, deploy key and secret can be found in [[4](https://eriksamuelsson.com/access-private-submodules-in-github-actions/)]. The secret name (e.g.` GH_ACTIONS_DEPLOY_KEY`) can then be referenced from the gh actions workflow YAML file as `${{ secrets.GH_ACTIONS_DEPLOY_KEY }}`

[1] https://github.community/t/automating-push-to-public-repo/17742/7
[2] https://github.community/t/push-built-files-to-repo-using-actions/18186
[3] https://dev.to/susomejias/use-token-for-github-actions-with-private-repositories-22m6
[4]  https://eriksamuelsson.com/access-private-submodules-in-github-actions/
[5] https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions

Issue #1726, opened 2020-10-06
Build test suite with GitHub Actions


lpugin:
It is now failing on building the doxygen documentation. Error message is
```
The job exceeded the maximum log length, and has been terminated.
```
Weird since it fails after about 5 minutes only. See https://travis-ci.org/github/rism-ch/verovio/builds/733209808

At this stage I would be quite happy to move to GitHub Actions. Help appreciated ;-)

musicEnfanthen:
The only way I see to fix this on Travis would be to make the log output less verbose. At the moment, doxygen build produces more than 28.000 lines of warnings what is terminating the build process. 

Two possible options: 
1) Configure doxygen in a way that it does not show the warnings; 
2) fix the warnings in Verovio code

Additionally, you could try to remove the verbose option in "`travis/ci_deploy-doxygen.sh`",

so `set -ev #` will become `set -e #`,

but my guess is that this is only a drop in the ocean in comparison to the huge amount of warnings.

musicEnfanthen:
Regarding GH actions I would be happy to help. Made a first draft some time ago, maybe it can be useful for further exploration, see https://github.com/musicEnfanthen/verovio/actions

C++ build is running through: https://github.com/musicEnfanthen/verovio/actions/runs/66312718/workflow

JS build is not: https://github.com/musicEnfanthen/verovio/actions/runs/66312719/workflow

musicEnfanthen:
Regarding GH actions I would be happy to help. Made a first draft some time ago, maybe it can be useful for further exploration, see https://github.com/musicEnfanthen/verovio/actions

C++ build is running through: https://github.com/musicEnfanthen/verovio/actions/runs/66312718/workflow

JS build is not: https://github.com/musicEnfanthen/verovio/actions/runs/66312719/workflow

lpugin:
Great, thanks! As a temporary fix I tried to remove the warnings from the doxygen output. Let's see.

Any idea what was not working for the JS build?

rettinghaus:
I also tried a bit with actions, but wasn't too successful so far. 
I thought, this could help with PRs:
https://github.com/HorstBaerbel/action-clang-format

musicEnfanthen:
Seems to work now without the doxygen output?

> Any idea what was not working for the JS build?

It was related to the precaching stage. Did not find emsdk folder with the current setup.

musicEnfanthen:
@rettinghaus Yes, could be good to run clang format checking/linting on every commit.

[...]

PR #1757, opened 2020-10-19
ci(gh-actions): Add dry run mode for gh actions

musicEnfanthen:
GH-Actions build workflow, as discussed in #1726 and implemented in #1751, deploys via `git push` to `<repository-owner>/verovio.org` and `<repository-owner>/verovio-doxygen` after finishing all steps. This is expected and desired behaviour on `rism-ch`, but could be undesired for verovio fork repos.

This PR adds a "dry-run mode" for verovio forks by adding a dry-run flag (`IS_DRY_RUN`) that is set to true by default. If so, `git push` commands will be executed as `git push -v --dry-run`, avoiding real push events to fork repos. When running on `rism-ch` repos, the `IS_DRY_RUN` flag is set programmatically to `false`, so push events behave normally. If forks need to actually push/deploy, the `IS_DRY_RUN` flag can be set manually to `false` in the `env` section at the top of the workflow.

Additionally, the PR mutes any other branches than `develop` to be run with the workflow, and uses gh-actions `working-directory: path/to/dir` option throughout the workflow instead of `cd /path/to/dir`.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -340,7 +340,7 @@ jobs:
 
     steps:
       - name: Prepare deployment for rism-ch
-        if: github.repository_owner == env.RISM_OWNER
+        if: ${{ github.repository_owner == env.RISM_OWNER }}
         run: |
           echo "Enable deploy steps on rism-ch..."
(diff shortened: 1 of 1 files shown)


musicEnfanthen:
Please wait with merging, I am checking a generic solution for #1758

PR #1760, opened 2020-10-20
ci(gh-actions): Mute automatic deployment from forks

### LINKED ISSUE #1758: only deploy on main repo
@musicEnfanthen is there a way in GitHub Actions to call the deploying only on the main repo? It's a bit annoying, that forks always fail here.

musicEnfanthen:
This PR addresses the issue raised in #1758 that on forks the gh-actions build workflow will not pass deployment steps in most cases due to missing permissions (secrets needed as described in #1751) or even missing forks of `.../verovio.org` and `.../verovio-doxygen` repos. These are the targets of the deployment from `rism-ch`, but can not be expected to exist on forks. Hence build workflow is likely to fail regularly on forks.

This PR introduces a check if the build workflow is running on `rism-ch` (set as `<MAIN_REPO_OWNER>`). If so, the deployment steps are automatically activated. For any other owner not matching `<MAIN_REPO_OWNER>` (forks), the deployment steps are skipped per default. They can be activated manually by setting the flag `DISABLE_DEPLOY_STEPS ` to `false`. Another flag, `IS_DRY_RUN` can be used together with `DISABLE_DEPLOY_STEPS: false` to call `git push` in dry-run mode in the deploy steps, so push event is only simulated, but not actually triggered. Different possible scenarios and their settings are reflected in the "test cases" of the single commits to this PR.

Fixes #1758.

Finishes #1757 .

Ready to merge.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -13,11 +13,11 @@ on:
 # globals
 env:
   # general settings
-  MAIN_REPO_OWNER: musicEnfanthen      # Main repo owner (default: rism-ch; should not be changed)
+  MAIN_REPO_OWNER: rism-ch      # Main repo owner (default: rism-ch; should not be changed)
                                 # If changed, owner needs deploy permission to <owner>/verovio.org and <owner>/verovio-doxygen (cf. Deploy jobs).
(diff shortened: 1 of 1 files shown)


lpugin:
Thanks for the detailed documentation.

PR #1770, opened 2020-10-21
Adjust mNum size

rettinghaus:
This PR adds support for `@fontsize` to `mNum` and adjusts the default size to score size instead of staff size. Before it was scaled with the top most staff size resulting in too tiny numbers. 

### Before:
mnum_old
### After:
mnum_new

### DIFF
--- a/src/view_page.cpp
+++ b/src/view_page.cpp
@@ -975,9 +975,33 @@ void View::DrawMNum(DeviceContext *dc, MNum *mnum, Measure *measure)
         // we set mNum to a fixed height above the system and make it a bit smaller than other text
         params.m_x = staff->GetDrawingX();
         params.m_y = staff->GetDrawingY() + 1.5 * m_doc->GetDrawingDoubleUnit(staff->m_drawingStaffSize);
-        params.m_pointSize = m_doc->GetDrawingLyricFont(100)->GetPointSize() * 4 / 5;
-
-        mnumTxt.SetPointSize(params.m_pointSize);
+        if (mnum->HasFontsize()) {
+            data_FONTSIZE *fs = mnum->GetFontsizeAlternate();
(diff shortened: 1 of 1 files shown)


src/view_page.cpp
```diff
@@ -975,9 +975,33 @@ void View::DrawMNum(DeviceContext *dc, MNum *mnum, Measure *measure)
...
+                switch (fs->GetTerm()) {
+                    case (FONTSIZETERM_xx_large): percent = 200; break;
+                    case (FONTSIZETERM_x_large): percent = 150; break;
+                    case (FONTSIZETERM_large): percent = 110; break;
+                    case (FONTSIZETERM_larger): percent = 110; break;
+                    case (FONTSIZETERM_small): percent = 80; break;
+                    case (FONTSIZETERM_smaller): percent = 80; break;
+                    case (FONTSIZETERM_x_small): percent = 60; break;
+                    case (FONTSIZETERM_xx_small): percent = 50; break;
+                    default: break;
```
lpugin, line 994: Maybe we could have this in a helper so we do not duplicate the hard-coded values? https://github.com/rism-ch/verovio/blob/3cbcfc8253dbd8ed059381004f224f10f95aa91c/src/view_text.cpp#L316-L328

We can have it directly in `data_FONTSIZE`. Something like `int data_FONTSIZE::GetPercentForTerm()`

rettinghaus: I thought about that too. But wouldn't it be better to have it in `options.cpp` as a map? That would make it more clear, that this is Verovio styling.

lpugin: These are pretty standard values. I agree it would be good to document them somewhere but I think having an option is over the top. When using xx-small users are deliberately leaving it to the UA.

rettinghaus: added proposed function

Issue #1767, opened 2020-10-20
Emscripten build breaks unpredictably on some GH action runs


musicEnfanthen:
For some build workflow runs, it seems that the emscripten config changes and clears the cache while running the build:

```
  shared:INFO: (Emscripten: Running sanity checks)
  [  4%] Building CXX object CMakeFiles/verovio.dir/home/runner/work/verovio/verovio/src/arpeg.cpp.o
  [  5%] Building CXX object CMakeFiles/verovio.dir/home/runner/work/verovio/verovio/src/artic.cpp.o
shared:INFO: (Emscripten: config changed, clearing cache)
shared:INFO: (Emscripten: config changed, clearing cache)
shared:INFO: (Emscripten: Running sanity checks)
  [  5%] Building CXX object CMakeFiles/verovio.dir/home/runner/work/verovio/verovio/src/att.cpp.o
Traceback (most recent call last):
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/em++.py", line 14, in <module>
  sys.exit(emcc.run(sys.argv))
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/emcc.py", line 753, in run
  shared.check_sanity(force=DEBUG)
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/tools/shared.py", line 599, in check_sanity
  Cache.erase()
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/tools/cache.py", line 81, in erase
  self.acquire_cache_lock()
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/tools/cache.py", line 54, in acquire_cache_lock
  self.filelock.acquire(60)
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/tools/filelock.py", line 240, in acquire
  self._acquire()
  File "/home/runner/work/verovio/verovio/emsdk-cache-310513115/emsdk-master/upstream/emscripten/tools/filelock.py", line 360, in _acquire
  fd = os.open(self._lock_file, open_mode)
[...]

PR #1781, opened 2020-10-23
ci(gh-actions): Fix emscripten caching

musicEnfanthen:
Another PR in regard to stabilize emscripten caching procedure (cf. #1767).

According to the kind trouble-shooting over there (https://github.com/mymindstorm/setup-emsdk/issues/9), it is recommended to set an explicit version of emscripten and to not use `runner_id` for run-specific cache naming.

This PR follows those steps, and, despite the (weird) fact, that emscripten still cleans the cache during build, the builds do not fail at least in tests: https://github.com/musicEnfanthen/verovio/runs/1296892725?check_suite_focus=true#step:7:52

Additionally, the total length of the script is reduced by about 70 lines by performing some housekeeping (removing unneeded steps, simplifying run calls).

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -7,6 +7,7 @@ on:
     branches:
       # Push events on develop branch
       - develop
+      - '**'
       # Push events on ci-test branch (uncomment if needed for testing purposes)
       # - ci-test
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_build.yml
```diff
@@ -38,8 +38,8 @@ env:
   DOXYGEN_BRANCH: master
 
   # emscripten
-  EMSCRIPTEN_VERSION: latest
-  EMSCRIPTEN_CACHE_FOLDER: emsdk-cache
+  EM_VERSION: 2.0.7
+  EM_CACHE_FOLDER: 'emsdk-cache'
```
lpugin, line 42: Do you think we can try without this first? I quite like using the latest so we are always up-to-date. This is also a good way for us to now straightaway when something breaks with a new version of the emdk. Noticing it straightway usually makes it easier to find where it breaks. (If we need to we can go with this solution then we should not forget to update regularly to a later version.)

musicEnfanthen: Yes, I totally agree, and asked about it over there. Meanwhile, I change back to `latest` here

PR #1791, opened 2020-10-29
increase thickness of non-solid curves

rettinghaus:
Instead of relying on SVG stroke width for non-solid slurs and ties use arithmetic mean of midpoint and endpoint thicknesses.


lpugin:
Could you add an example? And if it changes the current rendering a before / after of it? Thanks

rettinghaus:
based on `slur-002.mei` from the test suite:

### current
dashed_old

### new
dashed_new

PR #1794, opened 2020-10-30
Import MusicXML dynamics to MEI note@vel and dynam@val

earboxer:
Thanks @rettinghaus for [pointing out](https://github.com/rism-ch/verovio/issues/1793#issuecomment-719716709) that `dynam@val` is what we want to convert `direction/sound@dynamics` into.

The equation `std::min(int((dynamics * 90) / 100), 127)` is based on the MusicXML spec

> Dynamics (or MIDI velocity) are expressed as a percentage of the default forte value (90 for MIDI 1.0).

Note that values between [0, 1.111] will all become 0. Users should take care, as many MIDI programs do not treat velocity 0 the same as velocities 1-127. As 0 is a valid value for [data.MIDIVALUE](https://music-encoding.org/guidelines/v4/data-types/data.midivalue.html), I don't think this should be changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+* Support for `note@vel` and `dynamics@val` in MusicXML import (@earboxer)
 * Support for `multiRest@loc` (@rettinghaus)
 * Support for `mNum@fontsize` (@rettinghaus)
 * Support for `accidental-mark` in MusicXML import (@rettinghaus)
--- a/include/vrv/dynam.h
+++ b/include/vrv/dynam.h
@@ -8,6 +8,7 @@
 #ifndef __VRV_DYNAM_H__
 #define __VRV_DYNAM_H__
 
+#include "atts_midi.h"
 #include "controlelement.h"
 #include "textdirinterface.h"
 #include "timeinterface.h"
@@ -26,6 +27,8 @@ class Dynam : public ControlElement,
               public TimeSpanningInterface,
               public AttExtender,
               public AttLineRendBase,
+              public AttMidiValue,
+              public AttMidiValue2,
               public AttVerticalGroup {
 public:
     /**
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -329,7 +329,7 @@ private:
     std::string GetOrnamentGlyphNumber(int attributes) const;
 
     /*
-     * @name Methods for converting MusicXML string values to MEI attributes.
+     * @name Methods for converting MusicXML values to MEI attributes.
      */
     ///@{
     static data_ACCIDENTAL_WRITTEN ConvertAccidentalToAccid(const std::string &value);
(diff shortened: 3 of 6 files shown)


rettinghaus (approved):

craigsapp:
If you have an output adjusted MIDI velocity between 0.0 (exclusive) and 0.5, then this should be mapped to `1` and not to `0`.

The value 0 is a valid MIDI attack velocity, but it is a common standardized shorthand for a note off, so it is impossible for `0` to be used for a note on.   Instead the `0` means to turn off a previous note at the same pitch, and this meaning cannot be changed.  

So input values in the range `(0, 1.111]` that transform to `(0, 1.0]` should map to `1` and not `0` if you want correct MIDI attack velocities.

 In practical terms this will not be much of a problem since a MIDI velocity of `1` should be inaudible or nearly inaudible anyway.  If a `0` is converted to a non-zero, then that would be catastrophic (i.e., a stuck note on), but this will not be a problem since zero in both systems are the same.

lpugin:
@craigsapp anything to be changed to the PR?

earboxer:
@craigsapp, would you prefer if the code were

```c++
        float dynamics = node.attribute("dynamics").as_float(-1);
        if (dynamics > 0) {
            note->SetVel(std::max(std::min(int((dynamics * 90) / 100), 127)), 1);
        } else if (dynamics == 0) {
            note->SetVel(0);
        }
```

craigsapp:
Yes that looks good.  

It is difficult to read because there is a lot going on in a single line.  How about splitting it into two steps for clarity?:

```cpp
        float dynamics = node.attribute("dynamics").as_float(-1.0);
        if (dynamics > 0.0) {
            int mididynam = int(dynamics * 90.0 / 100.0)
            note->SetVel(std::max(1, std::min(127, mididynam)));
        } else if (dynamics == 0.0) {
            note->SetVel(0);
        }
```

I would also use explicit typecasting of the numbers, adding `.0` to clarify when floating-point calculations are intended.  Also I would use `double` rather than `float` as a data type for floating-point values (and `.as_double()` rather than `.as_float()`), but that is a stylistic thing that makes no real difference in this case.

You are flooring the values, and I would round them (but I don't know if one or the other is more correct without thinking about it more), and in any case the difference will be incredibly small perceptually:

```cpp
        float dynamics = node.attribute("dynamics").as_float(-1.0);
        if (dynamics > 0.0) {
            int mididynam = int(dynamics * 90.0 / 100.0 + 0.5)
            note->SetVel(std::max(1, std::min(127, mididynam)));
        } else if (dynamics == 0.0) {
            note->SetVel(0);
        }
```

This would round the range from 0.5 to 1.0 for `mididynam` to "1" rather than 0, but this is fixed by `max()` (there would still be the range from [0.0 to 0.5] that needs to be fixed by `max()` anyway).  Also you may not be able to hit MIDI dynamic value `127` exactly, but the weird 90% scaling confuses that issue.

earboxer:
Okay, @craigsapp, I've implemented it as suggested (though kept things as floats) in https://github.com/rism-ch/verovio/pull/1794/commits/b88e14b16530169951e7e85fcd4a15ca7eb5b8b9

(and refactored as a function `data_MIDIVALUE MusicXmlInput::ConvertDynamicsToMidiVal(const float dynamics)`)

craigsapp (approved):

lpugin:
Thanks!

mvayngrib:
does this mean midi should now have velocity changes if there are dynamics changes? I still see the same velocities regardless of `<dynamics>` 🤔

rettinghaus:
> does this mean midi should now have velocity changes if there are dynamics changes?

No

mvayngrib:
thanks @rettinghaus! Is that something on the roadmap?

PR #1801, opened 2020-11-03
* MusicXML Parsing: Don't create empty pgHead or pgFoot if there was …

earboxer:
…nothing added there.

When verovio is run with `--header auto` or `--footer auto`, it will automatically create a header or footer *if and only if* there is no existing pgFoot or pgHead.

This PR makes it so that musicXML won't create an empty pgHead if there was only a pgFoot credit (and vice versa).

Thus, you can add explicit footer text (via MusicXML) while keeping the automatically generated header.

Related to #1352

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -582,10 +582,8 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
     // generate page head
     pugi::xpath_node_set credits = root.select_nodes("/score-partwise/credit[@page='1']/credit-words");
     if (!credits.empty()) {
-        bool hasHead = false;
-        bool hasFoot = false;
-        PgHead *head = new PgHead();
-        PgFoot *foot = new PgFoot();
+        PgHead *head = NULL;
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -582,6 +582,8 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
     // generate page head
     pugi::xpath_node_set credits = root.select_nodes("/score-partwise/credit[@page='1']/credit-words");
     if (!credits.empty()) {
+        bool hasHead = false;
+        bool hasFoot = false;
         PgHead *head = new PgHead();
         PgFoot *foot = new PgFoot();
```
lpugin, line 588: This will cause a memory leak if one of them is empty. You don't need to add a `bool` but can simply initialize the pointers with `NULL` and check on that.

earboxer: Fixed in the latest commit. (You may prefer to squash-merge this pull request so the memory leak doesn't make it into the git history).

Issue #98, opened 2015-05-25
across staff stem direction


minamouse:
Another thing that seems to not be supported is the notation of having different stem directions across staves. For example, when a beamed group of notes has a note on another staff, all the stems go in the same direction, which means the beam always ends up above or below the notes. This means that they can't be placed in between the staves like this:
![screen shot 2015-05-25 at 3 02 12 pm](https://cloud.githubusercontent.com/assets/11529868/7801219/55ce5b02-02ef-11e5-8df3-b1001c895773.png)

craigsapp:
Here is another example:

![screen shot 2016-07-18 at 4 06 08 pm](https://cloud.githubusercontent.com/assets/3487289/16933570/64b5bc78-4d03-11e6-93d6-d6480ec8117d.png)

Here is the current verovio rendering:

![screen shot 2016-07-18 at 4 17 30 pm](https://cloud.githubusercontent.com/assets/3487289/16933598/c60ff3f8-4d03-11e6-9e03-2e45a3f90f08.png)

Here is the current encoding in MEI  (with a layer hack to get the clef change positioned properly):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/3.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="3.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt>
                <date>2016-07-18 16:16:47</date>
            </pubStmt>
        </fileDesc>
        <encodingDesc>
            <projectDesc>
                <p>Transcoded from Humdrum with Verovio version 0.9.12-dev-5395730</p>
            </projectDesc>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef xml:id="scoredef-000000200759646">
                        <staffGrp xml:id="m-000000004575334" symbol="brace">
                            <staffDef xml:id="staffdef-000000017739101" clef.shape="F" clef.line="4" key.sig="4f" meter.count="12" meter.unit="8" n="1" lines="5" />
[...]

PR #1831, opened 2020-11-12
Mixed direction beams

Monceber:
Resolves the problem with #98 

Implementation contains fix for both cross-staff beams with mixed stem direction as well as single staff beams.
For single staff mixed beam placement is considered only when it falls within staff boundaries (i.e. it doesn't overlap with ledger lines). In case where such place cannot be found beam is recalculated as ordinary one with ABOVE/BELOW placement, based on the predominant stem direction (with above taking priority).
image

There's no such limitations for the cross-staff beams, so generally it behaves the same as any other beam would, just with different placement.
image

<details><summary>Example:</summary>

```
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>Test for beams with mixed stem directions</title>
      </titleStmt>
      <pubStmt>
        <respStmt>
        </respStmt>
      </pubStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef meter.count="4" meter.unit="4" meter.sym="common">
            <staffGrp>
              <staffDef n="1" clef.line="2" clef.shape="G" lines="5"/>
            </staffGrp>
          </scoreDef>
          <section>
            <measure>
              <staff n="1">
                <layer n="1">
                  <beam>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down" breaksec="1"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up" breaksec="2"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up" breaksec="1"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down" breaksec="2"/>
                    <note pname="d" oct="6" dur="32" stem.dir="down"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                  </beam>
				  <beam>
                    <note pname="d" oct="7" dur="32" stem.dir="down"/>
                    <note pname="d" oct="4" dur="32" stem.dir="up" breaksec="1"/>
                    <note pname="d" oct="4" dur="32" stem.dir="up"/>
                    <note pname="d" oct="7" dur="32" stem.dir="down"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                    <note pname="d" oct="5" dur="32" stem.dir="down" breaksec="2"/>
                    <note pname="d" oct="5" dur="32" stem.dir="down"/>
                    <note pname="d" oct="3" dur="32" stem.dir="up"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="4" dur="32" stem.dir="up"/>
                    <note pname="d" oct="5" dur="32" stem.dir="down" breaksec="1"/>
                    <note pname="d" oct="4" dur="32" stem.dir="up"/>
                    <note pname="d" oct="5" dur="32" stem.dir="down"/>
                  </beam>
                  <beam>
                    <note pname="d" oct="5" dur="32" stem.dir="down"/>
                    <note pname="d" oct="4" dur="32" stem.dir="up" breaksec="2"/>
                    <note pname="d" oct="5" dur="32" stem.dir="down"/>
                    <note pname="d" oct="4" dur="32" stem.dir="up"/>
                  </beam>
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>

```

</details>


lpugin:
Looking at the test suite it seems that most of the changes are due to a shortening of stem length. This looks fine to me. Would you mind looking at test cross-staff-004 though? It seems that the beam should still be above for the last one of the first measure and following.

Monceber:
Yeah, I'll take a look. There also seems to be a problem with chords in mixed beam in test cross-staff-001, so I'll need to address that as well.

lpugin:
> There also seems to be a problem with chords in mixed beam in test cross-staff-001

It is not impossible that the encoding is non-sense.

lpugin:
Forget about cross-staff-001. I will fix the example

lpugin:
Regarding example cross-staff-004: I am currently looking at placing the beam on the staff below when all the notes are there. You only need to make sure that are stem up.

Monceber:
> Forget about cross-staff-001.

No, there might be a problem with example, but there's definitely problem with how new code handles chords in cross-staff beams - I've double checked it on my side and that seems to be the case. I'll need to fix that regardless.

Monceber:
@lpugin I have a fix for the beams, and they are even drawn properly without additional changes:
image

However, it works for beams only. In cross-staff-004 those are tuplets within beams, and their behavior is strange - stem direction is correct (based on horizontal positioning), but stem is drawn downwards:
image

lpugin:
Great! I think what we need to do is to activate the LayerElement::m_crossStaff pointer not only on DuratoinInterface element but on all LayerElement. At least we need it on Beam and Tuplet (Artic might be a different story).
Beam currently (kind-off) work because of the `@beam.with` attribute. However, there is a consensus that beams should shifted to the cross-staff without it when all the notes are cross-staff (see https://github.com/music-encoding/music-encoding/issues/714). I think the same applies to Tuplet.

A starting point would be to modify LayerElement::PrepareCrossStaffEnd and for Beam and Tuplet to activate the m_crossStaff when all the DurationInterface have a Object::m_crossStaff. (The should be the same Staff in most cases but we might want to check this just in case).

The we do not need to rely on Beam::HasBeamWith value. Basically we will have m_isCrossStaff indicating that we have a cross-staff situation for the Beam, and m_crossStaff indicating that actually the beam itself is moved.

Does it make sense? Can you give a try?

Monceber:
Yes, I think I get what you're talking about. Thank you for those pointers, I'll see what I can do about this,

Monceber:
Ok, or the time being I've pushed the changes that fix chord behavior and most of the beam behaviors.
As for tuplet problem - it turned to be just a regression introduced in #1722 that was interfering with proper placement. Actually this can be seen in the test run as well:
image

This also works even without @beam.with="below". I'm going to look into what you've suggested with m_crossStaff next - it might be a better approach compared to what I have atm, so I'll need to investigate it a bit. Additionally, I've yet to look at the experimental branch that you've mentioned before - there might be some improvements to be found there as well.

lpugin:
Great! Do you want me to merge this already?

Monceber:
I say, let it wait till Monday - I'll go through experimental branch that you've mentioned and your suggestions and see if anything could be changed/improved.

Monceber:
@lpugin I think for the time being this implementation is sufficient enough, but it can definitely be improved, especially positioning in some specific cases. For instance, we'd probably need something akin to suggested m_crossStaff for positioning when cross-staff elements span over several staves as opposed just two in the tested scenarios. There probably are some other edge cases that could benefit from it.

lpugin:
Great, thanks!

PR #1861, opened 2020-12-03
Flag position improvement

Monceber:
Slight improvement for the flag positioning. In some cases (e.g. with different fonts) calculated offset for the flags led to higher notes having flags positioned lower than lower notes.

<details><summary>Here some examples:</summary>

image
image
</details>
<details><summary>This fix makes those offsets go away:</summary>

image
image
</details>


lpugin:
The last commit fixed the bug. Thanks!

One question: don't we need the same adjustments for chords? (For chords, the stem is a child of the chord, not of the notes)

Monceber:
Yes, we actually do. I've added that last commit to prevent the problem, but we do need to extend this adjustments for chords.
I do have a change that should make it work for chords as well, I'll create PR for it in a couple of minutes.

PR #1862, opened 2020-12-03
Fix flag adjustment for chords

Monceber:
- fixed logic of flag adjustment to take into account other possible Parents (i.e. Chord), not just Notes


rettinghaus:
Looks good!

Before:
flaggedChords
After:
flaggedChordsFixed

PR #1864, opened 2020-12-04
Improve hairpin barline overlap

Monceber:
Fix for the issue https://github.com/rism-ch/verovio/issues/959

Example of rendering:
image


lpugin:
This PR craches on accid-006.mei. I suspect this is because of the bug introduced in https://github.com/rism-ch/verovio/pull/1861 but fixed since then. Please redo the PR on the latest develop. Thanks!

PR #1869, opened 2020-12-07
Crash fixes

Monceber:
A couple of fixes for crashes:
- fix MusicXML parsing to avoid null references for hairpin, harmonics bass-alter and measures with mNum
- adjusted comparator for accid to avoid crashing
- changed beam to log error instead of crashing when beam with `end` value is not found

### DIFF
--- a/include/vrv/accid.h
+++ b/include/vrv/accid.h
@@ -115,9 +115,12 @@ public:
         else if (first->GetDrawingY() > second->GetDrawingY()) {
             return false;
         }
-        // with unissons, natural should always be the last accidental
         else {
-            return (first->GetAccid() != ACCIDENTAL_WRITTEN_n);
+            // with unissons, natural should always be the last accidental (assuming there is a natural)
+            if ((first->GetAccid() == ACCIDENTAL_WRITTEN_n) || (second->GetAccid() == ACCIDENTAL_WRITTEN_n)) {
+                return (first->GetAccid() != ACCIDENTAL_WRITTEN_n);
+            }
+            return first->GetDrawingY() < second->GetDrawingY();
         }
     }
 };
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3341,6 +3341,11 @@ void MusicXmlInput::ReadMusicXmlBeamsAndTuplets(const pugi::xml_node &node, Laye
         // find start and end of the beam
         const auto beamStartIterator = std::find(currentMeasureNodes.begin(), currentMeasureNodes.end(), node);
         const auto beamEndIterator = std::find(beamStartIterator, currentMeasureNodes.end(), beamEnd);
+
+        if (beamEndIterator == currentMeasureNodes.end()) {
+            LogError("MusicXML import: Beam ending point has not been found for <'%s'>", layer->GetUuid().c_str());
+            return;
+        }
         // form vector of the beam nodes and find whether there are tuplets that start or end within the beam
         std::vector<pugi::xml_node> beamNodes(beamStartIterator, beamEndIterator + 1);
         bool isTupletStartInBeam
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2026,7 +2026,7 @@ void MusicXmlInput::ReadMusicXmlDirection(
             hairpin->SetTstamp(timeStamp);
             if (wedge->node().attribute("id")) hairpin->SetUuid(wedge->node().attribute("id").as_string());
             int staffNum = node.select_node("staff").node().text().as_int();
-            staffNum = !staffNum ? dynamic_cast<Staff *>(m_prevLayer->GetParent())->GetN() : staffNum;
+            staffNum = (!staffNum && m_prevLayer) ? dynamic_cast<Staff *>(m_prevLayer->GetParent())->GetN() : staffNum;
             if (staffNum != 0) {
                 hairpin->SetStaff(
                     hairpin->AttStaffIdent::StrToXsdPositiveIntegerList(std::to_string(staffNum + staffOffset)));
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2322,7 +2322,7 @@ void MusicXmlInput::ReadMusicXmlHarmony(pugi::xml_node node, Measure *measure, c
     if (bass) {
         harmText += "/" + GetContentOfChild(node, "bass/bass-step");
         pugi::xpath_node alter = node.select_node("bass/bass-alter");
-        harmText += ConvertAlterToSymbol(GetContent(alter.node()));
+        if (alter) harmText += ConvertAlterToSymbol(GetContent(alter.node()));
     }
     Harm *harm = new Harm();
     Text *text = new Text();
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1493,7 +1493,7 @@ void MusicXmlInput::ReadMusicXmlAttributes(
         // check if we have a staff number
         int staffNum = clef.node().attribute("number").as_int();
         staffNum = (staffNum < 1) ? 1 : staffNum;
-        Staff *staff = dynamic_cast<Staff *>(measure->GetChild(staffNum - 1));
+        Staff *staff = dynamic_cast<Staff *>(measure->GetChild(staffNum - 1, STAFF));
         assert(staff);
         pugi::xpath_node clefSign = clef.node().select_node("sign");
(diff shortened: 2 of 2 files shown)


rettinghaus (approved):

lpugin:
I think this PR yielded this change. 

image

Please follow-up if this is not looking as expected.

craigsapp:
I like the flat before the sharp (the after view at the bottom):

Screen Shot 2020-12-09 at 1 18 29 PM

But I do not know if there is a standardized position in such cases.  It seems that the order of the accidentals are reversed from their previous position based on the layer of the note.  But either way is seems OK to me (provided there is no standardized order that may be preferred).

The cautionary accidental cases has too much space between the notes and the accidental:

Screen Shot 2020-12-09 at 1 19 38 PM

The E-sharp closest to the notehead should be as close as the F-sharp sharp.  Probably the spacing for the accidentals are switched, so that the regular accidental is getting the bracketed accidental position, but there is no bracket to leave extra space for.

Monceber:
Correct me if I'm wrong, but I think that results shown are actually in the order after/before, because at least that's the comparable result I'm getting. With this change this is the rendering that I get:
image

Now, I'm not sure what is correct placement according to standards is - they are placed in the order they are defined in the MEI file (or at least this happens when there are 2 layers). For instance, if I were to move **flat** accidental from second layer to first and **sharp** - to second, this is what I would get:
image

Mind you, this is the same behavior as previously, but with small caveat - before the order of the accidentals would be _reversed_. Previously condition would always sort elements of the unison. It would ignore second accidental and always sort current element, unless it was natural.
```
else {
       return (first->GetAccid() != ACCIDENTAL_WRITTEN_n);
}
```

Regarding cautionary accidental there are two points there:
- it was being switched for the same reason (due to condition above) to the front (since it's encoded on second layer in MEI file)
- there is still the same gap between the accidental and note even if you define it on the first layer, where it implicitly would be placed in the front, so the problem is probably somewhere else

Additionally, I've also tried how it would look with more layers, but I have no idea which one looks better (or which one is correct for that matter):

<details><summary>Example</summary>

Before:
image

After:
image

File:
```
   <meiHead>
      <encodingDesc>
         <appInfo>
            <application version="1.0.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="0">
                     <staff n="1">
                        <layer n="1">
                           <note dur="4" oct="4" pname="c" accid="x" />
                        </layer>
                        <layer n="2">
                           <note dur="4" oct="4" pname="c" accid="s" />
                        </layer>
			<layer n="3">
                           <note dur="4" oct="4" pname="c" accid="f" />
                        </layer>
                        <layer n="4">
			   <note dur="4" oct="4" pname="c">
                              <accid accid="s" enclose="brack" />
                           </note>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>

```

</details>
Now, in the first place I did this change since it was causing a crash on the environment that I'm running (MSVS), however I think that it is applicable only to it and Debug mode only. If you think that previous behavior was better I'll create PR to revert this particular commit (7d646c7).

craigsapp:
> Now, I'm not sure what is correct placement according to standards is - they are placed in the order they are defined in the MEI file (or at least this happens when there are 2 layers). For instance, if I were to move flat accidental from second layer to first and sharp.

That is good.    There will be potential problems with natural signs when they are applied to unison chords.  In 19th century music notation and before, a natural was given to cancel out one sharp of a double-sharp (and similar cases for double flats):

Screen Shot 2020-12-10 at 3 18 51 AM

The second accidental is a `@accid="ns"`, where the natural sign cancels out one of the sharps in the double sharp.

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2020-12-10T03:18:32" version="3.1.0-dev-d63fe6c-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000002074647786">
    <score xml:id="score-0000000324987428">
     <scoreDef xml:id="scoredef-0000001465224137" midi.bpm="400">
      <staffGrp xml:id="staffgrp-0000000383501622">
       <staffDef xml:id="staffdef-0000001584982347" n="1" lines="5">
        <clef xml:id="clef-0000001549999091" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="staff-0000000620930526" n="1">
         <note xml:id="note-L3F1" dur="1" oct="4" pname="g" accid="x" />
         <note xml:id="note-L4F1" dur="1" oct="4" pname="g" accid="ns" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

--------------------

The current `develop` before this PR knows about this and avoids placing the natural before the sharp.  Here is an example where I switch the order of the G-sharp and G-natural in the two layers:

Screen Shot 2020-12-10 at 3 22 04 AM

How does this PR render the following MEI data?  Preferrably the natural should not come before the sharp, since this notation is historically ambiguous: it could be interpreted as a unison on G-sharp in 19th century style, or as two pitches G-sharp and G-natural in a more modern style.   In other words, it is not possible to distinguish between a single accidental `@accid="ns"` and two accidentals `@accid="n"` and `@accid="s"` that are given in the same order.

MEI data:

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2020-12-10T03:22:06" version="3.1.0-dev-d63fe6c-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000000587879174">
    <score xml:id="score-0000000440256168">
     <scoreDef xml:id="scoredef-0000000211333377" midi.bpm="400">
      <staffGrp xml:id="staffgrp-0000000378743852">
       <staffDef xml:id="staffdef-0000001499825985" n="1" lines="5">
        <clef xml:id="clef-0000000023046043" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="staff-0000000978214617" n="1">
         <note xml:id="note-L3F1" dur="1" oct="4" pname="g" accid="x" />
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L4">
       <staff xml:id="staff-L4F1N1" n="1">
        <layer xml:id="layer-L4F1N1" n="1">
         <note xml:id="note-L6F1" dur="1" oct="4" pname="g" accid="s" />
        </layer>
        <layer xml:id="layer-L6F2N2" n="2">
         <note xml:id="note-L6F2" dur="1" oct="4" pname="g" accid="n" />
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L7">
       <staff xml:id="staff-L7F1N1" n="1">
        <layer xml:id="layer-L7F1N1" n="1">
         <note xml:id="note-L8F1" dur="1" oct="4" pname="g">
          <accid xml:id="accid-L8F1" accid="n" func="caution" />
         </note>
        </layer>
        <layer xml:id="layer-L7F2N2" n="2">
         <note xml:id="note-L8F2" dur="1" oct="4" pname="g" accid="s" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

rettinghaus:
@craigsapp That is what wasn't change:

unisons

See also example 6 here: https://www.verovio.org/test-suite.xhtml?cat=accid

craigsapp:
Here is an interesting modern notation case:

Screen Shot 2020-12-10 at 3 51 21 PM

This is from Rain Tree Sketch II by Toru Takemitsu (published by Schott , 1992).  The first chord has a b-natural and b-flat, with the natural placed first (This is the only augmented unison that I see in the piece).   Rain Tree Sketch I has several augmented unisions, but the notation uses split stems in those cases (with the natural note placed first and the flat/sharp note placed second).

PR #1876, opened 2020-12-10
Improve hairpin barline overlap

Monceber:
Fix for the issue https://github.com/rism-ch/verovio/issues/959

Example of rendering:
image


lpugin:
Thanks!

PR #1882, opened 2020-12-14
CMake: run get_git_commit.sh from build directory.

matangover:
This fixes the build when running cmake from another directory, i.e. this now works:
`cd verovio && cmake -Hcmake -Btools`

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -90,6 +90,7 @@ endif()
 
 execute_process(
     COMMAND ../tools/get_git_commit.sh
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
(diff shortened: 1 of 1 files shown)


lpugin:
Thanks. Could you explain a bit more what issue this is actually fixing?

matangover:
Sure :) `get_git_commit.sh` relies on being run with `tools` as cwd. Since CMakeLists.txt used `execute_process` without specifying `WORKING_DIRECTORY`, it also relied on cmake itself being run with `tools` as cwd. However, there are cases when somebody wants to run CMake from an arbitrary working directory (specifying `-B` to choose the build directory).

In my case, I encountered this while using Verovio with the CMake Tools extension in Visual Studio Code. The extension doesn't run cmake from the build directory as CMakeLists.txt expected.

PR #1889, opened 2020-12-16
Improve score condensing

rettinghaus:
This PR

- adds support for `grpSym` in `scoreDef`
- changes MusicXML import accordingly
- replaces option `--condense-encoded` with new `--condense` option 
- improves reliability of auto condensing by checking count of `grpSym`
- adds value `auto` to `--system-divider` option for full control over rendering of dividers 

Condensing can now be turned off with `--condense none`.
It doesn't change default behaviour.


src/view_page.cpp
```diff
@@ -418,6 +374,65 @@ void View::DrawStaffDefLabels(DeviceContext *dc, Measure *measure, StaffGrp *sta
...
+    Staff *first = dynamic_cast<Staff *>(measure->FindDescendantByComparison(&comparisonFirst, 1));
+    AttNIntegerComparison comparisonLast(STAFF, groupSymbol->GetEndDef()->GetN());
+    Staff *last = dynamic_cast<Staff *>(measure->FindDescendantByComparison(&comparisonLast, 1));
+
+    if (!first || !last) {
+        LogDebug("Could not get staff (%d; %d) while drawing staffGrp - DrawStaffGrp",
+            groupSymbol->GetStartDef()->GetN(), groupSymbol->GetEndDef()->GetN());
+        return;
+    }
+
```
palgranum, line 396: You should move dc->StartGraphic after the possible return statement otherwise GrpSym node might never close.

PR #1891, opened 2020-12-18
SMuFL glyphs for all CMN rests

### LINKED ISSUE #1871: Use SMuFL glyphs for all CMN rests
At the moment alls rests longer than a quarter rest are drawn as simple rectangles. Even though the differences between fonts are small here, it could lead to inconsistent notational styles. 

Relates to #560

rettinghaus:
This PR brings a revision of drawing methods for `rest`, `mRest`, and `multiRest` to fully support SMuFL glyphs. 
Also, this adds support for rendering "old style" multiple measure rests: 
Kirchenpausen
(Rendered with Bravura font. Source file will be added to test suite.)

NB: #1890 has to be merged first.

fixes #1871


craigsapp:
How do you select old style vs. new style multibar rests?

rettinghaus:
With the `@block` attribute.

rettinghaus:
@craigsapp speaking of, we could add an additional option `--multi-rest-style` with the values `auto`, `block`, `encoded` and `glyphs`. What do you think @lpugin?

PR #1911, opened 2020-12-28
Minor improvements to test suite 

rettinghaus:
* Changes deprecated `swig_add_module` to `swig_add_library`
* allows path names with spaces for `testdir`


doc/diffTests.sh
```diff
@@ -44,7 +44,7 @@ if [ ! -z $build_dev ]; then
     cd python
     make -j8
 
-    $PYTHON ../../doc/test-suite.py $testdir $indir1
+    $PYTHON ../../doc/test-suite.py "$testdir" $indir1
```
lpugin, line 47: Put quotes for all paths

rettinghaus: agreed

PR #1912, opened 2020-12-28
keyAccid not available when keySig is attribute

rettinghaus:
This PR prevents `keySig` to be an attribute in MusicXML import, because when adding children to it MEI output will be invalid. 
@lpugin perhaps we should always check `keySig` for children when trying to make it an attribute?


rettinghaus:
Second commit cleans up handling of labels to assure valid MEI output.

PR #1923, opened 2021-01-04
Revert "CMake: run get_git_commit.sh from build directory."

lpugin:
Reverts rism-ch/verovio#1882 because this does not work anymore
```
laurent@laika verovio % cd bindings 
laurent@laika bindings % cmake ../cmake -DBUILD_AS_PYTHON=ON -B python
laurent@laika bindings % cd python 
laurent@laika python % make -j 8
```
Since `./include/vrv/git_commit.h` is not generated. @matangover could you fix it and redo a PR? Thanks

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -90,7 +90,6 @@ endif()
 
 execute_process(
     COMMAND ../tools/get_git_commit.sh
-    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
     OUTPUT_VARIABLE GIT_COMMIT
 )
 


rettinghaus:
Even without this line `get_git_commit.sh` doesn't work for me.

lpugin:
It works for me on macos 11 with python 3.9 and on ubuntu 20.04. On which setup are you building?

lpugin:
See before reverting https://github.com/lpugin/verovio/runs/1644534096?check_suite_focus=true and after https://github.com/lpugin/verovio/runs/1644681390?check_suite_focus=true

matangover:
OK I'll fix it

matangover:
I opened https://github.com/rism-ch/verovio/pull/1926

PR #1926, opened 2021-01-05
CMake: run get_git_commit.sh from CMake directory.

matangover:
This fixes the build when running cmake from another directory, i.e. this now works:
`cd verovio && cmake -Hcmake -Btools`

This supersedes https://github.com/rism-ch/verovio/pull/1882, which was reverted in https://github.com/rism-ch/verovio/pull/1923. To fix the problem, I changed it to now use the CMake _source_ directory instead of the _build_ directory, because the source directory is guaranteed to stay constant.


lpugin:
Perfect, thanks

PR #1940, opened 2021-01-11
MusicXML improvements

rettinghaus:
Fixes an issue with doubled IDs and doubled page breaks when importing MusicXML.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -741,6 +741,7 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
                         text->SetText(UTF8to16(groupName));
                         m_label->AddChild(text);
                         staffGrp->AddChild(m_label);
+                        m_label = NULL;
                     }
                 }
                 if (!groupAbbr.empty() && !m_labelAbbr) {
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -729,34 +729,34 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
                 const std::string groupAbbr
                     = GetContentOfChild(xpathNode.node(), "group-abbreviation[not(@print-object='no')]");
                 if (!groupName.empty()) {
-                    Label *label = new Label();
+                    m_label = new Label();
```
lpugin, line 732: Are you sure m_label is always NULL? You should assert it is. I don't know the details about what is being achieved here but this look quite like the door to a memory leak.

rettinghaus: You are right. Basically this should never happen, but better to be sure anyway.

lpugin: I think member pointers should be treated more systematically. That is:
* Set them to NULL in the constructor
* Assert that there are not NULL before assigning them to a new object
* Set them to NULL once added to the tree

rettinghaus: done.

Issue #1931, opened 2021-01-06
Encoded line breaks as suggested breakpoints


earboxer:
https://github.com/rism-ch/verovio/pull/1069 added the ability to use encoded `<sb>`s, but automatically determining the `<pb>`s

This works great, but when the score is resized (such as by changing `--unit` or `--page-width`), then it can look a little weird to have one system stretched out super long, or have a system compressed so much.

<img width="721" alt="Screen Shot 2021-01-06 at 15 48 31" src="https://user-images.githubusercontent.com/12452738/103818852-466e7700-5037-11eb-970d-0a8d42870a82.png">
<img width="703" alt="Screen Shot 2021-01-06 at 15 47 57" src="https://user-images.githubusercontent.com/12452738/103818861-4b332b00-5037-11eb-9279-e013b11e2a93.png">
<img width="510" alt="Screen Shot 2021-01-06 at 15 56 30" src="https://user-images.githubusercontent.com/12452738/103819157-dd3b3380-5037-11eb-8b82-0e8396517b68.png">

The stretching problem could be remedied by modifying the input file, and removing half of the `<sb>`s when `--unit` is set to a small enough value. That's not preferred, as it would be shifting more work onto the user, as they have to reload another file, instead of just redoing the layout.
The squeezing problem could be remedied by automatically switching to `breaks auto` mode when `--unit` is set to a big enough value. This also has the problem that somebody needs to hard-code a specific unit value when that option would be switched.

I suggest a "smart" line breaking mode, according to the following rules:
* Systems are not split unless they would be compressed too much by being on their own line.
* if two systems can fit together on one line, they will be.
* page breaks behave the same as `--breaks line` mode

craigsapp:
Can you post the data for the example score?

earboxer:
Examples were made with
* `verovio --breaks line --unit 8 1946trumpet.mei` -- too stretched out
* `verovio --breaks auto --unit 8 1946trumpet.mei` -- line break not used
* `verovio --breaks line --unit 20 --page-width 1500 1946trumpet.mei` -- too compressed

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
[...]

PR #1944, opened 2021-01-12
"smart" breaks mode: Use `<sb>` as suggested system breaks, occasionally ignore.

earboxer:
This is a proposed solution to https://github.com/rism-ch/verovio/issues/1931

The implementation is as following:

* Behaves like `--breaks=auto`, except for
* when the current system is over n% full, it will allow an encoded system break to be effective. (n is set to 66%)

(It's not exactly what I had in mind with #1931, but it might be better).

From the example file, it seems to behave as expected.

`tools/verovio --breaks smart --unit 8 ~/Downloads/1946trumpet.mei`
Screen Shot 2021-01-12 at 16 43 54

`tools/verovio --breaks smart --unit 20 --page-width 1500 ~/Downloads/1946trumpet.mei`
Screen Shot 2021-01-12 at 16 45 01

`tools/verovio --breaks smart --unit 8 --page-width 3500 ~/Downloads/1946trumpet.mei`
Screen Shot 2021-01-12 at 16 46 47

WIP Things to check
- `[ ]` ~~Make `<pb>` behave as expected.~~ (We can add "smart" behavior for this in the future).

### DIFF
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -260,10 +260,11 @@ public:
 
     /**
      * Casts off the entire document, with options for obeying breaks.
-     * @param useSb - 1 to use the sb from the document, 2 to sometimes us the sb from the document.
-     * @param usePg - true to use the pb from the document.
+     * @param useSb - true to use the sb from the document.
+     * @param usePb - true to use the pb from the document.
+     * @param smart - true to sometimes use encoded sb and pb.
      */
-    void CastOffDocBase(int useSb, bool usePb);
+    void CastOffDocBase(bool useSb, bool usePb, bool smart = false);
 
     /**
      * Casts off the running elements (headers and footer)
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -247,6 +247,12 @@ public:
     void CastOffDoc();
 
     /**
+     * Casts off the entire document, only using the document's system breaks
+     * if they would be close to the end in the normal document.
+     */
+    void CastOffSmartDoc();
+
+    /**
      * Casts off the entire document, using the document's line breaks,
      * but adding its own page breaks.
      */
@@ -254,10 +260,10 @@ public:
 
     /**
      * Casts off the entire document, with options for obeying breaks.
-     * @param useSb - true to use the sb from the document.
+     * @param useSb - 1 to use the sb from the document, 2 to sometimes us the sb from the document.
      * @param usePg - true to use the pb from the document.
      */
-    void CastOffDocBase(bool useSb, bool usePb);
+    void CastOffDocBase(int useSb, bool usePb);
 
     /**
      * Casts off the running elements (headers and footer)
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -827,15 +827,18 @@ void Doc::CastOffDoc()
 {
     Doc::CastOffDocBase(false, false);
 }
+
 void Doc::CastOffLineDoc()
 {
     Doc::CastOffDocBase(true, false);
 }
+
 void Doc::CastOffSmartDoc()
 {
-    Doc::CastOffDocBase(2, false);
+    Doc::CastOffDocBase(false, false, true);
 }
-void Doc::CastOffDocBase(int useSb, bool usePb)
+
+void Doc::CastOffDocBase(bool useSb, bool usePb, bool smart)
 {
     Pages *pages = this->GetPages();
     assert(pages);
@@ -857,14 +860,13 @@ void Doc::CastOffDocBase(int useSb, bool usePb)
     System *currentSystem = new System();
     contentPage->AddChild(currentSystem);
 
-    if (useSb == 1 && !usePb) {
+    if (useSb && !usePb && !smart) {
         CastOffEncodingParams castOffEncodingParams(this, contentPage, currentSystem, contentSystem, false);
 
         Functor castOffEncoding(&Object::CastOffEncoding);
         contentSystem->Process(&castOffEncoding, &castOffEncodingParams);
     }
     else {
-        bool smart = (useSb == 2);
         CastOffSystemsParams castOffSystemsParams(contentSystem, contentPage, currentSystem, this, smart);
         castOffSystemsParams.m_systemWidth
             = this->m_drawingPageContentWidth - currentSystem->m_systemLeftMar - currentSystem->m_systemRightMar;
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -831,7 +831,11 @@ void Doc::CastOffLineDoc()
 {
     Doc::CastOffDocBase(true, false);
 }
-void Doc::CastOffDocBase(bool useSb, bool usePb)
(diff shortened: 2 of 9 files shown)


include/vrv/doc.h
```diff
@@ -254,10 +260,10 @@ class Doc : public Object {
 
     /**
      * Casts off the entire document, with options for obeying breaks.
-     * @param useSb - true to use the sb from the document.
+     * @param useSb - 1 to use the sb from the document, 2 to sometimes us the sb from the document.
      * @param usePg - true to use the pb from the document.
      */
-    void CastOffDocBase(bool useSb, bool usePb);
+    void CastOffDocBase(int useSb, bool usePb);
```
lpugin, line 266: I would prefer to have an extra parameter `bool smart = false` rather than changing useSb to int because this is a bit cryptic.

src/doc.cpp
```diff
@@ -831,7 +831,11 @@ void Doc::CastOffLineDoc()
 {
     Doc::CastOffDocBase(true, false);
 }
-void Doc::CastOffDocBase(bool useSb, bool usePb)
+void Doc::CastOffSmartDoc()
```
lpugin, line 834: Add an empty line between methods.

src/sb.cpp
```diff
@@ -57,4 +58,26 @@ int Sb::CastOffEncoding(FunctorParams *functorParams)
     return FUNCTOR_SIBLINGS;
 }
 
+int Sb::CastOffSystems(FunctorParams *functorParams)
```
lpugin, line 61: Functor methods should always return a value (the PR currently fails because of this missing)

lpugin (changes requested):
Do not forget to run clang-format. Thanks!

include/vrv/doc.h
earboxer: Agreed (and changed in b7cf23747f8189402666411faf9b2e8f40b88609).

Though it seems a little strange to have three boolean variables for a function which (as implemented) only has three different execution paths (rather than `2^3`, that is, 8). I guess this is just one of the development consequences of modifying the function prototype before modifying any of the code.

lpugin:
> WIP Things to check
> * [ ]  Make <pb> behave as expected.

Thanks for the revisions. I'll merge once this is checked.

lpugin (approved):

src/sb.cpp
```diff
@@ -57,4 +58,27 @@ int Sb::CastOffEncoding(FunctorParams *functorParams)
...
+    CastOffSystemsParams *params = vrv_params_cast<CastOffSystemsParams *>(functorParams);
+    assert(params);
+    if (params->m_smart) {
+        // Get the last measure of the currentSystem
+        Measure *measure
+            = dynamic_cast<Measure *>(params->m_currentSystem->GetChild(params->m_currentSystem->GetChildCount() - 1));
+        if (measure != NULL) {
+            int measureRightX = measure->GetDrawingX() + measure->GetWidth() - params->m_shift;
+            // LogDebug("ratio: %f\n", (float)measureRightX / (float)params->m_systemWidth);
+            if (measureRightX > params->m_systemWidth * .66) {
```
rettinghaus, line 72: Instead of hard coding this we could use the value from  `--min-last-justification`, as it basically does the same for the last system as if you'd put another `sb` end the end of the MEI file.

earboxer: Thanks for bringing this up! I'm not sure `--min-last-justification` would be the right thing to change. In my application, I'm always setting `--min-last-justification` to 0, so that the last system is always justified to fill the full width. But I do see the logic in the argument.

But thinking about that value (`--encoded-sb-thresh`? and the [corollary](https://github.com/rism-ch/verovio/issues/1931#issuecomment-759147574) `--encoded-pb-thresh`) and whether it should be configurable, I think it starts to replace the following breaks modes:

* `--breaks=encoded` almost equivalent to `--breaks=smart --encoded-sb-thresh=0 --encoded-pb-thresh=0` (always use encoded breaks (notably different, as `--breaks=encoded` **will** overcompress systems like https://user-images.githubusercontent.com/12452738/103819157-dd3b3380-5037-11eb-8b82-0e8396517b68.png).
* `--breaks=line` almost equivalent to `--breaks=smart --encoded-sb-thresh=1 --encoded-pb-thresh=0` (same difference as above, with the overcompressing)
* `--breaks=smart`: `--breaks=smart --encoded-sb-thresh=0.66 --encoded-pb-thresh=0.66` (considering changing the 'smart' hard-coded value to 0.51; which is why I'd prefer this to be a variable)
* `--breaks=auto`: `--breaks=smart --encoded-sb-thresh=1 --encoded-pb-thresh=1` (only use encoded breaks when the page is filled up, i.e. never use encoded breaks).

So, then I think there are a couple of options (I guess I'll list from least desirable to most desirable)
1. Remove ability to overcompress music with encoded line breaks, remove/deprecate `--breaks` flag.
2. Use `--min-last-justification` for this value (possibly rename? to `--min-width-justification`?). Add `--min-height-justification` for the "smart" `<pb>` case. (maybe replace `--justify-vertically` flag? fix vertical justification issues?)
3. remove `--breaks=smart` mode, have `--encoded-sb-thresh` and `--encoded-pb-thresh` default to 1 for auto mode
4. Add `--smart-encoded-sb-thresh` (and `--smart-encoded-pb-thresh`) (defaulting to .6 or .5-ish): Flags which are only effective when you have `--breaks=smart`

craigsapp:
> --breaks=encoded almost equivalent to --breaks=smart --encoded-sb-thresh=0 --encoded-pb-thresh=0 (always use encoded breaks (notably different, as --breaks=encoded will overcompress systems like https://user-images.githubusercontent.com/12452738/103819157-dd3b3380-5037-11eb-8b82-0e8396517b68.png).

While not musically useful, I use this feature a lot for debugging music encodings, so make sure it does not get fixed :-).  Finale/Sibelius/MuseScore have problems with this and force soft system breaks when the hard returns would overly squash the music on a system, so then I have to make the music too small to see to keep all of the music for an encoded system on the same line.

src/sb.cpp
lpugin: I would prefer to go for option 4. I also find it useful to have options grouped alphabetically as much as possible, so I would suggest `--breaks-smart-sb` for the threshold value.

At this stage I am not absolutely clear why you would need a pb threshold since verovio does not compress pages vertically when there is not enough space in vertical justification. It does only expand. It means that anything below 1.0 would have no effect as it stands. So I would suggest to finalize the implementation with sb first.

earboxer:
> I would suggest --breaks-smart-sb for the threshold value.
Added!

> So I would suggest to finalize the implementation with sb first.
Alright, that's done. So we can leave the implementation with `<pb>` for some other time (probably for someone who has more of a use for that feature than I).

lpugin:
Thanks!

rettinghaus:
@lpugin @earboxer This PR actually broke `sb` import from ABC and MusicXML.

lpugin:
OK. Could you open and issue with more details about what is broken?

PR #1947, opened 2021-01-13
Improve automatic positioning for cross staff rests

Monceber:
Improvements for cross-staff rests, so that they follow defined rules for automatic placement.
image

<details><summary>Example</summary>

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Cross-staff rests</title>
            <respStmt />
         </titleStmt>
         <pubStmt><availability />
         </pubStmt>
      </fileDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffGrp symbol="brace" bar.thru="true">
                        <label>Piano</label>
                        <labelAbbr>Pno.</labelAbbr>
                        <instrDef midi.channel="0" midi.instrnum="0" midi.volume="78.00%" />
                        <staffDef n="1" lines="5" ppq="1" clef.shape="G" clef.line="2" key.sig="0" meter.count="3" meter.unit="4" />
                        <staffDef n="2" lines="5" ppq="1" clef.shape="F" clef.line="4" key.sig="0" meter.count="3" meter.unit="4" />
                     </staffGrp>
                  </staffGrp>
               </scoreDef>
               <section>
                  <pb />
		  <measure n="5">
                     <staff n="1">
                        <layer n="5">
                           <note dur="4" oct="4" pname="c" stem.dir="down" />
                           <note dur="4" oct="4" pname="d" stem.dir="down" />
                           <note dur="4" oct="4" pname="e" stem.dir="down" />
                        </layer>
                     </staff>
                     <staff n="2">
			<layer n="1">
                           <rest dur="4" staff="1" />
                           <rest dur="4" staff="1" />
                           <rest dur="4" staff="1" />
                        </layer>
                        <layer n="2">
                           <note dur="4" oct="2" pname="a" stem.dir="up" />
                           <note dur="4" oct="3" pname="c" stem.dir="up" />
                           <note dur="4" oct="3" pname="e" stem.dir="up" />
[...]

PR #1948, opened 2021-01-13
Automatic cross staff rest positining fix

Monceber:
As mentioned by @craigsapp in #1947, having `@layer="5"` on first layer is not exactly normal behavior. Not only that, previous code actually depended on having different layer numbers for cross-staff elements (which is clearly wrong). 

Commit ac4299a fixes this and works properly with correct numbering of the layers. Same example as in #1947, but with proper layer numbering renders properly.
image


Monceber:
I'll resubmit PR later to avoid conflicts with develop.

PR #1950, opened 2021-01-14
Refine automatic rest positioning

Monceber:
This change will limit extreme rest positioning in some cases and avoid large gaps in the rendering, where rest had to be placed far away.
Rendering before:
image

Rendering after:
image

This is not going to impact positioning of other rests, just the ones where offset was getting far too large.


lpugin:
It does affect this example
image
Could you double check?

rettinghaus:
I think the eighth rest should be positioned as before. 
layer-010
Otherwise it could be misread.

PR #1965, opened 2021-01-18
Develop staff spacing

lpugin:


lpugin:
The changes in the test-suite were expected. Double-stemmed (mixed) beam spacing need to be improved as already listed in some issues.

craigsapp:
Screen Shot 2021-01-18 at 12 00 33 PM

What is an efficient way to see the changes?  And is the overlay before/after view available?

The automatic github message says there are 5 changes in the Cross-staff category.  It would be useful to list which examples in that category have changed (i.e., list the names of the files that have the changes).   Also a page that lists only the examples that have changed would be interesting...

From here:

https://www.verovio.org/comparison.xhtml

I am only seeing non-overlaid examples:

Screen Shot 2021-01-18 at 12 00 20 PM

When I select from the dropdown menu, a selection is being done, but the menu displays a blank (in case you do not know about it):

Screen Shot 2021-01-18 at 12 04 39 PM

Also the lists of examples are seemingly in a random order?  Perhaps newest to oldest sorting would be good? (highest example number to lowest) on pages for each category:
https://www.verovio.org/test-suite.xhtml
and/or sorted by the verovio version numbers (most recent first, unknown last).

lpugin:
The best way is to download the `test-suite-diff` artifact and to open it locally. It contains the web-page with all the images.
This is the best we could do because of user access restrictions when running GH actions. When an external contributors makes a PR, the it is even not possible to add the link as a comment. But the artifact is available.

lpugin:
image

PR #1967, opened 2021-01-18
Allow redoLayout to work for --breaks smart mode

earboxer:
Analogous to https://github.com/rism-ch/verovio/pull/1416, but for https://github.com/rism-ch/verovio/pull/1944

In order for redoLayout to work in smart breaks mode, we need to store the `<sb>` internally, so it shouldn't get thrown out by `CastOffSystems`.

This also has the side effect of causing `verovio --breaks auto`, `verovio --breaks encoded`, and `verovio --breaks smart` to all have the same MEI output as each other (though they're different from `verovio --breaks line` :/ ).

### DIFF
--- a/src/sb.cpp
+++ b/src/sb.cpp
@@ -79,6 +79,8 @@ int Sb::CastOffSystems(FunctorParams *functorParams)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


rettinghaus (approved):
Looks good to me.

PR #1968, opened 2021-01-19
fix vertical aligner for grpSym

rettinghaus:


lpugin:
This is changing spacing for 76 test files. It looks OK but is this expected?

rettinghaus:
As far as I can see, it affects only grouped staves, so this is expected. Before the vertical aligner always returned 0 as default spacing.

Issue #1300, opened 2020-02-11
Expansion enhancements


craigsapp:
@wergo wrote:
> This is the intended behavior. Good idea to have a default behavior that should be requested as feature.

I am good at that 😛 

There are two useful enhancements for expansions:

(1) If `--expand` is given an empty string, then the first expansion list found in the file should be used:

```bash
verovio --expand "" file.mei -atmidi
```

(2) Using IDs to reference expansions is good in the case where you are aware of the contents, but related to discussions last year or so, there are two primary expansion categories that would be good to make easier to access: (1) Full-repeat expansions, and (2) No-repeat expansions.  

Perhaps giving formalized `@label` tags to these two cases would be good.  I use such labels to find the duration of a repertory in Humdrum when playing with full repeats or no repeats.  In this case I am not aware of the contents of the files, I just request the version of the file that should occur in both of those cases.  I currently use `@type="norep"` to indicate that the expansion is for the no-repeat version of the score (see example below).

Here are two example use cases of such a feature:

(1) A frontend interface to verovio could have a button that indicates if full repeats or no repeats should be taken for playback.

(2) For score analysis purposes it is often useful to not analyze sections of music that should be repeated in performance, so the no-repeat version is particularly useful for analysis.  

In addition, an expansion list for such cases is often necessary because the sequence of the music is important.  In other words, the juxtaposition of the first and second endings is synthetic and a score would not be performed literally in that sequence (but this can mostly be avoided through proper handling of `<ending>` in MEI data).

A simple analysis case would be to measure the amount of repetition in music caused by repeats in the score. 

For [Mozart piano sonatas](https://github.com/craigsapp/mozart-piano-sonatas), the printed duration of all sonatas is 2038.8 quarter notes, the duration with full repeats is 31664.3 quarter notes, and the duration without repeats is 20599.8 quarter notes.  The ratio of full-repeats to no-repeats is 1.537, so there is a 54% duplication caused by repeats in the music.

For [Beethoven piano sonatas](https://github.com/craigsapp/beethoven-piano-sonatas), the printed duration of all sonatas is 55863.4 quarter notes, the duration with full repeats is 71460.6, and the duration without repeats is 57297.9 quarter notes.  The ratio of full-repeats to no-repeats is 1.247, so there is a 25% duplication caused by repeats in the music.

And comparing the two repertories, there is more repeating of written material in the Mozart piano sonatas compared to the Beethoven piano sonatas.

[...]

PR #1981, opened 2021-01-20
Expose scoreTime in vrvToolkit

wergo:
This PR exposes `scoreTime` of a `note` with a given `xml:id` through the toolkit. This would be a useful method for using Verovio for an editor package.

(In order to make this PR work in javascript, the emscripten scripts would have to be updated, that I do not dare to touch.)

### DIFF
--- a/emscripten/buildToolkit
+++ b/emscripten/buildToolkit
@@ -175,6 +175,7 @@ $exports .= "'_vrvToolkit_getOptions',";
 $exports .= "'_vrvToolkit_getPageCount',";
 $exports .= "'_vrvToolkit_getPageWithElement',";
 $exports .= "'_vrvToolkit_getTimeForElement',";
+$exports .= "'_vrvToolkit_getTimesForElement',";
 $exports .= "'_vrvToolkit_getVersion',";
 $exports .= "'_vrvToolkit_loadData',";
 $exports .= "'_vrvToolkit_redoLayout',";
--- a/emscripten/verovio-proxy.js
+++ b/emscripten/verovio-proxy.js
@@ -51,6 +51,9 @@ verovio.vrvToolkit.getPageWithElement = Module.cwrap( 'vrvToolkit_getPageWithEle
 // double getTimeForElement(Toolkit *ic, const char *xmlId)
 verovio.vrvToolkit.getTimeForElement = Module.cwrap( 'vrvToolkit_getTimeForElement', 'number', ['number', 'string'] );
 
+// string getTimesForElement(Toolkit *ic, const char *xmlId)
+verovio.vrvToolkit.getTimesForElement = Module.cwrap( 'vrvToolkit_getTimesForElement', 'string', ['number', 'string'] );
+
 // char *getMIDIValuesForElement(Toolkit *ic, const char *xmlId)
 verovio.vrvToolkit.getMIDIValuesForElement = Module.cwrap( 'vrvToolkit_getMIDIValuesForElement', 'string', ['number', 'string'] );
 
@@ -186,6 +189,11 @@ verovio.toolkit.prototype.getPageWithElement = function ( xmlId )
 verovio.toolkit.prototype.getTimeForElement = function ( xmlId )
 {
     return verovio.vrvToolkit.getTimeForElement( this.ptr, xmlId );
+};
+
+verovio.toolkit.prototype.getTimesForElement = function ( xmlId )
+{
+    return JSON.parse( verovio.vrvToolkit.getTimesForElement( this.ptr, xmlId ) );
 };
 
 verovio.toolkit.prototype.getVersion = function ()
--- a/include/vrv/toolkit.h
+++ b/include/vrv/toolkit.h
@@ -243,6 +243,12 @@ public:
     std::string GetMIDIValuesForElement(const std::string &xmlId);
 
     /**
+     * Return the score time at which the element is the ID (xml:id) is notated.
+     * Returns 0 if no element is found.
+     */
+    int GetScoreTimeForElement(const std::string &xmlId);
+
+    /**
      * @name Set and get the scale
      */
     ///@{
--- a/include/vrv/toolkit.h
+++ b/include/vrv/toolkit.h
@@ -243,10 +243,12 @@ public:
     std::string GetMIDIValuesForElement(const std::string &xmlId);
 
     /**
-     * Return the score time at which the element is the ID (xml:id) is notated.
+     * Return a JSON object string with the following key values for a given note:
+     * scoreTimeOnset, scoreTimeOffset, scoreTimeTiedDuration,
+     * realTimeOnsetMilliseconds, realTimeOffsetMilliseconds, realTimeTiedDurationMilliseconds.
      * Returns 0 if no element is found.
      */
-    int GetScoreTimeForElement(const std::string &xmlId);
+    std::string GetTimesForElement(const std::string &xmlId);
 
     /**
      * @name Set and get the scale
--- a/src/toolkit.cpp
+++ b/src/toolkit.cpp
@@ -1538,7 +1538,7 @@ int Toolkit::GetTimeForElement(const std::string &xmlId)
     return timeofElement;
 }
 
-int Toolkit::GetScoreTimeForElement(const std::string &xmlId)
+std::string Toolkit::GetTimesForElement(const std::string &xmlId)
 {
     Object *element = m_doc.FindDescendantByUuid(xmlId);
 
@@ -1547,13 +1547,44 @@ int Toolkit::GetScoreTimeForElement(const std::string &xmlId)
         return 0;
     }
 
-    int scoreTimeofElement = 0;
+    double scoreTimeOnset = 0.0;
+    double scoreTimeOffset = 0.0;
+    double scoreTimeDuration = 0.0;
+    double scoreTimeTiedDuration = 0.0;
+    double realTimeOnsetMilliseconds = 0.0;
+    double realTimeOffsetMilliseconds = 0.0;
+
+    jsonxx::Object o;
     if (element->Is(NOTE)) {
+        if (!m_doc.HasMidiTimemap()) {
+            // generate MIDI timemap before progressing
+            m_doc.CalculateMidiTimemap();
+        }
+        if (!m_doc.HasMidiTimemap()) {
+            LogWarning("Calculation of MIDI timemap failed, time value is invalid.");
+            return 0;
+        }
         Note *note = vrv_cast<Note *>(element);
         assert(note);
-        scoreTimeofElement = note->GetScoreTimeOnset();
+        Measure *measure = vrv_cast<Measure *>(note->GetFirstAncestor(MEASURE));
+        assert(measure);
+        // For now ignore repeats and access always the first
+        realTimeOnsetMilliseconds = measure->GetRealTimeOffsetMilliseconds(1);
+        realTimeOffsetMilliseconds = realTimeOnsetMilliseconds + note->GetRealTimeOffsetMilliseconds();
+        realTimeOnsetMilliseconds += note->GetRealTimeOnsetMilliseconds();
+
+        scoreTimeOnset = note->GetScoreTimeOnset();
+        scoreTimeOffset = note->GetScoreTimeOffset();
+        scoreTimeDuration = note->GetScoreTimeDuration();
+        scoreTimeTiedDuration = note->GetScoreTimeTiedDuration();
     }
-    return scoreTimeofElement;
+    o << "scoreTimeOnset" << scoreTimeOnset;
+    o << "scoreTimeOffset" << scoreTimeOffset;
+    o << "scoreTimeDuration" << scoreTimeDuration;
+    o << "scoreTimeTiedDuration" << scoreTimeTiedDuration;
+    o << "realTimeOnsetMilliseconds" << realTimeOnsetMilliseconds;
+    o << "realTimeOffsetMilliseconds" << realTimeOffsetMilliseconds;
+    return o.json();
 }
 
 std::string Toolkit::GetMIDIValuesForElement(const std::string &xmlId)
--- a/src/toolkit.cpp
+++ b/src/toolkit.cpp
@@ -1541,20 +1541,20 @@ int Toolkit::GetTimeForElement(const std::string &xmlId)
 std::string Toolkit::GetTimesForElement(const std::string &xmlId)
 {
     Object *element = m_doc.FindDescendantByUuid(xmlId);
+    jsonxx::Object o;
 
     if (!element) {
         LogWarning("Element '%s' not found", xmlId.c_str());
-        return 0;
+        return o.json();
     }
 
-    double scoreTimeOnset = 0.0;
-    double scoreTimeOffset = 0.0;
-    double scoreTimeDuration = 0.0;
-    double scoreTimeTiedDuration = 0.0;
-    double realTimeOnsetMilliseconds = 0.0;
-    double realTimeOffsetMilliseconds = 0.0;
+    jsonxx::Array scoreTimeOnset;
+    jsonxx::Array scoreTimeOffset;
+    jsonxx::Array scoreTimeDuration;
+    jsonxx::Array scoreTimeTiedDuration;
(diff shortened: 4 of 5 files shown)


lpugin:
Thanks for the PR. Before you make changes I would be interested to hear from @craigsapp if he thinks there is an easy way to get the same from the output of the `renderToTimemap` function. In that case maybe we don't need this. 

If we do, then I think it would be more flexible to return a JSON object and have other values returned as well. It seems to me that all of these could be useful at some point
https://github.com/rism-ch/verovio/blob/8bbf0b0b7167eb0717d19bd47e8824761fc907b3/include/vrv/note.h#L207-L212

wergo:
Having all this info in a JSON object would be great too. Thanks.

craigsapp:
Here is a sample piece of music:

Screen Shot 2021-01-20 at 11 18 48 PM

And the command-line option to extract the JSON timemap which is used for MIDI file alignment with score highlighting:

```bash
verovio file.mei -t timemap
```

this produces for the sample MEI data at the bottom of this post:

```json
[
	{
		"tstamp":	0.000000,
		"qstamp":	0.000000,
		"tempo":	120,
		"on":	["note-L5F1"]
	},
	{
		"tstamp":	500.000000,
		"qstamp":	1.000000,
		"on":	["note-L6F1"],
		"off":	["note-L5F1"]
	},
	{
		"tstamp":	1000.000000,
		"qstamp":	2.000000,
		"on":	["note-L7F1"],
		"off":	["note-L6F1"]
	},
	{
		"tstamp":	1500.000000,
		"qstamp":	3.000000,
		"on":	["note-L8F1"],
		"off":	["note-L7F1"]
	},
	{
		"tstamp":	2000.000000,
		"qstamp":	4.000000,
		"off":	["note-L8F1"]
	}
]
```

This seems suitable for implementing an interface to an external timemap stored in the MEI file in a similar manner. There would need to be an input parameter to verovio to specify which timemap to build: MIDI timemap from the tempos in the score, or from the ID for an embedded timemap (there could be multiple timemaps stored in the file, right?)

"on" is a list of elements to start highlight at the given time, and "off" is a list of elements to unhighlight at the same time.  The `tstamp` value is probably equivalent to "real time", and the `qstamp` is equivalent (or close) to "score time".

The units of `tstamp` are milliseconds (from the start of the recording in your case), and the units of `qstamp` are quarter notes from the beginning of the score.  What units do you described "score time"?

This output should work for timings from an audio recording (or MIDI file performance).   This output from verovio is calculated as part of the MIDI generation process (or vice versa), so there would be a separate function to generate a similar output from an embedded timemap.  

I would think that for efficiency reasons, you would want to export the entire timemap before starting playback so that you would not need to extract the information.

Test MEI data:

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2021-01-20T23:39:02" version="3.2.0-dev-5450dc3-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000000546099368">
    <score xml:id="score-0000002094454345">
     <scoreDef xml:id="scoredef-0000002064863771">
      <staffGrp xml:id="staffgrp-0000001604977159">
       <staffDef xml:id="staffdef-0000001950224428" n="1" lines="5">
        <clef xml:id="clef-L3F1" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="4" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="staff-0000001427200775" n="1">
         <note xml:id="note-L5F1" dur="4" oct="4" pname="c" accid.ges="n" />
         <note xml:id="note-L6F1" dur="4" oct="4" pname="d" accid.ges="n" />
         <note xml:id="note-L7F1" dur="4" oct="4" pname="f" accid.ges="n" />
         <note xml:id="note-L8F1" dur="4" oct="4" pname="g" accid.ges="n" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

wergo:
Hi @craigsapp, thanks for pointing me to this timemap option that I was not aware of. 

However, what I was looking for is the scoreTime info. Thus, the beat number within a measure of a given note (starting with 1). This would be hard to extract from the timemap info. 

Or am I missing something?

craigsapp:
Hi

craigsapp:
You are not missing anything.  The measure number and MEI beat number in the measure would be good additions to add to the JSON timemap.  In that case time signature information whenever it appears in the score would also be useful to add.  In particular because the beat is dependent on  the bottom of the time signature.  If there is no time signature, then it is probably best to assume the quarter note is the beat.

In your timemaps, do you specify the time of every event, or is the timemap sparse? (such as only giving an alignment point every beat/measure? In other words it would be useful to have an interpolation function to estimate the realtime values for events that are not assigned a realtime value (this is one purpose of the qstamp, because otherwise the time signature information would be needed to do the interpolation, and the encoding would need to match the duration of a measure to the time signature (so interpolation in cadenzas, pickup measures the measures split by repeats become complicated to interpolate times through).

wergo:
Currently our alignment algorithm takes every score event to point to the time line in a recording. We even would love to see grace notes appear in there so that they get matched/aligned too. So, we did not have the necessity to interpolate at the moment.

However, for my current use case (https://atom.io/packages/mei-friend/), exposing the 6 variables that Laurent named in a JSON object for a given `xml:id` would be best. What about naming a new method `Toolkit::GetTimesForElement` so not to break to existing `Toolkit::GetTimeForElement`?

craigsapp:
I don't have a clear idea of how your system works, so that is why all of the annoying questions  (And if I don't understand it, then I cannot form an opinion).  And in particular how to deal with repeats in verovio (and do you deal with repeats?).

The alignment data is extracted related to your MEC Vienna presentation?  And is it generated dynamically, or is it prepared in advance and then stored in the MEI file?

Adding grace notes to the timemap output would be good to do.   Currently they are not exported by the MIDI conversion process since grace notes are ignored (and therefore are not assigned any times during the MIDI extraction process).   What do you consider the "scoreTime" of a grace note?  Is it the same as the next note that has duration?  In other words, I am wondering what you are using scoreTime for.  Are you using it to display the current measure and beat.  And if you are showing the beat, then what do you do in 6/8 meter, where there are two musical beats (at normal tempos) at the dotted quarter level, but the MEI beat is the eighth note (the bottom number of the time signature).

How do you deal with repeats in the score?  This is somewhat related to the proposed name `Toolkit::GetTimesForElement`.  In other words, what are you meaning by `s` in `Times`?  Both scoreTime and realTime?  If so, note there is also a complication that any note in a score can be played multiple times during a single performance.  

The last two paragraphs related to how a realTime timeline interfaces with the score.  During performance, realTime is monotonic (you cannot transport back in time), but scoreTime is not monotonic.  So the primary indexing for alignment is realtime, and it is not possible index in the general case by scoreTime since there are potentially many realTimes for every scoreTime.

This is also related somewhat to computational efficiency.  It is most efficient to have a list of time-sorted elements that the `verovio -t timemap` produces.  If you do a querySelector operation on verovio continuously for random access to the same data that is very computationally expensive (the system gets slower as there are more notes on the page).  So that is related to the question of what is the difference between the current timemap JSON data (with enhancements of measures and beats) and the method of data access that you need?   And related to this is that the current system needs to be modified to deal with performance repeats in the score, so it would be useful to think about this at the same time.

If the realTimes are store in the MEI data, then that would be useful to give a short (one measure sample) which may help me.   Also here is an example which forms my viewpoint that I am asking:

https://www.ccarh.org/haydn/op20n5/mvmt2

which has lots of repeats.  If you view the sort of that page, you can see the timemap that I created.  In this case I also store every single event in the timemap.  Here is the start of it (the pipe characters at the start of lines are irrelevant and caused by copy-and-paste from the view source page in Chrome):

```
  | {"measure":1,	"moffset":0,	"qstamp":0,	"tstamp":1.802},
  | {"measure":1,	"moffset":1,	"qstamp":1,	"tstamp":2.205},
  | {"measure":1,	"moffset":2,	"qstamp":2,	"tstamp":2.541},
  | {"measure":2,	"moffset":0,	"qstamp":3,	"tstamp":2.916},
  | {"measure":2,	"moffset":1,	"qstamp":4,	"tstamp":3.285},
  | {"measure":2,	"moffset":2,	"qstamp":5,	"tstamp":3.605},
  | {"measure":3,	"moffset":0,	"qstamp":6,	"tstamp":3.925},
  | {"measure":3,	"moffset":1,	"qstamp":7,	"tstamp":4.272},
  | {"measure":3,	"moffset":2,	"qstamp":8,	"tstamp":4.584},
  | {"measure":4,	"moffset":0,	"qstamp":9,	"tstamp":4.964},
```

The timemap has the realTime (tstamp) a `qstamp` which  the same as `qstamp` in the timemap output from verovio, but is also serves a redundant functionality to the `on` and `off` list in the timemap output.  I wanted this in the verovio output for the purpose of not needing to specifiy all event times in the time map (and if there is an event that does not have an entry, the qstamp is needed in order to interpolate a realTime between two adjacent events that are given).   I also add the measure number and the `moffset` which is the quarter-note duration from the start of the barline (which is analogous to the beat info that you want).

This list is index by `tstamp`, if you look down the list, you will see the scoreTime data repeating due to repeated sections of the music.  In addition there is a repeat of the minute (but without repeats).

wergo:
There are many questions to discuss.

1) Alignment: Our system (CLARA https://clara.mdw.ac.at/ will be updated soon) is actually storing the alignment info outside the MEI in RDF triples that are passed on to a Solid POD through the TROMPA CE. So for each note id there is a pointer to the timeline in the recording (actually the full data model is more complex). Storing this info inside the encoded score does not make sense to us since we are dealing with many recordings of the same score.

2) Repetitions: Regarding repetitions: we have introduced the `--expand` option (#1152) that reads an `expansion` element as in all your encodings and create a long MEI structure that can be exported as MIDI. To access the info about repetitions in the expanded timeMap, there is an `expansionMap` inside Verovio that can be accessed through `Toolkit::GetNotatedIdForElement()` and `Toolkit::GetExpansionIdsForElement()`. The latter returns a JSON string of all ids of the same note played multiple times (and thus represented multiple times and having different `xml:id`s in the expanded timemap), the former the first of this list which is the first (=notated) note. (We have discussed this in #1300.)

3) This Haydn movement is a great demo. Are you using the expand option to create your expanded timemap?

4) The purpose of the PR is much simpler: I want to insert a control element (`slur`, `hairpin` etc) at one or two selected notes and insert it with `@tstamp` rather than `@startid/endid` attributes.

5) The plural in the `Toolkit::GetTimesForElement` was just to make clear that this method would return multiple time formats and instances for a given note (`scoreTimeOnset`, `scoreTimeOffset`, `scoreTimeDuration`, `scoreTimeTiedDuration`, `realTimeOnsetMilliseconds`, `realTimeOffsetMilliseconds`). If it is clearer, we could have a method `Toolkit::GetScoreTimesForElement` and just return the first four items.

craigsapp:
I think I understand things now: 

There are two ways of dealing with repeated music.  I was thinking more about one case and Werner is thinking about the other.

Werner expands the repeats in a score for display with performance information (specifically to display tempo curves above the notation extracted from actual performances).   In this case the current MIDI rendering is very similar: each note in the score is played exactly once.

The second way of viewing the timing information is for an unexpanded score from which an expanded MIDI rendering is made (or a performance from MIDI/audio is aligned) in order to realize repeats.

For the expanded score version, there is a one-to-one mapping between scoreTime and realTime, and both are monotonic (always heading forward in time/space).

For the contracted score with an expanded MIDI/audio alignment, there is not a one-to-one mapping: realTime is monotonic, but scoreTime is not, since the same note in the score may be place more than once in the performance.

[By the way: what happens to `xml:ids` when producing an expanded score?  The IDs must be prefixed with a repetition marker such as `_2` or similar.]

So timing variables related to scoreTime should export as single values, and realTime variables should be exported in an array.   It would also be possible for a note to have zero realTime information, in which case an empty array should be returns for the realTime data.  This sort of case occurs if a performance takes only second endings, so there is no realTime information for the first endings.

craigsapp:
Also with regards to the timemap output: It should keep the same structure regardless of whether or not the MIDI is expanded or not when the score is compressed.  The main index is realTime, and there will be duplicates of scoreTime data when it returns later in a performance.

lpugin:
As I understand the current implementation of expansion is duplicating elements internally, so the timemap should work as it is already.

Going back to the PR I cannot think of a better name than `GetTimesForElement` even though I am not very happy about it. I would avoid `GetScoreTimesForElement` since the idea is also to include real time values.

@wergo: can you modify the PR for the function to return a JSON object? And if somebody has a better idea for the name we can change it to that.

wergo:
Ok, thanks for this @lpugin; In understand now. 

This PR now returns the six values in arrays. If not a note, it returns an empty JSON object, as in other similar methods.

craigsapp:
Yes, the current implementation for MIDI rendering does a single pass, ignoring expansions.  In the future there should be an option to render MIDI based on expansions encoded in the data.  This will also be useful for aligning audio to an unexpanded score, which allows for interactions such as in the demo:   [https://www.ccarh.org/haydn/op20n5/mvmt2](https://www.ccarh.org/haydn/op20n5/mvmt2), where the same note in the score has multiple performance times.  Also this sort of layout would be useful for performance analysis with tempo curves: you can overlay two different tempo curves for each repetition of the music to compare how the performance changes between the two repeats of the same music (so there could be potential use of this feature in CLARA as well).

One intermediate thing that would be nice to do would be to skip the first endings in the current MIDI rendering.  In that case the realTime variables for notes in the first ending would return empty arrays. (as it sounds very unmusical to listen to the first repeat  ending, and then immediately afterwards the second repeat ending -- although that is useful for proof-listening to all of the music in the score).

wergo:
> Yes, the current implementation for MIDI rendering does a single pass, ignoring expansions. In the future there should be an option to render MIDI based on expansions encoded in the data.

This is already there (since Verovio 2.5): When calling something like `verovio --expand expansion-default -t midi Beethoven_Op28_3-HenleUrtext.mei` you will get an expanded MIDI file with all repetitions as defined in the expansion element (prima/seconda volta, everything).

> This will also be useful for aligning audio to an unexpanded score, which allows for interactions such as in the demo: https://www.ccarh.org/haydn/op20n5/mvmt2, where the same note in the score has multiple performance times. Also this sort of layout would be useful for performance analysis with tempo curves: you can overlay two different tempo curves for each repetition of the music to compare how the performance changes between the two repeats of the same music (so there could be potential use of this feature in CLARA as well).

This is also possible with the current vervio; we will soon demonstrate it on CLARA: performed tempo curves on expanded verovio; multiple repeated tempo curves on original verovio. 

BTW: The Haydn demo is very nice; who is playing? And how did you realise the jumping if not with `--expand`?

craigsapp:
> This is already there (since Verovio 2.5)

Thanks for reminding me -- I still need to play around with the new (old) feature :-).  I put expansion lists into my humdrum-to-MEI conversions, but I have not utilized (or tested) them yet.

But I suspect that the current expansions first manipulate the MEI data to flatten it, and then the MIDI is calculated on that flattened score.   So I think it is not possible to load an unexpanded score to display, and generate expanded MIDI in a single step.  Is that right @lpugin?  In other words the displayed score linked to the MIDI file has to be expanded as well.

For the example I gave, it was created using SCORE notation renderings rather than verovio's.

lpugin:
> So I think it is not possible to load an unexpanded score to display, and generate expanded MIDI in a single step. Is that right @lpugin? In other words the displayed score linked to the MIDI file has to be expanded as well.

Correct. I can imagine cases where you want the MIDI (and times) aligned with the non expanded version of the score. This is where the arrays will be useful. One question is whereas we want to support this (e.g., internal MIDI repetition expansion) with and / or without encoded `<expansion>`. Both make sense and could be desirable. For simple cases, users probably would not expect to add `<expansion>`. For more complex ones these might be necessary.
Please add any follow-up comment here https://github.com/rism-ch/verovio/issues/235

PR #1992, opened 2021-01-26
refactor artic import from MusicXML

rettinghaus:


src/iomusxml.cpp
```diff
@@ -2846,51 +2848,24 @@ void MusicXmlInput::ReadMusicXmlNote(
...
-                    artics.push_back(ConvertArticulations(articulation.name()));
-                    if (!std::strcmp(articulation.name(), "detached-legato")) {
-                        artics.push_back(ARTICULATION_stacc);
-                        artics.push_back(ARTICULATION_ten);
-                    }
+                artics.push_back(ConvertArticulations(articulation.name()));
+                if (!std::strcmp(articulation.name(), "detached-legato")) {
+                    artics.clear();
+                    artics.push_back(ARTICULATION_stacc);
+                    artics.push_back(ARTICULATION_ten);
```
lpugin, line 2857: It seems that you are still setting multiple values to `@artic` attributes. So I assume this is still work-in-progress. Correct?

rettinghaus:
This is because stacc-ten was removed from values. Having these in two artic elements could lead to wrong positioning.
Anyway?

lpugin:
Having two artic elements will not be a problem for the positioning, so just split them up.

src/iomusxml.cpp
```diff
@@ -2852,8 +2852,15 @@ void MusicXmlInput::ReadMusicXmlNote(
...
                 if (!std::strcmp(articulation.name(), "detached-legato")) {
+                    // we need to split up this one
                     artics.clear();
                     artics.push_back(ARTICULATION_stacc);
+                    artic->SetArtic(artics);
+                    artic->SetColor(articulation.attribute("color").as_string());
+                    artic->SetPlace(artic->AttPlacement::StrToStaffrel(articulation.attribute("placement").as_string()));
+                    element->AddChild(artic);
+                    artics.clear();
+                    artic = new Artic();
```
lpugin, line 2863: You also need to set the color and the place to that one

rettinghaus: It is done 8 lines below.

lpugin: Ah yes, sorry, I overlooked it. I must say that the logic is now a bit cumbersome. It is a bit weird to add the value to `artics` (line 2853) and only after to check what it is and to remove it if it is a 'stacc ten'.

rettinghaus: Tried to improve it a bit.

Issue #2023, opened 2021-02-01
EditorToolkitCMN::Insert() xml:id with "#"


wergo:
When inserting a control element in `EditorToolkitCMN::Insert()`, the `@startid` and `@endid` attributes are inserted without the trailing "#". 

https://github.com/rism-ch/verovio/blob/8b76019f214b376055839faf4eab84bf8d47bc44/src/editortoolkit_cmn.cpp#L305-L306

Issue #2022, opened 2021-02-01
Toolkit::EditInfo() crashes on CMN


wergo:
When accessing `Toolkit::EditInfo()` through the npm package after successfully inserting a new `hairpin`element through `Toolkit::Edit()` and expecting an `xml:id` string ("hairpin-0000001234"), it throws this error:

`SyntaxError: Unexpected token h in JSON at position 0
    at JSON.parse (<anonymous>)
    at verovio.toolkit.editInfo`

It seems as if `verovio-proxy.js` tries to unsuccessfully JSONize the xml:id string returned by `Toolkit::EditorInfo()` and `EditorToolkitCMN::EditInfo()`:

https://github.com/rism-ch/verovio/blob/cb0871436633d118bcc4211350ef4e8ec04c0bbd/emscripten/verovio-proxy.js#L110

Shouldn't the JSON parsing expression be removed here?

(`EditorToolkitNeume::EditInfo()` seems to return a JSON string, however.)

PR #2027, opened 2021-02-02
Develop editor

wergo:
This PR addresses #2022 and #2023. 

For #2022, the `std::string m_editInfo` in `EditorToolkitCMN` was made a JSON object that now stores the `uuid` of the last edit operation and returns it in `EditorToolkitCMN::EditInfo()`. This is now more similar to the behaviour of `EditorToolkitNeume` where detailed error messages are stored in a JSON object `m_infoObject`. 

`tie`, `dir`, and `dynam` are now added as examples for control elements with only a startid or a startid and a endid.

Bevor merging, we should discuss, how the editor could grow.

### DIFF
--- a/emscripten/verovio-proxy.js
+++ b/emscripten/verovio-proxy.js
@@ -107,7 +107,7 @@ verovio.toolkit.prototype.edit = function ( editorAction )
 
 verovio.toolkit.prototype.editInfo = function ()
 {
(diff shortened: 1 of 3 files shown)


lpugin:
Looks good. I can merge it unless you really prefer to wait.

wergo:
Well, yes of course; let's merge. It works for me. (I was not sure whether in `EditorToolkitCMN::Chain` the potential multiple adding of the uuid would cause any trouble.)

PR #2030, opened 2021-02-03
Avoid space to be added for timestamp events

lpugin:
This PR produces more regular spacing when control events have timestamps pointing to positions where no musical event occurs. For example:

image

image

image

See the test-suite output below for additional examples. I am leaving the PR open for a day since this is a fairly significant change in case people have questions or comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Improved positioning of ornaments with multiple layers
 * Improved articulations in normal and cross-staff contexts
 * Improved spacing of clef changes
+* Improved timestamp events alignments by avoiding extra-space to be added
 * Embedding of Leland font from MuseScore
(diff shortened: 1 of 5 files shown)


craigsapp:
For the pedal example:

Screen Shot 2021-02-03 at 1 13 32 PM

I like the top rendering.  If the timestamp for the pedal-up is at the end of the measure, it should be right justified to the barline and not cross over it (note the up-pedal at the end of the music, which overlaps into the margin).

lpugin:
Yes, you are actually referring to this issue https://github.com/rism-ch/verovio/issues/949. Avoiding the timestamp additional space made it slightly worse but it is only a side effect that will go away once the pedal adjustment is fixed. So I am happy to ignore you comment ;-)

craigsapp:
😜

PR #2029, opened 2021-02-03
Improve vertical tempo alignment

Monceber:
closes #2013 
Improves positioning of the vertical tempo alignment - in case when there are accidentals present, tempo should be aligned with them, instead of the note.
image

This works with both @startid and @tmstamp

PR #2031, opened 2021-02-04
Vertical tempo alignment improvement

Monceber:
Addressed things, mentioned in #2029.
I left FindAllDescendantByComparison() since improved implementation requires all of the children, instead of just first in the list. This works better for chords like following, that have the same accidental for each note (and first accidental wouldn't necessarily be the leftmost):
image

### DIFF
--- a/include/vrv/horizontalaligner.h
+++ b/include/vrv/horizontalaligner.h
@@ -188,7 +188,7 @@ public:
     /* Return true and X position of the first accidental from the accid space of alignment reference for specific staff
      * (found by staffN parameter). Return false and 0 otherwise.
      */
-    std::pair<int, bool> GetAccidAdjustedPosition(Doc *doc, int staffN);
+    std::pair<int, bool> GetAccidAdjustedPosition(Doc *doc, int staffN, int staffSize);
 
     //----------//
     // Functors //
@@ -319,9 +319,9 @@ public:
     bool HasMultipleLayer() const { return (m_layerCount > 1); }
 
     /**
-     * Return first accidental from the accid space of the alignment reference
+     * Return const pointer to the accid space
      */
-    Accid *GetFrontAccid() const;
+    const std::vector<Accid *> *GetAccidSpace(bool docChildren = true) const { return &m_accidSpace; }
 
     //----------//
     // Functors //
--- a/include/vrv/horizontalaligner.h
+++ b/include/vrv/horizontalaligner.h
@@ -185,11 +185,6 @@ public:
      */
     bool HasTimestampOnly();
 
-    /* Return true and X position of the first accidental from the accid space of alignment reference for specific staff
-     * (found by staffN parameter). Return false and 0 otherwise.
-     */
-    std::pair<int, bool> GetAccidAdjustedPosition(Doc *doc, int staffN, int staffSize);
-
     //----------//
     // Functors //
     //----------//
@@ -317,11 +312,6 @@ public:
      * Return true if the reference has elements from multiple layers.
      */
     bool HasMultipleLayer() const { return (m_layerCount > 1); }
-
-    /**
-     * Return const pointer to the accid space
-     */
-    const std::vector<Accid *> *GetAccidSpace(bool docChildren = true) const { return &m_accidSpace; }
 
     //----------//
     // Functors //
--- a/src/horizontalaligner.cpp
+++ b/src/horizontalaligner.cpp
@@ -476,7 +476,7 @@ std::pair<int, bool> Alignment::GetAccidAdjustedPosition(Doc *doc, int staffN, i
     Accid *accid = vrv_cast<Accid *>(*iter);
     const int accidGlyphWidth = doc->GetGlyphWidth(accid->GetAccidGlyph(accid->GetAccid()), staffSize, false);
 
-    return { (*iter)->GetDrawingX() - accidGlyphWidth, true };
+    return { (*iter)->GetDrawingX() - accidGlyphWidth / 2, true };
 }
 
 bool Alignment::IsSupportedChild(Object *child)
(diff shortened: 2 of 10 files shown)


src/horizontalaligner.cpp
```diff
@@ -461,17 +461,22 @@ void Alignment::ClearGraceAligners()
...
-    Accid *accid = reference->GetFrontAccid();
-    if (!accid) return { 0, false };
+    const std::vector<Accid *> *alignmentAccidentals = reference->GetAccidSpace();
+    if (alignmentAccidentals->empty()) return { 0, false };
 
-    return { accid->GetDrawingX(), true };
+    auto iter = std::min_element(alignmentAccidentals->cbegin(), alignmentAccidentals->cend(),
+        [](const Accid *left, const Accid *right) { return left->GetDrawingX() < right->GetDrawingX(); });
+    Accid *accid = vrv_cast<Accid *>(*iter);
+    const int accidGlyphWidth = doc->GetGlyphWidth(accid->GetAccidGlyph(accid->GetAccid()), staffSize, false);
```
lpugin, line 477: I think you call `Accid->GetSelfLeft()` that will give you the exact x position you are looking for. You want to do the same in the `std::min_element` call above because the one that has the smallest drawing x position is not necessary the one with the smallest self x (depending on their accid glyph). However, calling `GetSelfLeft` would work only if this is not done in the drawing loop but in a separate alignment step. See my general remark.

include/vrv/horizontalaligner.h
```diff
@@ -319,9 +319,9 @@ class AlignmentReference : public Object, public AttNInteger {
     bool HasMultipleLayer() const { return (m_layerCount > 1); }
 
     /**
-     * Return first accidental from the accid space of the alignment reference
+     * Return const pointer to the accid space
      */
-    Accid *GetFrontAccid() const;
+    const std::vector<Accid *> *GetAccidSpace(bool docChildren = true) const { return &m_accidSpace; }
```
lpugin, line 324: What's the purpose of the `docChildren` parameter?

src/horizontalaligner.cpp
```diff
@@ -461,17 +461,22 @@ void Alignment::ClearGraceAligners()
     m_graceAligners.clear();
 }
 
-std::pair<int, bool> Alignment::GetAccidAdjustedPosition(Doc *doc, int staffN)
+std::pair<int, bool> Alignment::GetAccidAdjustedPosition(Doc *doc, int staffN, int staffSize)
```
lpugin, line 464: Instead of the `std::pair` you could return just `int` with `VRV_UNSET` when nothing is found. But you can leave it like this.

lpugin (commented):
If we really want to make the alignment perfect we should do it through an adjustment functor that sets the relative x position of each tempo FloatingPositioner. By doing it in an alignment adjustment step, we can call Accid::GetSelfLeft instead of Accid::GetDrawingX and substract half of the glyph width. This would also avoid the x position adjustment to be done during the drawing process (in DrawTempo) which is usually to be avoided. But I understand that changing this might be over the top. I can give you more detailed hints if you want to.

lpugin:
It just came to my mind that you can use https://github.com/rism-ch/verovio/blob/69cc9ed538cd379318bbb56aa5d7059ed994ec31/include/vrv/horizontalaligner.h#L135 instead of adding ` std::pair<int, bool> Alignment::GetAccidAdjustedPosition(Doc *doc, int staffN, int staffSize)`
(However, that would not work in the drawing process, so it would definitely be much cleaner to put it in a alignment step)

include/vrv/horizontalaligner.h
Monceber: Yeah, this should not be there. I'll remove it.

src/horizontalaligner.cpp
Monceber: Regarding this point, I think TEMPO don't even have FloatingPositioner set in the current implementation. They are drawn as FloatingObjects similar to the dynamics or mordents. To leverage FloatingPositioner, first a bunch of code reordering would be needed and I'm not sure that's not going to break current functionality.

Monceber: Nvm my previous comment, I think I see where FloatingPositioner is set.

lpugin:
The basic idea is:
* adjust the Tempo in Page::LayOutHorizontally with a new AdjustTempo functor
* add System::AdjustTempo functor implementation that passes the SystemAligner to the functor parameters
* add Tempo::AdjustTempo functor implementation that loops through all the FloatingPositioner on that system (one Tempo will have several FloatingPositioner objects if it is repeated on more than one staff). See Harm::AdjustHarmGrpsSpacing for one example, even though it will be much simpler than that. For each FloatingPositioner, get the left position of the content (i.e., chord) with Alignment::GetLeftRight() for the positioner's staff, and adjust it.

Monceber:
@lpugin I moved logic to AdjustTempo, as suggested, but I couldn't quite find a way to set relative offsets for the FloatingPositioners. This is different for Harm elements, since we offset actual elements that `<harm>` is linked to, whereas here we want to offset just `<tempo>` element. Since FloatingPositioners are reset between the calls and drawingRelativeX for them end up being reset to 0 as well. We can't set relative X for `<tempo>` as well, since it has different FloatingPositioners for different staves.
I went with new `m_drawingXRel` specifically for Tempo, but am I missing something here?

Monceber:
Drawing remains the same:
image

lpugin:
Great! I was not sure about the relative x of the FloatingPositioners. Actually is it the first time we are adjusting the relative x position of ControlElements this way. There is something similar for Arpeg, but because these are not expected to be repeated on multiple staves, the xRel shift is store in Arpeg (and not in its positioner). Your solution with `m_drawingXRel` is great. I would only rename it `m_drawingXRels` so it is clear it is not a single value.

(We can keep it in mind for the future. Maybe one change in the future could be to put this new member up in ControlElements so they are preserved. But we can think about it in due time.)

Thanks for all these revisions!

PR #2038, opened 2021-02-05
fix stem.mod on chords

rettinghaus:
Haven't found this case in real life, but should work anyway.

### DIFF
--- a/src/view_element.cpp
+++ b/src/view_element.cpp
@@ -1478,12 +1478,13 @@ void View::DrawStem(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
         stem->GetDrawingY(), stem->GetDrawingX() + m_doc->GetDrawingStemWidth(staff->m_drawingStaffSize) / 2,
         stem->GetDrawingY() - stem->GetDrawingStemLen());
 
-    if (stem->HasStemMod()) {
+    if (stem->HasStemMod() && (stem->GetDrawingStemLen())) {
         if (stem->GetStemMod() == STEMMODIFIER_sprech) {
             int yShift = m_doc->GetDrawingDoubleUnit(staff->m_drawingStaffSize) * 2;
             Note *note = vrv_cast<Note *>(stem->GetParent());
             if (!note) {
                 Chord *chord = vrv_cast<Chord *>(stem->GetParent());
+                assert(chord);
                 if (stem->GetDrawingStemLen() < 0) {
                     note = chord->GetTopNote();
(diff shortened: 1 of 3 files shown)


src/view_element.cpp
```diff
@@ -1482,6 +1482,15 @@ void View::DrawStem(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
         if (stem->GetStemMod() == STEMMODIFIER_sprech) {
             int yShift = m_doc->GetDrawingDoubleUnit(staff->m_drawingStaffSize) * 2;
             Note *note = vrv_cast<Note *>(stem->GetParent());
+            if (!note) {
+                Chord *chord = vrv_cast<Chord *>(stem->GetParent());
```
lpugin, line 1486: Add assert for chord. Thanks

lpugin:
Even less likely: a chord with `@stem.mod="spreech"` and no note. I think it will crash with your code. Pls double check this.

src/view_element.cpp
rettinghaus: done

rettinghaus:
Verovio automatically adds a note to empty chords, so everything works fine now:
Sprechakkord

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Sprechakkorde</title>
         </titleStmt>
         <pubStmt>
         </pubStmt>
      </fileDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffGrp>
                        <staffDef n="1" lines="5" clef.shape="G" clef.line="2" key.sig="0" meter.count="4" meter.unit="4" />
                     </staffGrp>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <chord dur="4" stem.mod="sprech">
                              <note />
                              <note />
                           </chord>
                           <chord dur="4" stem.mod="sprech">
                              <note />
                           </chord>
                           <chord dur="4" stem.mod="sprech">
                           </chord>
                           <chord dur="4" stem.len="0" stem.mod="sprech">
                              <note />
                              <note />
                           </chord>
                           <chord dur="4" stem.visible="false" stem.mod="sprech">
                              <note />
                              <note />
                           </chord>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

PR #2041, opened 2021-02-08
Develop fix grpsym

lpugin:
@Monceber please double check. I don't have complex GrpSym examples to test. Many changes are just renaming for better clarity and grouping methods by prefix. The idea is that setting the group symbols needs to be (re)done for each system. This is why it is now happening in `Doc::CastOffDocBase` and `Doc::CastOffEncodingDoc` and not in `Doc::PrerpareDrawing` anymore. ~~Note the there is no more call on Doc::m_scoreDef since this one is actually never drawn. We can add one call (in Doc::PrepareDrawing) if necessary.~~ [call is now added]

What is not optimal is that the resolving of `@startid`/`@endid` is performed for each System::m_darwingScoreDef but I do not expect this to have a noticeable impact. It can maybe be  optimized by resolving these once on Doc::PrepareDrawing (for Doc::m_scoreDef) and reusing them in ScoreDefSetGrpSym for System::m_drawingScoreDef. But I don't think it is necessary.

### DIFF
--- a/doc/test-suite.py
+++ b/doc/test-suite.py
@@ -63,6 +63,7 @@ if __name__ == "__main__":
 
             # filenames (input MEI and output SVG)
             meiFile = os.path.join(path1, item1, item2)
+            print(meiFile)
             name, ext = os.path.splitext(item2)
             svgFile = os.path.join(path2, item1, name + '.svg')
             pngFile = os.path.join(path2, item1, name + '.png')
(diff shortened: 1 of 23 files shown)


Monceber:
Yes, I think this looks good.
My only concern here is allowing `SetGroupSymbol` to reset group symbol. I understand the reasoning behind it, since we're going to call it for each System. However, this opens a possibility where `StaffGrp` could have several `GrpSym` defined, and basically the last one is going to overwrite all previous values.
However, I believe that having several group symbols for the same group/def is an encoding error anyway, so this might not be a big deal.

lpugin:
OK, thanks. I will merge in order to fix the issue. We can still see if the `GrpSym` overwrite needs to be fixed later on. I agree this would not make much sense in terms of encoding anyway - even though it is allowed in MEI.

PR #2049, opened 2021-02-10
Explanation and fix for neume custos rendering

JRegimbal:
Apologies for another PR for this function! I missed a case for F clefs previously and also realized that this approach would only hold for neume notation. I added a few more checks to fix these bugs and also a few comments to explain that these lines are included since MEI neume notation treats the C clef as C3 rather than C4.


lpugin:
Thanks. Do we have an issue open for this?

JRegimbal:
I don't believe there's an open issue for this.

PR #2056, opened 2021-02-17
Update Leipzig

rettinghaus:
This PR adds some more clefs, refines accidental parentheses and add ottava parentheses.


lpugin:
Thanks. Could you redo a PR with only the changes to
* Leipzig-5.2.sdf
* Leipgiz.svg
* supported.xsl

Then I can regenerate the files in the develop branch. This will make it easier to manage the changes. I am updating the wiki accordingly.

PR #2060, opened 2021-02-18
[WIP] Update swig in ubuntu

### LINKED ISSUE #2058: Verovio Python Toolkits are not equivalent across platforms (Mac, Linux)
I've installed the Python Verovio package, version 3.2.0 on an Ubuntu machine and on a Mac and unfortunately it seems like in the two binaries some methods have different signatures, which causes problems when calling it from the other platform.

### Mac
Screen Shot 2021-02-18 at 15 00 33
Screen Shot 2021-02-18 at 15 00 49
Screen Shot 2021-02-18 at 15 01 01

### Ubuntu
Screen Shot 2021-02-18 at 15 00 10
Screen Shot 2021-02-18 at 15 01 24
Screen Shot 2021-02-18 at 15 01 34

I've installed Verovio on both machines using `pip install verovio==3.2.0`

rettinghaus:
fixes #2058
[hopefully]

### DIFF
--- a/.github/workflows/python-ci-wheel.yml
+++ b/.github/workflows/python-ci-wheel.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       # Build the wheels for Linux, Windows and macOS for Python 3.9
       matrix:
-        os: [ macos-latest, windows-latest, ubuntu-18.04]
+        os: [ macos-latest, windows-latest, ubuntu-20.04]
         python-version: [3.6, 3.7, 3.8, 3.9]
         architecture: [ x86, x64 ]
         include:
@@ -37,13 +37,13 @@ jobs:
           - os: windows-latest
             architecture: x86
             platform_id: win32
-          - os: ubuntu-18.04
+          - os: ubuntu-20.04
             architecture: x64
             platform_id: manylinux_x86_64
         exclude:
           - os: macos-latest
             architecture: x86
-          - os: ubuntu-18.04
+          - os: ubuntu-20.04
             architecture: x86
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/python-ci-wheel.yml
```diff
@@ -92,7 +92,6 @@ jobs:
           CIBW_BUILD: ${{ env.CIBW_BUILD_IDENTIFIER }}
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
           CIBW_MANYLINUX_I686_IMAGE: manylinux2014
-          CIBW_BEFORE_ALL_LINUX: yum install swig -y
```
lpugin, line 95: have you tested this? We spent quite some time to get this action running, so I want to make sure it still builds before I merge.

rettinghaus: This was based on a comment by @musicEnfanthen but he tested it without success. I'll get back to this later.

musicEnfanthen:
Looks good to me. Test with recent manylinux image and ubuntu 20.04 succeeds: https://github.com/musicEnfanthen/verovio/runs/1934593395?check_suite_focus=true

I did not test the multiple Python versions for the sdist and the PyPi release.

.github/workflows/python-ci-wheel.yml
```diff
@@ -142,8 +142,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
-        python-version: [3.9]
+        os: [ubuntu-20.04]
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
lpugin, line 140: Actually this makes no sense to me. Why do you want different source distributions for each python version? It broke the action. So you should either remove it or fix the file name conflict it creates.

rettinghaus: ok, for which version then? I would guess 3.6, but you maybe prefer the latest?

lpugin: 3.9 used to work, so I don't see a reason to change this. Why do you want 3.6? In other words: what problem have you encountered with the source distribution?

musicEnfanthen: Beside the question of having sdists for multiple python versions, there seems to be a real issue with the naming of the artifacts: cf. #2069 (which wasn't introduced but revealed by this PR here)

lpugin: I set it back https://github.com/rism-ch/verovio/commit/633ee29af3fd1b1cff8759fe2bd1a78abad16c85
If there is a problem with the sdist open a dedicated issue.

PR #2061, opened 2021-02-20
Fix issues with the interpretation of the dur.quality attribute

martha-thomae:
This PR fixes issues regarding the interpretation of `@dur.quality` in the following cases:

1. When the **imperfection** (a perfect note switches to an imperfect value) / **augmentation** (an imperfect note switches to a perfect value, usually due to the presence of a dot of augmentation) happens at the level of the longa or semibrevis.
**Before:** we were only looking at the value of tempus to determine how to interpret the `imperfecta` / `perfecta` quality values. However, tempus only provides information about the brevis. So, **now** we are looking also at modusminor for the interpretation of the longa and at prolatio for the interpretation of the semibrevis.

2. When the `perfecta` value happens at a small note level (i.e., minima, semiminima, fusa, and semifusa). Here the notes should be augmented as well (changing from an imperfect to a perfect value).


lpugin:
Thanks @martha-thomae. Is this related to this issue https://github.com/rism-ch/verovio/issues/927?

PR #2062, opened 2021-02-22
Add support for MusicXML compressed files. Closes #2050

### LINKED ISSUE #2050: Support for MXL (compressed MusicXML) import
MXL is a compressed container format for MusicXML. It would be great if it were possible for Verovio to render a score from MXL.

Some examples of MXL files can be found here:
https://www.musicxml.com/music-in-musicxml/example-set/

And a schema can be found here:
https://github.com/w3c/musicxml/blob/gh-pages/schema/container.dtd

lpugin:
* MXL files are auto detected and uncompressed with the C++ and Python toolkit
* For the JS toolkit, the compressed files can be passed via `loadZipDataBase64` or `loadZipDataBuffer`
  * `loadZipDataBase64` takes a base64 encoded string as parameter
  * `loadZipDataBuffer` an `ArrayBuffer`


samuelbradshaw:
Thank you! I'm excited to try this out!

PR #2067, opened 2021-02-22
Fixed: Do not lowercase enum values

### LINKED ISSUE #2065: Enum Options in Python toolkit are lowercased
Currently, the enum options for Verovio are appearing as lowercased:

`verovio.pAE`, `verovio.mEI`, etc.

Looking at the Swig header file for Java I noticed that it's slightly different, and has the option to keep enum values uppercase:

```
// Change method names to lowerCamelCase
%rename("%(lowercamelcase)s") "";
// Ignore enum items (e.g., for fileFormat.PAE)
%rename("%(upper)s", %$isenumitem) "";
```

This change should be ported to the Python toolkit. It will probably break any existing code, though, so this change should be flagged.

ahankinson:
When building the Python toolkit, enum values were lowercased, leading to using things like `verovio.pAE` and `verovio.hUMDRUM`. This commit changes the swig interface file to not lowercase the enum values.

Fixes #2065

### DIFF
--- a/bindings/python/verovio.i
+++ b/bindings/python/verovio.i
@@ -7,6 +7,8 @@
 
 // Change method names to lowerCamelCase
 %rename("%(lowercamelcase)s") "";
+// Ignore enum items (e.g., for fileFormat.PAE)
+%rename("%(upper)s", %$isenumitem) "";
 
 // Method to ignore
 %ignore vrv::Toolkit::GetShowBoundingBoxes( );


ahankinson:
For reference, here is the output of the top-level `verovio` object after this change:

```
In [1]: import verovio

In [2]: dir(verovio)
Out[2]: 
['ABC',
 'AUTO',
 'DARMS',
 'ESAC',
 'HUMDRUM',
 'HUMMEI',
 'MEI',
 'MEIHUM',
 'MIDI',
 'MUSEDATAHUM',
 'MUSICXML',
 'MUSICXMLHUM',
 'PAE',
 'TIMEMAP',
 'UNKNOWN',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 '_verovio',
 'cvar',
 'resource_filename',
 'setDefaultResourcePath',
 'toolkit',
 'verovio']
```

PR #2077, opened 2021-02-25
New: Generate python3 type annotations

ahankinson:
This PR adds the '-py3' flag to the SWIG build process, which in turn generates Python 3.5+ type annotations.

Notably, this also removes python 2.7 support. No detection of the current version is done, so the behaviour of the -py3 flag with a Python 2.7 platform is not tested.

Please comment if you have any thumbs-up/thumbs-down on this.

@musicEnfanthen
@rettinghaus


rettinghaus:
Makes sense to me to do this. (And thanks for fixing the typos.)

musicEnfanthen:
Thanks, looks good. 

For future reference, could you add here some pointers to documentation of the flag `-py3` and describe a little bit the purpose of type annotations?

PS: There are some more "Verorio" typos in the CMAKE file.

ahankinson:
> For future reference, could you add here some pointers to documentation of the flag -py3 and describe a little bit the purpose of type annotations?

See, that's assuming that I read the documentation! 🤣 

Python type annotations mean you can annotate functions with their data types, so that you can know that the `loadData` method, for example, returns a boolean (that's what prompted the digging). Now in my IDE I can see this:

image

which tells me that `data` can be `Any` (a bit of a lie, since it's actually a `std::string const&`, but SWIG doesn't know how to convert it to a stricter Python `str` type), and the return value is a `bool`.

Type hints are described in PEP 484: https://www.python.org/dev/peps/pep-0484/

They're not strict typing, like in C++, and their function is primarily in writing Python, not running it. Most IDEs now will show them, and you can get type checkers like mypy that can analyze your code to see if you're passing or expecting incorrect types.

musicEnfanthen:
Perfect, thanks a lot @ahankinson from my current and future self 😄

rettinghaus (approved):

PR #2080, opened 2021-02-26
New: Flag to control xlink prefix on href

ahankinson:
After some extensive testing it seems that some browsers, in some cases, will not display SVG if the `href` attribute is prefixed with `xlink`, a change that was introduced with SVG 2.0. This is not yet a recognized specification, but the changes in browsers are being enacted nevertheless.

On the other hand, some strict SVG 1.1 clients will not render the SVG correctly if it *doesn't* include the `xlink:` prefix. In my testing this included Inkscape.

Given that the most likely target for Verovio output is browsers, but still wishing to retain backwards compatibility, this PR changes the default behaviour from producing `xlink:href` to simply `href`. Clients that need the full attribute name can set `--svg-include-xlink` which will insert the xlink prefix on the href attributes.

References previous discussions on this topic: #1158, #332

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -51,7 +51,7 @@ SvgDeviceContext::SvgDeviceContext() : DeviceContext()
     m_svgViewBox = false;
     m_html5 = false;
     m_formatRaw = false;
-    m_includeXlink = false;
+    m_removeXlink = false;
     m_facsimile = false;
     m_indent = 2;
 
@@ -858,9 +858,9 @@ void SvgDeviceContext::DrawMusicText(const std::wstring &text, int x, int y, boo
 
     int w, h, gx, gy;
 
-    // include the `xlink:` prefix for backwards compatibility with older SVG viewers.
+    // remove the `xlink:` prefix for backwards compatibility with older SVG viewers.
     std::string hrefAttrib = "href";
-    if (m_includeXlink) {
+    if (!m_removeXlink) {
         hrefAttrib.insert(0, "xlink:");
     }
 
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -51,6 +51,7 @@ SvgDeviceContext::SvgDeviceContext() : DeviceContext()
     m_svgViewBox = false;
     m_html5 = false;
     m_formatRaw = false;
+    m_includeXlink = false;
     m_facsimile = false;
     m_indent = 2;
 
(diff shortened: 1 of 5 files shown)


src/toolkit.cpp
```diff
@@ -1447,6 +1447,7 @@ std::string Toolkit::RenderToSVG(int pageNo, bool xml_declaration)
 
     svg.SetHtml5(m_options->m_svgHtml5.GetValue());
     svg.SetFormatRaw(m_options->m_svgFormatRaw.GetValue());
+    svg.SetXlinkOnHref(m_options->m_svgIncludeXlink.GetValue());
```
lpugin, line 1450: Rename to `SetIncludeXlink` for consistency

lpugin (changes requested):
Have you tested with the PDFKit js toolkit used in MEI Viewer?

ahankinson:
I've tested it and it seems to work as expected. I'm attaching a copy of a quick test file I put together; the verovio-toolkit.js was built with the source from this PR, but the other libraries are the same as those used with the viewer.

Easiest way to test it is to `cd` into this directory and then run a quick Python HTTP server: `python3 -m http.server 8007`, then visit `http://localhost:8007/test.html` in your browser. It should start to download a PDF file.

You can control the xlink behaviour by commenting / uncommenting the appropriate option in the test.html source. (It should be obvious). 

[test-pdfkit.zip](https://github.com/rism-ch/verovio/files/6049829/test-pdfkit.zip)

src/toolkit.cpp
ahankinson: Fixed.

rettinghaus:
Does it has to be independent from option `svgHtml5`? Or could those functionalities be combined in a `svg2` option?

ahankinson:
Laurent and I had a quick discussion about this, and decided that, since 1) SVG2 isn't official yet, but 2) browsers have started to implement it in piecemeal, that it's better to address the specific problem directly, rather than try to do a full SVG2 update. (Not to mention that there's not really a SVG 2 validator out there yet...)

ahankinson:
For reference, the full issue report is here: elm/svg#31 

The issue is that browsers only support `href` on SVG that is attached to the shadow DOM; it's not a problem (either way) if you are doing `innerHTML` and appending it directly to the DOM, but if you're using a JavaScript toolkit that does it's rendering calculations outside of the DOM and then doing diff updates, the ID references are not resolved.

ahankinson:
OK, after talking it over a bit more with @lpugin we decided to flip the behaviour of the tag; the old behaviour is now the default, and the new behaviour is enabled with a flag.

src/svgdevicecontext.cpp
```diff
@@ -858,9 +858,9 @@ void SvgDeviceContext::DrawMusicText(const std::wstring &text, int x, int y, boo
 
     int w, h, gx, gy;
 
-    // include the `xlink:` prefix for backwards compatibility with older SVG viewers.
+    // remove the `xlink:` prefix for backwards compatibility with older SVG viewers.
     std::string hrefAttrib = "href";
-    if (m_includeXlink) {
+    if (!m_removeXlink) {
```
ahankinson, line 863: FYI, I did it this way to avoid needing to strip characters off the string, and instead prepending to the string by default. This avoids needing to make a new copy of the string, or doing any index-based counting to remove the attribute.

lpugin (approved):

PR #2083, opened 2021-03-02
Improve beam notehead collision

Monceber:
Improved beam positioning to avoid collisions with noteheads in multi-layer situations.
If there might be overlapping elements on the other layers we do not shorten beam stems in that duration, if that's enough to avoid collision. Otherwise, we get additional offset for the beam, to avoid said elements.
Before:
image

After:
image

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -182,6 +182,12 @@ protected:
      */
     virtual void FilterList(ArrayOfObjects *childList);
 
+    /**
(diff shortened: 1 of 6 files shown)


ahankinson:
It looks like this has some regressions in the cross-staff tests...

lpugin:
Thanks!

lpugin:
Could we have the example added to the test suite in `layer`?

rettinghaus:
@lpugin I'll add it.

PR #2088, opened 2021-03-04
fix minor iOS issues

rettinghaus:


lpugin:
Thanks. Sorry for missing the Verovio.podspec when adding the zip folder

Issue #1453, opened 2020-05-07
Incorrect mRest duration assumption leads to pause in generated audio


earboxer:
In the following MEI file, there is an mRest in the bass clef of a shortened measure.

When the audio is played back, it pauses for two extra beats, so there is a gap of silence between "what a" and "day". Measure 4 should be only two beats long, so the mRest should only be 2 beats. (Though I don't know how it ought to know that).

This was converted from musicXML:

#### MusicXML

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
  <movement-title> When We All Get to Heaven</movement-title>
  <identification>
    <creator type="composer">MUSIC (HEAVEN 8.7.8.7.Ref): Emily D. Wilson (1865-1942)</creator>
    <creator type="poet">WORDS: Eliza E. Hewitt (1851-1920)</creator>
    <encoding>
      <software>MuseScore 3.4.2</software>
      <encoding-date>2020-05-07</encoding-date>
      <supports element="accidental" type="yes"/>
      <supports element="beam" type="yes"/>
      <supports element="print" attribute="new-page" type="no"/>
      <supports element="print" attribute="new-system" type="no"/>
      <supports element="stem" type="yes"/>
      </encoding>
    </identification>
  <defaults>
    <scaling>
      <millimeters>7.05556</millimeters>
      <tenths>40</tenths>
      </scaling>
    <page-layout>
      <page-height>1683.78</page-height>
      <page-width>1190.55</page-width>
      <page-margins type="even">
        <left-margin>56.6929</left-margin>
        <right-margin>56.6929</right-margin>
        <top-margin>56.6929</top-margin>
        <bottom-margin>113.386</bottom-margin>
        </page-margins>
      <page-margins type="odd">
        <left-margin>56.6929</left-margin>
        <right-margin>56.6929</right-margin>
        <top-margin>56.6929</top-margin>
        <bottom-margin>113.386</bottom-margin>
        </page-margins>
      </page-layout>
    <word-font font-family="FreeSerif" font-size="10"/>
    <lyric-font font-family="FreeSerif" font-size="11"/>
    </defaults>
  <credit page="1">
[...]

PR #2096, opened 2021-03-11
WIP:  Keep mRest duration converting from musicXML

earboxer:
This is an attempt to solve #1453, by changing how the following XML is converted:

```xml
        <rest/>
        <duration>4</duration>
        <voice>1</voice>
        </note>
```

Currently, it doesn't seem to work ~~(The generated MEI is not having the dur.ppq set in the mRest)~~ (The generated MIDI hasn't been changed yet)

This may also need more thought with respect to https://github.com/rism-digital/verovio/issues/1568

### DIFF
--- a/include/vrv/mrest.h
+++ b/include/vrv/mrest.h
@@ -9,6 +9,7 @@
 #define __VRV_MREST_H__
 
 #include "atts_shared.h"
+#include "durationinterface.h"
 #include "layerelement.h"
 #include "positioninterface.h"
 
@@ -22,10 +23,10 @@ namespace vrv {
  * This class models the MEI <mRest> element.
  */
 class MRest : public LayerElement,
+              public DurationInterface,
               public PositionInterface,
               public AttColor,
               public AttCue,
-              public AttFermataPresent,
               public AttVisibility {
 public:
     /**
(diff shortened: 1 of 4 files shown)


include/vrv/mrest.h
```diff
@@ -22,10 +23,10 @@ namespace vrv {
  * This class models the MEI <mRest> element.
  */
 class MRest : public LayerElement,
+              public DurationInterface,
               public PositionInterface,
               public AttColor,
               public AttCue,
-              public AttFermataPresent,
```
rettinghaus, line 28: Why do you want to remove this?

src/iomusxml.cpp
```diff
@@ -2639,6 +2639,7 @@ void MusicXmlInput::ReadMusicXmlNote(
             else {
                 MRest *mRest = new MRest();
                 element = mRest;
+                mRest->SetDurPpq(duration);
```
rettinghaus, line 2642: This will not turn up in the MEI file. You have to add i/o for this to `iomei.cpp`.

rettinghaus (changes requested):

include/vrv/mrest.h
earboxer: AttFermataPresent is included in DurationInterface

lpugin (changes requested):
It causes quite a few alignment problems to have the `DurationInterface` implemented in `mRest`. Since (at least for now) it is only to have `@dur.ppq` it would be simpler simply to implement `DurationGestural`. Please change this and also adjust MEIInput and MEIOutput accordingly

lpugin:
Please reopen with the requested changes.

Issue #1153, opened 2019-10-03
Option to hide measure numbers


earboxer:
We have "No header" and "No footer" options. It would also be nice to have a "No measure number" option.

craigsapp:
That would be useful to have.  There is an option called `--measure-number` which is not implemented to do anything yet (that I can tell).  The options are `system` which would number the first measure of each system (this is the default behavior), and `interval` which would number every fixed number of measures, such as every 5 or 10.  But there is no way current to set the interval?  

A better system would be to allow an integer to `--measure-number`.  When the option is not used, the default system numbering would be used, when the integer is 0, then no measure numbers would be displayed, and a positive integer would be the measure interval to display the numbers at.

A related issue ishttps://github.com/rism-ch/verovio/issues/909, where examples of forcing measure numbers to display are given.

Another solution is given in issue https://github.com/rism-ch/verovio/issues/374, where you remove the measure number via XSLT before sending to verovio.

rettinghaus:
You could even use `g.mNum {display:none}` in css.

craigsapp:
That is a good way, although there will be cases where notational elements are moved to avoid a measure number that is now invisible, so the notation may occasionally look strange.

earboxer:
> A better method would be to allow an integer to --measure-number. When the option is not used, the default system numbering would be used, when the integer is 0, then no measure numbers would be displayed, and a positive integer would be the measure interval to display the numbers at.

I'm not sure any values besides 0, 1, and null would be helpful.

An image search for sheet music shows two major practices

* showing measure numbers at the beginning of each system except the first (i.e. verovio's default numbering)
* showing no numbering

If more complicated numbering is desired, perhaps that would be something that should come from the MEI files.

Does MEI (or MusicXML) have anything related to hiding/showing measure numbers?

craigsapp:
> I'm not sure any values besides 0, 1, and null would be helpful.

A common third style is to display the measure number at every 10th measure, regardless of its position on the system.  Occasionally every 5th barline is numbered.  This system is useful for orchestral parts when rehearsal marks are not used since every part will have different system breaks.

Here is an example of this numbering style:

http://conquest.imslp.info/files/imglnks/usimg/7/7e/IMSLP04090-Beethoven_-_Piano_Sonatas_Lamond_-_27.pdf

[...]

PR #2100, opened 2021-03-16
Implement --mnum-interval for frequency of automatically placed measure numbers

earboxer:
see https://github.com/rism-digital/verovio/issues/1153#issuecomment-540940279

Examples

* `verovio --mnum-interval 1 file.mei` -- show every measure number
* `verovio --mnum-interval 10 file.mei` -- show every 10th measure number (like https://github.com/rism-digital/verovio/issues/1153#issuecomment-540050375)
* and with the value 0, it has the same behavior as master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+* Option --mnum-interval for changing frequency of automatically-placed measure numbers (@earboxer)
 
 ## [3.3.0] - 2021-02-25
 * Support for `@glyph.name`
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -59,8 +59,6 @@ enum option_FOOTER { FOOTER_none = 0, FOOTER_auto, FOOTER_encoded, FOOTER_always
 
 enum option_HEADER { HEADER_none = 0, HEADER_auto, HEADER_encoded };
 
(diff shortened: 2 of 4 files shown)


rettinghaus:
I don't think that this is a sufficient solution. What if you have non-integral measure numbers like `10a` an `10b`?
And how would you turn off measure number display? To me a value of 0 would be the choice to achieve that, but that will keep the current behaviour?

lpugin:
> And how would you turn off measure number display?

@rettinghaus you can do `<scoreDef mnum.visible="false">` or as you suggested https://github.com/rism-digital/verovio/issues/1153#issuecomment-538128676 use css.

Another approach would be to have `-1` as default yielding the current behaviour (meaning beginning of each system), and turn off measure number display when `0` is set. However, this is not very intuitive and would require to be well documented.

earboxer:
@rettinghaus In this implementation, as `std::atoi` will treat both "10a" and "10b" as 10, both would appear (for mnumInterval of 2, 5, or 10).
If you want to show every 10th measure number, you'd need to manually tell 10b to not show the measure number. This can be done by adding an empty mNum.

```xml
<measure n="9"></measure>
<measure n="10a"></measure>
<measure n="10b"><mNum></mNum></measure>
<measure n="11"></measure>
```

With `verovio --mnum-interval 10 file.mei`, you'd get "10a", "20", "30"...

lpugin:
Looks good to me as a start

Issue #858, opened 2018-05-11
Prefix font glyphs in SVG output


efreja:
Hello,

It seems that it's not possible to render with the JavaScript toolkit several music fragments with different fonts.
Whereas several options are taken into consideration (e.g scale) in each part, the font seems to be the one chosen for the first fragment as if it could not be overriden by a next setOptions, or option parameter of any method that accepts it like renderData.

In the example below, although the first incipit should be displayed with Leipzig and the second Bravura, it's Leipzig that is used for both. If we switch Leipzig and Bravura, it's Bravura that is used for both.

Thanks

-------------

```javascript
<?xml version="1.0" encoding="utf-8"?>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<script type='text/javascript' src='http://www.verovio.org/javascript/jquery.min.js'></script>
<script src="http://www.verovio.org/javascript/latest/verovio-toolkit.js" type="text/javascript"></script>
<!- Create the Vevorio toolkit instance -->
<script type="text/javascript">
var vrvToolkit = new verovio.toolkit();
</script>
</head>
<body>
<div id="output" width:100%;></div>
<script type="text/javascript">
  /* The Plain and Easy code to be rendered */
  var zoom = 60;
  var pageWidth = ($("#output").parent().width()) * 100 / zoom ;
  var pageHeight = pageWidth * 31/23.5 ;
  var border = 50;
  var spacingSystem = 2;
  var data = "@clef:G-2\n\
@keysig:xFCGD\n\
@timesig:3/8\n\
@data:'6B/{8B+(6B''E'B})({AFD})/{6.E3G},8B-/({6'EGF})({FAG})({GEB})/";
  /* Create the Vevorio toolkit instance */
  //var vrvToolkit = new verovio.toolkit();
  /* Render the data and insert it as content of the #output div */
  document.getElementById("output").innerHTML = vrvToolkit.renderData( 
      data, 
      { font:"Leipzig",
                     pageHeight: pageHeight,
                     pageWidth: pageWidth,
                     border: border,
                     scale: zoom,
                     spacingSystem: spacingSystem,
                     adjustPageHeight: 1,
                     ignoreLayout: 1
                      }
  );
  console.log(vrvToolkit.getLog());
</script>
<div id="output2" width:100%;></div>
<script type="text/javascript">
  /* The Plain and Easy code to be rendered */
  var zoom = 60;
  var pageWidth = ($("#output2").parent().width()) * 100 / zoom ;
[...]

PR #2105, opened 2021-03-18
Develop svg ids

lpugin:
This PR adds random IDs to glyphs IDs in the SVG. So instead of
```xml
<symbol id="E050" viewBox="0 0 1000 1000" overflow="inherit">
```
we will now have 
```xml
<symbol id="0000001310519730-E050" viewBox="0 0 1000 1000" overflow="inherit">
```
This avoids conflicts when several SVG examples are included in a web-page, particularly when different fonts are being used. It used not to work and the PR fixes that particular issue.

It can be merged unless we hear against it.

### DIFF
--- a/doc/verovio.conf
+++ b/doc/verovio.conf
@@ -28,13 +28,13 @@ DOXYFILE_ENCODING      = UTF-8
 # The PROJECT_NAME tag is a single word (or a sequence of words surrounded
 # by quotes) that should identify the project.
 
-PROJECT_NAME           = 
+PROJECT_NAME           = Verovio
 
 # The PROJECT_NUMBER tag can be used to enter a project or revision number.
 # This could be handy for archiving the generated documentation or
 # if some version control system is used.
 
-PROJECT_NUMBER         = develop
+PROJECT_NUMBER         = 
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description for a project that appears at the top of each page and should give viewer a quick idea about the purpose of the project. Keep the description short.
 
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -548,6 +548,8 @@ public:
 
     static void SeedUuid(unsigned int seed = 0);
 
+    static std::string GenerateRandUuid();
+
     static bool sortByUlx(Object *a, Object *b);
 
     //----------//
--- a/include/vrv/svgdevicecontext.h
+++ b/include/vrv/svgdevicecontext.h
@@ -203,7 +203,7 @@ public:
      * Set the SVG to have 'raw' formatting, with no extraneous whitespace or newlines.
      */
     void SetFormatRaw(bool rawFormat) { m_formatRaw = rawFormat; }
-    
+
     /**
      * Removes the xlink: prefex on href attributes, necessary for some newer browsers.
      */
@@ -288,6 +288,8 @@ private:
     bool m_removeXlink;
     // indentation value (-1 for tabs)
     int m_indent;
+    // prefix to be added to font glyphs
+    std::string m_glyphPrefix;
 };
 
 } // namespace vrv
--- a/include/vrv/svgdevicecontext.h
+++ b/include/vrv/svgdevicecontext.h
@@ -289,7 +289,7 @@ private:
     // indentation value (-1 for tabs)
     int m_indent;
     // prefix to be added to font glyphs
-    std::string m_glyphPrefix;
+    std::string m_glyphPostfixId;
 };
 
 } // namespace vrv
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -584,12 +584,7 @@ bool Object::DeleteChild(Object *child)
 
 void Object::GenerateUuid()
 {
-    int nr = std::rand();
-    char str[17];
-    // I do not want to use a stream for doing this!
-    snprintf(str, 17, "%016d", nr);
-
-    m_uuid = m_classid + std::string(str);
+    m_uuid = m_classid + Object::GenerateRandUuid();
 }
 
 void Object::ResetUuid()
@@ -919,6 +914,16 @@ void Object::SeedUuid(unsigned int seed)
     else {
         std::srand(seed);
     }
+}
+
+std::string Object::GenerateRandUuid()
+{
+    int nr = std::rand();
+    char str[17];
+    // I do not want to use a stream for doing this!
+    snprintf(str, 17, "%016d", nr);
+
+    return std::string(str);
 }
 
 bool Object::sortByUlx(Object *a, Object *b)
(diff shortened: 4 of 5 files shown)


lpugin:
@ahankinson that should solve the issue with the incipits

ahankinson:
OK! I'll check it out.

craigsapp:
Is there a reason for the unique number to be before the glyph number?  For other automatically generated IDs, the number followed the MEI-originating element name, such as:

```
id="keysig-0000000383861635"
```

-----

Is each glyph within a single SVG given the same or different random value?  Since they should otherwise be unique within the same SVG, then it would be useful to use the same random value for all glyphs in one SVG image.

An interesting idea would be to allow for a glyph prefix option to verovio when generating an SVG image of a page.   There would also have to be a page-prefix added automatically in such cases.  I do something similar for the note glyphs on this page:

http://www.ccarh.org/beethoven/op131/mvmt4

Where the original glyph IDs are P01, P02, etc., but annoyingly SCORE with not use the same ID for each glyph number for each output, so I prefix the IDs with the filename to ensure that they are unique between images (on the above page, each system is a separate SVG image).

--------

Related to issue https://github.com/rism-digital/verovio/issues/858

@lpugin said:
> This is due to conflicting ids for font glyphs in the SVG. (It is easily fixable by prefixing ids with a font identifier.)
 
If you prefix/postfix to SMuFL glyph number with the font name, this would also prevent the problem without the need for random numbers.  There could be multiple glyphs with the same ID on the same webpage, but they would all have the same content -- so it would not matter which one is being used by the SVG renderer since they are all the same.

Also note in that issue that using `` prevents the problem of cross-contamination between glyph ids in different images.   If you are creating dynamic SVG images within the webpage, then the svg content can be mime-encoded into `img@src`:

```javascript
var svgimage = "<img src='data:image/svg+xml;base64," + btoa(svgdata) + "'>";
```

-------------------

Another interesting possibility would be to have an option to not embed any music glyphs into the SVG images, and then have a separate image (or multiple images) that stores only the glyphs.   This would reduce the size of the individual incipits.

ahankinson:
> Another interesting possibility would be to have an option to not embed any music glyphs into the SVG images, and then have a separate image (or multiple images) that stores only the glyphs. This would reduce the size of the individual incipits.

We thought about this. The challenge is twofold:

 1. Getting Verovio to keep track of all the glyphs rendered in a given toolkit instance, so that it can output the image with those glyphs present, and 
 2. Figuring out how to stream the output such that you can embed that first image before all other incipits. 

We decided that the savings in bandwidth would be minimal, since glyph information is highly compressible and most web servers should have gzip turned on, so it's not particularly worth the added complexity.

ahankinson:
I'm not sure what your other comments mean?

> Is each glyph within a single SVG given the same or different random value? Since they should otherwise be unique within the same SVG, then it would be useful to use the same random value for all glyphs in one SVG image.

The value of any `id` attribute should be unique to the whole page, so if you're declaring an `id` value it should be a different random value from anything else. If you embed SVG in HTML, as we're doing, then no two elements should have the same ID value. (It doesn't actually matter if their graphical content is the same). 

If you the `` tag then you can't manipulate the SVG with JavaScript, and you can't apply a stylesheet external to your SVG. See this page: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Adding_vector_graphics_to_the_Web#adding_svg_to_your_pages

lpugin:
OK, I changed it to a postfix for consistency - even though they are not exactly the same thing.

To the questions:
* Yes, it is one ID per page, so all glyphs in one SVG page use the same postfix ID
* We could add an option to specify the ID, but I am not sure what that would bring. I don't see the extra bytes included as a problem.
* Yes, it used to work even with repeated definitions of the same glyph in one web page, but this is certainly not a good practice. They are cases where this is problematic. `` is a workaround and does yield other issues.

craigsapp:
This spec says that if there is more than one element with the same ID in the DOM, then the first one in tree order should be returned by `getElementById` in javascript (and is current in all html rendering engines):

https://dom.spec.whatwg.org/#interface-nonelementparentnode

Screen Shot 2021-03-18 at 9 20 53 AM

So if each glyph is prefixed/postfixed with the font name, it will not matter which one of the duplicate IDs are accessed to get the outline, although it will be the first one in the DOM in any case.


Here is a test of Chrome, which is treating IDs like classes when using `querySelectorAll`:

```html
<html>
<head>
<title>Duplicate IDs</title>
</head>
<body>

<p> A1: <span id="a"></span> </p>
<p> A2: <span id="a"></span> </p>
<p id="b1"> B1: <span id="b"></span> </p>
<p id="b2"> B2: <span id="b"></span> </p>
<p> C: <span id="c"></span> </p>

<script>
let spana  = document.querySelectorAll("#a");
let spanb1 = document.querySelectorAll("#b1 #b");
let spanb2 = document.querySelectorAll("#b2 #b");
let spanc  = document.querySelectorAll("#c");
for (const i of spana)  i.innerHTML = "a";
for (const i of spanb1) i.innerHTML = "b1.b";
for (const i of spanb2) i.innerHTML = "b2.b";
for (const i of spanc)  i.innerHTML = "c";
</script>
</body>
</html>
```

Produces this output:

Screen Shot 2021-03-18 at 9 36 23 AM

ahankinson:
You're linking to the (non-normative) DOM API spec; the (normative) HTML spec clearly says duplicate IDs in the tree is an error. The (normative) [SVG spec also says this is an error](https://www.w3.org/TR/SVG2/struct.html#Core.attrib). 

The W3C HTML validator says this your example has errors: Duplicate IDs:

image

We shouldn't produce bad SVG just because we can get away with it.

craigsapp:
To ensure unique IDs for glyphs that produce valid HTML, this PR is the only practical way.

One thought:  If you are creating multiple short incipits, these will takes less than one second to produce.  The random number generator is seeded with the time in seconds since 1 Jan 1970 (as an integer value without millisecond timings).   If so, then the random value used to distinguish between glyphs will only be updated to a new unique value if you restart verovio within a different second.   If you are running the verovio Toolkit in Javascript this should not be a problem because the toolkit is persistent, but on the command-line verovio is exited after each SVG is produced, so there would likely be problems with the glyph ids using server-side generation.

Second thought: If you do not have direct control the MEI/SVG source content or use [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier), then it would be possible for xml:ids between the different SVG images to collide.  Adding an option to include the glyph ID postfix on all IDs produced in the SVG image would be interesting.  This ID postfix should not be added by default in any case, since it would be a bother to otherwise remove the ID postfix to refer back to the source MEI xml:id.


A less practical method without this PR would be to insert the SVG images in `` tags.   Then iterate through the images one at a time, pulling them out of the img subtree into the main DOM tree and adding unique markers to the glyph IDs in the image one at a time.

ahankinson:
We're using the Python toolkit for generating the SVG server-side, and it's instantiated just once when the application instance starts. So I don't think there will be a problem with the RNG (there *may* be problems with memory usage, but I'm keeping an eye on that).

Issue #2106, opened 2021-03-19
SplitMultival crash


rettinghaus:
Assertion fails when artic itself has multiple values.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Articulation crash</title>
         </titleStmt>
         <pubStmt>
            <date isodate="2021-03-19">2021-03-19</date>
         </pubStmt>
      </fileDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" clef.shape="F" clef.line="4" />
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <chord dur="2">
                              <artic artic="stacc acc" />
                              <note oct="4" pname="c" accid.ges="n" />
                              <note oct="3" pname="b" accid="f" />
                              <note oct="3" pname="g" accid.ges="n" />
                              <note oct="3" pname="e" accid="n" />
                           </chord>
                           <rest dur="2" />
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

rettinghaus:
@lpugin this one's still crashing: 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0" ?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron" ?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Articulation crash</title>
         </titleStmt>
[...]

PR #2123, opened 2021-03-31
fix wrong parent issue for artics

rettinghaus:
proposed fix for #2106


lpugin:
Fix here https://github.com/rism-digital/verovio/commit/793d66a6165300169ca2714afae64330b8b90c6a (about the same). Thanks

Issue #920, opened 2018-07-22
Slur and stem direction interaction with chords


craigsapp:
In this example:

<img width="402" alt="screen shot 2018-07-22 at 12 37 15 pm" src="https://user-images.githubusercontent.com/3487289/43044708-ff2060d4-8dab-11e8-9acd-9dffd4a560f4.png">

there are two slurs between each pair of chords with no explicit `@place` information given.  For the first pair, both are visible.  For the second pair, they both overlap above the chord so only one slur can be seen.

MEI test data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2018-07-22T12:41:09" version="2.0.0-dev-d3c9630">
                    <name>Verovio</name>
                    <p>Transcoded from Humdrum</p>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title />
                </titleStmt>
            </work>
        </workDesc>
    </meiHead>
    <music>
        <body>
            <mdiv xml:id="mdiv-0000002002503144">
                <score xml:id="score-0000001060601612">
                    <scoreDef xml:id="scoredef-0000000610982995">
                        <staffGrp xml:id="staffgrp-0000000158075223">
                            <staffDef xml:id="staffdef-0000000039166534" clef.shape="G" clef.line="2" n="1" lines="5">
                                <label xml:id="label-0000002044673630" />
                            </staffDef>
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-L1F1">
                        <measure xml:id="measure-L1" right="invis" n="0">
                            <staff xml:id="staff-0000001616466207" n="1">
                                <layer xml:id="layer-L1F1N1" n="1">
                                    <chord xml:id="chord-L2F1" dur="4">
[...]

PR #2129, opened 2021-04-09
Connect slurs to note heads instead of chords in MusicXML import

rettinghaus:


lpugin:
Looks OK for me to attach slurs to notes. However, it is expected to yield a different rendering output - I have to check the implementation status of this. See https://github.com/rism-digital/verovio/issues/920#issuecomment-763618773
Will this what you expect?

rettinghaus:
Most of the cases should look the same, if they are encoded correctly. But when you have chords with three notes and slurs there is no way to connect the slurs to chords in a way it would make sense.

lpugin:
> Most of the cases should look the same, if they are encoded correctly. 

Sorry, I don't understand. The issue I referred to is showing that attaching slurs to notes and attaching slurs to chords should not look the same. For example, with a slur attached to the notes (one slur) you would expect this
image

With a slur attached to the chord this
image

> But when you have chords with three notes and slurs there is no way to connect the slurs to chords in a way it would make sense.

If a slur is connected to the chord, the number of notes it contains does not make any difference. So I don't understand this either.

rettinghaus:
Ok, what I meant is if there are chords with more than two noteheads and every notehead has a slur. 
However, slurs in MusicXML are _always_ connected to noteheads and there is no way to figure out if actually the chord it belongs to is meant. No word whatsoever in the documentation. So for now I think this would be the better implementation.

PR #2130, opened 2021-04-12
Improve mRest position for multiple layers

Monceber:
Improvement for the positioning of mRest, similar to how ordinary rests are positioned. mRest should be moved up/down, depending on whether it's in top/bottom layer.
Before the fix:
image
After the fix:
image

### DIFF
--- a/src/mrest.cpp
+++ b/src/mrest.cpp
@@ -86,8 +86,10 @@ int MRest::ResetHorizontalAlignment(FunctorParams *functorParams)
 
 int MRest::GetOptimalLayerLocation(Staff* staff, Layer* layer, int defaultLocation)
 {
-    Staff *parentStaff = vrv_cast<Staff *>(GetFirstAncestor(STAFF));
     if (!layer) return defaultLocation;
+    Staff *parentStaff = vrv_cast<Staff *>(GetFirstAncestor(STAFF));
+    assert(parentStaff);
+    
     // handle rest positioning for 2 layers. 3 layers and more are much more complex to solve
     if (parentStaff->GetChildCount(LAYER) != 2) return defaultLocation;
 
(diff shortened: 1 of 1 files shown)


src/mrest.cpp
```diff
@@ -79,4 +84,55 @@ int MRest::ResetHorizontalAlignment(FunctorParams *functorParams)
     return FUNCTOR_CONTINUE;
 }
 
+int MRest::GetOptimalLayerLocation(Staff* staff, Layer* layer, int defaultLocation)
+{
+    Staff *parentStaff = vrv_cast<Staff *>(GetFirstAncestor(STAFF));
+    if (!layer) return defaultLocation;
+    // handle rest positioning for 2 layers. 3 layers and more are much more complex to solve
+    if (parentStaff->GetChildCount(LAYER) != 2) return defaultLocation;
```
lpugin, line 92: Line 90 should come first. Also also `assert(parentStaff);`

src/mrest.cpp
```diff
@@ -79,4 +84,55 @@ int MRest::ResetHorizontalAlignment(FunctorParams *functorParams)
...
+
+    ListOfObjects layers;
+    ClassIdComparison matchType(LAYER);
+    parentStaff->FindAllDescendantByComparison(&layers, &matchType);
+    const bool isTopLayer = (vrv_cast<Layer *>(*layers.begin())->GetN() == layer->GetN());
+
+    auto otherLayerIter = isTopLayer ? std::prev(layers.end()) : layers.begin();
+    auto collidingElementsList = vrv_cast<Layer *>(*otherLayerIter)->GetLayerElementsForTimeSpanOf(this);
```
lpugin, line 100: Please do not use `auto` in such cases because it makes the code harder to read.

lpugin (commented):
Otherwise looks great!

PR #2133, opened 2021-04-13
Custos accidentals as attributes. Custos accidentals rendered

davidrizo:


rettinghaus (approved):

src/iomei.cpp
```diff
@@ -5031,6 +5031,20 @@ bool MEIInput::ReadCustos(Object *parent, pugi::xml_node custos)
...
+    AttAccidentalGestural accidentalGestural;
+    accidentalGestural.ReadAccidentalGestural(custos);
+    if (accidental.HasAccid() || accidentalGestural.HasAccidGes()) {
+        Accid *vrvAccid = new Accid();
+        vrvAccid->IsAttribute(true);
+        vrvAccid->SetAccid(accidental.GetAccid());
+        vrvAccid->SetAccidGes(accidentalGestural.GetAccidGes());
+        vrvCustos->AddChild(vrvAccid);
+    }
+    // -- end copy
```
lpugin, line 5046: Could you make a common `MEIInput::ReadAccidAttr(pugi::xml_node node, Object *object)` to be place there https://github.com/rism-digital/verovio/blob/7a392176fc9d49b97c0c6fac555113e7ffc0311e/include/vrv/iomei.h#L692-L701 ? Thanks

src/view_element.cpp
```diff
@@ -718,6 +718,9 @@ void View::DrawCustos(DeviceContext *dc, LayerElement *element, Layer *layer, St
     }
 
     dc->EndGraphic(element, this);
+
+    /************ Draw children (accidentals, etc) ************/
+    DrawLayerChildren(dc, custos, layer, staff, measure);
```
lpugin, line 723: Drawing the children should be done before ending the graphic. Otherwise the SVG tree will not match the MEI one

lpugin (changes requested):
Thanks! A few changes requested

src/view_element.cpp
davidrizo: Thanks for the comments. I've added those changes

src/iomei.cpp
```diff
@@ -5346,7 +5348,8 @@ bool MEIInput::ReadNote(Object *parent, pugi::xml_node note)
         vrvAccid->SetAccid(accidental.GetAccid());
         vrvAccid->SetAccidGes(accidentalGestural.GetAccidGes());
         vrvNote->AddChild(vrvAccid);
-    }
+    }*/
```
lpugin, line 5351: You can remove this

lpugin (changes requested):
Don't forget to run clang-format https://github.com/rism-digital/verovio/blob/master/doc/guidelines.md

src/iomei.cpp
davidrizo: Done

davidrizo:
Requested changes applied

PR #2137, opened 2021-04-14
Fix for crash with multivalued artics

Monceber:
Fix for another crash in #2106

Crash was happening due to adding artic to note/chord, while we were processing their children in Proces():
https://github.com/rism-digital/verovio/blob/d6dadf7f280f5b5207f9aef8cb353de8d8c411ce/src/object.cpp#L853-L858
This led to iterator invalidation and caused a crash. I've moved ConvertMarkupArticEnd() to the Layer class, so that we can process artics after we've gone through all notes/chords in the layer and can safely process them.

Renders just fine:
image


lpugin:
Thanks

PR #2138, opened 2021-04-15
Improve vertical rest position

Monceber:
A bit more of vertical rest position improvements:
1. Fix for rests overlapping with beams, should be working better now
image

2. Added limit on how high/low rest can be positioned, if elements from the same layer are taken as indicator. This should prevent rests from hanging above/below staves, which would make them look odd. Also added a change to make sure that top layer rests don't go below the center of the staff and vice versa for bottom layer rests. Like following:
image

3. Fixed this part of code from one of my previous PRs
`currentElementInfo.first = isTopLayer ? 4 : -4;`
The value has been invalid and caused some weird placement for top layer rests, so that should be gone now.


src/rest.cpp
```diff
@@ -370,10 +376,18 @@ int Rest::GetLocationRelativeToCurrentLayer(Staff *currentStaff, Layer *currentL
         ? GetElementLocation(nextElement, currentLayer, !isTopLayer).first
         : GetFirstRelativeElementLocation(currentStaff, currentLayer, false, isTopLayer);
 
-    if (VRV_UNSET == previousElementLoc) return nextElementLoc;
-    if (VRV_UNSET == nextElementLoc) return previousElementLoc;
+    int currentOptimalLocation = 0;
+    if (VRV_UNSET == previousElementLoc) currentOptimalLocation = nextElementLoc;
```
lpugin, line 380: Add {}

lpugin:
Thanks!

PR #2139, opened 2021-04-15
Initial implementation of French tablature

paul-bayleaf:


src/view_tab.cpp
```diff
@@ -139,7 +141,7 @@ void View::DrawTabRhythm(DeviceContext *dc, LayerElement *element, Layer *layer,
     int x = element->GetDrawingX();
     int y = element->GetDrawingY();
     y += m_doc->GetDrawingUnit(staff->m_drawingStaffSize) * 1.5;
-    int drawingDur = tabGrp->GetActualDur();
+    int drawingDur = (tabGrp->GetDurGes() != DURATION_NONE) ? tabGrp->GetActualDurGes() : tabGrp->GetActualDur();
```
lpugin, line 144: This does not make sense to me since we should not have to take `@dur.ges` into account for the drawing of the note. Anything particular you are trying to achieve?

paul-bayleaf: All the examples in <https://github.com/music-encoding/tablature-ig/tree/master/examples/lute> only use @dur.ges for the duration.  Whereas Tim's example judenkuenig-1523_2-elslein_liebes.xml uses @dur.   This code was an attempt to tolerate either @dur.ges or @dur or both.  Similar code appears in durationinterface.cpp line 69.   But if dur.ges does not affect the rhythm sign drawn, and @dur is always present then this change can be reverted and the examples fixed.   I'm still not sure I understand @dur.ges.

lpugin: Thanks for the clarification. The encodings are wrong. @reinierdevalk could you fix them?

lpugin: Please ignore my previous comment. Looking at this https://github.com/music-encoding/tablature-ig/blob/master/documentation/tablature-core.md it seems that we do expect `@dur.ges` on tabGrp and not `@dur`.

PR #2157, opened 2021-04-28
[WIP] get rid of shell script

rettinghaus:
As it is not best practice to start shell scripts from Python I replaced this in `setup.py`.  
@musicEnfanthen please have a look.
If we don't need `get_version.sh` for something else, I would remove it completely then.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,9 +8,9 @@ from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext as _build_ext
 from setuptools.command.sdist import sdist as _sdist
 from glob import glob
-import platform
+import git
 import os
-import subprocess
+import platform
 
 
 def get_commit():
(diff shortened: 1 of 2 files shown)


lpugin:
Have you tested it? This was added for versioning python wheels.

If you really want to get rid of shell scripts there is still be `get_git_commit.sh` that is used in many places but for which I could not find another way to do it.

rettinghaus:
Yes, I tested it, but to be sure please try yourself. That’s why I also asked Stefan. 

I‘ll have a look at the other script later.

lpugin:
It is difficult to test because it is triggered with GH actions and releases. I appreciate the effort, but I am not sure I understand the issue it fixes.

setup.py
```diff
@@ -55,15 +55,25 @@ def get_version():
...
+                        defines[definition[0]] = definition[1]
+                    except IndexError:
+                        pass
+            version = '.'.join(
+                (defines['VERSION_MAJOR'], defines['VERSION_MINOR'], defines['VERSION_REVISION']))
+            if defines['VERSION_DEV'] == 'true':
+                version += '.dev'
     if version.endswith(".dev"):
-        init_sha = subprocess.getoutput(
-            "git log -n 1 --pretty=format:%H -- bindings/python/.pypi-version")
```
lpugin, line 63: The `.devXXX` version number in the python wheels is currently based on the changes made since this file was changed for the last time. This means we can reset it when we make a release. Is this feature preserved?

ahankinson:
Using shell scripts in Python is generally fine; I’m not sure why you wouldn’t in this case.

setup.py
rettinghaus: Here we count commits in comparison to master. So as soon there is a release (from master branch) it is reset. [GitPython](https://gitpython.readthedocs.io/) makes this very easy.

lpugin: That does not work. If for some reason there is a commit to master (e.g., a change to the README file), it will reset it and that can create conflict on PyPi.

ahankinson:
GitPython itself seems to [call out to a subprocess](https://github.com/gitpython-developers/GitPython/blob/main/git/cmd.py#L726-L741). Isn't this just adding an additional third-party dependency to do the same thing?

musicEnfanthen:
Since you asked me at the beginning of this thread to have a look:

I am tending to agree with @ahankinson 's hesitations. From what I understand, the intention to introduce `subprocess` was indeed to provide a more reliable way to do exactly that: run subprocesses, shell scripts etc. (cf. https://docs.python.org/3/library/subprocess.html) 

The script as it was, is working, does not provoke any issues, and is much shorter and therefore probably easier to maintain (just to list the advantages). But maybe I am just not seeing the issue, too. @rettinghaus could you please provide more context about the underlying issue. Is there an official recommendation to not call bash scripts from Python scripts?

If there's really the need to change something in the original script, one could think of replacing the old legacy invocation functions of `subprocess` (like `getoutput`) with somewhat more dated approaches (cf. https://docs.python.org/3/library/subprocess.html#legacy-shell-invocation-functions)

lpugin:
OK, let's leave it as it is for now. Please open an issue first if it does need to be changed.

PR #2158, opened 2021-04-28
Develop format raw option

lpugin:


craigsapp:
What is Toolkit::SaveFile()?  I.e., this is a C++ function?  and if so, then how does this relate to the command-line and javascript toolkit interfaces?

lpugin:
> this is a C++ function? 

Yes. It is calling Toolkit::GetMEI and save the MEI data to a file. It is not available in the JS toolkit but is for the Python one

PR #2162, opened 2021-04-29
Use regex to detect XML file type.

JRegimbal:
This uses regular expressions to check for MEI and MusicXML inputs when not specified. It should have the same functionality as before, except it also supports checking if the opening tags include a line break (e.g., `<mei\n`). Checks are collapsed into two regular expressions, one for MEI and one for MusicXML.

### DIFF
--- a/src/toolkit.cpp
+++ b/src/toolkit.cpp
@@ -10,6 +10,7 @@
 //----------------------------------------------------------------------------
 
 #include <assert.h>
+#include <regex>
 
 //----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


src/toolkit.cpp
```diff
@@ -224,53 +225,13 @@ FileFormat Toolkit::IdentifyInputFrom(const std::string &data)
         // <score-timewise> == root node for time-wise organization of MusicXML data
         // <opus> == root node for multi-movement/work organization of MusicXML data
 
-        if (initial.find("<mei ") != std::string::npos) {
+        if (std::regex_search(initial, std::regex("<(mei|music|pages)[\\s\\n>]"))) {
```
ahankinson, line 228: Using a raw string means you don't have to double-escape the `\s` and `\n`:

`R"(<(mei|music|pages)[\s\n>])"`

craigsapp: I used to use raw strings in verovio code, but the clang-format program mangled them: it placed a space between the `R` and `"`.  So every time I ran clang-format, I had to go back and remove the space, since I could not find any option to suppress that behavior (and clang-format does not or did not understand raw string).

ahankinson: Oh, good to know!

ahankinson: Looks like you might be able to control this behaviour setting the LanguageStandard behaviour to C++11:

http://clang-developers.42468.n3.nabble.com/clang-format-raw-string-literals-td4042533.html

`LanguageStandard: c++11`

lpugin: I think you can do
```cpp
// clang-format off
do stuff
// clang-format on

Comment (ahankinson) [reply src/toolkit.cpp:228]:
```
// clang-format off
#include "vrv.h"
...

} // namespace vrv
// clang-format on
```

:-D

**Comment by craigsapp** [comment]
That works, although `LanguageStandard` seems to be a data type rather than a option name.  The option that works is:

```
Standard: c++11
```

I first added it to the top of the file and it did not work.  Then I found the option `Standard: Cpp03` further down in the file, at line 52.  After removing that line, raw strings started to work.

I will let @lpugin make the change to `verovio/.clang-format`, since he can consider any implications of changes to the formatting due to the new standard (but it seems like it needs to be done). Another change will be to not place a space between template endings in cases like this:

```
std::vector<std::vector<int> > tester;     // pre C++11 
std::vector<std::vector<int>> tester;      // C++11 allowed syntax (<C++11 does not understand)
```

Here is what is necessary to update `.clang-format`:

(1) Remove line 52 (or approximately that line if the file has changed recently): `Standard: Cpp03`
(2) Place `Standard: c++11` or `Standard: Cpp11` after the `Language: Cpp` option at the start of the file:

```
Language: Cpp
Standard: Cpp11
```

File: src/toolkit.cpp (hunk 1 above)
Comment (craigsapp) [reply src/toolkit.cpp:228]: Another consideration for the regular expression is that `\n` is one of the characters represented by `\s`, so the `\n` is redundant, and the regular expression can be simplified to:

```
"<(mei|music|pages)[\\s>]"
```

craigsapp: :-P.  See the new issue I am making in case my note below on this page gets lost.

rettinghaus:
The wiki build instructions have to be updated, because now regex is used outside PAE importer:
> For building it without Plain and Easy support (without regex.h)

lpugin:
> The wiki build instructions have to be updated, because now regex is used outside PAE importer:

Right. Updated in the upcoming documentation: https://github.com/rism-digital/verovio-reference-book/blob/master/_book/06-installing-or-building-from-sources/01-command-line.md (the wiki page will be removed)

Issue #1706, opened 2020-09-29
engravingDefaults not working in toolkit


rettinghaus:
The newly added support for `engravingDefaults` is not working in the toolkit. 
I don't know how to fix this, and the toolkit documentation is still under development.

lpugin:
@rettinghaus is this issue different than #1708 ? Please elaborate a bit if so. Otherwise can close it.

rettinghaus:
Yes, it is different. Passing `engravingDefaults: "non-existent-file"` does not through an error. Even if you pass a correct file, the function `Options::Sync` isn't called. 

And wouldn't it make sense for the toolkit to pass complete engravingDefaults JSON within the JSON object containing the options?

lpugin:
How are you passing the file? My understanding is that, as it stands, it cannot work. 

The way I think you can make it work would be to:
* make `OptionJson::SetValue` taking a JSON string as parameter and not a filename, and
* modify `main.cpp` to have it reading the file and passing its content instead of passing the filename - with a check on the option type similar to https://github.com/rism-ch/verovio/blob/375ee1e6a6e8765cf22467316a50741197259db9/tools/main.cpp#L280-L283.
For the JS toolkit, I think then JSON within JSON would work.

The problem I can see with this is that it would then not work with a file with other bindings (e.g., the Python one). So other people will complain, and I agree it would be confusing. So to make it clear, we could have a --engraving-defaults-file but only for the C++ toolkit, which would work as described above, and then have --engraving-defaults only supporting JSON with JSON for other bindings. (It should actually also work for the C++ toolkit)

rettinghaus:
This is also now implemented properly.

PR #2171, opened 2021-05-03
Engraving defaults

DavidBauer1984:
This PR addresses #1706 . In particular, the option `engravingDefaults` can now be used to pass options as JSON-string, while the new option `engravingDefaultsFile` is used to read JSON from a file.


src/options.cpp
```diff
@@ -39,8 +39,8 @@ std::map<int, std::string> Option::s_systemDivider = { { SYSTEMDIVIDER_none, "no
 constexpr const char *engravingDefaults
     = "{'engravingDefaults':{'thinBarlineThickness':0.15,'lyricLineThickness':0.125,"
       "'slurMidpointThickness':0.3,'staffLineThickness':0.075,'stemThickness':0.1,'tieMidpointThickness':0.25,"
-      "'hairpinThickness':0.1,'thickBarlineThickness':0.5,'tupletBracketThickness':0.1,'subBracketThickness':0.5,"
-      "'bracketThickness':0.5,'repeatEndingLineThickness':0.15, 'textEnclosureThickness': 0.2}}";
+      "'hairpinThickness':0.1,'thickBarlineThickness':0.5,'tupletBracketThickness':0.1,'subBracketThickness':0.1,"
```
ahankinson, line 42: Why are these values changed?

DavidBauer1984: The values should reflect default values of the corresponding options. Let's take an example. `thinBarlineThickness` is used to set the option `barLineWidth`. This happens in `Options::Sync`. The default value of `barLineWidth` is 0.3. There is always a conversion factor of 2 to MEI units involved. So `thinBarlineThickness` obtains the value 0.15. This scheme works fine for almost all entries in the list. It fails for `subBracketThickness`. This is used to set the option `subBracketThickness` which has default 0.2. If the old list value of 0.5 was used, then the option would be set to 1.0 (conversion factor). So I changed it to 0.1 in the list such that it works out. Similar for `repeatEndingLineThickness` and `textEnclosureThickness`.

lpugin:
Please resolve the conflicts. Thanks

Issue #1928, opened 2021-01-06
MIDI for performance/data


craigsapp:
It would be useful to have two MIDI output types from verovio: (1) MIDI for listening purposes, and (2) MIDI for data processing, such as computer analysis.

The listening form should be the default and labeled `midi` in the output options (`--to midi`).  The data form should be labeled `midi-data` in the output options (`--to midi-data`).  Currently there is one MIDI output format for verovio, which is basically equivalent to the new `midi-data` output format.

`midi` output should have realization of articulations (staccato, accents &ndash; including sforzandi) and ornaments as well as dynamics.  Ornaments need to be context-dependent on the tempo, for example an accacciatura could be played `x` milliseconds before the note which follows it (maybe making `x` an input option to verovio, such as `--ornament-duration-ms`).  Appoggiatura grace notes would be played at the time of the following notes, and the following note's start time would be delayed by the `@dur` of the preceding appoggiatura grace note.  Similarly for mordents (which will be the easiest ornament to implement, the first two notes could have a duration of `x` milliseconds, with the last one held for the rest of the duration of the non-ornamented note).  Long mordents would behave similarly but the first four notes would be short, and the 5th note would fill out the rest of the original notes duration.  Trills would be more complicated to implement, but a basic method would be to use the same `x` millisecond not between notes in the trill, adjusted so that the last note ends on the original note's pitch.  Grace notes before or after the trill note would alter the starting and ending notes of the trill (a tied/slurred note before the trill note means to start the trill on the primary note and not the auxiliary note).  Turns would place the ornamental notes immediately at the start of the note if placed above the notehead, or delayed to the end of the note if placed between notes.

`midi-data` would not realize ornaments or articulations that modify the duration of the note, such as staccatos, but this form of the MIDI output could implement dynamics an accent-based articulations.  In addition, MIDI Plus encoding of `@accid.ges` and slurring information could be used to preserve that information which is otherwise lost in a MIDI rendering of notation.  See [Beyond MIDI: A Handbook of Musical Codes](http://beyondmidi.ccarh.org/beyondmidi-600dpi.pdf), chapter 5 for the specification (which I will summarize in this issue for developing the algorithms).

craigsapp:
MidiPlus implementation notes:

[...]

PR #2172, opened 2021-05-03
Bowed tremolos for MIDI

DavidBauer1984:
This PR implements bowed tremolos for MIDI output.

The param object of the `generateMIDI` functor now stores a map of notes being replaced by note sequences. Hopefully, this will be useful, if someday we decide that we also want trills or fingered tremolandi in MIDI.


craigsapp:
See also issue https://github.com/rism-digital/verovio/issues/1928

Issue #2191, opened 2021-05-17
Reduce number of programming languages 


rettinghaus:
Around Verovio there are combinations of many different languages like JavaScript, Python, XSLT, and Shell scripts. This should be reduced in order to

1. improve maintainability 
2. improve cross-platform compatibility 
3. reduce dependencies
4. speed things up

lpugin:
Every programming language has specificities that quite likely were a reason why it was chosen. We can certainly changes things but if nothing is problematic and works, I generally prefer not to - at least I do not see this as a top priority. Please list specific things you would like to change and why.

rettinghaus:
- [ ] remove `get_version.sh`
- [ ] improve font generation

1. `get_version.sh` is slow and shows problems (e.g. [UUOC](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat)). It would be faster and more reliable to do it directly within `setup.py`.
2. font generation depends on the correct version of `Saxon` and `phantomjs`. Apart from `phantomjs` being outdated, it made problems before on different platforms.

ahankinson:
```
$ time ./get_version.sh                                                                                                                                                                                                                   develop*
3.5.0.dev
./get_version.sh  0.02s user 0.04s system 162% cpu 0.032 total
```

"slow" :-D

lpugin:
```bash
laurent@laika tools % time ./get_version.sh 
3.5.0.dev
./get_version.sh  0.01s user 0.04s system 109% cpu 0.048 total
```
It is not slow for me but if you want to remove it that is fine.

Font generation can be updated. What do you want to used instead of phantomjs?

rettinghaus:
I think the complete glyph extraction should be done with Python (i.e. no XSLT or JS).

lpugin:
That's fine with me. It would be good to see in advance the dependencies you will rely on because if they are just wrappers (e.g., https://pypi.org/project/phantomjs/) that will be pretty much useless.

lpugin:
This has been done.

PR #2192, opened 2021-05-17
Replace shell script in python build

rettinghaus:
This PR replaces `get_version.sh` with a native Python approach.
It addresses one part of #2191.


setup.py
```diff
@@ -55,9 +55,21 @@ def get_version():
...
-        version = subprocess.getoutput("bash -c 'cd tools; ./get_version.sh'")
+    if os.path.exists("./include/vrv"):
+        with open("./include/vrv/vrvdef.h") as header_file:
+            defines = {}
+            for line in header_file.readlines():
+                if line.startswith("#define"):
+                    line.rstrip()
+                    definition = line[8:].split()
+                    try:
+                        defines[definition[0]] = definition[1]
```
ahankinson, line 66: This adds all the definitions to the dictionary, which seems unnecssary given that we're only looking for three or four?

setup.py
```diff
@@ -55,9 +55,21 @@ def get_version():
         for line in lines:
             if line.startswith('Version:'):
                 return line[8:].strip()
-    if os.path.exists("./tools"):
-        print("Running tools/get_version.sh")
-        version = subprocess.getoutput("bash -c 'cd tools; ./get_version.sh'")
+    if os.path.exists("./include/vrv"):
+        with open("./include/vrv/vrvdef.h") as header_file:
+            defines = {}
+            for line in header_file.readlines():
```
ahankinson, line 61: You don't actually need "`readlines()`" -- `for line in header_file` should work.

setup.py
```diff
@@ -55,9 +55,21 @@ def get_version():
...
+                if line.startswith("#define"):
+                    line.rstrip()
+                    definition = line[8:].split()
```
ahankinson, line 64: There's not really a whole lot of difference between splitting from `#defines` and not -- you can just do stuff with a 3-size list, rather than a 2-size list.

setup.py
```diff
@@ -160,4 +173,4 @@ def get_version():
           'Bug Reports': 'https://github.com/rism-digital/verovio/issues',
           'Source': 'https://github.com/rism-digital/verovio',
       },
-)
+      )
```
ahankinson, line 176: I think this indentation level is not correct? It goes with `setup`, doesn't it?

ahankinson (changes requested):

setup.py
```diff
@@ -55,9 +55,21 @@ def get_version():
         for line in lines:
             if line.startswith('Version:'):
                 return line[8:].strip()
-    if os.path.exists("./tools"):
-        print("Running tools/get_version.sh")
-        version = subprocess.getoutput("bash -c 'cd tools; ./get_version.sh'")
+    if os.path.exists("./include/vrv"):
```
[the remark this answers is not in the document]
rettinghaus: removed

setup.py
```diff
@@ -55,9 +55,21 @@ def get_version():
...
+        with open("./include/vrv/vrvdef.h") as header_file:
+            defines = {}
+            for line in header_file.readlines():
+                if line.startswith("#define"):
+                    line.rstrip()
+                    definition = line[8:].split()
+                    try:
+                        defines[definition[0]] = definition[1]
+                    except IndexError:
+                        pass
```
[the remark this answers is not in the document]
rettinghaus: checking length now

setup.py
rettinghaus: Now skipping when `vrv_cast` is reached. Maybe we want to use more defines from there later on.

setup.py
rettinghaus: Done.

setup.py
rettinghaus: This comes from autopep8, so I guess it's correct.

rettinghaus:
@ahankinson Thanks for the review.

setup.py
```diff
@@ -55,9 +55,23 @@ def get_version():
...
-    if os.path.exists("./tools"):
-        print("Running tools/get_version.sh")
-        version = subprocess.getoutput("bash -c 'cd tools; ./get_version.sh'")
+    with open("./include/vrv/vrvdef.h") as header_file:
+        defines = {}
+        for line in header_file:
+            if not line.startswith("#define"):
+                continue
+            definition = line.strip().split()
+            if len(definition) > 2:
```
ahankinson, line 64: The 'else' is redundant; you can `continue` if you **don't** meet the requirements, then the code will stop processing there.

```
if len(segments) < 3:  # if it's fewer than three, something went wrong.
    continue

defines[definition[1]] = definition[2]
```

setup.py
```diff
@@ -55,9 +55,23 @@ def get_version():
         for line in lines:
             if line.startswith('Version:'):
                 return line[8:].strip()
-    if os.path.exists("./tools"):
-        print("Running tools/get_version.sh")
-        version = subprocess.getoutput("bash -c 'cd tools; ./get_version.sh'")
+    with open("./include/vrv/vrvdef.h") as header_file:
+        defines = {}
+        for line in header_file:
+            if not line.startswith("#define"):
```
ahankinson, line 61: You can be very specific about what you're looking for if you use a simple regex, rather than parsing the whole file and /or shortcutting at a specific line:

```
# find a needle in a haystack. :)
import re
needle = re.compile(r"VERSION_MAJOR|VERSION_MINOR|VERSION_REVISION|VERSION_DEV")
...

    for line in header_file:
        if not needle.search(line):
            continue
...
```

This also means you don't have to tie it to a specific order -- if (for some reason) `vrv_cast` moves above the version declaration, then this will stop functioning. Since the `re` module is built-in to Python it doesn't add any new dependencies, and the speed is comparable.

ahankinson (approved):
It looks OK, so I'll approve it, but there are still a few things that can be improved.

PR #2193, opened 2021-05-17
Develop improve font scripts

DavidBauer1984:
This PR addresses #2191 and replaces the XSLT-scripts by Python scripts. This removes the dependency on `Saxon`. It only requires the python package `lxml` which is quite common. Moreover, it speeds up the total runtime of `generate_all` (from 26s to 5s on my machine).


lpugin:
Can you redo the PR without the files generated? These make it difficult to follow. Thanks!

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+      dict[glyph.attrib["glyph-code"]] = glyph.attrib["smufl-name"]
+   return dict
+
+########################
+#  Parse Fontname.svg  #
+########################
+
+# Read svg file which is passed as argument and retrieve information
+def read_svg_file():
+   font_file_name = os.sys.argv[1]
```
ahankinson, line 63: There is no documentation about what this argument should be, so this will unhelpfully fail if you pass in the wrong argument. Consider using the `argparse` module to provide some built-in help.

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+
+######################
+#  Helper Functions  #
+######################
+
+# Open file in same directory as script and retrieve its content
+def get_file_content(filepath):
+   __location__ = os.path.realpath(os.path.dirname(__file__))
+   file = open(os.path.join(__location__, filepath), 'r')
+   content = file.read()
```
ahankinson, line 17: Is there any advantage to having this as a separate function, rather than just regularly opening the file? With context managers you don't need to worry about forgetting to close it anymore:

```
with open(filename, 'r') as myfile:
     # do something with myfile
# file is closed here.
```

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+
+#########################
+#  Parse supported.xsl  #
+#########################
+
+# Retrieve dictionary mapping glyph code to smufl name
+def get_supported_glyph_codes():
+   xml = get_file_content('supported.xsl')
+   glyphs = get_elements(xml, 'glyph')
+   dict = {}
```
ahankinson, line 52: `dict` [shadows a built-in type for Python](https://docs.python.org/3/library/functions.html#func-dict) --

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+      write_file_content(file_path, xml)
+
+# (3) Output bounding box svg
+root = create_svg_root()
+root.set("font-family", font_family)
+fontface = etree.SubElement(root, "font-face")
+fontface.set("units-per-em", units_per_em)
+for glyph in glyphs:
+   name = glyph.attrib["glyph-name"]
+   id = name.split("uni")[-1]
```
ahankinson, line 112: `id` [shadows a built-in Python name](https://docs.python.org/3/library/functions.html#id)

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+      font_svg_content = get_file_content(font_file_name)
+   except EnvironmentError:
+      print("Error opening font file %s!" % font_file_name)
+      os.sys.exit(1)
+   root = etree.fromstring(bytes(font_svg_content, encoding='utf-8'))
+   font_faces = get_svg_elements(root, "font-face")
+   if (len(font_faces) != 1):
+      print("Error: the file %s should have a unique font-face element!" % font_file_name)
+      print("Please check that the svg has correct namespace: %s" % svg_ns)
+      os.sys.exit(1)
```
ahankinson, line 74: `sys.exit(1)`

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+      path = etree.SubElement(root, "path")
+      path.set("transform", "scale(1,-1)")
+      if "d" in glyph.attrib:
+         path.set("d", glyph.attrib["d"])
+      file_path = f"../data/{font_family}/{id}-{supported_glyph_codes[id]}.xml"
+      xml = etree.tostring(root, pretty_print=True, encoding='unicode')
+      write_file_content(file_path, xml)
+
+# (3) Output bounding box svg
+root = create_svg_root()
```
ahankinson, line 106: You seem to only create a root element once, and you set some attributes in the function, and some here. Probably better to just inline the root element creation here and get rid of the separate function.

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
...
+   name = glyph.attrib["glyph-name"]
+   id = name.split("uni")[-1]
+   if (id in supported_glyph_codes):
+      path = etree.SubElement(root, "path")
+      path.set("name", supported_glyph_codes[id])
+      path.set("id", id)
+      horiz_adv_x = ""
+      if "horiz-adv-x" in glyph.attrib:
+         horiz_adv_x = glyph.attrib["horiz-adv-x"]
+      path.set("horiz-adv-x", horiz_adv_x)
```
ahankinson, line 120: `horiz_adv_x = glyph.attrib['horiz-adv-x"] if 'horiz-adv-x' in glyph.attrib else ""`

fonts/generate-smufl-code.py
```diff
@@ -0,0 +1,74 @@
...
+def get_supported_glyph_codes():
+   xml = get_file_content('supported.xsl')
+   glyphs = get_elements(xml, 'glyph')
+   dict = {}
+   for glyph in glyphs:
+      dict[glyph.attrib["glyph-code"]] = glyph.attrib["smufl-name"]
+   return dict
+
+##################
+#  Main program  #
```
ahankinson, line 48: Surely this can be collapsed into the previous script, and configured with some sort of mode flag? All of the 'helper functions' are exactly the same.

ahankinson: Also, this should be in a `if __name__ == "__main__"` block

ahankinson (changes requested):

fonts/extract-glyphs.py
```diff
@@ -0,0 +1,124 @@
+#! /usr/bin/env python3
+
+import os
+from lxml import etree
```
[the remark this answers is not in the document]
rettinghaus: Switched.

fonts/generate-smufl-code.py
DavidBauer1984: I would prefer to keep the scripts separate as they have different purpose.

PR #2200, opened 2021-05-20
Develop additive meters

rettinghaus:


rettinghaus:
Changes have been made here on LibMEI. We have to adjust LibMEI first and remove this part here again.

src/metersig.cpp
```diff
@@ -42,6 +43,12 @@ void MeterSig::Reset()
     ResetMeterSigVis();
 }
 
+int MeterSig::GetTotalCount() const
+{
+    const auto &summands = this->GetCount();
```
lpugin, line 48: Don't use `auto`

lpugin:
There are some documentation for the LibMEI stuff here https://book.verovio.org/contributing/generate-code-with-libmei.html Speak up if anything is not clear

PR #2201, opened 2021-05-20
Fixes possible use of uninitialised ntrack variable

rettinghaus:
FYI @craigsapp


craigsapp:
Thanks!  Although that will only prevent compiler warnings, since ntrack will otherwise always be initialized if used:

```cpp
int ntrack;
if (token) {
     ntrack = something; //  (the track number of the token)
}
while (token != NULL) {
    // do something with ntrack
    // which must be initialized in the above if-statement
}
```

PR #2202, opened 2021-05-20
adapt DrawArpeg to latest SMuFL specifications

rettinghaus:
To work with Leipzig font it has to be generated upfront with the changes from https://github.com/rism-digital/leipzig/pull/4. After that downward arpeggios will work correctly with Bravura and Leland.


lpugin:
A test file would be good, wouldn't it?

PR #2215, opened 2021-05-27
proper checks for version and header

### LINKED ISSUE #2214: MEI header version interaction
If you remove the `meiHead` from the tree Verovio parses it as a MEI3 file regardless of what the `meiversion` attribute states.

```xml
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" clef.shape="C" clef.line="3" />
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure right="dbl" n="1">
                     <staff n="1">
                        <layer n="1">
                           <note xml:id="note-01" dur="4" oct="4" pname="c" stem.dir="down" />
                           <note xml:id="note-02" dur="4" oct="4" pname="c" stem.dir="up" />
                           <note xml:id="note-03" dur="4" oct="4" pname="c" stem.dir="down" />
                           <note xml:id="note-04" dur="4" oct="4" pname="c" stem.dir="up" />
                        </layer>
                     </staff>
                     <mordent staff="1" startid="#note-01" />
                     <mordent staff="1" startid="#note-02" form="upper" />
                     <mordent staff="1" startid="#note-03" form="upper" long="true" />
                     <mordent staff="1" startid="#note-04" place="below" form="upper" long="true" />
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

rettinghaus:
fixes #2214

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -2912,7 +2912,7 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
         }
     }
     else {
-        // default to MEI 4
+        // default to MEI 5
         LogWarning("No meiversion found, falling back to MEI5 (dev)");
         m_version = MEI_5_0_0_dev;
     }
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -2895,16 +2895,21 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
 
     if (root.attribute("meiversion")) {
         std::string version = std::string(root.attribute("meiversion").value());
-        if (version == "5.0.0-dev")
+        if (version == "5.0.0-dev") {
             m_version = MEI_5_0_0_dev;
-        else if (version == "4.0.1")
+        }
+        else if (version == "4.0.1") {
             m_version = MEI_4_0_1;
-        else if (version == "4.0.0")
+        }
+        else if (version == "4.0.0") {
             m_version = MEI_4_0_0;
-        else if (version == "3.0.0")
+        }
+        else if (version == "3.0.0") {
             m_version = MEI_3_0_0;
-        else if (version == "2013")
+        }
+        else if (version == "2013") {
             m_version = MEI_2013;
+        }
     }
     else {
         // default to MEI 4
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -2893,26 +2893,35 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
         return false;
     }
 
-    current = root.child("meiHead");
-    if (current.empty()) {
-        LogWarning("No header found in the MEI data, trying to proceed...");
+    if (root.attribute("meiversion")) {
+        std::string version = std::string(root.attribute("meiversion").value());
+        if (version == "5.0.0-dev")
+            m_version = MEI_5_0_0_dev;
+        else if (version == "4.0.1")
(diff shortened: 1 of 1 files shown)


src/iomei.cpp
```diff
@@ -2893,26 +2893,35 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
...
+        if (version == "5.0.0-dev")
+            m_version = MEI_5_0_0_dev;
+        else if (version == "4.0.1")
+            m_version = MEI_4_0_1;
+        else if (version == "4.0.0")
+            m_version = MEI_4_0_0;
+        else if (version == "3.0.0")
+            m_version = MEI_3_0_0;
+        else if (version == "2013")
+            m_version = MEI_2013;
```
lpugin, line 2907: Please add `{}`

src/iomei.cpp
```diff
@@ -2893,26 +2893,35 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
...
-                m_version = MEI_4_0_1;
-            else if (version == "4.0.0")
-                m_version = MEI_4_0_0;
-            else if (version == "3.0.0")
-                m_version = MEI_3_0_0;
-            else if (version == "2013")
-                m_version = MEI_2013;
+        // default to MEI 4
+        LogWarning("No meiversion found, falling back to MEI4");
+        m_version = MEI_4_0_0;
```
lpugin, line 2912: We need to fallback to `5.0.0-dev`

src/iomei.cpp
rettinghaus: done

src/iomei.cpp
```diff
@@ -2893,26 +2893,40 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
...
+        }
+        else if (version == "4.0.0") {
+            m_version = MEI_4_0_0;
+        }
+        else if (version == "3.0.0") {
+            m_version = MEI_3_0_0;
+        }
+        else if (version == "2013") {
+            m_version = MEI_2013;
+        }
```
ahankinson, line 2912: Should this handle the case where it doesn't match any of these values?

rettinghaus: I guess not. Haven't thought about it though, as I just separated the version check from the header handling.

src/iomei.cpp
```diff
@@ -2893,26 +2893,40 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
...
-                m_version = MEI_5_0_0_dev;
-            else if (version == "4.0.1")
-                m_version = MEI_4_0_1;
-            else if (version == "4.0.0")
-                m_version = MEI_4_0_0;
-            else if (version == "3.0.0")
-                m_version = MEI_3_0_0;
-            else if (version == "2013")
-                m_version = MEI_2013;
+        // default to MEI 4
```
musicEnfanthen, line 2915:
```suggestion
        // default to MEI 5
```

src/iomei.cpp
rettinghaus: By default setting `m_version` is `MEI_UNDEFINED`. I think it makes sense to go with that if we find a nonsense value in `meiversion`.

rettinghaus:
@lpugin anything that needs to be change?

PR #2219, opened 2021-05-28
Detect clef slur collisions

DavidBauer1984:
This PR fixes collisions between slurs and clefs.

| Before change | After change |
| ------------- | ------------- |
| Before  | After |


craigsapp:
Ideally the slur end would fall in the middle of a space rather than on a line.

Behind Bars, page 110:

Screen Shot 2021-05-27 at 10 52 25 PM

DavidBauer1984:
@craigsapp Interesting! The slurs are adjusted in order to avoid collisions. I don't know if this rule is considered, at least for the initial slur position. But this should be addressed in a separate issue.

lpugin:
Please add this example to the test suite.

PR #2226, opened 2021-06-04
Option to prevent widow measure from happening

Monceber:
Added a new option `--fit-last-measure` to prevent pages with single measure. If such option is provided, whenever there is one measure left on the the last page, we would try to fit it to the previous page/system. In some cases that might lead to too tight placement in the system, hence by default this option is off.
Additionally, it honors `--breaks encoded` option, so in cases where there is single page after the page break, even if option is specified it would not try to move that measure.
I'm not quite sure if the name for the option is the best, so if there are any suggestions for the name - please let me know.

### DIFF
--- a/include/vrv/functorparams.h
+++ b/include/vrv/functorparams.h
@@ -1069,7 +1069,8 @@ public:
  * member 1: a pointer the document we are adding pages to
  * member 2: a pointer to the current page
  * member 3: the cummulated shift (m_drawingYRel of the first system of the current page)
- * member 4: the page height
+ * members 4-8: the page heights
+ * member 9: a pointer to the leftover system (last system with only one measure)
  **/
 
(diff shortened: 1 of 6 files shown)


lpugin:
> I'm not quite sure if the name for the option is the best, so if there are any suggestions for the name - please let me know.

I would suggest `--breaks-no-widow-measure` (See https://en.wikipedia.org/wiki/Widows_and_orphans) if we think that one day we want to have `--breaks-no-widow-system`. Otherwise just `--break-no-widow` or `--breaks-avoid-widow`

rettinghaus:
I think, trying squeeze in a whole system would do more harm than improving actually anything. So I would go with `--break-no-widow`.

lpugin:
so `--breaks-no-widow` (and not `--break-no-widow`, sorry for the initial typo)

PR #2229, opened 2021-06-04
WIP: MusicXML: Convert line-through attribute in lyrics into rend 

earboxer:
> line-through: Number of lines to use when striking through text. [0,3]

[musicxml spec for text](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/text/)

We will also want to use this for other places where `<text>` is used, so I have made this a method.

> The “combining long stroke overlay” (U+0336) results in an unbroken stroke across the text: 

[Wikipedia article for strikethrough](https://en.wikipedia.org/wiki/Strikethrough)

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2721,6 +2721,7 @@ void MusicXmlInput::ReadMusicXmlNote(
                     // std::string textColor = textNode.attribute("color").as_string();
                     std::string textStyle = textNode.attribute("font-style").as_string();
                     std::string textWeight = textNode.attribute("font-weight").as_string();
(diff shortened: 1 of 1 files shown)


samuelbradshaw:
The example on the Wikipedia page doesn't convince me that the combining character would work very well:
Screen Shot 2021-06-04 at 4 07 25 PM

rettinghaus:
A better approach would be to convert it to proper MEI with `<rend rend="line-through">` (see https://music-encoding.org/guidelines/v4/data-types/data.textrenditionlist.html) and then convert it to generic `SVG` with `text-decoration="line-through"` (see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-decoration).

PR #2230, opened 2021-06-07
fix and improve time conversion

rettinghaus:
This PR fixes a bug in time signature imports from MusicXML and cleans up the code a bit.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1229,21 +1229,24 @@ int MusicXmlInput::ReadMusicXmlPartAttributesAsStaffDef(pugi::xml_node node, Sta
                     else
                         meterSig->SetForm(METERFORM_norm);
                 }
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -1586,34 +1589,37 @@ void MusicXmlInput::ReadMusicXmlAttributes(
...
+            else if (time.child("senza-misura")) {
+                if (time.child("senza-misura").text()) {
+                    meterSig->SetSym(METERSIGN_open);
+                }
+                else {
+                    meterSig->SetForm(METERFORM_invis);
+                }
             }
             // add it if necessary
             if (meterSig) {
```
lpugin, line 1625: I don't think you need this check since it cannot be NULL

rettinghaus: Thanks for the hint!

PR #2243, opened 2021-06-11
Improve accent positioning in cross-staff situations

Monceber:
Changed code for accent positioning with cross-staff beams. Originally, they are properly rendered for the cross-staff notes, but for beams accent would be positioned in weird places on the original staff (as opposed to cross-staff).

Old rendering:
image

New rendering:
image

(Change applies only to the accent positioning, so disregard difference in beam rendering)


lpugin:
Please check artic-011 and add your example to the test suite

image

PR #2252, opened 2021-06-18
Improve rendering of the cross staff slurs

Monceber:
Mostly an attempt to improve rendering of cross-staff slurs.
Example, based on the `slur-014.mei` from the test suite.
Before:
image
After:
image

There are a lot of changes, but to put it simply, these are the main points changed:
1. Adjusted which elements were taken into consideration for the slur overlap. Previously, only elements from one staff/layer were used. With that. multi-layer or cross-staff slurs that crossed over notes/stems were overlapping with those elements, which led to a lot of collisions. Changes in `view_slur.cpp` address that, which should allow for better collision tracking in the cross-staff situations.
2. Main, and largest part - trying to make cross-staff slurs avoid aforementioned collisions. Default slurs, drawn as bezier curves with equally distanced control points do not really cut it - typically we end up with one side (or both) hanging way above the point. In some cases we end up having weird slurs like following:
image
With this change, such cases should be drawn in similar manner:
image


include/vrv/devicecontextbase.h
```diff
@@ -202,6 +202,66 @@ class Point {
...
 
+// ---------------------------------------------------------------------------
+// BezierCurve
+// ---------------------------------------------------------------------------
+
+/**
+ * Simple class for representing bezier cCurve
+ */
+
+class Doc;
```
lpugin, line 213: This goes at the top of the header file.

src/slur.cpp
```diff
@@ -168,126 +161,136 @@ int Slur::AdjustSlurCurve(Doc *doc, const ArrayOfCurveSpannedElements *spannedEl
...
-            if (posRatio && (dist != 0)) {
-                posX = itPoint->second.second.x - p1->x;
-                if (posX > dist / 2) posX = p2->x - itPoint->second.second.x;
-                if (dist != 0) posXRatio = (float)posX / ((float)dist / 2.0);
+bool Slur::AdjustSlurPosition(
+    Doc *doc, FloatingCurvePositioner *curve, BezierCurve &bezierCurve, float &angle, bool forceBothSides)
+{
+    bool isNotAdjustable = false;
+    const int margin = doc->GetDrawingUnit(100);
+    auto [maxShiftLeft, maxShiftRight]
```
lpugin, line 174: Don't use `auto`

src/slur.cpp
```diff
@@ -168,126 +161,136 @@ int Slur::AdjustSlurCurve(Doc *doc, const ArrayOfCurveSpannedElements *spannedEl
...
+            return true;
+        }
+        else {
+            Point points[4];
+            points[0] = bezierCurve.p1;
+            points[1] = bezierCurve.c1;
+            points[2] = bezierCurve.c2;
+            points[3] = bezierCurve.p2;
+            // Approximate bezier extrema and find time at which curve has highest/lowest Y value
+            const auto [time, yPos]
```
lpugin, line 207: Don't use `auto`

include/vrv/slur.h
```diff
@@ -52,21 +52,28 @@ class Slur : public ControlElement,
     curvature_CURVEDIR GetDrawingCurvedir() const { return m_drawingCurvedir; }
     void SetDrawingCurvedir(curvature_CURVEDIR curvedir) { m_drawingCurvedir = curvedir; }
     bool HasDrawingCurvedir() const { return (m_drawingCurvedir != curvature_CURVEDIR_NONE); }
-    void IsCrossStaff(bool isCrossStaff) { m_isCrossStaff = isCrossStaff; }
-    bool IsCrossStaff() const { return m_isCrossStaff; }
+    //void IsCrossStaff(bool isCrossStaff) { m_isCrossStaff = isCrossStaff; }
+    //bool IsCrossStaff() const { return m_isCrossStaff; }
```
lpugin, line 56: Remove them if we do not need it any more

include/vrv/slur.h
```diff
@@ -98,7 +105,7 @@ class Slur : public ControlElement,
      * This is not instantiated in Prepare::CrossStaff but in View::DrawSlurInitial
      * When set, the slur needs to be Re-drawn and re-adjusted after the vertical adjustment of the staves
      */
-    bool m_isCrossStaff;
+    //bool m_isCrossStaff;
```
lpugin, line 108: Can be removed in not necessary any more

lpugin (changes requested):
Looks great! Only minor changes.

src/slur.cpp
Monceber: I find structural bindings a little bit clearer, since we don't need to declare variables beforehand. But I'll change it to std::tie, since they cannot be used without 'auto'.

include/vrv/slur.h
```diff
@@ -105,7 +103,6 @@ class Slur : public ControlElement,
      * This is not instantiated in Prepare::CrossStaff but in View::DrawSlurInitial
      * When set, the slur needs to be Re-drawn and re-adjusted after the vertical adjustment of the staves
      */
```
lpugin, line 105: also remove the comment

lpugin (approved):
Great, thanks!

PR #2253, opened 2021-06-18
Thread safety

DavidBauer1984:
This PR improves thread safety by declaring static member variables either as const or thread_local. This should allow running several instances of Verovio concurrently.

One issue is the singleton object factory where element classes are registered. This is not used anywhere (!) for object construction. For now, we have commented out all declarations of factory variables. Other options would be to remove this code completely or to declare the singleton instance itself as thread_local.

### DIFF
--- a/src/neume.cpp
+++ b/src/neume.cpp
@@ -42,7 +42,7 @@ thread_local std::map<std::string, NeumeGroup> Neume::s_neumes
 // Neume
 //----------------------------------------------------------------------------
 
-// static ClassRegistrar<Neume> s_factory("neume", NEUME);
+static const ClassRegistrar<Neume> s_factory("neume", NEUME);
 
 Neume::Neume() : LayerElement("neume-"), ObjectListInterface(), AttColor()
 {
--- a/src/neume.cpp
+++ b/src/neume.cpp
@@ -32,7 +32,7 @@
 
 namespace vrv {
 
-thread_local std::map<std::string, NeumeGroup> Neume::s_neumes
+const std::map<std::string, NeumeGroup> Neume::s_neumes
     = { { "", PUNCTUM }, { "u", PES }, { "d", CLIVIS }, { "uu", SCANDICUS }, { "dd", CLIMACUS }, { "ud", TORCULUS },
           { "du", PORRECTUS }, { "ddd", CLIMACUS }, { "ddu", CLIMACUS_RESUPINUS }, { "udu", TORCULUS_RESUPINUS },
           { "dud", PORRECTUS_FLEXUS }, { "udd", PES_SUBPUNCTIS }, { "uud", SCANDICUS_FLEXUS },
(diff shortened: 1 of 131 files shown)


lpugin:
The object factory is not used yet because it's just been added. It is meant to facilitate object creation in the editor API. Basically we can do
```cpp
Object *note = ObjectFactory::GetInstance()->Create("note");
```
We need a solution that is not commenting / removing it.

lpugin (changes requested):
See comment on ObjectFactory

DavidBauer1984:
OK the factory is back. I also changed some members from thread_local morally const to const.

src/neume.cpp
```diff
@@ -117,7 +117,13 @@ NeumeGroup Neume::GetNeumeGroup()
...
         previous = current;
     }
-    return s_neumes[key];
+
+    try {
+        return s_neumes.at(key);
+    }
+    catch (std::out_of_range &e) {
+        return NEUME_ERROR;
+    }
```
lpugin, line 126: Why to you need to change this?

DavidBauer1984: s_neumes is now static const, so we must replace operator[] which only works for non-const maps. One way or the other we should check that the key is valid.

lpugin: OK. Then rather than a `try/catch` check with `.count` and return `NEUME_ERROR` if not found. We use `try/catch` only in quite specific places (e.g., input) and we should stick with this.

@JRegimbal does it work for you?

JRegimbal: This looks like it should be fine for us! @kemalkongar pinging you so other people are aware of this change.

DavidBauer1984: I guess the reason I preferred try/catch is that it is slightly more performant. Prechecking with find or count requires two binary searches per element access. It's not a big deal, since our map here is tiny small.

lpugin (approved):

alpha0010:
Hmm.  I guess this breaks my use case of passing toolkit instances between threads.

lpugin:
@alpha0010 would you be able to elaborate more? That would be necessary to see what can be done to fix what gets broken on your side.

alpha0010:
Worker thread pool creates toolkit instances.  Instances are cached (based on file loaded and render settings) for later use.  When a thread tries to use a toolkit (e.g. `RenderToSVG()`), it may not be the same thread on which the toolkit was created.  (Note that due to architecture constraints, although I can guarantee multiple threads will not access a toolkit instance simultaneously, I cannot force an instance to be used from a specific thread; nor even that the thread has the same lifespan as the toolkit.)

My solution is to execute the following right before I use any api on a toolkit instance.
```cpp
if (vrv::Resources::GetPath() != path) {
    vrv::Resources::SetPath(path);
    vrv::Resources::InitFonts();
}
```
It fixes the crash I have observed (though I have only done minimal testing, so unsure if there are other issues).

DavidBauer1984:
@alpha0010 We had a similar problem in our application (the resource path got lost for shared toolkit instances when switching between threads) and then I recommended to use one toolkit instance per thread. I did not really think that sharing toolkit instances makes sense, since this causes problems when done at the same time. But maybe it does (as in your case) if one is careful enough not to access the same toolkit instance simultaneously. Another scenario would be to do initialisation (i.e. setting font and resource path) in the main thread and then doing the layout and rendering in a worker thread to keep a GUI responsive.

What needs to be done is to move all vrv-resources into a non-static (and thus not thread-local) container object which is owned by the toolkit or document. Then the resource path and loaded fonts would survive thread-switching.

PR #2254, opened 2021-06-20
Minor improvements to Python scripts

rettinghaus:


doc/test-suite-diff.py
```diff
@@ -20,7 +21,7 @@ def start_webpage():
     head = etree.Element('head')
     html.append(head)
     title = etree.SubElement(head, 'title')
-    title.text = "Verovio test suite for Pull Request"
+    title.text = 'Verovio test suite for Pull Request'
```
ahankinson, line 24: Not sure why you’re changing to single quotes?

doc/test-suite-diff.py
```diff
@@ -130,11 +131,11 @@ def end_webpage(html, body, htmlOutFile):
             pngFileOut = os.path.join(path_out, item1, name + '.png')
             pngFile1Out = os.path.join(path_out, item1, name + '.after.png')
             pngFile2Out = os.path.join(path_out, item1, name + '.before.png')
-            print(pngFile1, pngFile2)
+            print(f'Comparing {name}')
```
ahankinson, line 134: This is actually changing the output of print, so it’s not simply a formatting change

fonts/generate_font_metadata.py
```diff
@@ -45,7 +45,7 @@ def validateGlyph(glyph):
             report += '- DUPLICATE UNICODE POINT'
 
         if report != '':
-            print('%s (%#0.2x) issues: %s' % (g.glyphname, glyph.unicode, report), log)
```
ahankinson, line 48: F strings are used elsewhere — why not here?

setup.py
```diff
@@ -12,6 +12,7 @@
 
 
 class build_ext(_build_ext):
+
```
ahankinson, line 15: Strange place for an empty line, especially with a docstring following it

ahankinson (changes requested):
Some questions about changes

setup.py
rettinghaus: PEP 257

ahankinson: PEP257 doesn't specify a blank line before the docstring, it specifies a blank line after.

doc/test-suite-diff.py
rettinghaus: Mixing single with double quotes is just bad style. (PEP8)

doc/test-suite-diff.py
rettinghaus: Correct, but simply printing the same file name twice with full paths isn't very informative, so it's still a kind of formatting change ;-)

setup.py
rettinghaus: See Error Code [D203](https://pep257.readthedocs.io/en/latest/error_codes.html) 

> 1 blank line required before class docstring

But we can leave it as it is.

fonts/generate_font_metadata.py
rettinghaus: This file originally comes from https://github.com/w3c/smufl repo. We removed all the log file stuff and this one was overlooked. I'm still wondering if it should be in line with the upstream version and if we should push further fixes/improvements there?

setup.py
ahankinson: D203 is deprecated and should be disabled if your editor doesn’t do this already:

See:

https://opendev.org/openstack/keystoneauth/commit/cf520d9ca45dfbb09ca9bab90fa3d615bc950326

rettinghaus: Yeah, I've seen that there has been a discussion. That's why I reverted the change already.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [unreleased]
+* Making Veroivo thread safe (@eNote-GmbH)
```
lpugin, line 4: Fix typo. Next time make a distinct PR with [skip-ci] for changelog updates. Thanks.

ahankinson (approved):

PR #2256, opened 2021-06-21
Perform nested alignment

lpugin:
This branch performs nested alignment of the layers instead of relying simply on the horizontal position. This improves layout with accidentals (or flags, etc.) or with multiple layers. 

layout-3 3 0

Some example improvements:

* dir-007 (before and after)
image
image

* unison-001
image
image

* tuplet-020
image
image

* layer-002
image

* accid-006
image


rettinghaus (approved):
Looks great! Thanks!

PR #2258, opened 2021-06-21
Further improvement of the nested alignment

lpugin:
Before and after

* unison-001
image

image

* tuplet-017
image
image

* layer-002
image


lpugin:
This change yields highly optimised layout, maybe too much in some cases. We can make it an option `--layer-compact-layout` (any ideas?) for it to be enabled explicitly.

This would be the default:
image


With `--layer-compact-layout`:
image

Any thoughts?

GBogdahn:
I think this looks very promising!

rettinghaus:
I would prefer the other way round, i.e. have a `--relaxed-spacing` option for a less-than-optimal spacing.

lpugin:
Latest change ensures there is a standard spacing between notes. This addresses #2261. See examples added there

rettinghaus:
Looking at `slur-002` this seems to fix #2259 too.

lpugin:
> Looking at `slur-002` this seems to fix #2259 too.

I doubt it. Probably just because the slur change hasn't be merge into this branch.

rettinghaus:
@lpugin to what exactly applies now `leftMarginAccid`? Decreasing the default value below 1 will lead to several collisions.

lpugin:
> @lpugin to what exactly applies now `leftMarginAccid`? Decreasing the default value below 1 will lead to several collisions.

I can look. Would it be possible to have some examples of the collisions you get when changing the values?

Issue #1803, opened 2020-11-06
Accidentals sequence in key signature change


cividini:
When `keysig.showchange="true"` leads to a key with fewer sharps or flats, the accidentals of the previous key signature are cancelled in Verovio AFTER the new key signature. According to “Behind bars”, p. 92, they have to be cancelled BEFORE the new key signature.

Example in flat keys: in the second key signature the remaining flat must appear after the four cancelled accidentals, and not before them.

Verovio Version 3.1.0-dev-5bdf089:
![accidential_key_change_Verovio_flat](https://user-images.githubusercontent.com/45197272/98370101-51f2f100-203a-11eb-8c2d-6cc10440232f.PNG)

Behind bars, p. 92:
![accidential_key_change_behind_bars](https://user-images.githubusercontent.com/45197272/98370134-60d9a380-203a-11eb-8a9d-54022f5fdef3.PNG)

MEI-Code:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei meiversion="4.0.0" xmlns="http://www.music-encoding.org/ns/mei">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title/>
   </titleStmt>
   <pubStmt/>
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2020-11-06" version="Version 3.1.0-dev-5bdf089">
     <name>Verovio</name>
     <p>example for key signature accidental change in flat keys</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title/>
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000001622068775">
    <score xml:id="score-0000001952486407">
     <scoreDef key.accid="f" key.mode="major" key.pname="d" key.sig="5f" meter.count="4" meter.unit="4" xml:id="scoredef-0000001921549423">
      <staffGrp xml:id="staffgrp-0000000077171473">
       <staffDef clef.line="2" clef.shape="G" lines="5" n="1" xml:id="staffDef_P1"/>
      </staffGrp>
     </scoreDef>
[...]

PR #2263, opened 2021-06-25
Order of accidentals in key changes

DavidBauer1984:
This PR addresses #1803 by grouping the naturals always in front at key changes.

KeyChange

### DIFF
--- a/src/view_element.cpp
+++ b/src/view_element.cpp
@@ -947,16 +947,20 @@ void View::DrawKeySig(DeviceContext *dc, LayerElement *element, Layer *layer, St
     // Show cancellation if show cancellation (showchange) is true (false by default)
     // This is not meant to make sense with mixed key signature
     if ((keySig->GetScoreDefRole() != SCOREDEF_SYSTEM) && (keySig->GetSigShowchange() == BOOLEAN_true)) {
-        // The type of alteration is different (f/s or f/n or s/n) - cancel all accid in the normal order
-        if (keySig->GetAccidType() != keySig->m_drawingCancelAccidType) {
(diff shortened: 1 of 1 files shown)


lpugin:
Looks good to me. However, since the issue was marked as "Waiting for MEI clarification" I'd like you make sure everybody is happy with this change. @cividini? I will merge it if nobody speaks up.

DavidBauer1984:
I think what needs to be clarified in MEI is the following part of the discussion:

> Currently key-signature cancellations are controlled by keySig@sig.showchange which is a boolean:
> 
> ```XML
> <keySig` xml:id="keysig-L13F1" sig="4f" sig.showchange="true" />
> ```
> 
> Perhaps MEI should change this to allow this to have three values:
> 
> (1) `@sig.showchange="none"` :: modern style (default if attribute not given)
> 
> (2) `@sig.showchange="before"` :: standard style
> 
> (3) `@sig.showchange="after"` :: old/French style
> 
> But in any case if the before/after options are not available, then the "standard style" should be used if @sig.showchange is true, instead of the "old style" version.

As I understand it, everybody agreed that the standard style should be the default and this is what has been done in the PR. Only further optional styles would need MEI attributes which have to be discussed, but these are not implemented in the PR.

Anyway, we can still wait a couple of days, if anybody has to say something on this.

craigsapp:
Sounds good to me.  The standard style should be the default in any case.  So before issue https://github.com/music-encoding/music-encoding/issues/727 is addressed, it is probably best to fix to `sig.showchange="true"` show the standard style (with the modern/old styles available when MEI is updated accordingly).

PR #2265, opened 2021-06-26
allow building as static and not changing source files during build

mockersf:
I'm currently trying to do [Rust bindings for Verovio](https://github.com/mockersf/rustovio), and met a few issues that this PR should fix:
* add a `BUILD_AS_STATIC_LIBRARY` option to build as a static library. Currently using `BUILD_AS_WASM` to have the same result
* add a `GET_GIT_COMMIT` option to be able to disable running the script `get_git_commit.sh`. This script modifies the source filesystem which is avoided as much as possible during rust builds - [example error this produces](https://docs.rs/crate/rustovio/0.1.0/builds/404489). I also added a default `include/vrv/git_commit.h` header file (by removing it from .gitignore and adding it back in) that will be overwritten during normal builds


lpugin (commented):
Thanks. Can't you allow the building process to write in `./include/vrv/`? Running cmake does generate other files (e.g., `./tools/Makefile`). Why is that one problematic?

Static linking: since Verovio is LGPL, static linking it possible only under some stricter conditions than dynamic linking. If you are making a wrapper that links Verovio statically, then it should be LGPL too - or LGPL compatible.

lpugin:
Re-open with answers if this is still desired.

PR #2272, opened 2021-07-02
Fix for rare beam crash

Monceber:
- changed code to adjust beams only if there are element coord references present. In rare cases with beams this could lead to a crash

### DIFF
--- a/src/beam.cpp
+++ b/src/beam.cpp
@@ -1421,7 +1421,7 @@ int Beam::AdjustBeams(FunctorParams *functorParams)
     AdjustBeamParams *params = vrv_params_cast<AdjustBeamParams *>(functorParams);
     assert(params);
 
-    if (this->HasSameas() || !this->GetChildCount()) {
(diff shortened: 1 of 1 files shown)


lpugin:
Do you know why / when it crashes? It might be good to know why this is happening because the problem might be somewhere else.

Monceber:
In this case, I've encountered a crash when a beam had only `<space>` children. This led to it not having any coordRef defined, since spaces are ignored, i.e. beam like this:
```XML
<beam>
  <space dur="16"/>
  <space dur="16"/>
  <space dur="16"/>
  <space dur="16"/>
</beam>
```
I mean, encoding is probably the main problem here, I think beam should have at least one valid child. However, Verovio should be crashing because of that. It should be checking if there are valid references to work with, before actually trying to access them.

lpugin:
> I mean, encoding is probably the main problem here, I think beam should have at least one valid child. However, Verovio should not be crashing because of that. It should be checking if there are valid references to work with, before actually trying to access them.

Thanks. I perfectly agree. I just wanted to make sure it was not crashing because of something no properly (re-)initialised.

PR #2278, opened 2021-07-06
Develop ligature as bracket

lpugin:


lpugin:
With the option `--ligature-as-bracket`

image

martha-thomae:
Looks awesome, thank you! 
Will talk with Juliette so that we can get this into the MP-Editor :)

PR #2292, opened 2021-07-15
Add GitHub-action for clang-format check

### LINKED ISSUE #1762: Check clang format on PR
It would be good to check that PRs are compliant with clang formatting. One simple way to do it would be to run clang-format on the code and fails if anything is not properly formatted. That is, if this yields any modifications.

Clang formatting call on the code should be:
```
clang-format -i -style=file include/vrv/*.h
clang-format -i -style=file src/*.cpp
```

One possible problem with this might be if the contributor uses a different version of clang-format resulting in different formatting output.

rettinghaus:
This PR fixes some formatting and adds an action to check clang-formatting on PRs.
Closes #1762


lpugin:
Could you keep a separate PR for the clang-format action? Thanks

PR #2294, opened 2021-07-16
Add github action for clang-format check

rettinghaus:


lpugin:
Aren't we going to have problems if a contributors runs a different version of clang-format that produces slightly different results?

rettinghaus:
I think, what could make problems is using clang-format version prior to 9. That's why we say we use version 10: https://book.verovio.org/contributing/guidelines.html
Anyway, it doesn't block merging, it just shows potential formatting problems.

rettinghaus:
(you opened #1762 😉)

lpugin:
> Anyway, it doesn't block merging, it just shows potential formatting problems.

Yes, that's right. Let's see how it works

PR #2303, opened 2021-07-19
Better support for engravingDefaults

rettinghaus:
This PR cleans up around the implementation of `engravingDefaults`. It removes the hard-coded default values here and adds the possibility to `ResetOptions` to the toolkit. This is useful for iterative rendering of files with different styles or different options. 
Also it adds support for SMuFL's `pedalLineThickness`.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -85,7 +85,7 @@ public:
         m_isCmdOnly = false;
     }
     virtual ~Option() {}
-    virtual void CopyTo(Option *option);
+    virtual void CopyTo(Option *option) = 0;
 
     void SetKey(const std::string &key) { m_key = key; }
     std::string GetKey() const { return m_key; }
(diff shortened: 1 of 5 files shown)


include/vrv/options.h
```diff
@@ -94,8 +94,10 @@ class Option {
     virtual bool SetValueDbl(double value);
     virtual bool SetValueArray(const std::vector<std::string> &values);
     virtual bool SetValue(const std::string &value);
-    virtual std::string GetStrValue() const;
-    virtual std::string GetDefaultStrValue() const;
+    virtual std::string GetStrValue() const = 0;
```
ahankinson, line 97: Maybe this is just me not understanding a C++ism, but why do you set a string to 0?

DavidBauer1984: In C++ that is how you define a _pure virtual function_ which must be overridden in derived classes. This makes Option an abstract class. It was somewhat required before by putting a runtime check `assert(false)` in the implementation, but it is much safer and standard to enforce this at compile time.

lpugin: Yes, there are a few places where we should add this.

PR #2312, opened 2021-07-27
Fix layout detection in MusicXML

rettinghaus:
Minor fix to correctly handle MusicXML files without breaks but other layout information.


99sedoy:
hi

Issue #1262, opened 2020-01-15
Render multiple <mdiv> at once


wergo:
Related to the discussion in #1141, #509, and #154, I wanted to raise a related issue, that is how to render more than one `<mdiv>` at once. 

Background are collection of pieces, such as Op. 119 by Beethoven where you would encode each of the 11 pieces as a separate `<mdiv>` within one MEI file, similar to an entire sonata or symphony, as suggested in the guidelines: https://music-encoding.org/guidelines/v4/content/shared.html#sharedDivisionsBody. 

Many of the pieces/mdiv end and start on the same page, as here No. 7 end, 8, and 9: 
<img width="884" alt="Beethoven_Op119-7-9" src="https://user-images.githubusercontent.com/45632600/72452214-76d15b80-37bd-11ea-9b92-60c6f567db4c.png">

It would be great to have an option to render all (or a selection of) mdivs of a MEI file to be able to get mdiv transitions within one page. I am aware that this might not be easy to implement in Verovio. 

Addendum: In Beethoven's variations on a Swiss Air WoO 64 in Breitkopf & Härtel you have even four `<mdiv>`s on one page; two of them even sharing the same line/system:
<img width="875" alt="Beethoven_WoO64" src="https://user-images.githubusercontent.com/45632600/72452321-a5e7cd00-37bd-11ea-8ac7-a160379cf61c.png">

palgranum:
Is there any hope that this feature will be prioritized in foreseeable future?
As @wergo points out, there seems to be no other way/workaround for engraving (parts of) multiple movements/mdivs on the same page.

In addition to needing to solve this problem, I would also like to be able to render any page from a (mei file) score that might contain multiple mdivs without having to switch mdivs and wait for the toolkit to reload and relayout. This switching operation can last several seconds and prevents a practical implementation of displaying the score interactively in real-time without employing some complex mdiv switching caching operation.

lpugin:
The problem with multiple mdiv is that you can have very different score definitions. For example, a movement with 20 instruments, followed by another one of 5, or the opposite. This will never work well with dynamic rendering. However, I am happy to guide you if you want help for trying to implement something.

kseniaosipova:
@lpugin yes, it can be different, but still required the possibility to render 2 or more movements at once. So, could we consider this as approved, if we will start to propose the implementation of Rendering multiple <mdiv> at once?

lpugin:
[...]

PR #2316, opened 2021-07-31
Implementation of section@restart

lpugin:
The implementation of section@restart allows for scoreDef changes to be displayed within a staff. This somehow addresses the case raised raised by @wergo [here](https://github.com/rism-digital/verovio/issues/1262#issue-550300604). However, it works only when all sections are in the same `<mdiv>`, so the rendering of multiple `<mdiv>` will have to be treated separately.

The current implementation expects the `staffGrp` with the label change to carry a `@n`. See the [test suite example](https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/section/section_001.mei) Cautionary scoreDef values (keySig, meterSig) should be properly handled.

image

### DIFF
--- a/include/vrv/functorparams.h
+++ b/include/vrv/functorparams.h
@@ -2111,7 +2111,9 @@ public:
  * member 3: the previous measure (for setting cautionary scoreDef)
  * member 4: the current system (for setting the system scoreDef)
  * member 5: the flag indicating whereas full labels have to be drawn
- * member 6: the doc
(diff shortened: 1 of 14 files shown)


rettinghaus:
Wow, this looks great. Is the extra offset on the first `label` intended?

lpugin:
Yes. The space of the brace was not taken into account. It was not noticeable on the initial scoreDef but is now corrected.

PR #2325, opened 2021-08-05
remove unwanted paths

### LINKED ISSUE #2320: Python build fails
Trying to build the Python binding fails with this:
```
./src/fig.cpp:63:5: error: unknown type name 'Svg'; did you mean 'svg'?
    Svg *svg = dynamic_cast<Svg *>(this->FindDescendantByType(SVG));
    ^~~
    svg
/usr/local/include/svg.h:32:16: note: 'svg' declared here
typedef struct svg svg_t;
               ^
./src/fig.cpp:63:29: error: unknown type name 'Svg'
    Svg *svg = dynamic_cast<Svg *>(this->FindDescendantByType(SVG));
                            ^
./src/fig.cpp:63:10: error: cannot initialize a variable of type 'svg *' with an rvalue of type 'vrv::Object *'
    Svg *svg = dynamic_cast<Svg *>(this->FindDescendantByType(SVG));
         ^                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./src/fig.cpp:64:28: error: member access into incomplete type 'svg'
    int width = (svg) ? svg->GetWidth() : 0;
                           ^
/usr/local/include/svg.h:32:16: note: forward declaration of 'svg'
typedef struct svg svg_t;
               ^
./src/fig.cpp:64:30: error: expected ':'
    int width = (svg) ? svg->GetWidth() : 0;
                             ^
                             : 
./src/fig.cpp:64:23: note: to match this '?'
    int width = (svg) ? svg->GetWidth() : 0;
                      ^
./src/fig.cpp:64:30: error: use of undeclared identifier 'GetWidth'; did you mean 'AttWidth'?
    int width = (svg) ? svg->GetWidth() : 0;
                             ^
./libmei/atts_shared.h:5984:7: note: 'AttWidth' declared here
class AttWidth : public Att {
      ^

```

 - macOS 11.5.1 with Python 3.9.6

Using `setup.py build_ext` without extra options.

rettinghaus:
As discussed this PR removes `/usr/local/include` as it can lead to build failures. It removes also the empty `./include` path.
Closes #2320


ahankinson:
`/use/local/include` is the location of any dependencies installed by Homebrew.

This can include newer versions of Python, SWIG, and maybe some other dependencies?

Was this change tested to still work correctly? For example, does it still find a newer Python if that is installed, over the system Python?

rettinghaus:
The thing is that Homebrew on newer machines (M1) is installed in `/opt/homebrew`, because `/usr/local/include` is handled more strict (see https://docs.brew.sh/FAQ#why-is-the-default-installation-prefix-opthomebrew-on-apple-silicon).
And from my understanding you should have Homebrew in your `$PATH` anyway, if you want to use it.

Anyway, this fixes the problems I had on my Apple Silicon machine.

rettinghaus:
CI works fine: https://github.com/rettinghaus/verovio/runs/3250495533

ahankinson:
Some libraries that conflict with built-in libraries (like Python) are made “keg only” which means that they are not automatically included into your system paths; you have to explicitly link against the Homebrew version.

It would be good to test this against an Intel installation as well. Unfortunately I don’t have my laptop at hand.

DavidBauer1984:
Maybe consider just changing the order of include directories (i.e. putting `/usr/local/include` at the end of the list). This should be enough to fix the build failure.

lpugin:
It works for me on an Intel machine. I am merging it.

rettinghaus:
Thanks. 
Just for documentation: I tried adding _both_ homebrew paths [here](https://github.com/rettinghaus/verovio/actions/runs/1102535258) and it made no problems. But as these are Mac-specific I would prefer to not have them in there by default.

PR #2329, opened 2021-08-06
Fix assertion

DavidBauer1984:
This PR fixes a wrong assertion. Interestingly, this was not observed because of the system function defined in cstdlib.


lpugin:
Well spotted! Quite surprising!

PR #2337, opened 2021-08-09
Include tools folder in clang-format check

rettinghaus:


tools/main.cpp
```diff
@@ -223,21 +223,19 @@ int main(int argc, char **argv)
...
-            { "stdin", no_argument, 0, 'z' },
-            { 0, 0, 0, 0 }
-        };
+    static struct option base_options[] = { //
+        { "all-pages", no_argument, 0, 'a' }, //
+        { "from", required_argument, 0, 'f' }, //
+        { "help", no_argument, 0, 'h' }, //
+        { "outfile", required_argument, 0, 'o' }, //
+        { "page", required_argument, 0, 'p' }, //
+        { "resources", required_argument, 0, 'r' }, { "scale", required_argument, 0, 's' }, //
```
lpugin, line 232: Why no `//` between these to options?

tools/main.cpp
```diff
@@ -223,21 +223,19 @@ int main(int argc, char **argv)
...
+        { "to", required_argument, 0, 't' }, //
+        { "version", no_argument, 0, 'v' }, { "xml-id-seed", required_argument, 0, 'x' }, //
```
lpugin, line 234: Same here

tools/main.cpp
rettinghaus: Missed those. Added.

PR #2347, opened 2021-08-17
Support for multiple mdiv elements

### LINKED ISSUE #1262: Render multiple <mdiv> at once

lpugin:
This PR implements support for multiple `<mdiv>` elements with different `<score>`. Fixes #1262 

For rendering all scores at once, the option --mdiv-all needs to be set. Below the rendering for the example [mdiv-001](https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/mdiv/mdiv-001.mei) from the test suite.

image

All scoreDef changes occurring when a new score starts (e.g., `@staff.spacing`) and pgHead2 and pgFoot/pgFoot2 changes should all be properly updated.

When rendering all mdivs with multiple scores without casting off (i.e., on one single long page), then the different scores are rendered below each other. When cast off, a new score will always start on a new system. 

The page-based output will need to be adjusted, but since nobody uses it, it should not be problematic. It will be addressed with a separate PR.


lpugin:
@rettinghaus any ideas about the changes appearing in the MusicXML test suite examples?

rettinghaus:
@lpugin I've no idea whatsoever. But I would assume you may proceed, because it doesn't look like a regression.

wergo:
Thanks @lpugin for this great PR!! 

On our TROMPA encoding repo, I have added WoO64 with mdivs per variation as example for mdivs starting in the same system: https://github.com/trompamusic-encodings/Beethoven_WoO64_BreitkopfHaertel

Issue #2344, opened 2021-08-12
Performance decrease between 3.4.1 and 3.5.0


lpugin:
**Describe the problem**
It seems that some changes between the two versions have yield a significant performance decrease. We need to investigate where it comes from.

**To Reproduce**
Very basic evaluation with
`
time ./verovio -r ../data -o output.svg --all-pages ~/Documents/verovio/gh-pages/examples/downloads/Hummel_Concerto_for_trumpet.mei
`

On my machine:
* Time with 3.4.1: **4.66s** user 0.65s system 77% cpu 6.860 total
* Time with 3.5.0: **6.99s** user 0.48s system 86% cpu 8.691 total

**Expected behavior**
Less of a performance decrease despite the improvements

**Input data**
https://github.com/rism-digital/verovio.org/blob/gh-pages/examples/downloads/Hummel_Concerto_for_trumpet.mei

**Verovio information**
Command-line

**Environment information (as appropriate)**
MacOS 11.5.1

PR #2351, opened 2021-08-20
Improve performance

DavidBauer1984:
This PR should improve total runtime by ~10%. The idea is not to retrieve class IDs via virtual functions, but to store them at construction in the Object base class.

Some examples that I have tested:

| Piece | Size | Runtime before | Runtime after |
| ----- | ---- | ---------------- | ------------- |
| Beethoven - Presto | 1.1 MB / 10 pages | 530ms | 475ms |
| Hummel - Trumpet Concerto | 5.7 MB / 128 pages | 4.52s | 3.94s |
| Holst - The Planets | 50.1 MB / 403 pages | 2min 8s | 1min 56s |

### DIFF
--- a/include/vrv/abbr.h
+++ b/include/vrv/abbr.h
@@ -29,7 +29,6 @@ public:
     virtual Object *Clone() const { return new Abbr(*this); }
     virtual void Reset();
     virtual std::string GetClassName() const { return "Abbr"; }
-    virtual ClassId GetClassId() const { return ABBR; }
     ///@}
 
 private:
--- a/include/vrv/accid.h
+++ b/include/vrv/accid.h
@@ -43,7 +43,6 @@ public:
     virtual Object *Clone() const { return new Accid(*this); }
     virtual void Reset();
     virtual std::string GetClassName() const { return "Accid"; }
-    virtual ClassId GetClassId() const { return ACCID; }
     ///@}
 
     virtual PositionInterface *GetPositionInterface() { return dynamic_cast<PositionInterface *>(this); }
--- a/include/vrv/add.h
+++ b/include/vrv/add.h
@@ -29,7 +29,6 @@ public:
     virtual Object *Clone() const { return new Add(*this); }
     virtual void Reset();
     virtual std::string GetClassName() const { return "Add"; }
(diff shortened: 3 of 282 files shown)


lpugin:
Looks good. My only concern is the introduction to Object::SetClassId because we should never have to change a ClassId. As far as I can see, this is only for BarLineAttr. Am I right? I think it would be better to have two distinct BarLineAttrLeft/Right classes so we can forget it. Can you look at this?

DavidBauer1984:
Actually I tried it first this way, but then I got worried, because `BarLineAttr` also impersonates itself as 'BarLine' when 'm_noAttr' is set. But I can have a look again and agree that it would be much nicer without Object::SetClassId.

DavidBauer1984:
@lpugin What do you think about having a bool option in Verovio --show-runtime that displays total runtime if desired. I would find it helpful as developer to keep an eye on runtime without having to launch instruments in Xcode (or other tools).

lpugin:
This makes me see that this is actually not good https://github.com/rism-digital/verovio/commit/2b909f61e50ca465ceba13b637d02012cd722106 and should probably not have made its way through.
ClassIds were introduced for speed optimization. The idea is that we do not do `dynamic_cast` of object type identification. Since now we even changed to `static_cast` (through `vrv_cast`) we should absolutely forbid ClassId to be changed.

Maybe we don't need BarlLineAttr at all and only rely on some flags in `BarLine`. I cannot remember why having `BarLineAttr` was the simplest option when it was introduced.

DavidBauer1984:
@rettinghaus You are right, I have reverted the change. This was the remnant of my fix for #2344 which has already been fixed by Laurent from another direction. So it has no big effect on runtime anymore, even though it still means a lot of unnecessary iteration, but we can optimize this another time.

@lpugin I have removed the BarLineAttr, but it was not straightforward. I guess the main reasons why it was introduced was to have another class ID to distinguish in:

- Doc::GetLeftMargin(ClassID id) / Doc::GetRightMargin(ClassID id)
- The class ID include list for the AdjustXPos functor.

To preserve the distinction I have added an additional overload in the first case and an additional param in the second case.

PR #2353, opened 2021-08-22
Fix Toolkit::RedoLayout() to not cast off when breaks option is set to 'none'

jumartin:
When calling `ToolKit::RedoLayout()`, casting off is performed even when the `breaks` option is `none`. This fixes it.


lpugin:
This looks fine. However, it does not really make sense to call `RedoLayout` when breaks are set to none. In which context do you do this?

jumartin:
I have a viewer which allows to switch between different types of layout : pages or panorama (no breaks).
When the user switches from one to the other, I set the appropriate options and call `RedoLayout`. 

By the way, thanks for this amazing library!

lpugin:
Thanks!

PR #2356, opened 2021-08-23
Add support for simple feature extraction

lpugin:
The feature extraction method is tailored for incipit search. It currently contains only pitches and interval and corresponding note ids for search result highlighting

Features can be retrieved through the `Toolkit::GetDescriptiveFeatures` that takes an empty stringified JSON object as parameter.

This PR also add the `Toolkit::ResetXmlIdSeed` method for changing the seed of the xml:id generator.


ahankinson:
Looks good to me. We should probably merge #2355 first, though, since this PR also incorporates the changes from there.

PR #2360, opened 2021-08-26
Cancel layout

DavidBauer1984:
This PR adds functionality to the toolkit to abort the layout in a thread-safe way. One use case is that a helper thread renders a score and meanwhile the user resizes the window. Then the current rendering must be aborted and restarted with new (page size) parameters.

### DIFF
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -77,8 +77,8 @@ public:
      * @name Access the abort flag
      */
     ///@{
-    void ScheduleAbort(bool value) { m_abort = value; };
-    bool AbortRequested() { return m_abort; };
+    void SetAbortMode(bool value) { m_abort = value; };
+    bool AbortRequested() const { return m_abort; };
     ///@}
 
     /**
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -8,6 +8,10 @@
 #ifndef __VRV_DOC_H__
 #define __VRV_DOC_H__
 
+#include <atomic>
+
+//----------------------------------------------------------------------------
+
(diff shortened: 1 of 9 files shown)


lpugin:
I will have a look at this. However, it does look intrusive and will need more explanations, especially on how to use it. How does it work? Can you give some example use cases?

DavidBauer1984:
The main use case is to provide the ability to cancel Verovio when running it from a helper thread such that the application / main thread can proceed faster with actions requested by the user. This only effects the rendering of larger files when otherwise we would have to wait 1 second or longer to proceed. Some user actions where this might be helpful:

- The user closes the application. Here we just want to abort the rendering.
- The user resizes the window which presents the score. We want to restart rendering with new options (changed page size).

In the implementation we add an atomic abort flag which is periodically checked by Verovio. If the application needs to abort rendering, then the main thread sets the abort flag to true. Verovio then shortcuts its current action without providing a useful result (i.e. returning an empty page from rendering). After abort is finished and the main thread still needs to render something, it can either create a new toolkit instance or reset the abort flag to false and continue with the old one.

On the one hand we don't want to pollute the codebase with plenty of calls to check for abort, but on the other hand aborting should be much faster than waiting for an action to complete. I've used a profiler to distribute the calls and have this balance as much as possible. One also has to be careful that each check creates a new execution path. If Verovio aborts at this point, then one must ensure that memory is cleaned up and the application does not get into a bad state and crashes. For example, if horizontal layout was not completed, then vertical layout must not be started.

DavidBauer1984:
This is getting dusty. Should I close it for the time being?

Issue #1991, opened 2021-01-26
Add an option to specify selected attributes to be added to the SVG


wergo:
It would be useful for faster DOM traversing with jQuery to have the `@n` attribute for `staff` and `layer` written into the SVG output. Currently, this information has to be found in the MEI directly which is less convenient (and probably slower).

So a staff element would then look like:
`<g id="staff-0001234" class="staff" n="2">`

(Feature request)

ahankinson:
If you are concerned about performance and are hitting bottlenecks, native JS lookups are much faster than using jQuery. If you need jQuery compatibility you can do DOM traversal in native JS, and then convert your results to jQuery objects when you find them.

lpugin:
@wergo this would unfortunately not be valid SVG. One simple and generic way around it to put in the MEI `@type` values you need to have in the SVG. Verovio will pass them as `@class` in the SVG. So if your MEI is `<staff n="1" type="n1">` you will get `<g id="staff-0001234" class="staff n1">`. Can this work for you?

wergo:
Thank you for your quick replies!

@ahankinson, could you give an example on how you would use "native JS lookups" for xml encodings (directly on the text?)? (Sorry, I am not quite native in JS myself.)

@lpugin This would require the `@type` attribute in the MEI code? I would rather prefer a solution that runs with the "normal" MEI encodings.

lpugin:
> This would require the @type attribute in the MEI code? 

Yes. You can add this easily with an XSLT.

> I would rather prefer a solution that runs with the "normal" MEI encodings.

I understand. But the problem is that this type of need is fairly project-specific. So you need `@n`, but somebody else will need `@pname`, or `@place`, or midi attributes, and so on. (We already had requests for all of these.) And we will end up replicating the entire MEI into the SVG ;-) I am happy to hear about other ideas, though.

ahankinson:
If you have the XML parsed into a DOM structure, you can use `document.querySelectorAll`:

https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll

I'm not suggesting that you drop jQuery automatically if that's what you're comfortable with, but if you are hitting bottlenecks in the DOM traversal and lookups you may want to look at optimizing those with native calls.

wergo:
[...]

PR #2361, opened 2021-08-26
Support for adding additional attributes to SVG elements

wergo:
This PR implements #1991, according to @lpugin's suggestions. 

A repeatable option `--svgAdditionalAttribute` has been added that accepts a string in the form `<elementName>@<attribute>` (e.g., `note@pname`, `layer@n`, `staff@follows`); these commands result in `data-*="..."` in the pertinent SVG elements, so for the examples above e.g.,  `<g class="note" data-pname="a" ...`,  `<g class="layer" data-n="2"...`, `<g class="staff" data-follows="note-123"...`. 

Repeatable means, this option may be used multiple times, so any number of attributes may be added to the SVG depending on individual demands.

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -978,9 +978,6 @@ void SvgDeviceContext::AppendIdAndClass(std::string gId, std::string baseClass, 
 
 void SvgDeviceContext::AppendAdditionalAttributes(Object *object)
 {
-    std::string baseClass = object->GetClassName();
-    std::transform(baseClass.begin(), baseClass.begin() + 1, baseClass.begin(), ::tolower);
-
     std::pair<std::multimap<ClassId, std::string>::iterator, std::multimap<ClassId, std::string>::iterator> range;
     range = m_svgAdditionalAttributes.equal_range(object->GetClassId()); // if correct class name...
     for (std::multimap<ClassId, std::string>::iterator it = range.first; it != range.second; ++it) {


src/svgdevicecontext.cpp
```diff
@@ -974,6 +976,33 @@ void SvgDeviceContext::AppendIdAndClass(std::string gId, std::string baseClass,
...
+                    m_currentNode.append_attribute(("data-" + attributeName).c_str()) = (*iter).second.c_str();
+                    found = true;
+                }
+            }
+            if (!found)
+                LogWarning(
+                    "No attribute '%s' in '%s', nothing added to SVG.", attributeName.c_str(), className.c_str());
+        }
+    }
+}
```
lpugin, line 1004: The lookup will be simpler here with the `<std::multimap>`. Something  like
```cpp
std::pair<auto first, auto second> range = m_svgAdditionalAttributes.equal_range(object->GetClassId());
for (auto it = range.first; it != range.second; ++it) {
    // it->second will be the attribute name
    // add it to the current note
}

**Review by lpugin — CHANGES_REQUESTED** [review]
Looks good. The change requests are basically all for optimization. Thanks!

File: include/vrv/svgdevicecontext.h [hunk 2]
```diff
@@ -209,6 +214,15 @@ class SvgDeviceContext : public DeviceContext {
      */
     void SetRemoveXlink(bool removeXlink) { m_removeXlink = removeXlink; }
 
+    /**
+     *  Copies additional attributes of defined elements to the SVG, each string in the form "elementName@attribute"
+     * (e.g., "note@pname")
+     */
+    void SetAdditionalAttributes(std::vector<std::string> additionalAttributes)
```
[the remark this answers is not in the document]
wergo: Thanks @lpugin for the throrough review! Find below the changes implemented.

src/svgdevicecontext.cpp
wergo: We still have to check whether the requested attribute exists in that element.

include/vrv/svgdevicecontext.h
```diff
@@ -220,7 +221,12 @@ class SvgDeviceContext : public DeviceContext {
      */
     void SetAdditionalAttributes(std::vector<std::string> additionalAttributes)
     {
-        for (std::string s : additionalAttributes) m_svgAdditionalAttributes.push_back(s);
+        for (std::string s : additionalAttributes) {
+            std::string className = s.substr(0, s.find("@")); // parse <element@attribute>, e.g., "note@pname"
+            std::string attributeName = s.substr(s.find("@") + 1);
+            ClassId classId = ObjectFactory::GetInstance()->GetClassId(className);
+            m_svgAdditionalAttributes.insert({ classId, attributeName });
+        }
```
wergo: Do we have to check whether the classId (or the attribute string) exists before inserting it to the multimap?

include/vrv/svgdevicecontext.h
```diff
@@ -225,6 +225,7 @@ class SvgDeviceContext : public DeviceContext {
             std::string className = s.substr(0, s.find("@")); // parse <element@attribute>, e.g., "note@pname"
             std::string attributeName = s.substr(s.find("@") + 1);
             ClassId classId = ObjectFactory::GetInstance()->GetClassId(className);
+            assert(classId);
```
wergo: Do we have to check whether the classId (or the attribute string) exists before inserting it to the multimap?

lpugin: No, you don't want an `assert` here because this is going to make Verovio crash if the user passes a parameter with a typo in the element name or an element name that is not supported by Verovio. I think what you can do is simply ignore it (i.e., not add it to the map) if `(classId == OBJECT)`

include/vrv/svgdevicecontext.h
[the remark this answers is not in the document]
lpugin: Don't forget this small change

src/svgdevicecontext.cpp
```diff
@@ -974,6 +976,29 @@ void SvgDeviceContext::AppendIdAndClass(std::string gId, std::string baseClass,
...
+        object->GetAttributes(&attributes);
+        for (ArrayOfStrAttr::iterator iter = attributes.begin(); iter != attributes.end(); ++iter) {
+            if (it->second == (*iter).first) // ...and attribute exists in class name, add it to SVG element
+            {
+                m_currentNode.append_attribute(("data-" + it->second).c_str()) = (*iter).second.c_str();
+                found = true;
+            }
+        }
+        if (!found)
+            LogWarning("No attribute '%s' in '%s', nothing added to SVG.", it->second.c_str(), baseClass.c_str());
```
lpugin, line 998: I don't think want a warning if an element does not have the attribute, do we? I think you can remove `bool found`

wergo: I removed this warning.

include/vrv/svgdevicecontext.h
wergo: Ok, great. Yes, wrong syntax is simply ignored now. Or, an error is thrown when the ClassId is not correct.

include/vrv/svgdevicecontext.h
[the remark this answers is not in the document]
wergo: Ups, yes, I forgot that, now in.

src/svgdevicecontext.cpp
```diff
@@ -974,6 +976,23 @@ void SvgDeviceContext::AppendIdAndClass(std::string gId, std::string baseClass,
     m_currentNode.append_attribute("class") = baseClass.c_str();
 }
 
+void SvgDeviceContext::AppendAdditionalAttributes(Object *object)
+{
+    std::string baseClass = object->GetClassName();
+    std::transform(baseClass.begin(), baseClass.begin() + 1, baseClass.begin(), ::tolower);
```
lpugin, line 982: As far as I can see you don't use this anymore. Please remove

wergo: True; have overlooked that. Removed now.

lpugin (approved):

Issue #1586, opened 2020-07-29
finalize page-based MEI 


rettinghaus:
As Verovio operates with it's own custoMEIzation, which you you even get out of Verovio, it would be good to have the possibility to re-feed those back to Verovio, i.e. Verovio could read paged-based MEI directly. Right now, this only produces errors. 

A proper customized schema would be nice to have, too.

lpugin:
This is a general feature request.

PR #2364, opened 2021-08-27
Fix export crash

DavidBauer1984:
Verovio currently crashes when doing a score based MEI export of the last page via `Toolkit::GetMEI()`. This seems to be a regression of #2347 .

This PR fixes the crash.


DavidBauer1984:
I close this one for now and see if there is a better fix.

lpugin:
OK. I was actually not aware that you actually rely on the page-based MEI. I am sorry about it. I though nobody used it since there was never any feedback / questions about it.

rettinghaus:
Actually there was #1586 ;-)

PR #2374, opened 2021-09-10
Fix some problems with MEI export

DavidBauer1984:
This PR fixes some of the problems with the score based, single page MEI export that occurred after merging #2346 . In particular, it fixes the crash when exporting the last page. The behaviour now is as follows:

- The score based full MEI export can be reloaded back into Verovio and even works with multiple mdivs.
- The single page, score based export looks exactly as before merging #2346 . As before, it cannot be reloaded and currently is not compatible with multiple mdivs.
- The page based export outputs the new milestone elements `systemElementEnd` and `pageElementEnd` to mark the end of a section, score or mdiv.


DavidBauer1984:
Evaluating the test suite has failed. Sounds like a python package problem?!

BuildFail

lpugin:
It seems to go through for the develop branch but to crash on the PR one. I suspect it fails on the slur example I added yesterday. Then cairo is crashing because the SVG is missing or corrupted. Could you try to merge develop first?

src/iomei.cpp
```diff
@@ -816,21 +820,34 @@ bool MEIOutput::WriteObject(Object *object)
...
-        return true;
+
+    if (m_scoreBasedMEI) {
+        if (object->Is({ PAGE, PAGES, SYSTEM })) {
+            return true;
+        }
+        if (object->Is({ MDIV, SCORE })) {
+            if (m_page >= 0) {
+                return true;
+            }
```
lpugin, line 835: Do you know when this is not true?

DavidBauer1984: This is only true when exporting a single page. The page number is zero based, so `m_page=0` means exporting page 1 and so on. When exporting all pages, then `m_page=-1` and the condition is false.

lpugin:
Looks good, thanks!

PR #2381, opened 2021-09-14
Check for missing staffDef

DavidBauer1984:
Verovio crashes when using staves with missing staffDef. This PR adds a simple check for `<staffGrp>` to ensure that it contains at least one `<staffDef>`.


lpugin:
As a side note: we do not want too many of these because preferably data should be checked beforehand.

PR #2388, opened 2021-09-21
Do not ignore system scoreDef clefs

### LINKED ISSUE #1960: --adjust-page-height in Verovio 3.1 not working properly
If you are just interested in the pure bounding box of a music example (without margins at all) I use these arguments with Verovio:
  `verovio -t svg -f pae --adjust-page-height --page-width 1500 --page-margin-bottom 0 --page-margin-top 0 --page-margin-left 0 --page-margin-right 0 -a incipit.pae`
 for this PAE code:
```c++
       @clef:C-1
       @keysig:bBEA
       @timesig:c
       @data:=3/2-(4-)'B/2''E4.'G8G/2B4E''E/ED-'B/
```
Unfortunately in Version 3.1 the music is wrongly positioned and lower parts of the soprano clef are cut off:

  Earlier versions of Verovio calculated the margin correctly. 
Here the output  of the same code and same options in Verovio 2.7:

  Is this a bug or do I have to use different options in Verovio 3.1?
  Thank you.

Monceber:
closes #1960

Fix for the issue that BB of scoreDef clefs is being ignored which, with options like `--adjust-page-height` might lead to clef truncation:
Before:
image
After fix:
image

However, given that BB is not ignored anymore leads placing first staff that has G-clef a bit lower (since upper part of the clef is now counted towards overlaps of the system):
image
After:
image

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -1826,10 +1826,7 @@ int Object::SetOverflowBBoxes(FunctorParams *functorParams)
         assert(currentLayer);
         // set scoreDef attr
         if (currentLayer->GetStaffDefClef()) {
-            // Ignore system scoreDef clefs - clefs changes withing a staff are still taken into account
-            if (currentLayer->GetStaffDefClef()->GetScoreDefRole() != SCOREDEF_SYSTEM) {
-                currentLayer->GetStaffDefClef()->SetOverflowBBoxes(params);
-            }
+            currentLayer->GetStaffDefClef()->SetOverflowBBoxes(params);
         }
         if (currentLayer->GetStaffDefKeySig()) {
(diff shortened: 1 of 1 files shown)


lpugin:
> However, given that BB is not ignored anymore leads placing first staff that has G-clef a bit lower (since upper part of the clef is now counted towards overlaps of the system):

Yes, disabling the scoreDef clef bounding boxes was a (bad) quick fix to avoid the staff position to move up or down when the clef is change, because this is weird and confusing. Let me have a look if we can find another solution.

rettinghaus:
@lpugin any new ideas on this?

lpugin:
Fixed with #2629

PR #2394, opened 2021-09-24
Implement text circle enclosure

Monceber:
Added support for the ```@rend="circle"```:
image


lpugin:
Could you add test file? Thanks!

Monceber:
Opened PR with test file:
https://github.com/rism-digital/verovio.org/pull/54

PR #2402, opened 2021-10-01
Verovio MIDI output for tablature

paul-bayleaf:
Verovio MIDI output for tablature

### DIFF
--- a/src/course.cpp
+++ b/src/course.cpp
@@ -21,10 +21,13 @@ namespace vrv {
 
 static ClassRegistrar<Course> s_factory("course", COURSE);
 
-Course::Course() : Object("course-"), AttNNumberLike()
+Course::Course() : Object("course-")
 {
+    RegisterAttClass(ATT_ACCIDENTAL);
     RegisterAttClass(ATT_NNUMBERLIKE);
-
+    RegisterAttClass(ATT_OCTAVE);
+    RegisterAttClass(ATT_PITCH);
+    
     Reset();
 }
 
@@ -33,7 +36,10 @@ Course::~Course() {}
 void Course::Reset()
 {
     Object::Reset();
+    ResetAccidental();
     ResetNNumberLike();
+    ResetOctave();
+    ResetPitch();
 }
 
 bool Course::IsSupportedChild(Object *child)
--- a/src/course.cpp
+++ b/src/course.cpp
@@ -21,13 +21,13 @@ namespace vrv {
 
 static ClassRegistrar<Course> s_factory("course", COURSE);
 
-Course::Course() : Object("course-")
+Course::Course() : Object("course-"), AttAccidental(), AttNNumberLike(), AttOctave(), AttPitch()
 {
     RegisterAttClass(ATT_ACCIDENTAL);
     RegisterAttClass(ATT_NNUMBERLIKE);
     RegisterAttClass(ATT_OCTAVE);
     RegisterAttClass(ATT_PITCH);
-    
+
     Reset();
 }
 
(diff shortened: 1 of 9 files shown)


src/course.cpp
```diff
@@ -21,10 +21,13 @@ namespace vrv {
 
 static ClassRegistrar<Course> s_factory("course", COURSE);
 
-Course::Course() : Object("course-"), AttNNumberLike()
```
rettinghaus, line 24: Why remove AttNumberLike?

paul-bayleaf: Explicit initialization of a base class with its default constructor is not necessary.  Base classes are implicitly initialized with their default constructor.   Explicit base class initialization is only required for for non-default constructors, e.g. Object("course-").   This class now has 5 base classes, but only 1 needs explicit initialization.   So rather than add another 3 unnecessary initializations I removed AttNumberLike.

lpugin: That is right, they are not necessary. However, we have them everywhere else - one reason being that they allow to see straight away that the calls to `RegisterAttClass` in the constructors are all there. So please add them for this reason and for code consistency.

src/tuning.cpp
```diff
@@ -65,4 +67,97 @@ int Tuning::CalcPitchPos(int course, data_NOTATIONTYPE notationType, int lines)
...
 
+int Tuning::CalcPitchNumber(int course, int fret, data_NOTATIONTYPE notationType)
+{
+    // Use <tuning><course>s, if available
+    // else use @tuning.standard, if available
+    // else use @notationType
+    
+    // Do we have the tuning for this course?
+    AttNNumberLikeComparison cnc(COURSE, std::to_string(course));
+    Course* const courseTuning{vrv_cast<Course*>(FindDescendantByComparison(&cnc))};
```
lpugin, line 78: We do
```cpp
Course *courseTuning = vrv_cast<Course *>(FindDescendantByComparison(&cnc));
```

src/tuning.cpp
```diff
@@ -65,4 +67,97 @@ int Tuning::CalcPitchPos(int course, data_NOTATIONTYPE notationType, int lines)
...
+    
+    if (courseTuning && courseTuning->HasPname() && courseTuning->HasOct()) {
+
+        // Distance in semitones from the octave's starting C to the given note
+        int midiBase{0};
```
lpugin, line 83: We usually do not use value initialization. Does it really matter?

src/tuning.cpp
```diff
@@ -65,4 +67,97 @@ int Tuning::CalcPitchPos(int course, data_NOTATIONTYPE notationType, int lines)
...
+        
+        // As this does not represent historical notation of any kind the only
+        // accidentals we should ever see are "s" and "f"
+        const int alter{(courseTuning->HasAccid())
+            ? (courseTuning->GetAccid() == ACCIDENTAL_WRITTEN_s)
+                ? 1
+                : (courseTuning->GetAccid() == ACCIDENTAL_WRITTEN_f)
+                ? -1
+                : 0
+            : 0};
```
lpugin, line 105: I find this hard to read... Wouldn't it be clearer with some `if/else` or a `switch` statement?

lpugin (changes requested):
Thanks! My comments are mostly coding style related. On this topic, please make sure your run clang-format. See https://book.verovio.org/contributing/guidelines.html#formatting

lpugin:
@paul-bayleaf I clarified the issue with `guitar.drop.D` being wrongly imported as `guitar-drop-D` in the develop branch and merged it here - the dotted version is the correct one. It is now fixed but that created a couple of small conflicts to resolve.

PR #2409, opened 2021-10-05
Clean up slur options

DavidBauer1984:
This PR cleans up the slur options:
* The options for `slurControlPoints`, `slurMinHeight`, `slurMaxHeight` and `slurHeightFactor` are removed.
* The range for `slurCurveFactor` is adjusted.
* There is a new option `slurMargin`. It can be used to set the safety distance in MEI units by which slurs curve around obstacles.

| Slur margin 1 (default) | Slur margin 2 |
| ---------------------- | ------------- |
| Margin-Standard | Margin2 |

### DIFF
--- a/include/vrv/boundingbox.h
+++ b/include/vrv/boundingbox.h
@@ -211,9 +211,10 @@ public:
     static Point CalcPointAtBezier(const Point bezier[4], double t);
 
     /**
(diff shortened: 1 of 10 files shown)


lpugin:
Looks good to me. I leave it open for another day in case somebody else has something to ask about the change for the options.

PR #2433, opened 2021-10-15
Fix cross-staff tuplet beams

rettinghaus:
This PR fixes an issue with cross-staff beams causing too much space because of the tuplet number positioning. 

### before
image

### after
image


lpugin:
Is the example from the test suite? If not please add it. Thanks

PR #2448, opened 2021-10-25
Allow lb in textual control elements

rettinghaus:
This PR allows the use of `lb` within `anchoredText`, `dynam`, `fing`, `harm`, and `tempo`.
(Also it moves `lv` to the correct place in Xcode.)


src/fing.cpp
```diff
@@ -43,7 +43,7 @@ void Fing::Reset()
 
 bool Fing::IsSupportedChild(Object *child)
 {
-    if (!child->Is({ REND, TEXT })) {
+    if (child->Is({ LB, REND, TEXT })) {
```
lpugin, line 46: This change flips the logic and is wrong. Please fix it, or even better make the check consistent with other `::IsSupportedChild` methods

rettinghaus: Thanks for the review. 
(Reminder to myself: Quick copy paste always needs checking.)

PR #2458, opened 2021-11-01
Add standard double whole note

Monceber:
- two glyphs added: `noteheadDoubleWholeSquare` and `noteheadDoubleWhole `
- added support for `dblwhole` notehead shape 
- changed code to use new `noteheadDoubleWholeSquare` glyph instead of drawing it with primitives

image


lpugin:
Please fix the conflict. Thanks

PR #2470, opened 2021-11-06
Make xml:ids of expanded sections predictable

wergo:
Currently, when executing an `--expand` command, cloned sections (or endings, lemma, readings) obtain new random `xml:ids`. However, in order to reproduce the `xml:id`s of expanded sections such as for previously matched  performances, we need to be able to predict them from the original MEI score. 

This PR introduces a predictable scheme for added `xml:id`s in the form `"[oldId]-rendx"` while `x` being the nth rendering of that section. E.g. a `note-000321` in the score of a Minuet would appear as `note-000321-rend2` in its expanded repetition, and as `note-000321-rend2` in its first da-capo appearance (and as `note-000321-rend3` if  repeated in da capo).

### DIFF
--- a/src/expansionmap.cpp
+++ b/src/expansionmap.cpp
@@ -72,18 +72,19 @@ void ExpansionMap::Expand(const xsdAnyURI_List &expansionList, xsdAnyURI_List &e
                 // clone current section/ending/rdg/lem and rename it, adding -"rend2" for the first repetition etc.
                 Object *clonedObject = currSect->Clone();
                 clonedObject->CloneReset();
+                GeneratePredictableIds(currSect, clonedObject);
                 clonedObject->SetUuid(currSect->GetUuid() + "-rend"
                     + std::to_string(GetExpansionIdsForElement(currSect->GetUuid()).size() + 1));
 
                 // get IDs of old and new sections and add them to m_map
                 std::vector<std::string> oldIds;
                 oldIds.push_back(currSect->GetUuid());
-                this->GetUuidList(currSect, oldIds);
+                GetUuidList(currSect, oldIds);
                 std::vector<std::string> clonedIds;
                 clonedIds.push_back(clonedObject->GetUuid());
-                this->GetUuidList(clonedObject, clonedIds);
+                GetUuidList(clonedObject, clonedIds);
                 for (int i = 0; (i < (int)oldIds.size()) && (i < (int)clonedIds.size()); i++) {
-                    this->AddExpandedIdToExpansionMap(oldIds.at(i), clonedIds.at(i));
+                    AddExpandedIdToExpansionMap(oldIds.at(i), clonedIds.at(i));
                 }
 
                 // go through cloned objects, find TimePointing/SpanningInterface, PListInterface, LinkingInterface
@@ -243,6 +244,18 @@ void ExpansionMap::GetUuidList(Object *object, std::vector<std::string> &idList)
     for (Object *o : *object->GetChildren()) {
         idList.push_back(o->GetUuid());
         GetUuidList(o, idList);
+    }
+}
+
+void ExpansionMap::GeneratePredictableIds(Object *source, Object *target)
+{
+    unsigned i = 0;
+    ArrayOfObjects targetObjects = *target->GetChildren();
+    for (Object *s : *source->GetChildren()) {
+        std::string id = s->GetUuid() + "-rend" + std::to_string(GetExpansionIdsForElement(s->GetUuid()).size() + 1);
+        targetObjects[i]->SetUuid(id);
+        GeneratePredictableIds(s, targetObjects[i]);
+        i++;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/expansionmap.cpp
```diff
@@ -72,18 +72,19 @@ void ExpansionMap::Expand(const xsdAnyURI_List &expansionList, xsdAnyURI_List &e
...
                 Object *clonedObject = currSect->Clone();
                 clonedObject->CloneReset();
+                GeneratePredictableIds(currSect, clonedObject);
                 clonedObject->SetUuid(currSect->GetUuid() + "-rend"
                     + std::to_string(GetExpansionIdsForElement(currSect->GetUuid()).size() + 1));
 
                 // get IDs of old and new sections and add them to m_map
                 std::vector<std::string> oldIds;
                 oldIds.push_back(currSect->GetUuid());
-                this->GetUuidList(currSect, oldIds);
```
lpugin, line 81: Keep `this->` (same elsewhere)

src/expansionmap.cpp
```diff
@@ -72,18 +72,19 @@ void ExpansionMap::Expand(const xsdAnyURI_List &expansionList, xsdAnyURI_List &e
                 // clone current section/ending/rdg/lem and rename it, adding -"rend2" for the first repetition etc.
                 Object *clonedObject = currSect->Clone();
                 clonedObject->CloneReset();
+                GeneratePredictableIds(currSect, clonedObject);
                 clonedObject->SetUuid(currSect->GetUuid() + "-rend"
                     + std::to_string(GetExpansionIdsForElement(currSect->GetUuid()).size() + 1));
```
lpugin, line 77: Can't you do this in `GeneratePredictableIds` before looping through the children? Then it is all in one place

src/expansionmap.cpp
```diff
@@ -246,4 +247,16 @@ void ExpansionMap::GetUuidList(Object *object, std::vector<std::string> &idList)
     }
 }
 
+void ExpansionMap::GeneratePredictableIds(Object *source, Object *target)
+{
+    unsigned i = 0;
+    ArrayOfObjects targetObjects = *target->GetChildren();
```
lpugin, line 253: It is not safe not to check that the number of children is the same.

src/expansionmap.cpp
```diff
@@ -246,4 +247,16 @@ void ExpansionMap::GetUuidList(Object *object, std::vector<std::string> &idList)
...
+    for (Object *s : *source->GetChildren()) {
+        std::string id = s->GetUuid() + "-rend" + std::to_string(GetExpansionIdsForElement(s->GetUuid()).size() + 1);
+        targetObjects[i]->SetUuid(id);
```
lpugin, line 256: Always use `.at()` instead of `[]`

lpugin (changes requested):

src/expansionmap.cpp
wergo: Ok.

src/expansionmap.cpp
wergo: Done.

src/expansionmap.cpp
wergo: I was assuming that a clone has the same size as its source, but now I am checking.

src/expansionmap.cpp
wergo: Sure, too much javascript... sorry.

src/expansionmap.cpp
```diff
@@ -243,18 +241,25 @@ void ExpansionMap::GetUuidList(Object *object, std::vector<std::string> &idList)
...
-    unsigned i = 0;
+    ArrayOfObjects sourceObjects = *source->GetChildren();
     ArrayOfObjects targetObjects = *target->GetChildren();
-    for (Object *s : *source->GetChildren()) {
-        std::string id = s->GetUuid() + "-rend" + std::to_string(GetExpansionIdsForElement(s->GetUuid()).size() + 1);
-        targetObjects[i]->SetUuid(id);
-        GeneratePredictableIds(s, targetObjects[i]);
+    if (sourceObjects.size() != targetObjects.size()) return;
+
+    target->SetUuid(
```
lpugin, line 254: You need to move this at the top of the method because it should be done anyway

src/expansionmap.cpp
```diff
@@ -243,18 +241,25 @@ void ExpansionMap::GetUuidList(Object *object, std::vector<std::string> &idList)
...
+        source->GetUuid() + "-rend" + std::to_string(this->GetExpansionIdsForElement(source->GetUuid()).size() + 1));
+
+    unsigned i = 0;
+    for (Object *s : sourceObjects) {
+        std::string id
+            = s->GetUuid() + "-rend" + std::to_string(this->GetExpansionIdsForElement(s->GetUuid()).size() + 1);
+        targetObjects.at(i)->SetUuid(id);
```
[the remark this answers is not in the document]
wergo: Thanks for spotting this! This is now done only at the beginning of the method. (And some local methods made private.)

PR #2478, opened 2021-11-10
Enforce minimal white space between systems

DavidBauer1984:
This PR ensures that there is a minimal white space of vertically one staff space between consecutive systems. This prevents the content below (pedal lines, slurs) from being very close to the content above of the following system (slurs, ottava lines). Before this sometimes happened (even with vertical justification).

| Before | After |
| ------ | ----- |
| Before | After |

### DIFF
--- a/src/clef.cpp
+++ b/src/clef.cpp
@@ -282,7 +282,7 @@ int Clef::AdjustClefChanges(FunctorParams *functorParams)
 
     // This should never happen because we always have at least barline alignments - even empty
     if (!previousAlignment || !nextAlignment) {
-        LogDebug("No aligment found before and after the clef change");
+        LogDebug("No alignment found before and after the clef change");
         return FUNCTOR_CONTINUE;
     }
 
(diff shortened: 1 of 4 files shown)


craigsapp:
It would be very useful if there were a user controllable option for this minimal spacing.  Currently the minimum distance is hardwired to 2 diatonic steps.  Perhaps allowing a range from 0 to 100 with a default of 2.

Maybe `--staff-min-gap` and `--system-min-gap` might be good names for such an option, and ossibly also `--group-min-gap` (or maybe `--gap-min-staff` and `--gap-min-system` would be more consistent names with existing option names).   

The reason for having a separate staff and system minimum gap is that having an adjustable value would be useful in particular for the system gap, but less so for the staff gap within a system.    For piano music (the test piece is a good example), I have to set the  `--spacing-staff` value quite high to avoid visual collisions that could be solved with `--gap-min-system` set larger that would allow `--spacing-staff` to be set smaller that I currently do.

rettinghaus:
@craigsapp basically we have this option already in place with `spacingSystem`. This mimics scoreDefs `spacing.system` attribute (see https://music-encoding.org/guidelines/v4/attribute-classes/att.spacing.html).
However, the MEI description is not totally clear if this should apply to the spacing between the outer staves (like `spacing.staff`) or indeed the *complete* system including slurs, octave lines, etc.
In the end it is desirable to control both these approaches, but I'd prefer to check back with the MEI community first. You may open an issue over there.

DavidBauer1984:
@lpugin OK thank you, I see the problem. I'll check if we can rescue this by not unnecessarily
moving any systems.

PR #2475, opened 2021-11-09
Rendering portato slurs

DavidBauer1984:
This PR improves the rendering of portato slurs. Works with stem side or centered staccato as well as with other articulations.

| Before | After |
| ------ | ----- |
| Before1 | Example1 |
| Before2 | Example2 |
| Before3 | Example3 |

<details>
  <summary>Show MEI example</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title />
            <respStmt />
         </titleStmt>
         <pubStmt><date isodate="2021-07-30" type="encoding-date">2021-07-30</date>
         </pubStmt>
      </fileDesc>
      <encodingDesc xml:id="encodingdesc-xwopln">
         <appInfo xml:id="appinfo-a209u3">
            <application xml:id="application-oqlg5f" isodate="2021-10-07T15:48:50" version="3.7.0-dev-ab9a2ad">
               <name xml:id="name-54q3e8">Verovio</name>
               <p xml:id="p-9bogng">Transcoded from MusicXML</p>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv xml:id="mlj4v68">
            <score xml:id="sdie2kd">
               <scoreDef xml:id="sqosio4">
                  <staffGrp xml:id="s2kukk6">
                     <staffGrp xml:id="S1-ty9q8jPDVsCWNSEmZ4EMBA" bar.thru="true">
                        <staffDef xml:id="sbx4ten" n="1" lines="5" ppq="4">
                           <clef xml:id="clef-1" shape="G" line="2" />
                           <keySig xml:id="key-1" sig="2f" />
                           <meterSig xml:id="time-1" count="4" sym="common" unit="4" />
                        </staffDef>
                        <staffDef xml:id="spk2jp7" n="2" lines="5" ppq="4">
                           <clef xml:id="clef-2" shape="F" line="4" />
                           <keySig xml:id="key-2" sig="2f" />
[...]

PR #2479, opened 2021-11-10
Fix articulation size

rettinghaus:
As @craigsapp pointed out in #2475 staccato dots are too small, which caught my attention, because this is not a font issue. 
Interestingly all articulation was hard coded to cue size, which I removed now.

See example `artic-001` from the test suite with Leland music font:
### before
artic-leland
### after
artic-leland-new


craigsapp:
I had a vague feeling that the other articulations seems a bit small as well... Now we know why.

lpugin:
Well spotted!

wergo:
Thanks a lot @rettinghaus, I always wanted to mention this! Looks great now.

PR #2481, opened 2021-11-10
Update ranges in supported.xsl

rettinghaus:
This adds several missing ranges and new glyphs from SMuFL 1.4 to `supported.xsl` and formats it. 
It doesn't add new glyphs to Verovio directly, but some of those might come in handy soon (or were strongly missed) like the ones from `Medieval and Renaissance prolations supplement`.

### DIFF
--- a/fonts/supported.xsl
+++ b/fonts/supported.xsl
@@ -3,1005 +3,1005 @@
     <xsl:variable name="supported">
         <staffBracketsAndDividers>
             <description>Staff brackets and dividers</description>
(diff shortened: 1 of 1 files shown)


lpugin:
@rettinghaus any ideas why this changes triggered changes in so many test-suite examples? https://github.com/rism-digital/verovio/actions/runs/1445808864

lpugin:
Forget it. They were caused by this one https://github.com/rism-digital/verovio/pulls?q=is%3Apr+is%3Aclosed

PR #2486, opened 2021-11-15
Adjust ledger flag overlap

Monceber:
In some case flags on notes with up stem direction would overlap with ledger lines still:
image

Made small adjustment to make sure that this no longer happens - instead of taking middle point of flag (anchor point on the stem), endpoint of the flag is now being compared to the ledger lines. Previously it was done only for notes with down stem direction and now we doing same calculation for both directions.

No more overlaps:
image

### DIFF
--- a/src/elementpart.cpp
+++ b/src/elementpart.cpp
@@ -15,6 +15,7 @@
 
 #include "btrem.h"
 #include "chord.h"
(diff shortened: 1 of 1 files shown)


lpugin:
It looks that stems up and down do not have the same 

image

PR #2498, opened 2021-11-18
Refactor DrawBarLine and update fonts

rettinghaus:
This PR refactors the calculation and drawing of bar lines. We now use glyphs for repetition bar line dots as suggested in the [SMuFL implementation notes](https://www.w3.org/2019/03/smufl13/tables/repeats.html#implementation-notes) (although the differences are subtle at best). 
For that the needed glyphs were added to [Leipzig](https://github.com/rism-digital/leipzig/pull/10/commits/d10c3b2d204a03f77b7e5e240a47faad7f9b80ee) and Leland is updated to latest version 0.73. 
It also fixes a small bug in distance of repetition bar lines.


setup.py
```diff
@@ -72,9 +72,9 @@ def get_version():
             version += '.dev'
     if version.endswith(".dev"):
         init_sha = subprocess.getoutput(
-            "git log -n 1 --pretty=format:%H -- bindings/python/.pypi-version")
+            'git log -n 1 --pretty=format:%H -- bindings/python/.pypi-version')
         count = subprocess.getoutput(
-            "git rev-list --count HEAD \"^{}\"".format(init_sha))
+            'git rev-list --count HEAD \"^{}\"'.format(init_sha))
```
lpugin, line 77: If you change to single quotes, why are you still escaping double quotes?

src/view_page.cpp
```diff
@@ -873,52 +871,47 @@ void View::DrawBarLine(
...
-    }
-    else if (form == BARRENDITION_dbldotted) {
-        DrawVerticalSegmentedLine(dc, x, line, barLineWidth, dotLength);
-        DrawVerticalSegmentedLine(dc, x2, line, barLineWidth, dotLength);
-    }
-    else {
-        // Use solid barline as fallback
-        LogWarning("%s bar lines not supported", barLine->AttBarLineLog::BarrenditionToStr(barLine->GetForm()).c_str());
-        DrawVerticalSegmentedLine(dc, x, line, barLineWidth);
+    switch (form) {
```
lpugin, line 875: May I ask: what is the reason for changing to a `switch`?

setup.py
rettinghaus: Thanks for pointing me on this. This slipped in this PR, so I tried now to get a more consistent use of single quotes. (And fixed this here.)

src/view_page.cpp
rettinghaus: I find it more pleasant to read. Also we have this in many other places when it comes to data types.

lpugin: I find it less pleasant, but that's not important.

PR #2521, opened 2021-12-01
update and sort imports for iOS [skip-ci]

rettinghaus:


lpugin:
@rettinghaus when I rebuild the framework in xCode the order of the header files is changed. This means that the changes made in d57ffef015d62affa7550392500e03d69e925cff will eventually be overwritten.

PR #2531, opened 2021-12-06
Develop beamlet fixes

### LINKED ISSUE #2518: beam flag direction
Verovio choses the side of the stem that beam flags (beamlets) are placed.  Usually this is done correctly, but occasionally not, sometimes because how to do it may be ambiguous.

Here is an example target rendering:

image

and here is the actual rendering:

Screen Shot 2021-11-30 at 12 59 40 PM

I propose to use the `note@breaksec` to control the direction of the flag on the third note.  In the following data that note has `@breaksec="1"` which means to break one secondary beam.  In this case this causes the flag on the note.  The rule should be to point the flag to the left when there is a breaksec on the note.   There could be counter-example which would not work, but I cannot think of any at the time.

Alternatively (or in addition), the presence of the rest on one side of the note should be used to decide that the flag points towards the rest (it does not make sense to point a flag towards another note in the beam, particularly if it has the same rhythm) unless it is a dotted note).

<details>
<summary>
Click to view MEI data for above example
</summary>

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2021-11-30T12:56:42" version="3.8.0-dev-b3abbd0-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m9kuzp0">
    <score xml:id="syt13mc">
     <scoreDef xml:id="sqp5ho8">
      <staffGrp xml:id="sp5ywqz">
       <staffDef xml:id="symv2zb" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1" right="invis">
       <staff xml:id="sml0oii" n="1">
         <beam xml:id="beam-L3F1-L8F1">
          <note xml:id="note-L3F1" dur="16" oct="6" artic="acc" pname="f" accid="s" />
          <note xml:id="note-L4F1" breaksec="1" dur="16" oct="6" pname="e" accid.ges="n" />
          <rest xml:id="rest-L5F1" dur="16" ploc="a" oloc="5" />
          <note xml:id="note-L6F1" breaksec="1" dur="16" oct="6" pname="e" accid.ges="n" />
          <note xml:id="note-L7F1" dur="16" oct="6" pname="f" artic="acc" accid.ges="s" />          <note xml:id="note-L8F1" dur="16" oct="6" pname="e" accid.ges="n" />
         </beam>
        </layer>
       </staff>
       <slur xml:id="slur-L3F1-L4F1" staff="1" startid="#note-L3F1" endid="#note-L4F1" />
       <slur xml:id="slur-L7F1-L8F1" staff="1" startid="#note-L7F1" endid="#note-L8F1" />
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

-----------------------

Unrelated issues:

(1) The beam should be horizontal according to standard music typesetting rules (note closest to the beam occurs more than once).

(2) The stems are too long and it would be better to be the same as the target rendering.  The problem is caused by trying to avoid the secondary beam, but that beam is not present under the 16th note rest.  For example, if the rest were moved up a bit, the beam position would become correct:

Screen Shot 2021-11-30 at 1 06 46 PM

(3) A more difficult thing is that the accent mark should go above the slur in this sort of case, because otherwise the slur has an exaggerated slope.  But I think it is not possible to control the stacking order of slurs/articulations yet in verovio (on a case by case basis)?

Monceber:
closes #2518

image

Fixes an issue with invalid beam slope when first notes of the beam have shorter duration but are have only 1 ledger line, and last notes have 2 or more:
image
After the fix:
image

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -250,7 +250,6 @@ public:
     int m_overlapMargin;
     int m_maxShortening; // maximum allowed shortening in half units
     bool m_centered; // beam is centered on the line
-    bool m_shortened; // stem is shortened because pointing oustide the staff
     data_BEAMPLACE m_beamRelativePlace;
     char m_partialFlags[MAX_DURATION_PARTIALS];
     data_BEAMPLACE m_partialFlagPlace;
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -68,6 +68,9 @@ public:
     int GetAdjacentElementsDuration(int elementX) const;
 
 private:
+    // Helper to adjust beam positioning with regards to ledger lines (top and bottom of the staff)
+    void AdjustBeamToLedgerLines(Doc *doc, Staff *staff, BeamDrawingInterface *beamInterface);
+
     void CalcBeamInit(Layer *layer, Staff *staff, Doc *doc, BeamDrawingInterface *beamInterface, data_BEAMPLACE place);
 
     bool CalcBeamSlope(
@@ -79,8 +82,6 @@ private:
 
     // Helper to adjust position of starting point to make sure that beam start-/endpoints touch the staff lines
     void CalcAdjustPosition(Staff *staff, Doc *doc, BeamDrawingInterface *beamInterface);
-
-    void CalcStemLenInHalfUnitsgth(Layer *layer, Staff *staff, Doc *doc, BeamDrawingInterface *beamInterface);
 
     void CalcBeamPlace(Layer *layer, BeamDrawingInterface *beamInterface, data_BEAMPLACE place);
 
--- a/include/vrv/drawinginterface.h
+++ b/include/vrv/drawinginterface.h
@@ -122,17 +122,17 @@ public:
     bool HasOneStepHeight();
 
     /**
+     * Get total beam width with regards to the shortest duration of the beam (counting both beams and whitespace
+     * between them)
+     */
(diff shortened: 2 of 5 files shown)


lpugin:
Looks great, thanks. Please double check beam-036 and a few others that are now too long.

image

craigsapp:
There is a problem with the rest vertical position in this example:

Screen Shot 2021-12-07 at 9 04 34 AM

The rest needs to do up or down one diatonic step (MEI vu).  Probably down is best.

Monceber:
Changed implementation a bit, should be more universal and fix both the problem and retain previous functionality:
image
beam-036
image
beam-031 and beam-032
image
image


@craigsapp Regarding your point, this was but a matter of encoding for the rest in MEI file - rest had implicitly specified position via @oloc/@ploc. With automatic placement it looks as following:
image

lpugin:
The beams are now much better. There is a side effect on cross-staff ones, though. See cross-staff-010

image

PR #2543, opened 2021-12-15
Change floating objects order (place fingerings before hairpins)

Monceber:
Changed from:
image
To:
image


lpugin:
Looks good to me! Other opinions?

PR #2544, opened 2021-12-16
Filtering in the score based export

DavidBauer1984:
This PR enhances the filtered score based export:
- The (filtered) score based export now always generates a valid MEI which can be loaded back in Verovio. In particular, this is compatible with multiple mdivs.
- New filtering options: it is now possible to filter a page range (provide `firstPage` and `lastPage` in the json as int), by measure range (provide `firstMeasure` and `lastMeasure` in the json as string containing the measures uuids), by mdiv (provide `mdiv` in the json as string containing the mdiv uuid), or by a combination of those. For backward compatibility I kept the possibility to provide `pageNo` to filter a single page.
- Adapted scoredef: when exporting a part in the middle of a score, the scoredef is adjusted to reflect changes in clef, key signature, etc. Also the instrument labels are removed or replaced by abbreviations. The following example shows the rendered export of single measures:

| After clef change | After key signature change |
| ----------------- | --------------------------- |
| M77 | M99 |

- It is possible to ignore the MEI header, which likely makes sense when exporting a part in the middle of a score. For this provide `ignoreHeader` in the json with true.

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -31,6 +31,7 @@
 #include "choice.h"
 #include "chord.h"
 #include "clef.h"
+#include "comparison.h"
 #include "corr.h"
 #include "course.h"
 #include "custos.h"
@@ -257,10 +258,10 @@ std::string MEIOutput::GetOutput()
 
 bool MEIOutput::WriteObject(Object *object)
 {
-    return this->WriteObject(object, true);
+    return this->WriteObject(object, true, false);
 }
 
-bool MEIOutput::WriteObject(Object *object, bool handleScoreBasedFilter)
+bool MEIOutput::WriteObject(Object *object, bool handleScoreBasedFilter, bool useCustomScoreDef)
 {
     if (this->IsScoreBasedMEI() && handleScoreBasedFilter) {
         // Update current page etc.
@@ -270,6 +271,7 @@ bool MEIOutput::WriteObject(Object *object, bool handleScoreBasedFilter)
             // Transition Before => Here
             if (m_filterMatchLocation == MatchLocation::Before) {
                 m_filterMatchLocation = MatchLocation::Here;
+                m_firstFilterMatch = object;
                 this->WriteStackedObjects();
             }
         }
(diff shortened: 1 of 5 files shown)


src/iomei.cpp
```diff
@@ -171,81 +175,58 @@ bool MEIOutput::Export()
...
+        decl.append_attribute("href") = schema.c_str();
+        decl.append_attribute("type") = "application/xml";
+        decl.append_attribute("schematypens") = "http://relaxng.org/ns/structure/1.0";
+
+        // schematron processing instruction - currently not working for page-based MEI
+        if (!this->IsPageBasedMEI()) {
             decl = meiDoc.append_child(pugi::node_declaration);
             decl.set_name("xml-model");
-            decl.append_attribute("href") = schema.c_str();
+            decl.append_attribute("href") = "https://music-encoding.org/schema/dev/mei-all.rng";
```
lpugin, line 213: Why not use `schema.c_str()` ?

lpugin:
Will it work to apply a filter if the data has never been rendered? I am thinking of the scoreDef adjustments based on the system drawing scoreDef. Could you double check?

DavidBauer1984:
It is required that layout was calculated, but the document must not necessarily have been rendered. The scoreDefs that are used are calculated in the `ScoreDefSetCurrent` functor which is called at the begin of `CastOffDocBase`. As far as I know it is not possible to load a document without doing the layout from the toolkit API. So as long as the document is loaded, we should be on the safe side here.

lpugin:
Could you explain the logic with the `m_objectStack`? I would expect that when no filter is applied saving would remain straight-forward. However, when looking at this 
```cpp
bool MEIOutput::WriteObject(Object *object)
{
    if (this->IsScoreBasedMEI()) {
        if (!this->ProcessScoreBasedFilter(object)) {
            return true;
        }
    }
    return this->WriteObjectInternal(object, false);
}
```
and then at what is happening in `bool MEIOutput::ProcessScoreBasedFilter(Object *object)` I am not sure it does. I would expect that you would simply check if a filter is applied at all, and if not return false in order to proceed as before (basically for `WriteObject` to call `WriteObjectInternal`). Or is this what is happening?

DavidBauer1984:
When no filter is set, then `IsMatchingFilter` always returns true. To see this, note that in the first call of `UpdateFilter` the match locations for measure and mdiv are set to a matching state (L1017-1019 & L1055-1057).

Hence in the first call of `ProcessScoreBasedFilter` the `m_filterMatchLocation` already performs the transition from before to here. At this point the object stack is empty, so no stacked objects are written. `m_filterMatchLocation` then remains on here for the rest of the export. The object stack runs silently in the background (pushing and popping objects), but is never used to write anything.

Having no filter somehow corresponds to the edge case of a very large filter containing the whole document. I decided to keep the object stack silently running in the background, since maybe it's useful for other purposes. If you want, I could switch it off by checking `this->HasFilter()`?

DavidBauer1984:
Anyway, I have added an additional check, if a filter is present. This should make this part of the code more clear.

PR #2560, opened 2022-01-04
Use short int in MusicXML import

rettinghaus:
This updates the MusicXML importer by using short int wherever safely possible. 
Also it now respects the `print-object` value for metronome marks.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run clang-format style check for C/C++ programs.
-        uses: jidicula/clang-format-action@v3.4.0
+        uses: jidicula/clang-format-action@v4.4.1
         with:
           clang-format-version: "11"
           check-path: ${{ matrix.path['check'] }}
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -192,8 +192,8 @@ private:
      */
     ///@{
     bool ReadMusicXmlPart(pugi::xml_node node, Section *section, short int nbStaves, const short int staffOffset);
-    bool ReadMusicXmlMeasure(
-        pugi::xml_node node, Section *section, Measure *measure, short int nbStaves, const short int staffOffset, int index);
+    bool ReadMusicXmlMeasure(pugi::xml_node node, Section *section, Measure *measure, short int nbStaves,
+        const short int staffOffset, int index);
     ///@}
 
     /*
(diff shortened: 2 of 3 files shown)


lpugin (changes requested):
Please double check the test suite evaluation which fails on the MusicXML import

rettinghaus:
Yes, checking right now.

lpugin:
I really have bad memories of problems arising with with this type of unsigned / signed and type conversions and casts. When there is a problem they are a real pain to debug.

Can you examplain why `short int staffNum = clef.attribute("number").as_int();` does not work and why you changed to `short int staffNum = clef.attribute("number").as_uint();` I am not sure I understand because `short int` is signed and it does not look clean to me.

(For this optimization to have a real impact (assuming it does not have a CPU overhead) you would need to convert hundreds of files in parallel with hundreds of slurs in each of them, and I am pretty sure you will have memory issues elsewhere way before.)

DavidBauer1984:
Here are my 50 cents to the discussion.

**1. Using unsigned variables**
Please don't use unsigned variables (local or class members) to indicate that the variables content is semantically nonnegative. This introduces very subtle bugs since any check `number >= 0` suddenly always is true by definition. The Google C++ style guide recommends avoiding unsigned integers except in situations that definitely require it (for example file formats sizes).
[Some further reading on using unsigned](https://wesmckinney.com/blog/avoid-unsigned-integers/) 
I wouldn't worry too much about conversions from unsigned to signed types. We already use them a lot when considering the size of a container (`size_t` is unsigned). Example:
`const int layerCount = (int)objects.size();`

**2. Using short instead of int**
This has a tiny benefit on memory, it saves you 2 bytes. On the other hand, it adds some headache, because each time you have to ask yourself: should I use short or int? Is it save to convert between the two of them? Using short in arithmetic operations is not faster, since the int type in C is usually the size of your standard CPU register which matters for all calculations.
[Some further reading on using short](https://stackoverflow.com/questions/24371077/when-to-use-short-over-int)
I think it should be ok to use short here and there, but we shouldn't mix short and int on a large scale in the codebase.

lpugin:
Yes, `(int)objects.size()` is one case where we do it all the time. It is not great but it is a clearly identified case. The idea behind it is to reduce the number of types we use. Essentially `int`, and `char` for cases where that is clearly appropriate.  But we stayed away from `short`as much as possible for the different reasons given in this thread. 

I fully agree that mixing `int` and `short` in the codebase would not be as a right move, and so wouldn't using `unsigned` variables.

I am still interested to know why changing to unsigned `as_uint()` fixes the issue introduced with the switch to `short`. Any ideas? I don't think we want to leave it like this.

rettinghaus:
The issue had nothing to do with `short`/`uint`. It came from switching to `std::stoi` to avoid extra casting a `cstr`. But that failed with `GetN` because n can be empty. All the other cases it checks if the string is empty.

lpugin:
OK. Have you removed them? There were here https://github.com/rism-digital/verovio/commit/9499d0cbde4c14d560e57aa457bd25fb7d07fef0 but now I cannot see them any more... Sorry if I am overlooking something

rettinghaus:
@lpugin yes, I dropped the last commit with `as_uint()`.

lpugin:
I added a note to the contribution guidelines.
https://book.verovio.org/contributing/guidelines.html#integer-data-types

lpugin (approved):

PR #2567, opened 2022-01-12
Fix missing hyphens

### LINKED ISSUE #2565: Hyphens missing between lyric syllables (regression from 3.7.0)
**Describe the problem**
I have several MXL files (compressed MusicXML) that I convert to MEI and render on the fly using Verovio's JavaScript toolkit. Recently, I've noticed that many of the hyphens between lyric syllables aren't being rendered. This does not occur when I manually set the Verovio version to 3.7.0, so it seems to be a regression from a change that occurred after 3.7.0.

Example page where sheet music is being rendered:
https://singpraises.net/collections/en/hymns/arise-o-glorious-zion?tab=sheet-music-interactive

Original MXL file:
https://assets.ldscdn.org/66/20/66206ef3a936eba658d11bdaec2299eaa960586f/arise_o_glorious_zion.mxl

Screenshot, rendered with Verovio 3.7.0:
Screen Shot 2022-01-09 at 4 49 02 PM

Screenshot, rendered with Verovio 3.8.0:
Screen Shot 2022-01-09 at 4 48 48 PM

(I did notice that the hyphen on the first word, "Arise," is missing with both Verovio 3.7.0 and 3.8.0 – I'm not sure why – but there are several more hyphens missing in 3.8.0.)

**To Reproduce**
1. Load the attached MXL or MusicXML file into Verovio.
2. View the rendered output.

**Expected behavior**
Several hyphens should appear in the output lyrics. On the first line of the first verse, there should be hyphens in "Arise," "glorious," "Zion," and "latter."

**Actual behavior**
Of the words mentioned above, only "glorious" has a hyphen.

**Input data**
I'm not sure if the hyphens are lost before or after converting to MEI, but for convenience, a ZIP file is attached with the original MXL file, an uncompressed MusicXML file, and the converted MEI file:
[arise-o-glorious-zion.zip](https://github.com/rism-digital/verovio/files/7836227/arise-o-glorious-zion.zip)

**Verovio information**
 - Verovio 3.8.0 (latest release) and 3.9.0-dev-2f4dccb
 - JavaScript toolkit

 - OS: macOS 12.1 Monterey (Intel)
 - Browser: Chrome, Safari

The MEI viewer on the Verovio website appears to have the same issue, so I don't think it's specific to my website code, but I can provide code if needed.

DavidBauer1984:
This PR fixes three problems causing missing syllable hyphens and closes #2565 .

1. The time spanning order check is skipped for syllables (red).
2. The musicxml importer can now deal with multiple `syllabic` elements within lyrics (blue).
3. The start and end of the time spanning interface is properly set for syllables between chords (green).
After

### DIFF
--- a/include/vrv/chord.h
+++ b/include/vrv/chord.h
@@ -207,6 +207,11 @@ public:
     int PrepareLayerElementParts(FunctorParams *functorParams) override;
 
     /**
+     * See Object::PrepareLyrics
+     */
+    int PrepareLyrics(FunctorParams *functorParams) override;
+
+    /**
      * See Object::CalcOnsetOffsetEnd
      */
     int CalcOnsetOffsetEnd(FunctorParams *functorParams) override;
--- a/include/vrv/functorparams.h
+++ b/include/vrv/functorparams.h
@@ -1896,7 +1896,8 @@ public:
 
 /**
  * member 0: the current Syl
- * member 1: the last Note
+ * member 1: the last Note or Chord
+ * member 2: the penultimate Note or Chord
  **/
 
 class PrepareLyricsParams : public FunctorParams {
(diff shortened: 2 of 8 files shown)


samuelbradshaw:
Thank you, @DavidBauer1984!

lpugin:
Example added to the test-suite https://github.com/rism-digital/verovio.org/commit/ccbbcae26a052e41e5698f562d1288b84cfc0da7

PR #2568, opened 2022-01-12
Implement MIDI for tablature

paul-bayleaf:
Implement MIDI for tablature.
(I had a pull request this earlier but seems to have disappeared, probably my error.)


lpugin:
I am not sure what happen. That one was merged https://github.com/rism-digital/verovio/pull/2402/files

As far as I can see, this PR contains some code that we discussed. Namely the https://github.com/rism-digital/verovio/pull/2568/files#diff-17a7bcdc2c2e792707144ffd6d70cbfb0b3129d832f9c9f4d1271767e64c5f87R718-R727 loop for adding notes of the previous staff. That caused the last note(s) not to be played. As far as I remember I passed you some information on how to change this and do it with and functor end.

lpugin:
@paul-bayleaf are you sure this PR is still valid? I think MIDI rendering has been merged already and that this code is obsolete. Please double check.

paul-bayleaf:
Yes this PR is still valid.  The develop-tab branch does not have the 'rule of holds' logic to extend a note's noteOff time beyond the note's rhythm.  This PR also implements the various new COURSETUNING_ values.

src/staff.cpp
```diff
@@ -716,4 +718,22 @@ int Staff::GenerateMIDI(FunctorParams *functorParams)
...
+        if (held.m_pitch > 0) {
+            params->m_midiFile->addNoteOff(params->m_midiTrack, held.m_stopTime * params->m_midiFile->getTPQ(),
+                params->m_midiChannel, held.m_pitch);
+        }
+    }
+
+    params->m_heldNotes.clear();
+
+    return FUNCTOR_CONTINUE;
+}
```
lpugin, line 737: This should be moved the the Layer. That is, move it to a `Layer::GenerateMIDIEnd(FunctorParams *functorParams)` with probably exactly the same code. I don't think we have seen encoding with multiple layers, but if we have some then the code should be triggered at the end of each layer and not of each staff (only).

lpugin (changes requested):
Great, thanks for checking.

lpugin (approved):

lpugin:
Thanks! And sorry for delaying the merge.

PR #2581, opened 2022-01-18
Merge latest development for tablature

lpugin:
* Adjustments to the Leipzig font and to the glyph rendering size
* Beam shortest values for `tab.lute.french` and `tab.lute.italian`


rettinghaus:
I think text shouldn't be scaled up when using tablature notation. 
image

From what I see in Bravura Lute tablature has a scaling of 1.5 between staff lines. Again, there are no _Implementation notes_ on tablature notation in SMuFL.

PR #2583, opened 2022-01-19
Adjust short slurs

### LINKED ISSUE #2418: Slur + flag interaction
In the new slur rendering code, there is a problem with slurs that start on the stem sides of notes with flags:

Screen Shot 2021-10-09 at 2 00 11 AM

When the slur detects the flag, it move upwards as illustrated in the second slur in the example above.  However, this slur looks ridiculous.  Either the right end of the slur must also be moved up on the ending note, or the starting point of the slur should be moved back down and perhaps moved slightly to the right to avoid the flag.

There are two components to the ridiculous look of the slur (1) the slur is angled down, while the melodic interval is going up, so the slur is giving contrary information compared to the notes. (2) the angled slur's mid control point is not being calculated correctly (more on that further below).

MEI test data:

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2021-10-09T02:03:40" version="3.7.0-dev-09f2846">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mxeyhck">
    <score xml:id="sv907i4">
     <scoreDef xml:id="sl8qgx0">
      <staffGrp xml:id="sdgjf7r">
       <staffDef xml:id="svd5gvg" n="1" lines="5">
        <clef xml:id="cf7idae" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="3" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="s3vl5ge" n="1">
         <note xml:id="note-L4F1" dur="4" oct="4" pname="f" accid.ges="n" />
         <note xml:id="note-L5F1" dur="4" oct="4" pname="g" accid.ges="n" />
         <note xml:id="note-L6F1" dur="8" oct="4" pname="f" accid.ges="n" />
         <note xml:id="note-L7F1" dur="8" oct="4" pname="g" accid.ges="n" />
        </layer>
       </staff>
       <slur xml:id="slur-L4F1-L5F1" type="placed" staff="1" startid="#note-L4F1" endid="#note-L5F1" curvedir="above" />
       <slur xml:id="slur-L6F1-L7F1" type="placed" staff="1" startid="#note-L6F1" endid="#note-L7F1" curvedir="above" />
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

Also note that the slur thickness with asymmetric slurs looks bad and is incorrect.  The bezier curve should be calculated at the angle created by the end points so that the thickness is more consistent.   In other words, rotation the coordinate system so that the curve is calculated as if horizontal, then calculate the bezier curves (presuming there are two separate curves used to create the thickness, and then rotate the result back.  Or another way of saying this: The mid-slur control point needs to move on a perpendicular line to the line connecting the endpoints.  In this example it seems to be moving vertically rather than perpendicularly.

It is quite important for slurs to appear symmetric as much as possible, and the new code is messing around a lot with that rule.

Another fine point is that the slur is ending very close to a staff line.  Since it is nearly perpendicular with the line this is not a big problem, but since  there is a lot of leeway in the placement of the slur on the stem side of a note, the right endpoint of the second slur should be further from a staff line.  The same comment goes for the end of the first slur which is also unnecessarily too close to the staff line.

DavidBauer1984:
This PR fixes #2418 . It introduces some extra handling such that short slurs are adjusted as a whole (rather than shifting endpoints individually) in order to avoid awkward slur shapes.
After

### DIFF
--- a/include/vrv/slur.h
+++ b/include/vrv/slur.h
@@ -150,7 +150,8 @@ private:
     void FilterSpannedElements(FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin);
 
     // Calculate the vertical shift of the slur end points
-    std::pair<int, int> CalcEndPointShift(FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin);
+    std::pair<int, int> CalcEndPointShift(
+        FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin, int unit);
 
     // Calculate the horizontal control point offset
     std::tuple<bool, int, int> CalcControlPointOffset(
@@ -176,6 +177,9 @@ private:
     ///@{
     // Shift end points for collisions nearby
     void ShiftEndPoints(int &shiftLeft, int &shiftRight, double ratio, int intersection) const;
+
+    // Rebalance shifts to avoid awkward tilting of short slurs
+    void RebalanceShifts(int &shiftLeft, int &shiftRight, double distance, int unit) const;
 
     // Rotate the slope by a given number of degrees, but choose smaller angles if already close to the vertical axis
     // Choose doublingBound as the positive slope value where doubling has the same effect as rotating:
--- a/src/slur.cpp
+++ b/src/slur.cpp
@@ -313,7 +313,7 @@ void Slur::AdjustSlur(Doc *doc, FloatingCurvePositioner *curve, Staff *staff)
     bezier.UpdateControlPointParams(curve->GetDir());
 
     const int unit = doc->GetDrawingUnit(100);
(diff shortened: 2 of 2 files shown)


lpugin:
Example added to the test-suite https://github.com/rism-digital/verovio.org/commit/ee30397dc6a72b3a5294d2b25ecb3e9b69b202da

PR #2587, opened 2022-01-21
Simplify staff search

DavidBauer1984:
This PR simplifies staff search by introducing a helper function for layer elements. This can operate in two modes: looking for the ancestor staff only or resolving cross staff additionally.

### DIFF
--- a/include/vrv/layerelement.h
+++ b/include/vrv/layerelement.h
@@ -185,7 +185,7 @@ public:
     /**
      * Find the ancestor or cross staff
      */
-    Staff *FindStaff(StaffSearch strategy);
+    Staff *FindStaff(StaffSearch strategy, bool assertExistence = true) const;
 
     /**
      * Look for a cross or a a parent LayerElement (note, chord, rest) with a cross staff.
--- a/include/vrv/layerelement.h
+++ b/include/vrv/layerelement.h
@@ -183,9 +183,9 @@ public:
     Alignment *GetAlignment() const { return m_alignment; }
 
     /**
-     * Find the ancestor or cross staff
+     * Get the ancestor or cross staff
      */
-    Staff *FindStaff(StaffSearch strategy, bool assertExistence = true) const;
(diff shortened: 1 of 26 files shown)


DavidBauer1984:
**Remark:** Initially I wanted to name this function `GetStaff`. However this triggers a compilation error since there is also a `GetStaff` function generated in `AttStaffIdent`. I was surprised that this is not sorted out by overload resolution, but it turns out that this (i.e. overload resolution through subclasses) does not work in C++ due to a procedure called name lookup. Anyway, the error could be fixed by including something like 
```C++
using LayerElement::GetStaff; 
using AttStaffIdent::GetStaff;
``` 
in the header of the note class. Such code must be added for any class that derives from LayerElement and TimePointInterface and uses `GetStaff`. I did not like this, hence changed the name into `FindStaff`.

lpugin:
Looks good. I agree that `GetStaff` would have been more intuitive, and I remember have tried to solve compiler resolutions of that type. In the end I thought that it is better for human reading to have a different names anyway. 

Another name idea would be:
```
Staff  *GetAncestorStaff(StaffSearch strategy = ANCESTOR_ONLY, bool assertExistence = true) const;
```
with a default `strategy` so then you would do mostly `GetAncestorStaff()` or `GetAncestorStaff(RESOLVE_CROSSSTAFF)` with the idea the `Find` prefixed methods would be (mostly?) for retrieving elements lower in the hierarchy?

include/vrv/layerelement.h
```diff
@@ -29,6 +29,9 @@ class MeterSig;
 class Staff;
 class StaffAlignment;
 
+// Helper enums
+enum StaffSearch { ANCESTOR_ONLY = 0, RESOLVE_CROSSSTAFF };
```
paul-bayleaf, line 33: Suggest RESOLVE_CROSS_STAFF
As cross staff is two words.

lpugin: addressed here https://github.com/rism-digital/verovio/compare/17250a23c6b8...6a58106de5a2

PR #2588, opened 2022-01-21
Fix for the stem length on beams with mixed duration

### LINKED ISSUE #2561: Beam + 64th note interactions with 8ths and 16ths
In the following example, only the third beamed group of notes is correct:

Screen Shot 2022-01-07 at 12 07 40 AM

The first two beamed groups have beams too close to the note heads, and in the second measure the beams are too far from the notes (but better than the first measure case).   The same problem does not occur when the 64th notes are replaced with 32nd notes or 128th notes.

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-01-07T00:06:06" version="3.9.0-dev-5935d76-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m4csume">
    <score xml:id="sdd79i9">
     <scoreDef xml:id="sivuqxo">
      <staffGrp xml:id="sxuz77z">
       <staffDef xml:id="sjvkr99" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="s2uil4s" n="1">
         <beam xml:id="beam-L4F1-L6F1">
          <note xml:id="note-L4F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L5F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L6F1" dur="8" oct="5" pname="e" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L7F1-L9F1">
          <note xml:id="note-L7F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L8F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L9F1" dur="16" oct="5" pname="e" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L10F1-L12F1">
          <note xml:id="note-L10F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L11F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L12F1" dur="32" oct="5" pname="e" accid.ges="n" />
         </beam>
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L13" right="invis" n="2">
       <staff xml:id="staff-L13F1N1" n="1">
        <layer xml:id="layer-L13F1N1" n="1">
         <beam xml:id="beam-L14F1-L16F1">
          <note xml:id="note-L14F1" dur="8" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L15F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L16F1" dur="64" oct="5" pname="e" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L17F1-L19F1">
          <note xml:id="note-L17F1" dur="16" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L18F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L19F1" dur="64" oct="5" pname="e" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L20F1-L22F1">
          <note xml:id="note-L20F1" dur="32" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L21F1" dur="64" oct="5" pname="e" accid.ges="n" />
          <note xml:id="note-L22F1" dur="64" oct="5" pname="e" accid.ges="n" />
         </beam>
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

Monceber:
- changed code to take the shortest note stem length as uniform in beams with multiple notes on the same location (e.g. horizontal beams)
closes #2561
image

This problem also occurred for 128th notes, but since beam overlapped with staff top it was adjusted then (which was not the case for 64th notes). This change makes rendering of beams with notes of different duration consistent with rendering of beams with notes of same duration, as well as beams with beamlets.

Before:
image

After:
image

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -259,7 +259,7 @@ public:
 
     int m_x;
     int m_yBeam; // y value of stem top position
-    int m_dur; // drawing duration
+    short m_dur; // drawing duration
     int m_breaksec;
     int m_overlapMargin;
     int m_maxShortening; // maximum allowed shortening in half units
--- a/src/beam.cpp
+++ b/src/beam.cpp
@@ -267,8 +267,8 @@ bool BeamSegment::NeedToResetPosition(Staff *staff, Doc *doc, BeamDrawingInterfa
 {
     const int unit = doc->GetDrawingDoubleUnit(staff->m_drawingStaffSize);
     // find shortest duration for above/below beams for the sake of calculating overlap with additional beams
-    int topShortestDur = DUR_8;
-    int bottomShortestDur = DUR_8;
+    short topShortestDur = DUR_8;
+    short bottomShortestDur = DUR_8;
     std::for_each(m_beamElementCoordRefs.begin(), m_beamElementCoordRefs.end(), [&](BeamElementCoord *coord) {
         if (coord->m_partialFlagPlace == BEAMPLACE_above) {
             topShortestDur = std::max(topShortestDur, coord->m_dur);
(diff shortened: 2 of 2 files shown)


include/vrv/beam.h
```diff
@@ -259,7 +259,7 @@ class BeamElementCoord {
 
     int m_x;
     int m_yBeam; // y value of stem top position
-    int m_dur; // drawing duration
+    short m_dur; // drawing duration
```
lpugin, line 262: Please keep `int`. Thanks

craigsapp:
The stem lengths are excessively long.  The second closest stem should be an octave or less away from the notehead, but it is about two diatonic steps further away than expected.

Example from Behind Bars, p. 18 for 32nd notes:

Screen Shot 2022-01-21 at 3 26 03 AM

And some more examples of 16ths and 64th notes as well on the following page:

Screen Shot 2022-01-21 at 3 27 36 AM

Monceber:
I believe that's exactly how these beams are drawn currently:
image

I think using the stem length of the shortest note allows for better consistency among the beams, that have notes of that duration (e.g. 128th). If we used stem length of the longest note, that might lead to jumps in stem length between different beams, as was seen in original issue:
image

src/beam.cpp
```diff
@@ -1058,14 +1058,17 @@ void BeamSegment::CalcBeamStemLength(Staff *staff, data_BEAMPLACE place, bool is
         }
     }
 
+    short minDuration = DUR_4;
```
paul-bayleaf, line 1061: Suggest use int.   coord->m_dur is type int.

Monceber: I ended up missing one place. Thanks.

PR #2591, opened 2022-01-24
Add horizontal alignment cache for faster layout redoing

lpugin:
The branch adds a layout cache for the horizontal alignment in order not to repeat its complete calculation every time a document is un-cast off and cast off. The `Toolkit::RedoLayout` method now takes an optional JSON string parameter with a boolean `resetCache` key. When set to `false` (default is `true`), the cached horizontal layout is used.

The principle is to restore the Measure x and width and the LayerElement x and y and to use these for casting off the system. That works well. However casting off the pages can sometimes yield slightly different results. The reason is that  not all drawing positions are (and can be?) cached (e.g., stem length?), I think. As a result, when doing the vertical alignment based on the cache, the height of a system can be a bit higher than with the fully calculated layout. Slurs seem to be some of the elements triggering this. Because of this difference, the system might not fit on the page and be moved to the next one.

However, it still seems to be a valuable direction to follow because the speed gain when redoing the layout is about 30% on small documents and goes up to 50% or large ones.

Console timing in Chrome with the Hummel concerto example. Using cache is nearly twice as fast.
image

I am making this as a draft for now. Feedback and ideas are welcome.


craigsapp:
> Using cache is nearly twice as fast.

That sounds great.  

This would be used if no edits have been made, but you need to redo layout?  Such as for a webpage resize, or generating a PDF from music currently viewed on a webpage?.  When fully implemented, it would be useful to list in the documentation which layout parameters can be changed with this feature turned on (i.e., which layout parameters are used at/after the cast-off processing that do not affect initial layout calculations).   For example `scale`, width, height, margins can be used with this option, and it looks like most rendering options are only needs at the cast-off stage. (maybe spacing options will influence application of object avoidances such as a slur and a turn?)

A related question:  will "`resetCache`" remain "`false`" until set to true again?   There is a general problem of using an option temporarily in the Javascript toolkit, because you have to remember to revert any temporary options.  This is complicated when there are more than one independent processing sequences (screen versus PDF layout).  I get around the problem by externally managing options in an external object, then I can ensure that I know the current state of all options by checking that object.

It might be nice to add some functions to the javascript toolkit (and other persistent toolkit implementations):

(1) `vrvToolkit.getOptions()` == return a JSON object containing all options
(2) `vrvToolkit.getOptionOverrides()` == return a JSON object containing all options which are not equal to the default values
(3) `vrvToolkit.resetOptions()` == set all options to their default (initial) values.

If this is interesting, I can create an issue for it.

rettinghaus:
We have a resetOptions.

craigsapp:
Thanks! Is that new (I remember requesting it a year or two ago, but I could not find an issue related to it today).

lpugin:
See also:
* https://book.verovio.org/toolkit-reference/toolkit-methods.html#getavailableoptions
* https://book.verovio.org/toolkit-reference/toolkit-methods.html#getoptions (for current and default options)

> A related question: will `resetCache` remain `false` until set to true again? 

No, the option is not stored. You need to set it to `false` for every call where you do not want to reset the cache when redoing the layout.

I have to think about the options that can change. Essentially, it is meant to be used when resizing the score without changing the content. A few thoughts:
* changing spacing values would not work because the spaced used by a measure would change
* changing margins should work because casting off is still redone
* changing font should mostly work, even though not exactly

DavidBauer1984:
This is great and considerably faster. I have tested changing the font and it looked great even though the layout was a bit different. Code also is very clean and looks good.

src/measure.cpp
```diff
@@ -1198,8 +1202,10 @@ int Measure::CastOffSystems(FunctorParams *functorParams)
     CastOffSystemsParams *params = vrv_params_cast<CastOffSystemsParams *>(functorParams);
     assert(params);
 
-    // Check if the measure has some overflowing control elements
-    int overflow = this->GetDrawingOverflow();
+    const bool hasCache = this->HasCachedHorizontalLayout();
+    int overflow = hasCache ? this->m_cachedOverflow : this->GetDrawingOverflow();
+    int width = hasCache ? this->m_cachedWidth : this->GetWidth();
+    int drawingXRel = this->m_drawingXRel;
```
DavidBauer1984, line 1208: Why not use the cache for `drawingXRel`? Is this desired?

include/vrv/arpeg.h
```diff
@@ -116,6 +121,11 @@ class Arpeg : public ControlElement,
      * (See View::DrawArpeg that sets the FloatingPositioner)
      */
     int m_drawingXRel;
+
+    /**
+     * The cached value for m_darwingXRel for caching horizontal layout
```
DavidBauer1984, line 126: Typo here and below: m_drawingXRel

include/vrv/page.h
```diff
@@ -99,6 +99,13 @@ class Page : public Object {
      */
     void LayOutHorizontally();
 
+    /**
+     * Layout out the measures horizontally using the cached values.
+     * This should be done in preparation of cast-off which it based on measure positioning.
+     * The content of the measures it not layout out and keeps previous relative positioning.
+     */
```
DavidBauer1984, line 106: Typo twice it=is

src/measure.cpp
lpugin: It is (at least it should be?) the same because it had be restored by `Measure::HorizontalLayoutCache` at this stage.

include/vrv/page.h
```diff
@@ -99,6 +99,13 @@ class Page : public Object {
      */
     void LayOutHorizontally();
 
+    /**
+     * Layout out the measures horizontally using the cached values.
```
ahankinson, line 103: Typo? "Layout out the measures..."

include/vrv/toolkit.h
```diff
@@ -570,8 +570,11 @@ class Toolkit {
      *
      * This can be called once the rendering option were changed, for example with a new page (sceen) height or a new
      * zoom level.
+     *
+     * @param jsonOptions A stringified JSON object with the action options
+     * resetCache: true or false; true by default;
```
ahankinson, line 575: Does this mean that the argument is a string with the value of `"true"` or `"false"` (instead of actual boolean options `true` or `false`)?

src/system.cpp
```diff
@@ -1003,6 +1003,8 @@ int System::CastOffPages(FunctorParams *functorParams)
         currentShift += params->m_pgHead2Height + params->m_pgFoot2Height;
     }
 
+    // LogMessage("System: %d", this->GetHeight());
```
ahankinson, line 1006: Do you want to keep this in the commit?

include/vrv/toolkit.h
lpugin: No, it is only the entire JSON object that is a string, not the values. In C++, you will do
```cpp
toolkit.RedoLayout("{ 'resetCache': true }");
```
In JS
```js
toolkit.redoLayout({ 'resetCache': true });
```
That is how it is for all options passed as JS. That is, they are a string in the C++ codebase, but a JSON object in the JS interface.
Can you propose a better wording?

lpugin: (basically  the same here https://github.com/rism-digital/verovio/pull/2606)

ahankinson (approved):
Looks good to me, but I didn't review the actual functioning of the code.

PR #2594, opened 2022-01-25
Update deprecated FindPythonLibs in CMake

rettinghaus:
Updates the deprecated [FindPythonLibs](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) and adds suggested paths to use on Apple Silicon.


lpugin:
Since this is currently not a top priority I am closing it. Please just re-open it once it is ready. Thanks

Issue #2554, opened 2021-12-22
Add support for @stem.sameas


o-sapov:
`@stem.sameas` is a new MEI attribute for single stem voices. Cf. the MEI discussion [here](https://github.com/music-encoding/music-encoding/issues/551).

Here is an example.

![551-01_m9_ob-fg-corno](https://user-images.githubusercontent.com/24490898/147090667-4a281e7b-e045-4ce4-b6bb-dba85addd886.png)
*W. A. Mozart, Neue Mozart Ausgabe, K. 551/01, m. 9, oboi, fagotti & corni*

<details>
<summary>Click to see the MEI file.</summary>
<p>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.1/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.1/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei meiversion="4.0.1" xmlns="http://www.music-encoding.org/ns/mei">
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>stem.sameas</title>
      </titleStmt>
      <pubStmt/>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv n="1" xml:id="mdiv_72">
        <score xml:id="score_78">
          <scoreDef key.mode="major" key.pname="c" key.sig="0" meter.count="4" meter.sym="common" meter.unit="4" optimize="false" xml:id="scoreDef_01">
            <staffGrp bar.thru="false" xml:id="staffGrp_01">
              <staffGrp bar.thru="true" symbol="bracket" xml:id="staffGrp_02">
                <staffDef clef.line="2" clef.shape="G" lines="5" n="2" xml:id="staffDef_P1">
                  <label xml:id="label_P1">Oboi</label>
                  <layerDef n="1" xml:id="layerDef_01"/>
                  <layerDef n="2" xml:id="layerDef_02"/>
                </staffDef>
                <staffDef clef.line="4" clef.shape="F" lines="5" n="3" xml:id="staffDef_P2">
                  <label xml:id="label_P2">Fagotti</label>
                  <layerDef n="1" xml:id="layerDef_03"/>
                  <layerDef n="2" xml:id="layerDef_04"/>
                </staffDef>
              </staffGrp>
[...]

PR #2596, opened 2022-01-25
Support for stem.sameas

lpugin:
This PR adds support for the `@stem.sameas` for orchestral notation as requested here https://github.com/rism-digital/verovio/issues/2554

This is the result for the test file https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/stem/stem-014.mei
image

There are a few encoding requirements with the current implementation:
* `@stem.sameas` is expected to be given on the notes of the second layer
* the stem direction is up by default but `@stem.dir="down"` can be given on the notes of the first layer
* when the stem direction is indicated and the notes are in a beam, is has to be given on each note

Treatment of articulations needs to be clarified. The current implementation does not properly handle unison or second intervals. However, these are pretty rare.

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -220,8 +220,12 @@ protected:
     int CalcLayerOverlap(Doc *doc, Object *beam, int directionBias, int y1, int y2);
 
 private:
-    /** A flag indicating that we have notes with stem shared and that it should not be drawn */
-    bool m_stemSameAsNotes;
+    /**
+     * A pointer to the beam with which stem are shared.
+     * The pointer is bi-directional (each beam points to the other one.
+     * It is the in Note::ResovleStemSameas()
+     */
+    Beam *m_stemSameas;
 
 public:
     /** */
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -77,11 +77,33 @@ public:
     int GetStartingY() const;
     ///@}
 
+    /**
+     * @name Getters for the stem same as role.
+     */
+    ///@{
+    bool StemSameas() const { return (m_stemSameasRole != SAMEAS_NONE); }
+    bool StemSameasIsUnset() const { return (m_stemSameasRole == SAMEAS_UNSET); }
+    bool StemSameasIsPrimary() const { return (m_stemSameasRole == SAMEAS_PRIMARY); }
+    bool StemSameasIsSecondary() const { return (m_stemSameasRole == SAMEAS_SECONDARY); }
+    ///@}
+
+    /**
+     * @name Methods to intialize and update the role for stem.sameas context
+     * Before the drawing beam place has been calculated, the role is marked as unset.
+     * Then it is marked as primary or secondary depending on the beam place.
+     */
+    ///@{
+    void InitSameasRoles(Beam *sameasBeam, data_BEAMPLACE &drawingPlace);
+    void UpdateSameasRoles(data_BEAMPLACE place);
+    ///@}
+
 private:
     // Helper to adjust beam positioning with regards to ledger lines (top and bottom of the staff)
     void AdjustBeamToLedgerLines(Doc *doc, Staff *staff, BeamDrawingInterface *beamInterface);
 
     void CalcBeamInit(Layer *layer, Staff *staff, Doc *doc, BeamDrawingInterface *beamInterface, data_BEAMPLACE place);
+
+    void CalcBeamInitForNotePair(Note *note1, Note *note2, Staff *staff, int &yMax, int &yMin);
 
     bool CalcBeamSlope(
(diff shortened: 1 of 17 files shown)


lpugin:
Articulations work as expected with this example https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/artic/artic-017.mei

For the pairs in the first measure we expect to see the articulation sign on both sides because it is given explicitly on both notes, and in the second measure only once above the articulation in the note with `@stem.sameas` is encoded with `@sameas`.

image

```xml
<measure n="1">
   <staff n="5">
      <layer n="1">
         <rest xml:id="rest_162" dur="4" />
         <note xml:id="note_168" dur="4" oct="4" pname="g">
            <artic artic="spicc" />
         </note>
         <note xml:id="note_181" dur="4" oct="5" pname="g" stem.dir="down">
            <artic artic="spicc" />
         </note>
      </layer>
      <layer n="2">
         <rest sameas="#rest_162" />
         <note dur="4" oct="3" pname="g" stem.sameas="#note_168">
            <artic artic="spicc" />
         </note>
         <note dur="4" oct="4" pname="g" stem.sameas="#note_181">
            <artic artic="spicc" />
         </note>
      </layer>
   </staff>
</measure>
<measure n="2">
   <staff n="5">
      <layer n="1">
         <rest xml:id="rest_163" dur="4" />
         <note xml:id="note_180" dur="4" oct="4" pname="g">
            <artic xml:id="artic_192" artic="spicc" />
         </note>
         <note xml:id="note_182" dur="4" oct="5" pname="g" stem.dir="down">
            <artic xml:id="artic_193" artic="spicc" />
         </note>
      </layer>
      <layer n="2">
         <rest sameas="#rest_163" />
         <note dur="4" oct="3" pname="g" stem.sameas="#note_180">
            <artic sameas="#artic_192" />
         </note>
         <note dur="4" oct="4" pname="g" stem.sameas="#note_182">
            <artic sameas="#artic_193" />
         </note>
      </layer>
   </staff>
</measure>

```

craigsapp:
Great!

> the stem direction is up by default but `@stem.dir="down"` can be given on the notes of the first layer

Preferably this will become more automated in the future since the resulting chord is notationally treated as being on a single-layer staff.   The current situation would make converting between sounding and written scores difficult.  The algorithm should be:

(1) Use the stem direction on the note that does not use a sameas stem.
(2) If there is no stem direction on that primary note, then create a virtual chord with the primary note and the sameas note(s) and use the existing stem direction calculations for chords (in a single-layer staff context).
(3) Similarly for beamed notes: group into chords and then run through stem assignment analysis for beams.

> when the stem direction is indicated and the notes are in a beam, is has to be given on each note

Is this true or does just the first and last note of the beam need to be given (with the notes in the middle assumed to be in the same direction as the outer primary notes' stem directions.  If no stem directions are given, then should not the default of up-stem be used (as described in your description above for non-beamed notes)?

I am wondering about how beams affect merging of sameas stems?  Merging two beam groups has a problem if not all notes in the sameas beam group do not have a sameas stem.    This seems to imply that there should be stem direction attributes for beams, and sameas stems for beams as well...


> Treatment of articulations needs to be clarified.

I would use only the articulation of the primary note and ignore that of the sameas note(s).  Merging notes from different instruments into a single chord implies that the all have the same articulations.     If the parts have different articulations, then they should not be merged into chords and instead remain as separate layers.  If someone merges notes using stem sameas but the notes have different articulations, then it is the responsibility/fault of the encoder, who should not have done that.

Dealing with slurs is more complicated: maybe some editions would merge the slur (this seems to be the best default), while others may show a slur both above and below (allowing merging of rhythms, but preserving different phrasings for the same rhythms), but I don't think I have seen the latter case.

craigsapp:
There are cases of three or four parts merged into one layer (most notably for French horns) that would not work easily with this system (since there are only two sides to the staff).  But someone can comment on the relative amount each style is used in orchestral scores.

lpugin:
I re-worked the PR. Stem directions are no longer needed and will be automatically calculated. They can still be encoded if desired. However, it has to be in the notes / beams of the first layer, but I think this is an acceptable limitation. For beams, usual place indication work (`beam@place` or `note@stem.dir`, which can be on a single note)

Also, `@stem.sameas` is not longer expected to be given on the notes of the second layer, but it is probably a good idea still to do so. At least it would make sense to be consistent and have always the links going in one direction. On this topic, it looks that it could be nice to be able to do `beam@stem.sameas` instead of having to do it for every note in the same way we have `beam@sameas`. However, this is not supported in MEI. If somebody is interested by this feature, please follow-up in the MEI repository directly.

Nothing has changed for the articulation. These will be adjusted in a separate issue / PR if necessary.

PR #2600, opened 2022-01-26
Regenerate fonts

rettinghaus:
This PR's only purpose is to remove and re-generate all SMuFL glyphs with the updated script.


rettinghaus:
There seems to be something off with cutouts now.

lpugin:
Something goes wrong when building the test suite

rettinghaus:
Yes, but I guess that's not my fault:
```shell
E: Unable to locate package openresolv
E: Unable to locate package wireguard
```

lpugin:
Please try to make a new PR.

PR #2605, opened 2022-01-27
Handle multibar rests in MIDI

### LINKED ISSUE #2604: Multibar rests not handled correctly in midi and timemap output
**Describe the problem**
When exporting midi or timemap information (using `-t midi` or `-t timemap` flags), multibar rest durations are not taken in account - they are treated as a single bar, with a single bar pause duration.

When exporting svg with the input data below including `<measure-style><multiple-rest>`, everything works fine (of course):
MultibarRest-SVG

However, when exporting the same input data as midi, and then importing that into another program, the multibar rest information is lost: 
MultibarRest-MIDI

The same happens when exporting timemap data - the multibar rest duration (in this case three bars) is left out causing the pause duration between the notes to be the length of one bar instead of three:
MultibarRest-JSON

**To Reproduce**
1. Load the music-xml input data below
2. Generate midi or timemap output (using `-t midi` or `-t timemap` flags)
3. Not that the information for multibar rests is lost - instead they are treated as a single bar rest.

**Expected behavior**
Multibar rest time information must be handled correctly

**Input data**
The smallest example (MEI unless about another input format) to reproduce the problem
``` xml
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
    <part-list>
        <score-part id="P1">
            <part-name>Piano</part-name>
            <part-abbreviation>Pno.</part-abbreviation>
        </score-part>
    </part-list>
    <part id="P1">
        <measure number="1" width="188.93">
            <attributes>
                <divisions>1</divisions>
                <key>
                    <fifths>0</fifths>
                </key>
                <time>
                    <beats>4</beats>
                    <beat-type>4</beat-type>
                </time>
                <clef>
                    <sign>G</sign>
                    <line>2</line>
                </clef>
            </attributes>
            <note default-x="69.27" default-y="-30.00">
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <duration>4</duration>
                <voice>1</voice>
                <type>whole</type>
            </note>
        </measure>
        <measure number="2" width="203.57">
            <attributes>
                <measure-style>
                    <multiple-rest>3</multiple-rest>
                </measure-style>
            </attributes>
            <note>
                <rest measure="yes"/>
                <duration>4</duration>
                <voice>1</voice>
            </note>
        </measure>
        <measure number="3" width="203.57">
            <note>
                <rest measure="yes"/>
                <duration>4</duration>
                <voice>1</voice>
            </note>
        </measure>
        <measure number="4" width="203.57">
            <note>
                <rest measure="yes"/>
                <duration>4</duration>
                <voice>1</voice>
            </note>
        </measure>
        <measure number="5" width="138.67">
            <note default-x="10.00" default-y="-30.00">
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <duration>4</duration>
                <voice>1</voice>
                <type>whole</type>
            </note>
            <barline location="right">
                <bar-style>light-heavy</bar-style>
            </barline>
        </measure>
    </part>
</score-partwise>
```

**Verovio information**
 - Version of Verovio: 3.9.0-dev
 - Binding: command line

 - OS: Linux subsystem on Windows

DavidBauer1984:
This PR adds handling of multirests in MIDI and the time map output. Closes #2604 .

### DIFF
--- a/src/tempo.cpp
+++ b/src/tempo.cpp
@@ -127,4 +127,27 @@ int Tempo::ResetDrawing(FunctorParams *functorParams)
     return FUNCTOR_CONTINUE;
 }
 
+int Tempo::CalcMaxMeasureDuration(FunctorParams *functorParams)
+{
+    CalcMaxMeasureDurationParams *params = vrv_params_cast<CalcMaxMeasureDurationParams *>(functorParams);
+    assert(params);
+
+    if (this->HasMidiBpm()) {
+        params->m_currentTempo = this->GetMidiBpm();
+    }
+    else if (this->HasMm()) {
+        double mm = this->GetMm();
+        int mmUnit = 4;
+        if (this->HasMmUnit() && (this->GetMmUnit() > DURATION_breve)) {
+            mmUnit = pow(2, (int)this->GetMmUnit() - 2);
+        }
+        if (this->HasMmDots()) {
+            mmUnit = 2 * mmUnit - (mmUnit / pow(2, this->GetMmDots()));
+        }
+        params->m_currentTempo = mm * 4.0 / mmUnit + 0.5;
+    }
+
+    return FUNCTOR_CONTINUE;
+}
+
 } // namespace vrv
--- a/include/vrv/functorparams.h
+++ b/include/vrv/functorparams.h
@@ -992,6 +992,7 @@ public:
  * member 1: the current time in seconds
  * member 2: the current tempo
  * member 3: the tempo adjustment
+ * member 4: factor for multibar rests
  **/
 
 class CalcMaxMeasureDurationParams : public FunctorParams {
@@ -1002,11 +1003,13 @@ public:
         m_currentRealTimeSeconds = 0.0;
         m_currentTempo = 120.0;
         m_tempoAdjustment = 1.0;
+        m_multiRestFactor = 1;
     }
     double m_currentScoreTime;
     double m_currentRealTimeSeconds;
     double m_currentTempo;
     double m_tempoAdjustment;
+    int m_multiRestFactor;
 };
 
 //----------------------------------------------------------------------------
--- a/include/vrv/functorparams.h
+++ b/include/vrv/functorparams.h
@@ -988,9 +988,10 @@ public:
 //----------------------------------------------------------------------------
 
 /**
- * member 0: std::vector<double>: a stack of maximum duration filled by the functor
- * member 1: double: the duration of the current measure
- * member 2: the current bpm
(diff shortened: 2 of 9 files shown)


src/tempo.cpp
```diff
@@ -127,4 +127,27 @@ int Tempo::ResetDrawing(FunctorParams *functorParams)
...
+        }
+        if (this->HasMmDots()) {
+            mmUnit = 2 * mmUnit - (mmUnit / pow(2, this->GetMmDots()));
+        }
+        params->m_currentTempo = mm * 4.0 / mmUnit + 0.5;
+    }
+
+    return FUNCTOR_CONTINUE;
+}
+
```
lpugin, line 152: If I remember correctly the reason why we did not do that is because putting it there will make the tempo change be parsed after the content of the measure, because the data structure is typically:
```xml
<measure>
   <staff/>
   <staff/>
   <tempo/>
</measure>
```
This is why we had a tempo `lookup` in Measure. Could you double check? Thanks

cambiata: Yay! 👍

DavidBauer1984: Yes that's why I introduced an end functor. Now the content of the measure (including tempo changes and multi measure rests) is evaluated and after that all the updates (`m_currentTempo`, `m_currentScoreTime`, `m_currentRealTimeSeconds`) are set in `Measure::CalcMaxMeasureDurationEnd`.

If we enhance the example in the issue with
```xml
<tempo midi.bpm="60" staff="1" tstamp="1.000000"></tempo>
```
in the second measure (the one with the multi measure rest), then the time map becomes:
```json
[
	{
		"tstamp":	0.000000,
		"qstamp":	0.000000,
		"tempo":	120.000000,
		"on":	["n61e36o"]
	},
	{
		"tstamp":	2000.000000,
		"qstamp":	4.000000,
		"off":	["n61e36o"]
	},
	{
		"tstamp":	14000.000000,
		"qstamp":	16.000000,
		"tempo":	60.000000,
		"on":	["n6cte68"]
	},
	{
		"tstamp":	18000.000000,
		"qstamp":	20.000000,
		"off":	["n6cte68"]
	}
]
```
This looks good to me, since now we spend 4*3 seconds on the multi measure rest.

lpugin: Ah yes, I overlooked that change. Excellent, thanks!

PR #2612, opened 2022-02-01
Fixes outstanding issues in PAE importer

lpugin:
### Pseudo neumatic notation

Pseudo neumatic notation coded with dotted 128th notes is supported again
```
7.EDCDEEE
```
image

### Dots with mensural notation

With a `G+2` clef
```
1.2'GGBA''DC'BB
```
image

### Nested beams

Nested beams are now supported when the second one is within a grace group
```
6{''D'Bqq{3''C'BA}r8B}''gD{8C'6AF}
```
image

However, they will not be converted when the second one is place outside the grace group
```
6{''D'B{qq3''C'BAr}8B}''gD{8C'6AF}

[Warning] PAE: A beam cannot be started with '{' before closing the previous one. (character 7)
[Warning] PAE: An extra '}' to close a beam is present. (character 21)
```
image

Note that if there is no nesting of the beams, a beam outside a grace group remains valid
```
6''D'B{qq3''C'BAr}8B''gD{8C'6AF}
```
image

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -3956,17 +3956,23 @@ bool PAEInput::ConvertDuration()
     std::string paeStr;
     bool isChord = false;
 
-    for (auto &token : m_pae) {
-        if (token.IsVoid()) continue;
+    // Here we need an iterator because we might have to add a mensural dots
+    std::list<pae::Token>::iterator token = m_pae.begin();
+    while (token != m_pae.end()) {
+        if (token->IsVoid()) {
+            ++token;
+            continue;
+        }
 
         // Extract duration string we can then convert in one go
-        if (this->Is(token, pae::DURATION)) {
+        if (this->Is(*token, pae::DURATION)) {
             if (!durationToken) {
-                durationToken = &token;
+                durationToken = &(*token);
                 paeStr.clear();
             }
-            paeStr.push_back(token.m_char);
-            token.m_char = 0;
+            paeStr.push_back(token->m_char);
+            token->m_char = 0;
+            ++token;
             continue;
         }
         // We have reach the end of a duration string - convert it, including patterns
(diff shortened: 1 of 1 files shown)


lpugin:
Tagging @jenniferward and @BaMikusi

src/iopae.cpp
```diff
@@ -3978,24 +4027,49 @@ bool PAEInput::ConvertDuration()
...
-            DurationInterface *interface = dynamic_cast<DurationInterface *>(token.m_object);
+            DurationInterface *interface = dynamic_cast<DurationInterface *>(token->m_object);
             assert(interface);
             interface->SetDur(currentDur->first);
             if (currentDur->second) {
-                interface->SetDots(currentDur->second);
+                if (interface->GetDur() == DURATION_128 && token->Is(NOTE)) {
+                    Note *note = vrv_cast<Note *>(token->m_object);
+                    assert(note);
+                    note->SetDur(DURATION_4);
```
rettinghaus, line 4056: Why set a duration here at all? We wouldn't have to set any stem attributes.

BaMikusi:
Sorry for the layman's question: will Verovio indeed put the dots half a mile away from any mensural notehead, as shown above?

ahankinson:
Yes, that's standard for Verovio's mensural notation rendering of dots-of-division: https://www.verovio.org/test-suite.xhtml?cat=mensural

lpugin:
> Sorry for the layman's question: will Verovio indeed put the dots half a mile away from any mensural notehead, as shown above?

Yes, that is how Verovio renders dots in mensural notation. 

One thing we can do is to enable even spacing of notes with mensural notation. It would look like this
image

src/iopae.cpp
```diff
@@ -2318,7 +2319,8 @@ const std::map<int, std::string> PAEInput::s_errCodes{
     { ERR_055_KEYSIG_CHANGE, "The key signature cannot be changed more than once in a measure." },
     { ERR_056_TIMESIG_CHANGE, "The time signature cannot be changed more than once in a measure." },
     { ERR_057_MENSUR_CHANGE, "The mensur sign cannot be changed more than once in a measure." },
-    { ERR_058_FERMATA_MREST, "A fermata on measure with extra '%s' is invalid." }
+    { ERR_058_FERMATA_MREST, "A fermata on measure with extra '%s' is invalid." },
```
ahankinson, line 2322: Is this supposed to have a measure number? "on measure <something>"?

Or should it be "A fermata on a measure rest with extra '%s' is invalid."?

lpugin: "rest" was missing. I fixed it.

ahankinson (approved):

BaMikusi:
I must admit the version with even spacing reminds me considerably more of the actual notation of contemporary sources than the middle-of-nowhere dotting of the first version.

PR #2614, opened 2022-02-02
Arpeggio playback in MIDI

DavidBauer1984:
This PR implements arpeggio playback in MIDI. We use the same time interval as for unaccented grace notes.

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -1273,6 +1273,8 @@ public:
 
     /**
      * Prepare the MIDI export
+     * Captures information (i.e. from control elements) for MIDI interpretation
+     * This information is usually required beforehand in GenerateMIDI
      */
     ///@{
     virtual int PrepareMIDI(FunctorParams *) { return FUNCTOR_CONTINUE; }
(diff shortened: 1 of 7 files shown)


include/vrv/object.h
```diff
@@ -1270,6 +1271,13 @@ class Object : public BoundingBox {
     virtual int ResolveMIDITies(FunctorParams *) { return FUNCTOR_CONTINUE; }
     ///@}
 
+    /**
+     * Prepare the MIDI export
```
lpugin, line 1275: Could you add a comment about what this prepares?

src/doc.cpp
```diff
@@ -383,13 +383,20 @@ void Doc::ExportMIDI(smf::MidiFile *midiFile)
             filters.push_back(&matchStaff);
             filters.push_back(&matchLayer);
 
+            Functor prepareMIDI(&Object::PrepareMIDI);
+            PrepareMIDIParams prepareMIDIParams;
+            prepareMIDIParams.m_transSemi = transSemi;
+            prepareMIDIParams.m_currentTempo = tempo;
+            this->Process(&prepareMIDI, &prepareMIDIParams, NULL, &filters);
+
```
lpugin, line 391: Question: this is performed for each staff/layer separately. Is this an issue for arpeggios on two staves / layers?

DavidBauer1984: Right, we should pull this out of the loop.

PR #2616, opened 2022-02-02
Add support for ABC tuplets

Monceber:
Added logic to parse and import tuplets from the ABC format.

image

<details><summary>Example ABC</summary>

```
X:6
T:Farewell Manchester
%%VWML:Yarkers1797-5234-p6-0
F:http://www.vwml.org/record/Yarkers1797/5234/p6
Z:Transcribed by David Jacobs Sept 2015
L:1/8
M:2/4
K:G
B2A2|GcB2|AGFG|A/G/F/E/D2|B2A2|GcB2|AGFG|A4:|
d2c2|Bge2|dGc2|Bge2|dG cB/A/|BA/G/ Dc|B2A2|G4:|
B/A/G/B/ A/G/F/A/|G/e/d/c/B2|AB/G/ FG|A/G/F/E/D2|B/A/G/A/ A/G/F/G/|G/e/d/c/ B2|AB/G/ FG|A4:|
d/c/B/d/ c/B/A/c/|B/A/G/g/ e2|d/c/B/d/ c/B/A/c/|B/A/G/g/ e2|d/c/A/d/ c/B/A/c/|B/A/G/B/ D/G/A/"C"c/|B2A2|G4:|
(3BDB (3ADA|(3GDcB2|(3ABA FG|(3ABG (3FED|(3BDB (3ADA|(3GDcB2|(3ABG FG|A4:|
"Cont"(3dBd (3cAc|(3BGge2|(3dBd (3cAc|(3BGge2|(3dBd (3cAc|(3BAG (3DGc|G2A2|G3:|
```

</details>

### DIFF
--- a/src/ioabc.cpp
+++ b/src/ioabc.cpp
@@ -347,6 +347,11 @@ void ABCInput::AddLayerElement()
             m_layer->AddChild(*iter);
         }
     }
+    // clean-up leftover data, if any
+    if (beam && !beam->GetChildCount()) delete beam;
+    if (m_containerElement.m_element && !m_containerElement.m_element->GetChildCount())
+        delete m_containerElement.m_element;
+
     m_containerElement = {};
     m_noteStack.clear();
 }
--- a/src/ioabc.cpp
+++ b/src/ioabc.cpp
@@ -311,24 +311,35 @@ void ABCInput::AddLayerElement()
     if (!m_noteStack.size()) return;
     // if just one note in the stack - add it do the layer directly
     if (m_noteStack.size() == 1) {
-        m_layer->AddChild(m_noteStack.back());
+        if (m_containerElement.m_element && (ElementType::Tuplet == m_containerElement.m_type)) {
+            m_containerElement.m_element->AddChild(m_noteStack.back());
+            if (!--m_containerElement.m_count) {
+                m_layer->AddChild(m_containerElement.m_element);
+                m_containerElement = {};
+            }
+        }
+        else {
+            m_layer->AddChild(m_noteStack.back());
+        }
         m_noteStack.clear();
+        return;
     }
     // otherwise we can have beam or tuplet (for now)
-    LayerElement *element = NULL;
-    // if container pair key is set to Tuplet - proceed with it
-    if (m_containerPair.second && (ElementType::Tuplet == m_containerPair.first)) {
-        element = m_containerPair.second;
-    }
-    // otherwise default to it being beam
-    else {
-        element = new Beam();
-    }
+    Beam *beam = new Beam();
     // add stacked notes to the current element
     for (auto iter = m_noteStack.begin(); iter != m_noteStack.end(); ++iter) {
-        element->AddChild(*iter);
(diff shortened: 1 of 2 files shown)


craigsapp:
In case you don't notice: the triplets in this example should be beamed (according to the input data).

Monceber:
Thanks. Added changes to allow for both beamed and unbeamed tuplets:
image

src/ioabc.cpp
```diff
@@ -304,44 +305,100 @@ void ABCInput::CalcUnitNoteLength()
...
-            for (auto iter = m_noteStack.begin(); iter != m_noteStack.end(); ++iter) {
-                m_layer->AddChild(*iter);
-            }
+            element = beam;
+        }
+        m_layer->AddChild(element);
+    }
+    else {
+        for (auto iter = m_noteStack.begin(); iter != m_noteStack.end(); ++iter) {
+            m_layer->AddChild(*iter);
```
lpugin, line 347: Could you double check that we don't have a memory leak if beam has no descendant NOTE? I think if we end up here beam is never used or deleted

src/ioabc.cpp
```diff
@@ -304,44 +305,100 @@ void ABCInput::CalcUnitNoteLength()
...
+    // List of tuplets with default base of 3
+    const std::unordered_set<int> threeBase = { 2, 4, 8, 9 };
+    if (!tupletNumbase) {
+        tupletNumbase = threeBase.count(tupletNum) ? 3 : 2;
+    }
+    // Get number of elements supposed to be in the tuplet _:_:9
+    // Ignore this for the time being
+    tuplet->SetNum(tupletNum);
+    tuplet->SetNumbase(tupletNumbase);
+    m_containerElement = { ElementType::Tuplet, tuplet, tupletNum };
```
lpugin, line 398: Woundn't it be necessary to check that we do not have anything else in m_containerElement before assigning this? Or is it done before?  I guess that could happen only with nested tuplets, but it seems to me that in such a case `tuplet` would be lost and we would have a memory leak, wouldn't we?

Monceber: Yes, thanks for pointing this out. We definitely want to clean up previous data if we reach this point without element ever being assigned. Regarding nested `tuplet`, now, that's a different question, since I'm not even sure how such encoding would be achieved in ABC format if supported.

src/ioabc.cpp
```diff
@@ -347,6 +347,11 @@ void ABCInput::AddLayerElement()
             m_layer->AddChild(*iter);
         }
     }
+    // clean-up leftover data, if any
+    if (beam && !beam->GetChildCount()) delete beam;
```
lpugin, line 351: Sorry, but this is still cumbersome. If the beam a has a clef child but no note (hypothetical but not necessarily impossible) it would still not be deleted. I think you could set it to NULL above once it is added to m_layer or assigned to m_element, and then you can forget about the `!beam->GetChildCount()`. Or you do the same check as above `beam->FindDescendantByType(NOTE)`, but I think setting it to NULL would be clearer.

PR #2617, opened 2022-02-02
allow more accidentals in keySig

rettinghaus:
This PR updates the regex for keySig and allows moving further in the circle of fifths. 

keysig

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Key signatures with more than seven accidentals</title>
         </titleStmt>
         <pubStmt>
            <respStmt>
               <persName role="encoder">Klaus Rettinghaus</persName>
               <corpName role="funder">Enote GmbH</corpName>
            </respStmt>
            <date isodate="2022-02-02">2 Feb 2022</date>
         </pubStmt>
         <seriesStmt>
         </seriesStmt>
         <notesStmt>
            <annot>When using more than seven accidentals in keySig Verovio uses double accidentals.</annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="3.9.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" clef.shape="G" clef.line="2" key.sig="11s" />
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure right="invis">
                     <staff n="1">
                        <layer n="1">
                           <space dur="2" />
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

There are different ways to draw these key signatures. This implementation takes the easiest way, others are documented here: https://en.wikipedia.org/wiki/Theoretical_key
For allowing different renderings we should think about something on MEI side first.

### DIFF
--- a/src/att.cpp
+++ b/src/att.cpp
@@ -312,7 +312,7 @@ data_KEYSIGNATURE Att::StrToKeysignature(const std::string &value, bool logWarni
     int alterationNumber = 0;
     data_ACCIDENTAL_WRITTEN alterationType = ACCIDENTAL_WRITTEN_NONE;
 
-    std::regex test("mixed|0|[1-7][s|f]");
+    std::regex test("mixed|0|([1-9]|1[0-2])[f|s]");
     if (!std::regex_match(value, test)) {
         if (logWarning) LogWarning("Unsupported data.KEYSIGNATURE '%s'", value.c_str());
         return { -1, ACCIDENTAL_WRITTEN_NONE };
@@ -322,7 +322,7 @@ data_KEYSIGNATURE Att::StrToKeysignature(const std::string &value, bool logWarni
         return { VRV_UNSET, ACCIDENTAL_WRITTEN_NONE };
     }
     else if (value != "0") {
-        alterationNumber = atoi(value.substr(0, 1).c_str());
+        alterationNumber = std::stoi(value);
         alterationType = (value.at(1) == 's') ? ACCIDENTAL_WRITTEN_s : ACCIDENTAL_WRITTEN_f;
     }
     else {
--- a/src/keysig.cpp
+++ b/src/keysig.cpp
@@ -179,7 +179,7 @@ void KeySig::FillMap(MapOfPitchAccid &mapOfPitchAccid)
 
     int i;
     data_ACCIDENTAL_WRITTEN accidType = this->GetAccidType();
-    for (i = 0; i < this->GetAccidCount(); i++) {
+    for (i = 0; i < this->GetAccidCount(); ++i) {
         mapOfPitchAccid[KeySig::GetAccidPnameAt(accidType, i)] = accidType;
     }
 }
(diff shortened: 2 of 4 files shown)


craigsapp:
I see you have adjusted the transposition code.  Does it handle the theoretical keys now?

rettinghaus:
@craigsapp no, not yet. I just changes to pre-increment, as you lately reminded me of being pedantic in this matter. 😉

craigsapp:
😉 

I don't expect any significant problems with the transpose code to deal with extended circle-of-fifths.  Probably just need to remove checks that would otherwise prevent them.   A good basic test would be to transpose by "A1" (augmented unison).  This should transpose G major into G-sharp major.  Likewise, "d1" (diminished unison) should transpose F major into F-flat major.

The main caveat is that if you plan on processing any notes with triple sharps/flats, the `Transposer::m_maxAccid` variable needs to be changed from 2 to a larger number.   The theoretical key signatures don't need them, but the likelihood of triple sharps/flats increases for notes inside of these keys.

In the Transposer constructor, the current limit is set to 2 (double sharp/flat):

```cpp
Transposer::Transposer()
{
    // Initialize with base-40 system by default:
    this->SetMaxAccid(2);
}
```

This could be set to "3" if any possibility of triple sharps/flats should be allowed.

I have a convenience function that sets the maximum to 42 sharps/flats, which is useful for extreme headroom:

```cpp
void Transposer::SetBase600()
{
    this->SetMaxAccid(42);
}
```

There is no loss of computer efficiency using the larger transposition space, so switching to "Base-600" is just as good as "Base-40" (600 chosen as an arbitrary allowed base just because it is a multiple of 100).  But since triple-sharp/flats were not expected, I used the more classical restricted base of 40 (meaning up to double sharps/flats).

`m_maxAccid = 3` is equivalent to "base-54", where octaves are assigned to the integer 54 because there are 7 accidental states for each of the 7 diatonic notes, plus 5 holes in the system to allow for intervals to be represented as constant integers:

```
n = 3
base = (2n+1)*7+5 = 7*7+5 = 54
```

craigsapp:
Note that the key signatures are not completely right yet:  The double sharp/flat is added after the single sharps/flats, but the same diatonic single sharp/flat needs to be removed when the same double sharp/flat is given.

See the example on https://en.wikipedia.org/wiki/Theoretical_key

Screen Shot 2022-02-02 at 8 57 01 AM

rettinghaus:
@craigsapp you picked one specific example. As I stated above there are (AFAIK) no definite rules for the layout. 

> For allowing different renderings we should think about something on MEI side first.

craigsapp:
There are no rules for the layout since they are theoretical, but the current display is 100% wrong:

Screen Shot 2022-02-02 at 9 46 37 AM

(You cannot have both an f-sharp at the left and an f-double-sharp on the right).

rettinghaus:
Damn you, Max Reger, you got it all wrong!!! 😠 

image

craigsapp:
And the example does not contain a b-flat or b-double-flat.j

PR #2618, opened 2022-02-02
Update glyphs

rettinghaus:
This PR updates some names in the `supported.xsl`. 
As far as I can see nobody needs reversed or turned clefs right now, so I removed them.


lpugin:
Any particular issue with reversed clefs? It would be better to leave some in case somebody uses them.

rettinghaus:
No issue, I'm just pretty sure I switched them on when I added the glyphs in Leipzig, although they weren't needed anywhere. I leave the extracted glyphs in for now.

lpugin:
I guess we also need not to remove them from `supported.xsl` otherwise they will not be usable any more.

PR #2595, opened 2022-01-25
Change font generation process (remove JS dependency)

Monceber:
Part 1 of the changes for improvement of handling fonts in the Verovio (support for more fonts without the need of re-compilation/re-generation).

This part addresses improvements to the font generation scripts, namely the following:
- removing PhantomJS dependency and replacing it with Python code
- simplifying parts of the generation scripts and removing parts of the redundant code
- added script to generate XML files for plain text fonts (previous added and reverted as part of #1691)

Changes that were done overall
-
- changed **generate_all.sh** script by removing dependency on PhantomJS as well generate-bbox.js script that was called with it
- combined two steps from **generate_all.sh** (extract glyphs and generate bboxes) into one steps, that extracts all glyphs
- glyphs are no longer extracted from font.SVG into **tmp/** folder - they are read in the **extract-bounding-boxes.py** and extracted into glyph.XML files
- added two dependencies to the Python code: `lxml` as direct replacement for the ElementTree (to allow ordering of attributes in the XML files and support of pretty print feature for readability); `svgpathtools` as tool to replace funcitonality done in **Js** part of the code (converting `@path` attribute from SVG file into bounding box to get coordinates, height, width, etc. about glyph)
- font.XML that is generated as result is pretty much the same as before, save for the addition of `@n` attribute which contains name of the glyph. As result, glyph.XML file that are generated in the **Font/** folder are generated without the glyphname.
- font files for all fonts are re-generated as well all glyphs for all fonts (glyphs are basically the same, but with different name)
- changed **generate_woff.sh** script - removed PhantomJS and long overdue Saxon9ee dependencies. Changed script to use plain font generation for the VerovioText.

Files added and removed in this PR
-
- added [fonts/extract-bounding-boxes.py](https://github.com/rism-digital/verovio/commit/0cc7805c7fa120b98dc30ac89504760857791e04#diff-0c586ae72ebb06284150493599ca55d0f48b7ccff3308975424b92699addc183)
- added [fonts/generate_plain_text.py](https://github.com/rism-digital/verovio/commit/13ce9bcbcdbcd88dbd7e3aef66b17f4d4a834cb8#diff-a134973c399ebba5be8bed2f52883bc1c311a6bb0e6e8048cd6c5d16e7347f7b)
[...]

PR #2626, opened 2022-02-08
Improve handling of the SMUFL fonts

Monceber:
Part 2 for the improvements for the handling fonts in Verovio (follow-up for #2595).

This part addresses handling and parsing of the generated fonts in the Verovio code itself:
- removing generation of smufl.cpp file and the file itself - this allows us to extend the map of id-to-glyph values without having to regenerate .cpp file each time
- removing the check for the SMUFL_COUNT would allow to have glyphs parsed from font.xml, even if they are not in supported.xml (e.g. font generated separately with additional glyphs)
- removing parsing of the filenames from the directory with glyphs xmls in favor of just parsing font.xml (which contains name of the glyph)
- changing from map to unordered_map should slightly improve performance, given the fact that order is not important when looking up for glyphs


lpugin (commented):
Looks good. I am not sure I understand the issue with `SMuFL_COUNT` because it checks only if the number of glyph in **smaller** than expected. If you have additional glyphs, this should not be an issue, should it?

Monceber:
Now that you mention it, yes, you're absolutely correct.
I'll add it back.

Issue #1955, opened 2021-01-14
Mixed beam layout refinement


craigsapp:
Here is a funny mixed-beam example:

<img width="333" alt="Screen Shot 2021-01-14 at 9 15 06 AM" src="https://user-images.githubusercontent.com/3487289/104624941-061b8400-5649-11eb-9aa0-b838e94f3396.png">

which should be renderd as:

<img width="292" alt="Screen Shot 2021-01-14 at 9 04 24 AM" src="https://user-images.githubusercontent.com/3487289/104624461-737ae500-5648-11eb-98d3-d514c63f6fdb.png">

Part of the problem is related to issue https://github.com/rism-ch/verovio/issues/1032, but there is a separate issue which is specific to mixed beams:  The spacing of mixed beams are controlled by stems rather than note heads.  The goal for spacing would be to space the stems on the mixed beam evenly.  In this case the middle note should have a stem touching the beam exactly at its middle (although other layout contexts could perturb that default spacing).

<details>
<summary>
Click to view MEI data for example.
</summary>

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2021-01-14T09:21:18" version="3.1.0-dev-0361cac-dirty">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
  </meiHead>
 <music>
  <body>
   <mdiv xml:id="mdiv-0000001676760591">
    <score xml:id="score-0000000715297447">
     <scoreDef xml:id="scoredef-0000000652114363" midi.bpm="120">
      <staffGrp xml:id="staffgrp-0000001472696674" bar.thru="true" symbol="brace">
       <staffDef xml:id="staffdef-0000000896913537" n="1" lines="5">
[...]

PR #2628, opened 2022-02-08
Improve mixed beam layout

Monceber:
This addresses a part of the issue described in #1955.

Mixed beams on one staff now should have equal distance between stems, instead of the notes. This, does not change how cross-staff beams are placed as they are handled a bit differently.
image

When there are other elements on the same duration as mixed beam (like in other layers or staves), new changes are ignored in favor of default logic, to avoid cases shown [here](https://github.com/rism-digital/verovio/issues/1955#issuecomment-763798390). 
image

We'd have to rethink the whole logic behind alignments and element placement if we were to separate their positioning based on layer/staff/etc.

### DIFF
--- a/src/layerelement.cpp
+++ b/src/layerelement.cpp
@@ -948,8 +948,8 @@ std::pair<int, int> LayerElement::CalculateXPosOffset(FunctorParams *functorPara
                 && (element->GetDrawingY() < staffBottom) && (element->GetDrawingY() < this->GetDrawingY())) {
                 verticalMargin = this->GetDrawingY() - element->GetDrawingY();
             }
-            overlap = std::max(
-                overlap, boundingBox->HorizontalRightOverlap(this, params->m_doc, margin, verticalMargin));
+            overlap
+                = std::max(overlap, boundingBox->HorizontalRightOverlap(this, params->m_doc, margin, verticalMargin));
         }
         else {
             overlap = std::max(overlap, boundingBox->HorizontalRightOverlap(this, params->m_doc, margin));
@@ -1985,7 +1985,7 @@ void LayerElement::AdjustMixedBeamXPos(Doc *doc, int drawingUnit, int &upcomingP
     if (restCount || (this->Is(NOTE) && (noteCount != 1))
         || (this->Is(CHORD) && this->GetChildCount(NOTE) != noteCount))
         return;
-    
+
     StemmedDrawingInterface *stemInterface = this->GetStemmedDrawingInterface();
     DurationInterface *duration = this->GetDurationInterface();
     if (stemInterface && duration) {


lpugin:
Wouldn't it be better to do the alignment correction within a separate functor call dedicated to it? When looking at it I thought this should be done right after the `AlignHorizontally` call. That way it should be possible to handle cases on different staves more easily. What do you think? 

The logic looking at other staves / layers looks good and you should keep it. Maybe one option would be to move them half-way when we have other elements and share the displacement between them? I am not sure how that would look.

include/vrv/functorparams.h
```diff
@@ -1332,6 +1332,30 @@ class ConvertToUnCastOffMensuralParams : public FunctorParams {
...
+ **/
+
+class CountClassIdsParams : public FunctorParams {
+public:
+    CountClassIdsParams()
+    {
+        m_elementClassIdCount.clear();
+        m_include.clear();
+        m_exlude.clear();
+    }
```
paul-bayleaf, line 1352: These are redundant as the members have default constructors.  Suggest
CountClassIdsParams() = default;

include/vrv/functorparams.h
```diff
@@ -1332,6 +1332,30 @@ class ConvertToUnCastOffMensuralParams : public FunctorParams {
...
+
+    MapOfClassIdCount m_elementClassIdCount;
+    std::vector<ClassId> m_include;
+    std::vector<ClassId> m_exlude;
```
paul-bayleaf, line 1356: typo, m_exclude

Monceber:
@lpugin Yeah, I'll have to think about this. My biggest concern with this is that even if this is done right after `AlignHorizontally`, we'd still to need to make changes to `AdjustXPos`. All elements will reach it eventually, and we'd need to make sure, that elements are not shifted further to the right, even if they overlap horizontally (as in case with mixed beams).

Regarding sharing displacement when there are other element, the biggest problem is how they rendered as a result. Even if they have shared displacement, they still would end up looking weird (since distance between stem positions for mixed and ordinary beams would differ). One way would be to have separate alignments for each staff, so that mixed beams can have their own positioning regardless of elements on the other staves, but I'm sure that would look weird as well (since notes on different staves/layers would not align).

Monceber:
> Could you illustrate this?

I'm talking about this case:
image

In mixed beams, notes from opposite sides might overlap horizontally due to placing stems equally. `AdjustXPos` as it is right now would try to space notes so that there is no such overlap and push last note to the right.

lpugin:
OK, now I see, and it makes sense. Maybe we can do both? First make a alignment correction, and then add the exception you added in `AdjustXPos` for beams within the staff.

PR #2636, opened 2022-02-11
Fixed: Deprecate Python 3.6, add 3.10

### LINKED ISSUE #2635: Remove wheel build for Python 3.6, add support for 3.10
Python 3.6 support ended on 23 Dec 2021, and 3.10 was released on 04 Oct 2021. I suggest we remove 3.6 from the PyPI build matrix, and add 3.10. 

Users still using 3.6 will still be able to build from source.

Thoughts on this @musicEnfanthen @musicog @rettinghaus ?

ahankinson:
Can wait a bit to see if anyone has any objections to the change, but this will be ready to go if/when approved.

Fixes #2635

### DIFF
--- a/.github/workflows/python-ci-wheel.yml
+++ b/.github/workflows/python-ci-wheel.yml
@@ -24,7 +24,7 @@ jobs:
       # Build the wheels for Linux, Windows and macOS for Python 3.9
       matrix:
         os: [macos-latest, windows-latest, ubuntu-20.04]
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: [3.7, 3.8, 3.9, '3.10']
         architecture: [x86, x64]
         include:
           - os: macos-latest
--- a/.github/workflows/python-ci-wheel.yml
+++ b/.github/workflows/python-ci-wheel.yml
@@ -24,7 +24,7 @@ jobs:
       # Build the wheels for Linux, Windows and macOS for Python 3.9
       matrix:
         os: [macos-latest, windows-latest, ubuntu-20.04]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
         architecture: [x86, x64]
         include:
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


lpugin:
I guess we will only see if everything works once merged into the `develop-pypi` branch.

.github/workflows/python-ci-wheel.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       # Build the wheels for Linux, Windows and macOS for Python 3.9
       matrix:
         os: [macos-latest, windows-latest, ubuntu-20.04]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
rettinghaus, line 27: I had problems with "3.10" in here, does it work locally for you?

rettinghaus (approved):

.github/workflows/python-ci-wheel.yml
ahankinson: I can merge it to develop-pypi to test.

rettinghaus: see https://github.com/rettinghaus/verovio/runs/5154734671?check_suite_focus=true

ahankinson: Looks like just quoting the value works? https://github.com/luxonis/depthai-python/actions/runs/1335184627/workflow#L111

ahankinson: I've added it to just 3.10, and pushed to develop-pypi to see, but I can also quote all the values too

musicEnfanthen: Yes, seems like version numbers should be quoted: https://github.com/actions/setup-python/issues/160#issuecomment-724485470 )see also the README examples in that repo)

ahankinson: It looks like we're still using a pinned and custom version of cibuildwheel, even though the fix is in the latest version. Should I also update cibuildwheel in the GH action?

(I wonder where this rabbit hole goes...)

lpugin: @musicEnfanthen is the only magician I know who can fix it...

ahankinson: @musicEnfanthen do you want to take this task, or should I?

rettinghaus: @ahankinson I think I figured it out, see #2638

PR #2642, opened 2022-02-14
Fix Xcode warning

DavidBauer1984:
This PR fixes the warning regarding copying the container in the loop. Additionally fixes a wrong assertion which was triggered when running the test suite.


DavidBauer1984:
@lpugin Could you have a look at stem-015 from the test library? This crashes on the dev branch. I included a fix for this in the PR, but now it's impossible to run the test suite evaluation check. Also the rendering still looks wrong to me:
stem-15
There should be flags in the second staff, right?

lpugin:
I fixed the issue. Could you revert https://github.com/rism-digital/verovio/pull/2642/commits/7e593407103cfdf5facf79f75fef0f76d7a73a78 ? Thanks

Issue #2525, opened 2021-12-03
Tablature: Augmentation dots on rhythm-signs


TimCrawford:
(Both Italian and French tab) Augmentation dots on the rhythm-signs are too small and close to the signs themselves
![Aug_dot_on_Italian _lute_tab_rhythmsign](https://user-images.githubusercontent.com/2544314/144839045-37a42888-c981-4519-a213-e1ccb4518951.jpg)

![Aug_dot_on_French _lute_tab_rhythmsign](https://user-images.githubusercontent.com/2544314/144839095-b7f887ee-a1fc-493e-ab6b-bc468de0d14a.jpg)

MEI for these examples is the same for Italian and French lute tablature, but with attribute 'notationtype' changed appropriately on line 17
```xml
<?xml version="1.0" standalone="no"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title>Augmentation dot on tablature rhythm sign - French lute tablature</title>
            </titleStmt>
            <pubStmt />
        </fileDesc>
    </meiHead>
    <music>
        <body>
            <mdiv n="1">
                <score>
                    <scoreDef>
                        <staffGrp>
                            <staffDef n="1" lines="6" notationtype="tab.lute.french">
                                <tuning>
                                    <course n="1" pname="g" oct="4" />
                                    <course n="2" pname="d" oct="4" />
                                    <course n="3" pname="a" oct="3" />
                                    <course n="4" pname="f" oct="3" />
                                    <course n="5" pname="c" oct="3" />
                                    <course n="6" pname="g" oct="2" />
                                    <course n="7" pname="f" oct="2" />
                                    <course n="8" pname="e" oct="2" />
                                    <course n="9" pname="d" oct="2" />
                                    <course n="10" pname="c" oct="2" />
                                </tuning>
                            </staffDef>
                        </staffGrp>
                    </scoreDef>
                    <section n="1">
                        <measure n="1">
                            <staff n="1">
                                <layer n="1">
                                    <tabGrp dur="2" dots="1">
                                        <tabDurSym />
                                        <note tab.course="2" tab.fret="0" />
                                    </tabGrp>
                                    <tabGrp dur="4">
                                        <tabDurSym />
                                        <note tab.course="2" tab.fret="2" />
                                    </tabGrp>
                                </layer>
                            </staff>
                        </measure>
                    </section>
                </score>
[...]

PR #2643, opened 2022-02-14
Fix Tablature: Augmentation dots on rhythm-signs #2525

paul-bayleaf:
An original example of augmentation dots from A Varietie of Lute Lessons
07_fantasie_7_dowlandj_VLL

Verovio rendered before
07_fantasie_7_dowlandJ ft3 mei svg crop

Verovio rendered after
07_fantasie_7_dowlandJ ft3 mei-2525 svg crop

These changes only affect lute tablature, not guitar.

### DIFF
--- a/src/view_tab.cpp
+++ b/src/view_tab.cpp
@@ -197,10 +197,23 @@ void View::DrawTabDurSym(DeviceContext *dc, LayerElement *element, Layer *layer,
         if (tabDurSym->GetDrawingStem()) {
             y = tabDurSym->GetDrawingStem()->GetDrawingY();
         }
-        y += m_doc->GetDrawingUnit(glyphSize) * 0.5 * stemDirFactor;
-        x += m_doc->GetDrawingUnit(glyphSize);
+
+        int dotSize = 0;
+
+        if (staff->IsTabGuitar()) {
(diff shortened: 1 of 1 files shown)


src/view_tab.cpp
```diff
@@ -197,10 +197,23 @@ void View::DrawTabDurSym(DeviceContext *dc, LayerElement *element, Layer *layer,
...
+
+        if (staff->IsTabGuitar()) {
+            y += m_doc->GetDrawingUnit(glyphSize) * 0.5 * stemDirFactor;
+            x += m_doc->GetDrawingUnit(glyphSize);
+            dotSize = glyphSize * 2 / 3;
+        }
+        else {
+            // lute tablature
+            y += m_doc->GetDrawingUnit(glyphSize) * stemDirFactor * 3 / 2;
+            x += m_doc->GetDrawingUnit(glyphSize) * 3 / 2;
```
rettinghaus, line 211: Instead of hard coding the horizontal offset it would be better to take the flag glyph width into account, because every font will look differently.

rettinghaus (changes requested):

rettinghaus:
Also the hard coded vertical shift may lead to a collision with shorter durations
tabDots

paul-bayleaf:
Thanks, I'll have to address these in another pull request as this one is closed.

PR #2647, opened 2022-02-14
Augmentation dots, allow for font width and number of flags.  Add 5 flag

paul-bayleaf:
Augmentation dots, take into account font width and the number of flags.
Added 5 flag glyph SMUFL_EBAC_luteDuration32nd, some tablature need 5 flags.

test mei
.

### DIFF
--- a/src/view_tab.cpp
+++ b/src/view_tab.cpp
@@ -173,20 +173,21 @@ void View::DrawTabDurSym(DeviceContext *dc, LayerElement *element, Layer *layer,
     int y = element->GetDrawingY();
 
     const int glyphSize = staff->GetDrawingStaffNotationSize();
+    const int drawingDur = (tabGrp->GetDurGes() != DURATION_NONE) ? tabGrp->GetActualDurGes() : tabGrp->GetActualDur();
 
     // For beam and guitar notation, stem are drawn through the child Stem
     if (!tabGrp->IsInBeam() && !staff->IsTabGuitar()) {
-        const int drawingDur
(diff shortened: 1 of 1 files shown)


paul-bayleaf:
@rettinghaus to address the issues raised in my previous pull request

src/view_tab.cpp
```diff
@@ -200,15 +201,21 @@ void View::DrawTabDurSym(DeviceContext *dc, LayerElement *element, Layer *layer,
 
         int dotSize = 0;
 
-        if (staff->IsTabGuitar()) {
+        if (tabGrp->IsInBeam() || staff->IsTabGuitar()) {
```
lpugin, line 204: Don't you want to have the same size for dots within and without beams in lute tablature? It seems that those within will now have guitar tab size (and position.)

paul-bayleaf: I'll look out some beamed tablatures with dots, but you are right we will probably need to adjust those as well.

lpugin: That is fine. Do you want me to merge this already, or wait? I am fine either way.

(PS I have to fix the additional stem showing up together with the glyph for lute tablature. I introduced the bug when working on the guitar tab. I can do that right after the merge)

paul-bayleaf: I can't find a single example in my tablature facsimiles of a dotted note under a beam!  But I've only looked at a small sample.  Experimenting using the same size and distancing for dots under a beam that I've implemented for flags looks too big and too far from the stem.   In any case as the beams don't currently look like renaissance tablature "gridiron" beams I think we can leave dots under beams as they are for later consideration.   I think it's ok to merge.

test mei

Issue #2648, opened 2022-02-15
Misplaced repetition dots in tablature


rettinghaus:
Repetition barlines in lute tablature are expected to show three dots between lines. 
![lute_repetition](https://user-images.githubusercontent.com/7693447/154040639-88a0d961-f0ed-40ad-8daa-dab204d85a16.png)

Compare the SMuFL glyph in https://w3c.github.io/smufl/latest/tables/renaissance-lute-tablature.html

In German tablature I often see five dots.
![image](https://user-images.githubusercontent.com/7693447/154038163-447f5632-5b11-48dc-9c45-7954a0d1ddc7.png)

```xml
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title>Repetions in tablatures</title>
        <respStmt>
          <persName role="encoder">Klaus Rettinghaus</persName>
        </respStmt>
      </titleStmt>
      <pubStmt>
        <date isodate="2022-02-15">2022-02-15</date>
      </pubStmt>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef>
            <staffGrp>
              <staffDef n="1" notationtype="tab.lute.french" lines="6" />
            </staffGrp>
          </scoreDef>
          <section>
            <measure left="rptstart" right="single" n="1">
              <staff n="1">
                <layer n="1">
                  <multiRest num="3" />
                </layer>
              </staff>
            </measure>
            <measure left="rptboth" right="rptend" n="1">
              <staff n="1">
                <layer n="1">
                  <multiRest num="7" />
                </layer>
              </staff>
            </measure>
            <measure right="rptstart" n="1">
              <staff n="1">
                <layer n="1">
                  <multiRest num="5" />
                </layer>
              </staff>
            </measure>
          </section>
        </score>
      </mdiv>
    </body>
  </music>
</mei>
```

PR #2653, opened 2022-02-18
Fix Misplaced repetition dots in tablature #2648

paul-bayleaf:
Fix issue #2648 

Odd number of staff lines use 2 dots, even number of staff lines use 3 dots.

4 line staff
test4 mei svg

5 line staff
test5 mei svg

6 line staff
test6 mei svg


rettinghaus (approved):
Thanks!

PR #2661, opened 2022-02-21
Use Python3 by default

lpugin:
* Specific Python version can be selected with -DPYTHON_VERSION=X.X (e.g. -DPYTHON_VERSION=2.7)
* Require CMake 3.12 or later


ahankinson (approved):

ahankinson:
Looks good. I think we should probably deploy this to the develop-pypi branch too, just to make sure it doesn't break there.

PR #2672, opened 2022-02-24
Const correctness for tree search

DavidBauer1984:
So far any element method such as `LayerElement::IsGraceNote` that (directly or indirectly) relies on descendant search cannot be made const even though it changes nothing. This PR resolves this by providing const versions for the major tree search methods. Since descendant search relies on functors we further provide the possibility to implement functors as const. The PR also includes various applications on how to use tree search in a const setting.

**Technical remarks:**

- If a function returns a pointer, then its const overload should usually return a const pointer.

- In order to avoid code duplication for const overloads, the const version of the method should be implemented and the non-const version should rely on the const version via the [following pattern:](https://stackoverflow.com/questions/123758/how-do-i-remove-code-duplication-between-similar-const-and-non-const-member-func)
```cpp
return const_cast<T *>(std::as_const(*this).f(args));
```

- Other usages of `const_cast` often indicate that something is wrong with const correctness. There are two cases where `const_cast` was introduced in the PR: (1) To access interfaces from comparisons. This is temporary and will be resolved in a later PR that improves const correctness for interfaces. (2) To update the current score in the document from a const functor. I am a bit unhappy about this, but didn't find a good solution for this now. I think that storing the current score at one central point leads to trouble when running functors (i.e. descendant search) within functors and should therefore be reconsidered.

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -287,7 +287,7 @@ public:
      * Return true if the beam has a tabGrp child.
      * In that case, the ObjectList will only have tabGrp elements. See Beam::FilterList
      */
-    bool IsTabBeam();
+    bool IsTabBeam() const;
 
     /**
      * @name Checker, getter and setter for a beam with which the stems are shared
--- a/include/vrv/btrem.h
+++ b/include/vrv/btrem.h
@@ -53,7 +53,7 @@ private:
     /**
      * Calculate the duration of an individual note in a measured tremolo
      */
-    data_DURATION CalcIndividualNoteDuration();
+    data_DURATION CalcIndividualNoteDuration() const;
 
 public:
     //
--- a/include/vrv/comparison.h
+++ b/include/vrv/comparison.h
@@ -241,7 +241,7 @@ public:
         if (!MatchesType(object)) return false;
         // This should not happen, but just in case
         if (!object->HasAttClass(ATT_NINTEGER)) return false;
-        const AttNInteger *element = vrv_cast<const AttNInteger *>(object);
+        const AttNInteger *element = dynamic_cast<const AttNInteger *>(object);
         assert(element);
         return (element->GetN() == m_n);
     }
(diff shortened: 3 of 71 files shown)


lpugin:
Thanks. Could you have a look at this https://ci.appveyor.com/project/LaurentPugin/verovio-r1t6l/builds/42696017?

lpugin:
> To update the current score in the document from a const functor. I am a bit unhappy about this, but didn't find a good solution for this now. I think that storing the current score at one central point leads to trouble when running functors (i.e. descendant search) within functors and should therefore be reconsidered.

Yes, I am not sure what to do with this. Keeping a current score was the best solution I could find when adding support for multiple scores. And that meant resetting it in the functor, which, as I understand, is this issue here.

Have you evaluated if this yields a significant increase of the executable size? This is something we need to keep under control for the JS build, but I don't think the increase should be dramatic. So just to make sure.

DavidBauer1984:
For the cmake build the increase in the executable size is 0.06% by this PR. I guess this can be neglected. The size is now 10,6 MB. How much can we afford? I guess a fair amount of the size comes from the standard library and third party libs.

lpugin:
The size increase can definitely be neglected. This is a good improvement. We can think about the score issue. Maybe we can store the current score in the functor object, have a Doc::m_currentFunctor and retrieve the current score through it (if any)? I don't know if that would work and would be any better.

lpugin:
There is one case where `Doc::SetCurrentScore` is used outside functors, which is in `Page::LayoutVertically` for getting the right header. I think it would be easy to get around it.

One issue with the current score stored in the functor might be with functors called from functors. I know we have a few, but these usually only process a sub-tree only with no score change to be expected. It should not be a problem, I think.

DavidBauer1984:
Similarly I thought about always storing it as a functor param and redirect it into sub-functors if needed. The (functor param) current score would have to be passed as argument into all the functions where it is needed. I think it's not an urgent issue since most MEIs have only one score anyway and I am not aware of any bugs because of this. But we might have to change this at some point.

lpugin:
OK, we can think about it.I am not sure my proposal of using a Doc::m_currentFunctor would work because that still means changing the Doc within const methods. 

Maybe another idea would be to store a global map of the current score corresponding *Doc object. It could be a stack (list) so each Functor object would have its corresponding current score. So something like
```cpp
std::map<Doc *, std::list<std::map<Functor *, Score *> > > vrv::m_docCurrentScore;
```
Then in `Doc::GetCurrentScore` would the correponding list from `vrv::m_docCurrentScore` and return the Score on top of the list (if any). The Functor destructor would be responsible for removing itself (map entry) from the list.

However, this is only to document the idea. We can merge it already. Thanks!

PR #2673, opened 2022-02-25
Add option to limit page justification

Monceber:
Add option `--max-vertical-justification` to limit justification of the systems on page to certain ratio.
It specifies ratio of justifiable height to page height, and enforces justification to that value. This is useful for cases where there is only one system per page which ends up being stretched to fit height of the page.
Value of 1 pretty much disables this feature, returning previous functionality.
Value of 0 functions in the same way, as if there was no justification enabled.
Default value is set to **30% (0.3)** of the page.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -594,6 +594,7 @@ public:
     OptionBool m_mensuralToMeasure;
     OptionDbl m_midiTempoAdjustment;
     OptionDbl m_minLastJustification;
+    OptionDbl m_maxSingleSystemJustification;
     OptionBool m_mmOutput;
     OptionIntMap m_footer;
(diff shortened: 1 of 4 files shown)


include/vrv/options.h
```diff
@@ -594,6 +594,7 @@ class Options {
     OptionBool m_mensuralToMeasure;
     OptionDbl m_midiTempoAdjustment;
     OptionDbl m_minLastJustification;
+    OptionDbl m_maxVerticalJustification;
```
lpugin, line 597: Please move to the general layout group and rename to `m_justificationMaxVertical` for better grouping (the one above is unfortunately badly placed / named)

PR #2696, opened 2022-03-09
Add support for rendering of non-arp arpeggios

Monceber:
Added support for the `order="nonarp"` to arpeggios:
image


lpugin (commented):
Thanks!

The logic of `View::DrawArpeg` is a bit weird. Could you re-arrange it to do:

```cpp
dc->StartGraphic(arpeg, "", arpeg->GetUuid());

if (order == arpegLog_ORDER_nonarp)
...
else
...

dc->EndGraphic(arpeg, this);
```
and avoid the `return` in the middle? Thanks

PR #2699, opened 2022-03-10
Fix beam stem calculation

Monceber:
Previous logic with stem length calculation had a problem with following condition:
```
if ((coord->m_closestNote->GetDrawingLoc() == relevantNoteLoc)
   || (!isHorizontal && (std::abs(m_uniformStemLength) < 13)))
```
In some case (where last note had the shortest stem) code would assign stem length based on the last note. However, on the next pass-through we would enter this condition regardless, since value of the `m_uniformStemLength` would be less than 13. It didn't really impact rendering of the beams so far, but calculation was invalid.

This change fixes that and makes this part more stable while retaining the same rendering.

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -137,6 +137,9 @@ private:
     // Helper to calculate max/min beam points for the relative beam place
     std::pair<int, int> CalcBeamRelativeMinMax(data_BEAMPLACE place) const;
 
+    // Helper to calculate location and duration of the note that would be setting highest/lowest point for the beam
+    std::pair<int, int> CalcStemDefiningNote(Staff *staff, data_BEAMPLACE place);
+
     // Calculate positioning for the horizontal beams
     void CalcHorizontalBeam(Doc *doc, Staff *staff, BeamDrawingInterface *beamInterface);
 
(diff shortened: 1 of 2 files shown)


lpugin:
Thanks. Please double check 

gracenote-001
image

beam-063 (we should avoid to have the beam in-between staff lines)
image

Monceber:
image
image

craigsapp:
See flags in this case 😉 

Screen Shot 2022-03-11 at 9 05 11 AM

craigsapp:
Here is the only allowable positioning of the beam in this case:

Screen Shot 2022-03-11 at 9 10 58 AM

craigsapp:
Below are the three options possible for this case:

Screen Shot 2022-03-11 at 9 18 45 AM

Screen Shot 2022-03-11 at 9 18 34 AM

In preferential order: middle, then last, then first.   The first should not be used except for some strange layout case.   The middle one is overall the best:  it is better to slightly shorten stems where there are only 1 to 2 subbeams.   The adjustment is 1.5 diatonic steps (MEI vus), which is the maximum that is acceptable in normal cases (so the first beam is way too short).   

The last one is slightly worse than the middle one, but much more acceptable than the first beam.  In this case the beam is not adjusted up/down from the octave stem lengths.  But perhaps my objection to it feeling too high is due to the middle note's stem length is 1 diatonic step higher than the octave.   Another minor thing is that the beam is tangent to the top line of the staff, making the two visually merge much more than the middle example.   

Another problem is that there is much more visual activity in the top space of the staff under the beam for the last beam compared to the middle example, making it a little harder to read the beam compared to the middle one.  In other words the white spaces under the beam in the last example are smaller than in the middle example, making these negative spaces more noticeable.   (Maybe there is a general rule to try to avoid placing beamlets on the centerline of staff lines?)

There could be other factors related to the wider musical context, such as keeping a consistent look across the system or even page that might affect the best choice between the second and third option.  

Good luck implementing computer algorithms for all these sorts of decisions 😜

Monceber:
image

Issue #2676, opened 2022-02-27
Automatic expansion of left/right margins to avoid clipping


craigsapp:
It seems useful to adjust the behavior of the left page margin option (`pageMarginLeft`) to account for system brackets and braces to the left of the staves.

Here is a demonstration of the problem that would be useful to fix:

<img width="637" alt="Screen Shot 2022-02-27 at 9 31 07 AM" src="https://user-images.githubusercontent.com/3487289/155893165-ec4137b9-2509-4538-8b9d-f8fb170554c3.png">

<details>
<summary>
Click to view MEI data for above example.
</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-02-27T09:37:40" version="3.9.0-dev-d663d28-dirty">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
  </meiHead>
 <music>
  <body>
   <mdiv xml:id="ml227v8">
    <score xml:id="spfso9w">
     <scoreDef xml:id="spomzvt">
      <staffGrp xml:id="s2yj4s7" bar.thru="true" symbol="bracket">
       <staffGrp xml:id="syoofuc" bar.thru="false" symbol="brace">
        <staffDef xml:id="swxvf7p" n="1" lines="5">
         <label xml:id="label-L4F3">Violino 1</label>
         <clef xml:id="clef-L6F3" shape="G" line="2" />
         <keySig xml:id="keysig-L7F3" sig="1f" />
         <meterSig xml:id="metersig-L8F3" count="4" sym="common" unit="4" />
        </staffDef>
        <staffDef xml:id="s26kznc" n="2" lines="5">
         <label xml:id="label-L4F2">Violino 2</label>
[...]

PR #2700, opened 2022-03-10
Clef space to thick bar line

DavidBauer1984:
This PR keeps some space between system clefs and thick bar lines (rendered via `--bar-line-width`).

| Before | After |
| ------ | ----- |
| Before | After |

### DIFF
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -1059,6 +1059,11 @@ int Measure::AdjustXPos(FunctorParams *functorParams)
         StaffAlignment *staffAlignment = system->m_systemAligner.GetStaffAlignmentForStaffN(staffN);
         params->m_staffSize = (staffAlignment) ? staffAlignment->GetStaffSize() : 100;
 
+        // Prevent collisions of scoredef clefs with thick barlines
+        if (this->IsFirstInSystem()) {
+            params->m_upcomingMinPos = params->m_doc->GetDrawingBarLineWidth(params->m_staffSize);
+        }
+
         filters.clear();
         // Create ad comparison object for each type / @n
         std::vector<int> ns;
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -1060,7 +1060,7 @@ int Measure::AdjustXPos(FunctorParams *functorParams)
         params->m_staffSize = (staffAlignment) ? staffAlignment->GetStaffSize() : 100;
 
         // Prevent collisions of scoredef clefs with thick barlines
-        if (this->IsFirstInSystem() && (system->GetDrawingScoreDef()->GetSystemLeftline() != BOOLEAN_false)) {
+        if (this->IsFirstInSystem() && system->GetDrawingScoreDef()->HasSystemStartLine()) {
             params->m_upcomingMinPos = params->m_doc->GetDrawingBarLineWidth(params->m_staffSize);
         }
 
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -238,6 +238,18 @@ void Measure::SetDrawingXRel(int drawingXRel)
     m_drawingXRel = drawingXRel;
 }
 
+bool Measure::IsFirstInSystem() const
+{
+    assert(this->GetParent());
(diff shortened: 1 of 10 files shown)


lpugin:
That looks good. However, as I understand the issue is only with the scoreDef line and not all bar lines. So I wonder if we should not instead fix https://github.com/rism-digital/verovio/issues/2676 and at the same time place the system scoreDef line on the left of the system? That way we would not need to take into account its width for the placement of the clef.

image

rettinghaus:
I'm not sure I see the connection of this fix with #2676.

lpugin:
> I'm not sure I see the connection of this fix with https://github.com/rism-digital/verovio/issues/2676.

What I am trying to say it that instead of having the scoreDef line overlapping with the staff lines we should have it placed on the left of the staff lines as illustrated with the image above. Then we would not have to adjust the clef position depending on the presence of a line in the scoreDef and of its width, which in my opinion would be better.

If we change https://github.com/rism-digital/verovio/blob/3749e205a6249bf28f6e6c5ef229b08194d49efa/src/view_page.cpp#L363 to:
```cpp
this->DrawVerticalLine(dc, yTop, yBottom, x - barLineWidth / 2, barLineWidth);
```
Then rendering changes from
image
to
image

Obviously the second one is not perfect because the height of the bar does not take into account the staff line width and the bar then overlaps with the brace. This would still need to be fixed.

The reason this relates to #2676 is that if we do this, the line is actually drawn in the margin and will not show up if the margin is set to 0:

image

rettinghaus:
Ok, thanks for the explanation. But then we would have different behaviors for the left and right (bar) lines of each system, right? I'm not sure this would be a good approach.

lpugin:
ScoreDef lines are not bar lines. If the previous system had no bar line at the end you would still draw the scoreDef line. This is actually the left barline of the first measure
image
So I don't see the problem of treating them differently. 

We can very easily adjust the position of the first measure in `System::AlignMeasures` to make sure that the score Def line is shown if the margin is 0. 

Anyway, it does not really matter because. I'll review it further.

src/measure.cpp
```diff
@@ -1047,6 +1059,11 @@ int Measure::AdjustXPos(FunctorParams *functorParams)
         StaffAlignment *staffAlignment = system->m_systemAligner.GetStaffAlignmentForStaffN(staffN);
         params->m_staffSize = (staffAlignment) ? staffAlignment->GetStaffSize() : 100;
 
+        // Prevent collisions of scoredef clefs with thick barlines
+        if (this->IsFirstInSystem() && system->GetDrawingScoreDef()->HasSystemStartLine()) {
+            params->m_upcomingMinPos = params->m_doc->GetDrawingBarLineWidth(params->m_staffSize);
+        }
```
lpugin, line 1065: You do now want to check if the measure is the first of the system and if the scoreDef has a staff line for every staff in the measure. That will be quite inefficient for an orchestral score. Move this outside the loop and use a variable for resetting `m_upcomingMinPos`

src/view_page.cpp
```diff
@@ -353,14 +353,14 @@ void View::DrawStaffGrp(
...
-            || (m_doc->GetCurrentScoreDef()->GetSystemLeftline() == BOOLEAN_true))) {
-        // int barLineWidth = m_doc->GetDrawingElementDefaultSize("bracketThickness", staffSize);
-        const int barLineWidth = m_doc->GetDrawingBarLineWidth(staffSize);
-        this->DrawVerticalLine(dc, yTop, yBottom, x + barLineWidth / 2, barLineWidth);
+    if (topStaffGrp) {
+        ScoreDef *scoreDef = vrv_cast<ScoreDef *>(staffGrp->GetFirstAncestor(SCOREDEF));
+        if (scoreDef && scoreDef->HasSystemStartLine()) {
+            const int barLineWidth = m_doc->GetDrawingBarLineWidth(staffSize);
+            this->DrawVerticalLine(dc, yTop, yBottom, x + barLineWidth / 2, barLineWidth);
+        }
```
lpugin, line 361: Since the parent scoreDef is used below in the function move it up and use the same local variable

PR #2717, opened 2022-03-16
Const interface getter

DavidBauer1984:
This PR does some maintenance for interface getters:
- Add const overloads
- Switch to vrv_cast which is faster

### DIFF
--- a/include/vrv/accid.h
+++ b/include/vrv/accid.h
@@ -52,11 +52,8 @@ public:
      * @name Getter to interfaces
      */
     ///@{
-    PositionInterface *GetPositionInterface() override { return dynamic_cast<PositionInterface *>(this); }
-    const PositionInterface *GetPositionInterface() const override
-    {
-        return dynamic_cast<const PositionInterface *>(this);
-    }
+    PositionInterface *GetPositionInterface() override { return vrv_cast<PositionInterface *>(this); }
+    const PositionInterface *GetPositionInterface() const override { return vrv_cast<const PositionInterface *>(this); }
     ///@}
 
     /** Override the method since alignment is required */
--- a/include/vrv/accid.h
+++ b/include/vrv/accid.h
@@ -48,7 +48,16 @@ public:
     /** Override the method since it is align to the staff */
     bool IsRelativeToStaff() const override { return (this->HasLoc() || (this->HasOloc() && this->HasPloc())); }
 
+    /**
+     * @name Getter to interfaces
+     */
+    ///@{
     PositionInterface *GetPositionInterface() override { return dynamic_cast<PositionInterface *>(this); }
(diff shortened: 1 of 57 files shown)


DavidBauer1984:
@lpugin No checks were executed for this PR.

rettinghaus:
Now they are.

lpugin:
We need to double check the change from `dynamic_cast` to `vrv_cast`. It is OK only if the `HasInterface` check is done before hand. It seems to be fine in most cases but it seems that we have a few ones where it is not. See

https://github.com/rism-digital/verovio/blob/632d70705c7e72cefbce9fafd3ac84cfec1203b5/include/vrv/comparison.h#L129

https://github.com/rism-digital/verovio/blob/632d70705c7e72cefbce9fafd3ac84cfec1203b5/include/vrv/comparison.h#L153

These are easy to adjust, but make sure there are none somewhere else.

DavidBauer1984:
So far we had a mix of overloading (virtual functions) and `dynamic_cast`. If the interface does not exist, then the base implementation of the interface getter already returns `NULL`. So it seems unnecessary to additional check with `dynamic_cast` and replacing it by `vrv_cast` should never change the behaviour nor require additional checks.

I only see a problem if an element class is not derived from an interface, but still the interface getter is implemented. But I would consider this an implementation error.

So in the lines above the getter returns `NULL` if the interface is not present (even with `vrv_cast`). And the result is checked for `NULL` in the next line. Seems safe to me. Am I missing something?

DavidBauer1984:
Actually when I looked into this I was wondering why we don't use `dynamic_cast` once and for all in the base implementation. Something like
```cpp
BeamDrawingInterface *GetBeamDrawingInterface() { return dynamic_cast<BeamDrawingInterface*>(this); }
```
in `object.h` without the need to override the getters anywhere.

But it turned out that this is slower: I guess this was the reason to make the interface getters virtual in the first place?!

lpugin:
> So far we had a mix of overloading (virtual functions) and dynamic_cast. If the interface does not exist, then the base implementation of the interface getter already returns NULL. So it seems unnecessary to additional check with dynamic_cast and replacing it by vrv_cast should never change the behaviour nor require additional checks.

Ah, yes, you are right. I haven't thought about that. 

> I only see a problem if an element class is not derived from an interface, but still the interface getter is implemented. But I would consider this an implementation error.

Yes, this would be an error and not likely to happen.

> Actually when I looked into this I was wondering why we don't use dynamic_cast once and for all in the base implementation. 

Yes, that would be much slower. It was actually like this at the beginning and the the `Object::m_type` was added in order to optimize the speed. I did something similar for the interfaces. I am now realizing that it is somehow duplicated with the virtual interface accessors and the `Object::HasInterface`. We can leave it like this in my opinion.

PR #2718, opened 2022-03-17
Fix beamspan stem length issue during justification

Monceber:
Fix for the vertical justification issue that happens with `beamSpan` that have cross-staff elements:
image

This is caused by the fact, that `beamSpan` is control element and is rendered after all layer elements. This means, that stem adjustments that occur inside beam code happen after stems were already rendered and do not impact final result.
With this change this problem should be gone without impacting `beamSpan` performance:
image

### DIFF
--- a/include/vrv/view.h
+++ b/include/vrv/view.h
@@ -198,6 +198,7 @@ protected:
      * Defined in view_page.cpp
      */
     ///@{
+    void DrawBeamspanOnly(DeviceContext *dc, Object *parent, Measure *measure, System *system);
     void DrawPageElement(DeviceContext *dc, PageElement *element);
     void DrawSystem(DeviceContext *dc, System *system);
     void DrawSystemList(DeviceContext *dc, System *system, const ClassId classId);
--- a/src/view_page.cpp
+++ b/src/view_page.cpp
@@ -1188,6 +1188,15 @@ void View::DrawMNum(DeviceContext *dc, MNum *mnum, Measure *measure, int yOffset
     }
 }
 
+void View::DrawBeamspanOnly(DeviceContext *dc, Object *parent, Measure *measure, System *system)
+{
+    for (auto current : parent->GetChildren()) {
+        if (!current->Is(BEAMSPAN)) continue;
+        this->DrawControlElement(dc, dynamic_cast<ControlElement *>(current), measure, system);
+    }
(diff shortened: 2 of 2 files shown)


lpugin:
Have you checked that it works over a system break?

Monceber:
Yes, I've tested it with other example that I had and rendering seems fine:
image
image

lpugin:
There are a few problems with the PR. 

First of all, if the `beamSpan` is within editorial markup, then the drawing calls will not follow the same path. `View::DrawBeamspanOnly` will still be called after drawing the staves, which is what you are trying to address. I noticed that when the `beamSpan` is on one system, it can still work (in some cases?) because `View::DrawBeamspanOnly` will still call `DrawSystemList(dc, system, BEAMSPAN)` at the end of the *first* measure, which is still before the staves of the next measure. But this will certainly yields some inconsistencies.

This leads to another issue, namely that `DrawSystemList` is called for each measure. It is meant to be call only once per system. As a result, the beams are (re-)drawn for each measure, and are duplicated in the SVG if the beam spans over two measures.

Finally, the order of the SVG does not mirror the one of the MEI. We get
```xml
<g class="measure">
   <g class="beamSpan"/>
   <g class="staff"/>
   <g class="staff"/>
</g>
```

I am looking at it and will try to make another PR to fix the issue.

lpugin:
There are a few problems with the PR. 

First of all, if the `beamSpan` is within editorial markup, then the drawing calls will not follow the same path. `View::DrawBeamspanOnly` will still be called after drawing the staves, which is what you are trying to address. I noticed that when the `beamSpan` is on one system, it can still work (in some cases?) because `View::DrawBeamspanOnly` will still call `DrawSystemList(dc, system, BEAMSPAN)` at the end of the *first* measure, which is still before the staves of the next measure. But this will certainly yields some inconsistencies.

This leads to another issue, namely that `DrawSystemList` is called for each measure. It is meant to be call only once per system. As a result, the beams are (re-)drawn for each measure, and are duplicated in the SVG if the beam spans over two measures.

Finally, the order of the SVG does not mirror the one of the MEI. We get
```xml
<g class="measure">
   <g class="beamSpan"/>
   <g class="staff"/>
   <g class="staff"/>
</g>
```

I am looking at it and will try to make another PR to fix the issue.

lpugin:
Fixed here https://github.com/rism-digital/verovio/pull/2724

PR #2733, opened 2022-03-23
Fix overlap of accidentals with ledger lines

Monceber:
Fix overlap that happens between ledger lines and accidentals, especially when `--ledger-line-extension` option is specified:
image
After fix:
image


src/accid.cpp
```diff
@@ -152,7 +154,17 @@ void Accid::AdjustX(LayerElement *element, Doc *doc, int staffSize, std::vector<
     const int unit = doc->GetDrawingUnit(staffSize);
     int horizontalMargin = doc->GetRightMargin(ACCID) * unit;
     // Reduce spacing for successive accidentals
-    if (element->Is(ACCID)) horizontalMargin *= 0.66;
+    if (element->Is(ACCID))
+        horizontalMargin *= 0.66;
+    else if (element->Is(NOTE)) {
```
lpugin, line 159: Add `{}`

PR #2738, opened 2022-03-25
Support for ligature in PAE

lpugin:
Undocumented coding of ligatures in mensural PAE uses the `+` sign. There are currently about 3,500 of them in the RISM dataset.

This PR implements support for ligature.
```
@clef:C+1
@data:1DFD+F9(F)D(F)+D0F(D)F+(D)
```
image

The importer checks that the durations are valid durations for a ligature and that there are now consecutive notes with the same pitch.

Examples
```
@clef:C+1
@data:1GG+/BG/
```
```
[Warning] PAE: To indicate a ligature, a '+' must be followed by a note. (character 4)
```
image

```
@clef:C+1
@data:1'A+2.''C4C2CC
```
```
[Warning] PAE: The duration in a ligature cannot be shorter than a semibreve. (character 8)
```
image

Limitation: there is no check the the ligature is actually valid (e.g., a pair SB B would still be imported as ligature)

### DIFF
--- a/include/vrv/iopae.h
+++ b/include/vrv/iopae.h
@@ -571,6 +571,7 @@ private:
     bool ConvertTuplet();
     bool ConvertDuration();
     bool ConvertTie();
+    bool ConvertLigature();
     bool ConvertAccidGes();
     ///@}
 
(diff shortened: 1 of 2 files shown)


rettinghaus:
As PAE is supposed to be an international standard this coding practice should be documented (and thus made official).

ahankinson (approved):

PR #2741, opened 2022-03-28
Add stylistic option for beam rendering

Monceber:
Added option `--beam-no-stem-extension` for stylistic change of beam rendering. With this option, stems inside beams will stop at first sub-beam.

image
image

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -98,6 +98,9 @@ public:
     ///@}
 
 private:
+    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-extend-stems" is set)
+    void AdjustBeamStemExtension(BeamDrawingInterface *beamInterface);
+
     // Helper to adjust beam positioning with regards to ledger lines (top and bottom of the staff)
     void AdjustBeamToLedgerLines(Doc *doc, Staff *staff, BeamDrawingInterface *beamInterface);
 
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -98,8 +98,8 @@ public:
     ///@}
 
 private:
-    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-no-stem-extension" is set)
-    void AdjustBeamStemExtension(BeamDrawingInterface *beamInterface);
+    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-french-style" is set)
+    void AdjustBeamToFrenchStyle(BeamDrawingInterface *beamInterface);
 
     // Helper to adjust beam positioning with regards to ledger lines (top and bottom of the staff)
     void AdjustBeamToLedgerLines(Doc *doc, Staff *staff, BeamDrawingInterface *beamInterface);
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -98,7 +98,7 @@ public:
     ///@}
 
 private:
-    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-extend-stems" is set)
+    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-no-stem-extension" is set)
     void AdjustBeamStemExtension(BeamDrawingInterface *beamInterface);
 
     // Helper to adjust beam positioning with regards to ledger lines (top and bottom of the staff)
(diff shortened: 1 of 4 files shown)


lpugin:
Thanks. Would it be possible to show the examples without the option? What it does is not very clear.

craigsapp:
> Thanks. Would it be possible to show the examples without the option? What it does is not very clear.

That is because your are French(ish) so the notation looks normal to you 😉 

In the standard style of stems connecting to beams (i.e., German), the stem ends at the furthest part of the beam from the notehead.  In this case the stem is ending at the first part of the beam that it encounters (except at the start/end of the beam).

In SCORE terminology, this is called "French stemming".  Here is the version 4 update manual for SCORE where it was added:

Screen Shot 2022-03-28 at 11 45 13 PM

In other words, the AD command is used to connect the stems to the furthest position on the beam, while the ADF (for "ADjust French") to move the stems to the closest position on the beam.

craigsapp:
Here is an example of French stemming from a French edition of Beeethoven's piano sonatas:

Beethoven, Sonatas Volumes 1 & 2 edited by Paul Dukas. Édition classique a Durand & fils, No. 9327. 1915.

https://github.com/craigsapp/beethoven-piano-sonatas/blob/master/reference-edition/sonata08-1.pdf

Screen Shot 2022-03-28 at 11 54 58 PM

Compare to 

[Ludwig van Beethovens Werke, Serie 16, Nr.131](https://imslp.org/wiki/Ludwig_van_Beethovens_Werke_(Beethoven,_Ludwig_van)) (pp.121-36)
Leipzig: [Breitkopf und Härtel](https://imslp.org/wiki/Breitkopf_und_H%C3%A4rtel), n.d.[1862]. Plate B.131.

https://s9.imslp.org/files/imglnks/usimg/e/ee/IMSLP243128-PMLP01410-Beethoven,_Ludwig_van-Werke_Breitkopf_Kalmus_Band_20_B131_Op_13_scan.pdf

Screen Shot 2022-03-28 at 11 57 02 PM

Screen Shot 2022-03-28 at 11 57 22 PM

lpugin:
> That is because your are French(ish) so the notation looks normal to you

Yeah.. and commenting before my coffee (never do that) and thinking primarily in terms of beam layout / positioning rather than styling 😀

Shouldn't we go for a more explicit option name such as `--beam-french-style` ?

src/beam.cpp
```diff
@@ -431,6 +432,59 @@ bool BeamSegment::NeedToResetPosition(Staff *staff, Doc *doc, BeamDrawingInterfa
...
+
+    // set to store durations of relevant notes (it's ordered, so min duration is going to be first)
+    std::set<int> noteDurations;
+    // lambda check whether coord has element set and whether that element is CHORD or NOTE
+    const auto isNoteOrChord = [](BeamElementCoord *coord) {
+        return (coord->m_element && coord->m_element->Is({ CHORD, NOTE }));
+    };
+    // iterators
+    using VectorIt = ArrayOfBeamElementCoords::iterator;
+    using VectorReverseIt = ArrayOfBeamElementCoords::reverse_iterator;
```
lpugin, line 447: MINOR I would rename to something more specific. `CoordIt`? or `BeamCoordIt`?

craigsapp:
> Shouldn't we go for a more explicit option name such as `--beam-french-style`?

I like that since I think it is not common outside of France (from 1900 +/- 50 years approximately).

src/options.cpp
```diff
@@ -1147,6 +1147,11 @@ Options::Options()
     m_beamMinSlope.Init(0, 0, 0);
     this->Register(&m_beamMinSlope, "beamMinSlope", &m_generalLayout);
 
+    m_beamNoStemExtension.SetInfo("No stem extension in beams",
+        "For notes in beams, stems will stop at first outermost sub-beam without crossing it");
+    m_beamNoStemExtension.Init(false);
+    this->Register(&m_beamNoStemExtension, "beamNoStemExtension", &m_generalLayout);
+
```
lpugin, line 1154: I would prefer a more explicit option name. I would suggest `beamFrenchStyle`

include/vrv/beam.h
```diff
@@ -98,6 +98,9 @@ class BeamSegment {
     ///@}
 
 private:
+    // Helper to adjust stem length to extend only towards outmost subbeam (if option "--beam-no-stem-extension" is set)
+    void AdjustBeamStemExtension(BeamDrawingInterface *beamInterface);
```
lpugin, line 102: Should somehow match the option name if that one is changed

lpugin (approved):

PR #2742, opened 2022-03-28
Minor fix for beam positioning

Monceber:
- changed code to make sure that beams with unit height are not adjusted up/down which results in hanging beam ends

Fix for some cases where beam were still not touching staff lines:
image
image

After fix:
image
image


Monceber:
It doesn't seem like build failure is caused by the changes in the PR. 
slur-015.mei is rendered properly with them:
image

lpugin:
This is weird. I also wonder why we get all the `"transposeMdiv" is ignored when "transpose" is set as well. Please use only one of the two options.` warnings. Any ideas?

DavidBauer1984:
@lpugin The problem is that the test suite dumps all options into json and then reloads them. This triggers the transposition warning since it relies on `IsSet`. Currently working on a fix for this...

lpugin:
It crashes on the `tuplet-014.mei` in `BeamSegment::CalcAdjustPosition`. Please double check.

lpugin (approved):

PR #2724, opened 2022-03-22
Fix cross-staff beamspan stems

lpugin:
Another approach for the fix proposed in https://github.com/rism-digital/verovio/pull/2718

Before
image

After
image

<details>
<summary>MEI code</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Beam span</title>
         </titleStmt>
         <pubStmt>
            <respStmt>
               <persName role="editor">Laurent Pugin</persName>
               <persName role="encoder">Andriy Makarchuk</persName>
            </respStmt>
            <date isodate="2022-02-15">2022-01-15</date>
            <pubPlace>
               <ref target="https://github.com/rism-digital/verovio/pull/2640" />
            </pubPlace>
         </pubStmt>
         <seriesStmt>
            <title>Verovio test suite</title>
         </seriesStmt>
         <notesStmt>
            <annot>BeamSpans can be cross-staff and cross-measure.</annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="3.9.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef midi.bpm="120" key.mode="major" key.sig="2f" meter.count="4" meter.unit="2" meter.sym="common">
                  <staffGrp symbol="brace" bar.thru="true">
                     <staffDef n="1" lines="5" clef.shape="G" clef.line="2">
                        <instrDef midi.channel="0" midi.instrnum="0" />
                     </staffDef>
                     <staffDef n="2" lines="5" clef.shape="F" clef.line="4">
                        <instrDef midi.channel="1" midi.instrnum="0" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure metcon="false" n="222">
                     <staff n="1">
                        <layer n="1">
[...]

PR #2750, opened 2022-03-30
Adjust System::AdjustCrossStaffYPos functor.

lpugin:
* Addresses https://github.com/rism-digital/verovio/pull/2724#issuecomment-1082894019


lpugin:
Even though the test suite evaluation does not show it, it is fixed

image

The problem is that the change is too small to trigger a diff evaluation. Probably the very reason why the bug was not noticed.

PR #2754, opened 2022-03-31
Add new feature to select a portion of a score

lpugin:
This PR add a new `Select` method to the toolkit. It allows to select a portion of a score for rendering.

The `Select` method takes one stringified JSON object (or a JSON object for the JS toolkit) parameter.

The selection is measure-based. It can be made either with `start` and `end` xml:id, or with a `measureRange`. The measure range syntax is similar to the one of the [EMA](https://github.com/music-addressability/ema/blob/master/docs/api.md) API. However, because only one range is supported, the key is `measureRange` and not `measureRanges`.

Example with the http://verovio.humdrum.org/?file=chopin-first-editions/010-1b-Sm-012.krn

It is now possible to do (Python toolkit)
```python
tk = verovio.toolkit()
tk.select("{ 'measureRange': '2-3' }")
tk.loadFile('/Users/laurent/xcode/010-1b-Sm-012.mei')
svg = tk.renderToSVG()
```

image

Note that slurs starting or ending outside the selection are rendered properly. This also applies to other control events.

The selection does not have to be made before a file is loaded. It can be done after, or it can be changed. For example, we can change the selection to something else. It applies once `redoLayout` is called.
```python
tk.select("{ 'measureRange': '82-end' }")
tk.redoLayout()
svg = tk.renderToSVG()
```
image

Single measure selection is also possible
```python
tk.select("{ 'measureRange': '38' }")
```
image

Selection with xml:id
```python
tk.select("{'start': 'measure-L337', 'end': 'measure-L355'}")
```

Once a selection has been made, it can be rendered like an entire file. For example, a selection of several measures can be rendered on multiple pages, and rendering options can be changed. Redoing the layout will be much faster because only the selection will be re-laid out.

Retrieving the MEI always return the entire MEI file.

A selection can be cancelled with an empty string or an empty JSON object
```
tk.select("")
```


src/object.cpp
```diff
@@ -1255,7 +1255,7 @@ bool Object::IsPreOrdered(const Object *left, const Object *right)
     if (commonParent) {
         return (commonParent->GetChildIndex(*iterPair.first) < commonParent->GetChildIndex(*iterPair.second));
     }
-    return false;
+    return true;
```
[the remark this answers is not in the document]
DavidBauer1984: I think it's fine. If I remember correctly, this is only relevant if `left` and `right` live in separate trees. This should not be relevant in the other cases where the function is applied.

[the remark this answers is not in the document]
lpugin: OK. Let's remember this change if the weird slurs show up again.

DavidBauer1984:
This is really cool! I have played around a bit.

- When selecting two measures, they are rendered across the whole page width. For example, in Landscape mode I get
Select1

- Key signature changes that appear outside (before) the selection are missed:
Select2

- Before processing the selection I have a `Page` with children `Mdiv, Score, System, PageMilestoneEnd, PageMilestoneEnd`. After `InitSelection` selecting measures 2-3 the tree structure becomes `Page` with children `Score, System`. Is it really safe for the rest of the layout to operate on a tree without `Mdiv` and without corresponding end milestone for the score?

craigsapp:
Looks great!  

Your example has an end-of-movement marker (`end`):

> tk.select("{ 'measureRange': '82-end' }")

You don't give an example at the start.  Is there a symbolic start marker such as `begin`?    Such a thing is necessary due to pickup measures (which might technically be labeled as measure "0").

lpugin:
> This is really cool! I have played around a bit.

Yes, I think this approach as a pretty good potential. It still needs to be polished, though.

>     * When selecting two measures, they are rendered across the whole page width. For example, in Landscape mode I get
> 

You should be able to change the layout options and have it rendered as desired. Could you be more specific if there there any issues?

> 
> Select1
> 
>     * Key signature changes that appear outside (before) the selection are missed:

I'll check this.

> 
>     * Before processing the selection I have a `Page` with children `Mdiv, Score, System, PageMilestoneEnd, PageMilestoneEnd`. After `InitSelection` selecting measures 2-3 the tree structure becomes `Page` with children `Score, System`. Is it really safe for the rest of the layout to operate on a tree without `Mdiv` and without corresponding end milestone for the score?

`mdiv` should not be an issue because they are not structuring. I was more concerned about the `score` not having a corresponding milestone. However, I could not find problems so far. I also tried with `--mdiv-all` and multiple `mdiv`s and it did work. But you are right that this is a potential problem we need to look at.

lpugin:
> > tk.select("{ 'measureRange': '82-end' }")
> 
> You don't give an example at the start. Is there a symbolic start marker such as `begin`? Such a thing is necessary due to pickup measures (which might technically be labeled as measure "0").

The beginning can be indicated with `start`. See the [EMA specifications](https://github.com/music-addressability/ema/blob/master/docs/api.md#measures). Tagging @raffazizzi for pickup measures. Maybe he can advise us?

raffazizzi:
@lpugin awesome to see that this will be a feature in Verovio. It'll be so useful! The EMA specification addresses measures by their index in a sequence of contiguous measures in a document, starting at `1`. In a strict interpretation, a pick-up measure is the first measure in a document, albeit incomplete, and would be addressed by and index of `1` (these indexes need not match human readable labels). 

This interpretation would be my preference. If, however, this seems "wrong" to Verovio developers, there is room to adjust the EMA specification since it does not mention pick-up measure explicitly. I am keen on keeping things aligned and can publish a new version of the specification (e.g. 1.0.1) with a clarification on pick-up measures if need be.

DavidBauer1984:
**Stretching issue**
The issue (in my view) is that when using selection the last measure is always stretched to full page width. If I render a piece consisting of a single measure, then the output is not stretched to full page width:
WithoutSelection
However, after selecting a single measure in a piece with several measures, the output is stretched:
WithSelection
The problem also appears here where I select measures 5-11. Measures 5-10 are placed on the first system, while measure 11 goes into the second. However, then measure 11 should not be stretched to full system width:
WithSelection2
Or is there a special rendering option to prevent this behaviour when using selection?

**Tree structure**
The only place where I know that we rely on the pairing of milestone elements and their end markers is the MEI export. And this is not effected by the selection. Maybe it's ok. Also milestone elements have only been recently introduced, so we don't have a lot of functors using them. I think it's still helpful to clarify which tree structures are valid and which are not. Otherwise it easily happens that we make certain assumptions in functor code that do not hold true in general.

lpugin:
@DavidBauer1984 the missing `mdiv` milestone was causing the stretching issue. That should be fixed now.

@RichardFreedman you would need to build the Python toolkit within your virtual environment. If you do so, then yes, you should be able to select and render a range of measure. If you cannot build it locally, hopefully the feature will be ready for the next release!

RichardFreedman:
@lpugin Thanks for this great new feature, Laurent.  It seems I will need to wait until the next update on pypi.  I am able to install from there, but installing directly from git is not working for me.  No worries!  There is plenty else to do!  (But I would like to show you CRIM Intervals in Jupyter Notebooks some time, and now Verovio will be a part of those!)

DavidBauer1984:
Here is another improvement that would be nice. After selecting `100-end` in some piece I obtain the output:
Selection
So apparently the output starts at measure 97 instead of 100. The reason for this is that the current implementation picks the 100th measure encoded in the piece and not the measure with `n="100"`. Usually those are not the same because of pick-up measures or different endings. 
It would be nice if for a selection `A-B` we check if there actually is a measure with encoded number `n="A"` or `n="B"`. If so prefer this one. Otherwise we could still fall back to the A-th or B-th measure in the piece.

include/vrv/docselection.h
```diff
@@ -0,0 +1,60 @@
...
+    ///@{
+    DocSelection();
+    virtual ~DocSelection(){};
+    ///@}
+
+    /**
+     * Parse the stringified JSON string a stores the values.
+     * Priority is given to measureRange selection over start / end ids
+     */
+    bool Parse(const std::string selection);
```
DavidBauer1984, line 39: (MINOR) Better pass arg as const reference.

include/vrv/object.h
```diff
@@ -905,7 +905,7 @@ class Object : public BoundingBox {
     /**
      * Calculate the slur direction
      */
-    virtual int PrepareSlurs(FunctorParams *) { return FUNCTOR_CONTINUE; }
+    virtual int CalcSlurs(FunctorParams *) { return FUNCTOR_CONTINUE; }
```
DavidBauer1984, line 908: Can we rename this into `CalcSlurDirection`? That's all it does for now. The actual slur calculation happens during drawing and in the `AdjustSlurs` functor.

rettinghaus:
@DavidBauer1984 I'm not sure about the expected result when selecting `100-end`. There are too many ways to encode things like measure numbers, so that you'll at some point always miss the thing you want to get. As @raffazizzi pointed out the first measure in the MEI is actually _the first measure_. `measure@n` is `att.nNumberLike`, so what to do with "71a" and "71b". Upbeat measures may have a "0" value when it's at the beginning of a piece, but what would you expect for mid-piece upbeat measure? Should that one have a `@n` at all? You could check for `metcon="false"`, but that's not fully reliable also, because then you maybe miss a cadenza (and also not everyone is using this attribute). Furthermore there are possible discrepancies between the `@n` value and the string in `mNum`. Which one should be selected then?
The only way to solve this is probably a full XPath selector, but this is too much for this specific feature here. This should be done before hand if you want to use this feature, because you're probably the only one who knows your MEI's structure.

src/docselection.cpp
```diff
@@ -0,0 +1,169 @@
...
+    doc->m_selectionEnd = "";
+
+    if (m_selectionRangeStart != VRV_UNSET) {
+        std::string selectionStartId = "";
+        std::string selectionEndId = "";
+
+        ListOfConstObjects measures;
+        ClassIdComparison matchType(MEASURE);
+        // Depth is this->pages->page-system->measure
+        doc->FindAllDescendantsByComparison(&measures, &matchType, 4);
```
DavidBauer1984, line 128: Could also use `FindAllDescendantsByType(MEASURE, false)` here. Then you don't have to worry about search depth.

lpugin:
> It would be nice if for a selection A-B we check if there actually is a measure with encoded number n="A" or n="B". If so prefer this one. Otherwise we could still fall back to the A-th or B-th measure in the piece.

As @rettinghaus pointed out, that would work only for some simple cases. I would recommend to do the selection with `start` and `end` with `measure@xml:id`s instead of `measureRange` when that one is not ideal.

ahankinson (approved):

craigsapp:
> @DavidBauer1984 I'm not sure about the expected result when selecting 100-end.

I would expect it to be from the measure labeled as n="100" through the end of the movement;  Not the 100th `<measure>` through the end of the piece. This feature is for humans looking at a score, seeing measure numbers on the score and wanting to extract a musical example based on that.

> It would be nice if for a selection A-B we check if there actually is a measure with encoded number n="A" or n="B". If so prefer this one. Otherwise we could still fall back to the A-th or B-th measure in the piece.

That sounds good.   If there is more than one measure labeled n="A", then the first one should be selected.   And then n="B" would be the first measure labeled "B" after that.  If either n="A" or n="B" does not exist, then fall back to counting `<measure>`.

> so what to do with "71a" and "71b".

When `@n` numbers are not sufficiently precise, then that is the time to use the `xml:id` system.

If you select measures 70–72, then the measure sequence extracted would be `70, 71a, 71b, 72, 73`.  If the range is starting or ending on `71a` or `71b` then that will depend on how it is implemented (the best thing is for publishers to not use the stupid `a/b` system for labeling measures in first/second endings, thus avoiding the problem).  A problem can occur if you allow `71a`, because then you have to worry about someone using `71-a` as a measure number, which will break the system:

```
tk.select("{ 'measureRange': '71-b-82' }")
```

`measure@n` used to be only a number, but was switched to be "number-like" (to allow for cases such as `71a` and `71b` to be encoded in `@n`).  This allows for correct parsing of the measure number as long as there is no ASCII dash in the string:

```
tk.select("{ 'measureRange': '71b-82' }")
```

A complication occurs if you specify the range `71-82`.  Ideally that should include `71a` and `71b` (or at least start at `71a`).  But that would be too smart to handle such cases.

> Furthermore there are possible discrepancies between the @n value and the string in mNum. Which one should be selected then?

I would be in favor of limiting it to `@n`, and if that is not sufficient due to some complexity in the measure numbering, then use `@xml:id` addressing.

---------------

Another complexity is what to do if the measure numbers for the range are both present, but in reverse order.   The measure extraction tool in VHV allows such a case, but the range is not corrected, and the order of the extracted measures is reversed (this is more for compositional and analytic use rather than extracting examples):

https://verovio.humdrum.org/?file=chorales/chor010.krn
Screen Shot 2022-04-05 at 4 56 50 AM

https://verovio.humdrum.org/?file=chorales/chor010.krn&filter=myank%20-m%20%24-0
Screen Shot 2022-04-05 at 4 57 54 AM

raffazizzi:
To avoid these issues, the EMA specification disregards labels and simply uses the measure index. This allows the processing to be simple and selection tools can be created to aid the user in picking the right measures. I think trying to account for human readable labels can get too complex and more edge cases will emerge.

Perhaps a solution would be implementing something akin to `GetPageWithElement` in order to get the correct measure index number. E.g. `getMeasureWithElement`. The user can use that method to obtain the correct measure(s) index and use the returned values with `select`

PR #2757, opened 2022-04-01
Trim leading and trailing spaces from attributes

### LINKED ISSUE #2671: Handling booleans
Verovio cannot handle all allowed encodings of booleans in MEI.

Things like `num.visible="false"` work fine. 
What doesn't work (and throws a warning) is `num.visible=" false "`, although this is valid.

I think LibMEI is what would be needed to change here, but we can also deal with this while loading a file. Any thoughts?

Monceber:
closes #2671

Fix for an issue with spaces in attributes without relying on MEI changes. This removes trailing and leading spaces that might be present in any of the attributes in MEI file, while leaving any other spaces (like the ones in @staff or @plist) intact.

### DIFF
--- a/include/vrv/iomei.h
+++ b/include/vrv/iomei.h
@@ -807,6 +807,11 @@ private:
     bool IsEditorialElementName(std::string elementName);
 
     /**
+     * Normalize attributes of xmlElement, removing white spaces if necessary
+     */
+    void NormalizeAttributes(pugi::xml_node &xmlElement);
+
+    /**
      * Read score-based MEI.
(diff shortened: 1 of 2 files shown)


lpugin:
Please rebase. Thanks

ahankinson:
@craigsapp I think this is just for values in attributes: `<foo bar=" stuff ">`

craigsapp:
OK, that is good (but it would be good to only filter spaces and tabs from attributes as well).  I see from the removal of std::isspace that non-breaking spaces will not be removed (but it might be useful to also remove tabs if they are somehow added to attributes).

PR #2774, opened 2022-04-06
Code refactoring for clarifying the role and scope of the functors.

lpugin:
Following the comment in https://github.com/rism-digital/verovio/pull/2754#issuecomment-1087608170 the PR refactors the code in order to clarify the role and the scope of the functors.

The PR is only about renaming and moving methods in the code and should yield no change in behaviour. The only exception is in https://github.com/rism-digital/verovio/commit/e12068007be046eba88ef25ea02b523905b77e7b with the move of the processing of `ResolveSpanningBeamSpans` (now `CalcSpanningBeamSpans`) from `Page::LayOutHorizontally` to its call to `Page::ResetAligners`, which should make not difference.

The PR aims at clarifying the following structure in Verovio rendering process

#### Step 1 

Preparing the data before processing it. This includes typically resolving pointers such as `@startid`, `@endid`, `@sameas` or calculating `@tstamp` or `@tsamp2` positions. This is performed when the file is loaded and should be redone if the content is changed (e.g., through the editor API). If a selection has been made, it should be deactivated before doing it again and can be reactivated after.

#### Step 2 

Calculating and adjusting the horizontal and the vertical layouts. This includes grouping events appearing at the same time (e.g. across staves) and adjusting their position for avoiding collision detection.

#### Step 3 

Justification of the content. That includes horizontal and vertical justification.

#### Step 4

Drawing the content

### Code structure

Obviously, there are not just a succession of steps. For example, calculating the layout does require content to be drawn at intermediate stages in order to have access to the bounding boxes for collision detection. Along the same lines, casting off the content requires different successions / repetitions of layout calculation and redrawing.

Nonetheless, the PR tidies up the code by adjusting the functors related to theses steps with some naming conventions.

Functors for **Step 1** are called from `Doc:PrepareData` - previously `PrepareDrawing` but renamed for better distinguishing its role in the process. All the functor names should be prefixed with `Prepare` (e.g., `Object::PrepareTimestamps`)
See https://github.com/rism-digital/verovio/blob/develop-doc/include/vrv/object.h#L1194-L1365

Functors for **Step 2** are called from `Page::LayoutHorizontally` / `Page::LayoutVertically` and related methods. The functor names should be prefixed with `Calc`, `Adjust` or `Align` depending on their roles in the process.
See https://github.com/rism-digital/verovio/blob/develop-doc/include/vrv/object.h#L801-L984

Functors for **Step 3** are only a few and should be prefixed with `Justify`

**Step 4** is essentially loop-based and does not have a dedicated set of functors. 

There are many additional functors that can have a quite generic (e.g., `Object::FindAllByComparison`) or on the contrary a very specific scope (e.g.,`Object::GenerateMIDI`). The prefixes used above should be avoided. Typically, the PR renames `PrepareMIDI` to `InitMIDI`.
See https://github.com/rism-digital/verovio/blob/develop-doc/include/vrv/object.h#L1435-L1492


include/vrv/linkinginterface.h
```diff
@@ -70,23 +70,23 @@ class LinkingInterface : public Interface, public AttLinking {
...
-     * classes explicitely. See FloatingObject::FillStaffCurrentTimeSpanning for an example.
+     * classes explicitely. See FloatingObject::PrepareStaffCurrentTimeSpanning for an example.
      */
 
     /**
-     * See Object::FillStaffCurrentTimeSpanning
+     * See Object::PrepareStaffCurrentTimeSpanning
      */
-    virtual int InterfaceFillStaffCurrentTimeSpanning(FunctorParams *functorParams, Object *object);
+    virtual int InterfacePrepareStaffCurrentTimeSpanning(FunctorParams *functorParams, Object *object);
```
ahankinson, line 79: Most of the renamed functions start with a verb -- "Calc", "Init", "Prepare", etc. This one starts with a noun. 

I'm not 100% sure what else to call it, but thought I might flag this as standing out.

include/vrv/object.h
```diff
@@ -829,36 +889,44 @@ class Object : public BoundingBox {
...
+
+    /**
+     * End Functor for Object::AdjustLayers
+     */
     virtual int AdjustLayersEnd(FunctorParams *) { return FUNCTOR_CONTINUE; }
-    ///@}
 
     /**
-     * @name Lay out the X positions of the grace notes looking at the bounding boxes.
+     * Lay out the X positions of the grace notes looking at the bounding boxes.
```
ahankinson, line 910: You have "LayOut" as a function prefix that you've used, but this function is called "AdjustGraceXPos". Maybe for terminology you can fix the docstring? "Lay out the X positions" -> "Adjust the X postitions"?

include/vrv/object.h
```diff
@@ -871,41 +939,47 @@ class Object : public BoundingBox {
...
     virtual int AdjustXOverflow(FunctorParams *) { return FUNCTOR_CONTINUE; }
+
+    /**
+     * End Functor for Object::AdjustXOverflow
+     */
     virtual int AdjustXOverflowEnd(FunctorParams *) { return FUNCTOR_CONTINUE; }
-    ///@}
 
     /**
      * Lay out the X positions of the staff content looking at the bounding boxes.
```
ahankinson, line 953: Same here -- "Lay out" -> "Adjust"

include/vrv/object.h
```diff
@@ -871,41 +939,47 @@ class Object : public BoundingBox {
...
      * Calculate the x position of tuplet brackets and num
      */
     virtual int AdjustTupletsX(FunctorParams *) { return FUNCTOR_CONTINUE; }
 
     /**
-     * Calculate the slur direction
+     * Cache or restore cached horizontal layout for faster layout redoing
      */
-    virtual int CalcSlurDirection(FunctorParams *) { return FUNCTOR_CONTINUE; }
+    virtual int LayOutHorizontallyCache(FunctorParams *) { return FUNCTOR_CONTINUE; }
```
ahankinson, line 982: "Horizontally" is an adverb, but it's modifying a noun here; Perhaps either `LayOutHorizontalCache` or `LayOutHorizontallyCached`

If this is only for caching, perhaps it would be clearer to move `Cache` to the front?

`CacheHorizontalLayOut`?

include/vrv/object.h
```diff
@@ -1330,6 +1423,13 @@ class Object : public BoundingBox {
      */
     virtual int UnCastOff(FunctorParams *) { return FUNCTOR_CONTINUE; }
 
+    /**
+     * CasttOff a document to selection.
```
ahankinson, line 1427: Extra `t` in `Cast`

ahankinson (changes requested):
Just a few comments and a spelling mistake.

include/vrv/linkinginterface.h
lpugin: That is because this is not an Object functor but instead a Interface method redirected from a functor. A bit complicated to explain, but they are all prefixed with `Interface` to make that clear. Should be fine as is.

include/vrv/object.h
lpugin: Yes, I was scratching my head for this one. This is what we call in `Doc::CastOffDocBase`
https://github.com/rism-digital/verovio/blob/6e5213b585ef63278a5da9c13b0c1de483b45d5e/src/doc.cpp#L911-L921

`Page::LayOutHorizontallyCachePage` is used for caching but also restoring the cache and calls this functor. I would suggest to:
* Rename `Page::LayOutHorizontallyCachePage` to Page::LayOutHorizontallyWithCache`
* Rename the functor `LayOutHorizontallyCache` to `CacheHorizontalLayout`

Does it look any clearer?

ahankinson: Yes, that looks better!

ahankinson: "Layout" and "Lay out" -- Although English usage isn't 100% consistent, the former is a noun, and the latter is a verb ("to lay out").

lpugin: Yes, the idea was to have a verb

PR #2780, opened 2022-04-07
Use glyphs for dynamics enclosure

Monceber:
Adjust code for drawing enclosing symbols for dynamics:
- use SMUFL glyphs to draw opening and closing symbols when there are only dynamic symbols
- use VerovioText to when dynamics are mixed with text (instead of adding children to Dynam just add symbols to text to be rendered)

Before:
image

After:
image

### DIFF
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -1652,8 +1652,7 @@ void View::DrawDynam(DeviceContext *dc, Dynam *dynam, Measure *measure, System *
         // editorial element within the dynam as it would with text. Also, it is center only if it is a symbol.
         if (isSymbolOnly) {
             dc->SetFont(m_doc->GetDrawingSmuflFont((*staffIter)->m_drawingStaffSize, false));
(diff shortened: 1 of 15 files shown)


src/view_control.cpp
```diff
@@ -1669,6 +1670,54 @@ void View::DrawDynam(DeviceContext *dc, Dynam *dynam, Measure *measure, System *
...
+
+    // calculate width of the symbol-only dynamic; generally it consists of only one symbol, but in case when it's
+    // combination of different dynamics glyphs we need to get total width of all elements
+    const int left = m_doc->GetGlyphLeft(dynamSymbol.at(0), staff->m_drawingStaffSize, false);
+    int width = 0;
+    for (int i = 0; i < dynamSymbol.size(); ++i) {
+        if (i == dynamSymbol.size() - 1)
+            width += m_doc->GetGlyphRight(dynamSymbol.at(i), staff->m_drawingStaffSize, false);
+        else
+            width += m_doc->GetGlyphAdvX(dynamSymbol.at(i), staff->m_drawingStaffSize, false);
```
lpugin, line 1693: Add `{}`

Issue #2770, opened 2022-04-05
detached slurs [regression]


frakel:
From 3.7 onwards Verovio renders slurs in certain contexts apparently worse
  than previous versions. Here is an example from Mozart K.475 m.144.146:
  
![475_m144_detached-slurs_compare](https://user-images.githubusercontent.com/5627140/161792758-6fb3342c-74c2-4a73-966c-7c2222e620fc.png)

  <details><summary>Click here for the MEI code:</summary>
<p>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/4.0.1/mei-CMN.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/4.0.1/mei-CMN.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei meiversion="4.0.0" xmlns="http://www.music-encoding.org/ns/mei">
  <meiHead>
    <fileDesc>
      <titleStmt>
        <title label="NMA digital" type="unit">K475 m.144.146 slur regression</title>
        <contributor>dime/fk</contributor>
      </titleStmt>
      <pubStmt/>
    </fileDesc>
  </meiHead>
  <music>
    <body>
      <mdiv>
        <score>
          <scoreDef key.mode="major" key.pname="f" key.sig="1f" meter.count="2" meter.sym="cut" meter.unit="2" xml:id="scoreDef_01">
            <staffGrp bar.thru="true" symbol="brace" xml:id="staffGrp_01">
              <staffDef clef.line="2" clef.shape="G" lines="5" n="1" xml:id="staffDef_P1"/>
              <staffDef clef.line="4" clef.shape="F" lines="5" n="2" xml:id="staffDef_P2"/>
            </staffGrp>
          </scoreDef>
          <section>
            <measure n="144" xml:id="m144_k475_001">
              <staff n="1" xml:id="staff_40458">
                <layer n="1" xml:id="layer_40464">
                  <rest dur="4" oloc="5" ploc="a" xml:id="rest_40470"/>
                  <rest dur="8" oloc="5" ploc="a" xml:id="rest_40476"/>
                  <beam xml:id="beam_40482">
                    <chord dur="8" xml:id="chord_40488">
                      <supplied xml:id="supplied_40494">
                        <artic artic="spicc" xml:id="artic_40500"/>
                      </supplied>
[...]

Issue #2730, opened 2022-03-23
spacing + slur interaction


craigsapp:
In the following two examples, the musical spacing is the only change between them:
<img width="287" alt="Screen Shot 2022-03-22 at 10 46 49 PM" src="https://user-images.githubusercontent.com/3487289/159632359-279a1dd7-6551-42af-8b68-952a4dbb1518.png">
<img width="381" alt="Screen Shot 2022-03-22 at 10 47 15 PM" src="https://user-images.githubusercontent.com/3487289/159632368-7f0ffc81-54a3-4bd5-b856-384e552b5c2d.png">

<details>
<summary>
Click to view MEI data for above examples
</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-03-22T22:47:17" version="3.10.0-dev-fe12a79-dirty">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mnyz661">
    <score xml:id="skiekyj">
     <scoreDef xml:id="s4ykmx2">
      <staffGrp xml:id="slleks2">
       <staffDef xml:id="sifvf81" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="F" line="4" />
        <keySig xml:id="keysig-L3F1" accid="f" pname="b" mode="major" sig="2f" />
        <meterSig xml:id="metersig-L5F1" count="2" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
     <section xml:id="section-L1F1">
      <measure xml:id="measure-L1" n="2">
       <staff xml:id="s9qyfbg" n="1">
[...]

PR #2781, opened 2022-04-07
Slur improvements

DavidBauer1984:
This PR does some improvements for slurs:

**1. Detached slurs**
We fix the regressions observed in #2730 and #2770 by reverting the changes in PR #2583. 
Fixed3
Fixed1
 @craigsapp Please check if #2418 should be reopened again. Note that due to recent changes in slur thickness the example doesn't look as ridiculous anymore.
Unfixed
However the directional problem (slur going down while notes going up) still persists. In retrospect it probably makes more sense to address this in the initial slur calculation instead of collision avoidance.

**2. Slurs moving outwards at end points**
We enforce that the slur direction at the end points is vertical in the worst case.
| Before | After |
| ------ | ----- |
| Before | After |

<details>
<summary> Show MEI example </summary>

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-03-25T05:47:52" version="3.8.0-dev-c01277e">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mu8bh4q">
    <score xml:id="s78qk69">
     <scoreDef xml:id="s62kb8i">
      <staffGrp xml:id="sqks0o8">
       <staffDef xml:id="sum0tzs" n="1" lines="5">
        <clef xml:id="cydwnv2" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="4" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="spibhi9" n="1">
         <beam xml:id="beam-L12F1-L13F1">
          <note xml:id="note-L12F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L13F1" type="placed" dur="16" oct="5" pname="g" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L14F1-L15F1">
          <note xml:id="note-L14F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L15F1" type="placed" dur="16" oct="5" pname="a" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L16F1-L17F1">
          <note xml:id="note-L16F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L17F1" type="placed" dur="16" oct="5" pname="b" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L18F1-L19F1">
          <note xml:id="note-L18F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L19F1" type="placed" dur="16" oct="6" pname="c" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L21F1-L22F1">
          <note xml:id="note-L21F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L22F1" type="placed" dur="16" oct="6" pname="d" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L23F1-L24F1">
          <note xml:id="note-L23F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L24F1" type="placed" dur="16" oct="6" pname="e" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L25F1-L26F1">
          <note xml:id="note-L25F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L26F1" type="placed" dur="16" oct="6" pname="f" stem.dir="down" accid.ges="n" />
         </beam>
         <beam xml:id="beam-L27F1-L28F1">
          <note xml:id="note-L27F1" type="placed" dur="16" oct="4" pname="b" stem.dir="down" accid.ges="n" />
          <note xml:id="note-L28F1" type="placed" dur="16" oct="6" pname="g" stem.dir="down" accid.ges="n" />
         </beam>
        </layer>
       </staff>
       <slur xml:id="slur-L12F1-L13F1" staff="1" startid="#note-L12F1" endid="#note-L13F1" />
       <slur xml:id="slur-L14F1-L15F1" staff="1" startid="#note-L14F1" endid="#note-L15F1" />
       <slur xml:id="slur-L16F1-L17F1" staff="1" startid="#note-L16F1" endid="#note-L17F1" />
       <slur xml:id="slur-L18F1-L19F1" staff="1" startid="#note-L18F1" endid="#note-L19F1" />
       <slur xml:id="slur-L21F1-L22F1" staff="1" startid="#note-L21F1" endid="#note-L22F1" />
       <slur xml:id="slur-L23F1-L24F1" staff="1" startid="#note-L23F1" endid="#note-L24F1" />
       <slur xml:id="slur-L25F1-L26F1" staff="1" startid="#note-L25F1" endid="#note-L26F1" />
       <slur xml:id="slur-L27F1-L28F1" staff="1" startid="#note-L27F1" endid="#note-L28F1" />
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```
</details>

### DIFF
--- a/include/vrv/boundingbox.h
+++ b/include/vrv/boundingbox.h
@@ -189,6 +189,11 @@ public:
         const Point bezier[4], bool isMaxExtrema, int approximationSteps = BEZIER_APPROXIMATION);
 
     /**
+     * Calculate the euclidean distance between two points
+     */
+    static double CalcDistance(const Point &p1, const Point &p2);
+
+    /**
      * @return true if the distance between the points does not exceed margin
      */
     static bool ArePointsClose(const Point &p1, const Point &p2, int margin);
--- a/include/vrv/devicecontextbase.h
+++ b/include/vrv/devicecontextbase.h
@@ -270,6 +270,12 @@ public:
     void UpdateControlPoints();
     ///@}
 
+    /**
+     * Estimate the curve parameter corresponding to the control points
+     * Based on the polyline P1-C1-C2-P2
+     */
+    std::pair<double, double> EstimateCurveParamForControlPoints() const;
+
 private:
     // Control point X-axis offset for both start/end points
     int m_leftControlOffset = 0;
--- a/include/vrv/slur.h
+++ b/include/vrv/slur.h
@@ -218,8 +218,7 @@ private:
     void FilterSpannedElements(FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin);
 
     // Calculate the vertical shift of the slur end points
-    std::pair<int, int> CalcEndPointShift(
-        FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin, int unit);
+    std::pair<int, int> CalcEndPointShift(FloatingCurvePositioner *curve, const BezierCurve &bezierCurve, int margin);
 
     // Calculate the horizontal control point offset
     std::tuple<bool, int, int> CalcControlPointOffset(
@@ -242,9 +241,6 @@ private:
     ///@{
     // Shift end points for collisions nearby
     void ShiftEndPoints(int &shiftLeft, int &shiftRight, double ratio, int intersection, bool isBelow) const;
-
-    // Rebalance shifts to avoid awkward tilting of short slurs
-    void RebalanceShifts(int &shiftLeft, int &shiftRight, double distance, int unit) const;
 
     // Rotate the slope by a given number of degrees, but choose smaller angles if already close to the vertical axis
     // Choose doublingBound as the positive slope value where doubling has the same effect as rotating:
--- a/src/boundingbox.cpp
+++ b/src/boundingbox.cpp
@@ -763,9 +763,14 @@ Point BoundingBox::CalcPositionAfterRotation(Point point, float alpha, Point cen
     return point;
 }
 
+double BoundingBox::CalcDistance(const Point &p1, const Point &p2)
+{
+    return std::hypot(p1.x - p2.x, p1.y - p2.y);
+}
+
 bool BoundingBox::ArePointsClose(const Point &p1, const Point &p2, int margin)
 {
-    return (hypot(p1.x - p2.x, p1.y - p2.y) <= margin);
+    return (BoundingBox::CalcDistance(p1, p2) <= margin);
 }
 
 double BoundingBox::CalcSlope(const Point &p1, const Point &p2)
--- a/src/devicecontext.cpp
+++ b/src/devicecontext.cpp
@@ -98,6 +98,20 @@ void BezierCurve::UpdateControlPoints()
     c2.y = p2.y + sign * m_rightControlHeight;
 }
 
+std::pair<double, double> BezierCurve::EstimateCurveParamForControlPoints() const
+{
+    const double dist1 = BoundingBox::CalcDistance(p1, c1);
+    const double dist2 = BoundingBox::CalcDistance(c1, c2);
+    const double dist3 = BoundingBox::CalcDistance(c2, p2);
+    const double distSum = dist1 + dist2 + dist3;
+    if (distSum > 0.0) {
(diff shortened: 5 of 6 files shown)


craigsapp:
> We enforce that the slur direction at the end points is vertical in the worst case.

That is a tiny change, but it does seem to improve readability now that the start of the slurs point directly to the lower notehead.   Increasing the curvature a little bit more would help avoid the ending notehead and ledger lines.

> Note that due to recent changes in slur thickness the example doesn't look as ridiculous anymore.

Yes, it is much better so we can keep it closed for now.  Preferably if one slur is moved to the top of the stem, the other would as well.   Also the slurs are not symmetric for any particular reason.  Here is where I place them manually (with a more symmetric look):

Screen Shot 2022-04-07 at 6 46 01 AM

DavidBauer1984:
> Also the slurs are not symmetric for any particular reason.

The asymmetry stems from the fact that there is an obstacle (flag / accidental) only on one side of the slur.

The initial slur is drawn like this:
Initial
Then collision avoidance kicks in and moves the front of the slur upwards, thus breaking symmetry.

We could address this by picking a different start/endpoint based on the presence of i.e. flags. Also could try to fix symmetry during shape adjustment at the very end. Anyway, this is a different issue and should be fixed in another PR.

@craigsapp I'll make a note, but feel free to open a separate issue for this.

craigsapp:
In this sort of situation, the slur (Bezier curve) should be rotated.  What you are doing here is something different.  The initial slur control points are calculated correctly to get a symmetric slur.  Is there a way to recalculate the Bezier control points given the new endpoint based on the original placement algorithm?

craigsapp:
You are moving the Bezier control points only vertically to adjust the slur, but there will need to be some horizontal motion of the control points as well when rotation is done.  Rotation will also fix the problem with the thickness of the slur.

If I were implementing rotation for this sort of case, I would first calculate the slur along the horizontal axis with the desired curvature and distance between the endpoints.   Then choose a rotation point (left, center point, or right side of the slur) to rotate to the desired angle, and then translation from the rotation point to the final position in the music.

(In other words I would throw away all control points and recalculate new ones)

DavidBauer1984:
The problem is that any rigid motion (translation, rotation) will move the endpoints. In general this is bad since the whole point of the collision avoidance algorithm (see #2384) is to avoid collisions without moving the endpoints a lot. Also symmetrizing the slur with respect to its end points after collision avoidance only makes sense in certain cases. First of all, there is always the change of producing new collisions by symmetrisation. Second, there are cases where non-symmetric slurs are much better. Think of a example like this:
Non-symmetric
 
Anyway I have added consideration of flags during initial slur drawing as suggested in my previous post:
Example1
Example2

craigsapp:
For this slur, I would put the peak of the slur over the highest note:

Screen Shot 2022-04-08 at 7 09 57 AM

The update (c) looks good:

Screen Shot 2022-04-08 at 7 34 27 AM


 Notice that it has the same shape as the original (a), as if it were a rotation of the original, but longer.  In particular the curvature at both ends of slurs (a) and (c) are the same.  Slur (b) does not have matching curvatures at each end, and the end highlighted in red is extremely objectionable.

DavidBauer1984:
@lpugin Issues #2730 and #2770 can be closed.

PR #2783, opened 2022-04-08
Compare to correct value

Monceber:
Follow-up for changes in #2782 - in one conditions there is incorrect value (wrong enum) that is being compared to.


rettinghaus:
Fixed with 0f39834af6bee3ea97bf115ff919bb9af42e9ab8

PR #2789, opened 2022-04-13
Extrtact SMUFL alternates from fonts

Monceber:
Changed font generation process to extract SMUFL alternates (`glyphsWithAlternates` in the metadata file).

I'll open separate PR for the newly generated glyphs.

### DIFF
--- a/src/accid.cpp
+++ b/src/accid.cpp
@@ -80,7 +80,7 @@ std::wstring Accid::GetSymbolStr(const data_NOTATIONTYPE notationType) const
     }
     // If there is glyph.name (second priority)
     else if (this->HasGlyphName()) {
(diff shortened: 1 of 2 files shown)


src/accid.cpp
```diff
@@ -80,7 +80,7 @@ std::wstring Accid::GetSymbolStr(const data_NOTATIONTYPE notationType) const
     }
     // If there is glyph.name (second priority)
     else if (this->HasGlyphName()) {
-        wchar_t code = Resources::GetGlyphCode(this->GetGlyphName());
```
lpugin, line 83: Thanks for fixing this. I am surprised if was never noticed !?

lpugin:
Please rebase it. Thanks

PR #2790, opened 2022-04-13
Add sanity checks to avoid crashes in several places

Monceber:
Just a few code changes to avoid crashes that might happen in several cases.
- check that `m_staffAlignment` is initialized before using the pointer
- make sure that glyph code for tremolo is valid
- make sure that `beamSpan` has elements before doing any calculations

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -2072,7 +2072,7 @@ int Object::ScoreDefSetCurrent(FunctorParams *functorParams)
         LayerElement *element = vrv_cast<LayerElement *>(this);
         assert(element);
         Clef *clef = vrv_cast<Clef *>(element->ThisOrSameasAsLink());
-        assert(clef);
+        if (!clef->Is(CLEF)) return FUNCTOR_CONTINUE;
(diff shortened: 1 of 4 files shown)


src/object.cpp
```diff
@@ -2072,7 +2072,7 @@ int Object::ScoreDefSetCurrent(FunctorParams *functorParams)
         LayerElement *element = vrv_cast<LayerElement *>(this);
         assert(element);
         Clef *clef = vrv_cast<Clef *>(element->ThisOrSameasAsLink());
-        assert(clef);
+        if (!clef->Is(CLEF)) return FUNCTOR_CONTINUE;
```
lpugin, line 2075: Sorry for being pedantic, but looks weird. Please do
```cpp
LayerElement *elementOrLink = element->ThisOrSameasAsLink();
if (!elementOrLink || !elementOrLink->Is(CLEF)) return FUNCTOR_CONTINUE;
Clef *clef = vrv_cast<Clef *>(elementOrLink);
```

PR #2806, opened 2022-04-22
Making resources non-static

### LINKED ISSUE #2343: Make resources non-static
Following the discussing in #2253 we should move all vrv-resources into a non-static container object which is owned by the toolkit or document. This would allow passing a toolkit instance between different threads without loosing resource path and loaded fonts.

DavidBauer1984:
This PR makes resources non-static. Instead they are now a member of the document. This has the following advantages:
- Resources don't get lost when transferring a toolkit instance into a different thread.
- It is now possible to set up different toolkit instances with different resources in the same thread.

This closes #2343 .

**Technical remarks:**
(1) The default resource path remains static to keep code like

DefaultResourcePath

working. It is copied to the non-static resource path when constructing toolkit instances. In particular, changing the default resource path has no effect on toolkit instances that already have been constructed.

(2) The device contexts now store a pointer to the resources. This is set up automatically in `View::DrawCurrentPage` and fine if the resources are used in the device context only for drawing. If a device context uses resources for other purposes, then it must ensure that the resource pointer is properly set. One example is the SvgDeviceContext that uses resources in its `Commit` function. Here we explicitly set the resource pointer after construction, see `Toolkit::RenderToSVG`.

(3) This PR implies that all custom device contexts must be adapted. This should be pretty straightforward. One has to replace static resource access
```cpp
const Glyph *glyph = Resources::GetGlyph(c);
``` 
by
```cpp
const Resources *resources = this->GetResources();
assert(resources);
...
const Glyph *glyph = resources->GetGlyph(c);
```


include/vrv/doc.h
```diff
@@ -72,8 +73,18 @@ class Doc : public Object {
...
-    void SetOptions(Options *options) { (*m_options) = *options; };
+    void SetOptions(Options *options) { (*m_options) = *options; }
+    ///@}
+
+    /**
+     * Getter for the resources
+     */
+    ///@{
+    const Resources &GetResources() const { return m_resources; }
+    Resources &ModifyResources() { return m_resources; }
```
lpugin, line 86: It would be clearer if renamed to `GetResourcesForModification` or `GetModifiableResources`  because the call to the methods does not actually modify the resources.

PR #2813, opened 2022-04-25
Implement support for extended MeterSig symbols

Monceber:
Updated libMEI and added support for the MeterSig count values, that contain symbols `\`, `*` and `-`.
image

### DIFF
--- a/include/vrv/att.h
+++ b/include/vrv/att.h
@@ -207,9 +207,6 @@ public:
     std::string ProlatioToStr(data_PROLATIO data) const;
     data_PROLATIO StrToProlatio(const std::string &value, bool logWarning = true) const;
 
-    std::string SummandListToStr(data_SUMMAND_List data) const;
-    data_SUMMAND_List StrToSummandList(std::string value) const;
-
     std::string TempusToStr(data_TEMPUS data) const;
     data_TEMPUS StrToTempus(const std::string &value, bool logWarning = true) const;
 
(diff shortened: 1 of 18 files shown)


lpugin (commented):
I am not convinced to store `MeterSig::m_count` as a string. Applying a regex every time `MeterSig::GetTotalCount` is called through its call to `Meter::GetMeterCounts()` does not look very efficient and a step back. Can you think of a way to avoid this?

Monceber:
@lpugin I've changed code to cache calculated values, so regex will be used only on the first call. On all other subsequent call we'll use stored value instead.

lpugin:
Yes, that is better. However, thinking about it it seems to me that having a
 ```cpp
typedef std::pair<std::vector<int>, MeterCountSign> data_METERCOUNT_pair
```
and do the splitting / reassembly in 
```cpp
std::string Att::MetercountPairToStr(data_METERCOUNT_pair data) const;
data_METERCOUNT_pair StrToMetercountPair(std::string value) const;
```
would be more inline with what we have elsewhere.

It would also avoid possible side effects with the cache not being reset when the `m_count` value changes.

What do you think?

Monceber:
Yes, I guess that would make code cleaner and closer to what we have now.
I was avoiding doing changes to the generated LibMEI code since that is going to be changed some day.
We might just as well ignore changes to the file how we did this previously. but with schema changes at some point we'd have to address it anyway.

I guess, if you're okay with changing generated parts of the code then I'll move my changes there, that would indeed solve this problem.

PR #2821, opened 2022-04-26
Read annot text children instead of storing them

Monceber:
Right now `<annot>` element is simply being ignored and empty `<desc>` element added to SVG (see annot-001.mei)
Instead of reading content to internal storage, text children should be parsed which would fix the issue.

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -6927,6 +6927,7 @@ bool MEIInput::ReadAnnot(Object *parent, pugi::xml_node annot)
         return true;
     }
     else {
+        vrvAnnot->m_content.remove_children();
         return this->ReadTextChildren(vrvAnnot, annot, vrvAnnot);
     }
 }
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -6911,16 +6911,9 @@ bool MEIInput::ReadAnnot(Object *parent, pugi::xml_node annot)
     vrvAnnot->ReadPlist(annot);
     vrvAnnot->ReadSource(annot);
 
-    vrvAnnot->m_content.reset();
-    // copy all the nodes inside into the document
-    for (pugi::xml_node child = annot.first_child(); child; child = child.next_sibling()) {
-        vrvAnnot->m_content.append_copy(child);
-    }
-
     parent->AddChild(vrvAnnot);
     this->ReadUnsupportedAttr(annot, vrvAnnot);
-    // for Annot we do not load children because they preserved in Annot::m_content
-    return true;
+    return this->ReadTextChildren(vrvAnnot, annot, vrvAnnot);
 }
 
 bool MEIInput::ReadApp(Object *parent, pugi::xml_node app, EditorialLevel level, Object *filter)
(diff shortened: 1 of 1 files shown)


lpugin:
What is the use-case for this? Depending on what we have in `<annot>` we can have side effects if we load the content. This is why it was only kept separately. We can still put it in the SVG output, but I am not sure why / how you would do this. As it stands, you can easily go back to the MEI data through its `@xml:id` if needed.

rettinghaus:
This was working at some point (see #645) and then it broke at some point (which was unnoticed because we don't check for changes in the SVG in the CI). This is just a small fix.

lpugin:
Sounds OK. What about re-saving the document. Will the original structure be preserved with both types of `annot` content?

lpugin:
I had a quick look, and there is problem in the MEI output with text only annotations because the content is loaded twice, once as `Annot::m_content`, and once as child. Annotations with tags are properly preserved.

Input:
```xml
<layer n="1">
   <annot>Text only</annot>
</layer>
```
Output:
```xml
<layer n="1">
   <annot>Text onlyText only</annot>
</layer>
```

<details>
<summary>Show MEI</summary>

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Annotations</title>
         </titleStmt>
         <pubStmt>
            <date isodate="2020-03-30">2020-03-30</date>
         </pubStmt>
         <seriesStmt>
         </seriesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="2.0.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="3">
                        <clef shape="perc" line="2" />
                        <meterSig count="4" unit="4" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <annot>Text only</annot>
                        </layer>
                        <layer n="2">
                           <annot>Text <rend>and</rend> tags</annot>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```
</details>

src/iomei.cpp
```diff
@@ -6829,16 +6829,24 @@ bool MEIInput::ReadAnnot(Object *parent, pugi::xml_node annot)
...
-    parent->AddChild(vrvAnnot);
     this->ReadUnsupportedAttr(annot, vrvAnnot);
-    // for Annot we do not load children because they preserved in Annot::m_content
-    return true;
+    // Unless annot has only text we do not load children because they are preserved in Annot::m_content
+    if (hasNonTextContent) {
+        return true;
+    }
+    else {
+        return this->ReadTextChildren(vrvAnnot, annot, vrvAnnot);
```
lpugin, line 6848:
```suggestion
        vrvAnnot->m_content.remove_children();
        return this->ReadTextChildren(vrvAnnot, annot, vrvAnnot);
```

lpugin (approved):

PR #2826, opened 2022-04-28
Tiny comparison fix

rettinghaus:


src/att.cpp
```diff
@@ -128,7 +128,7 @@ data_BEATRPT_REND Att::StrToBeatrptRend(const std::string &value, bool logWarnin
 std::string Att::BulgeToStr(const data_BULGE &data) const
 {
     std::ostringstream ss;
-    for (int i = 0; i < data.size(); ++i) {
+    for (size_t i = 0; i < data.size(); ++i) {
```
lpugin, line 131: For consistency do `(int)data.size()` instead

src/att.cpp
```diff
@@ -149,7 +149,7 @@ data_BULGE Att::StrToBulge(const std::string &value, bool logWarning) const
     // Convert entries to numerical values
     data_BULGE bulge;
     Att converter;
-    for (int i = 0; i < entries.size() - 1; i += 2) {
+    for (size_t i = 0; i < entries.size() - 1; i += 2) {
         const double distance = converter.StrToDbl(entries[i]);
         const double offset = converter.StrToDbl(entries[i + 1]);
```
lpugin, line 154: Please also change this to `.at()`

src/chord.cpp
```diff
@@ -917,9 +917,6 @@ int Chord::ResetData(FunctorParams *functorParams)
 
 int Chord::PrepareDataInitialization(FunctorParams *functorParams)
 {
-    PrepareDataInitializationParams *params = vrv_params_cast<PrepareDataInitializationParams *>(functorParams);
-    assert(params);
-
```
lpugin, line 922: You also need to remove the `functorParams` argument name (also in the definition) because some compilers will complain about the argument not being used.

src/att.cpp
rettinghaus: For consistency I changed it to `size_t`. Or we change all other occurrences in `att.cpp`.

lpugin: Yes, change them.

src/chord.cpp
rettinghaus: As they're overrides Xcode complains about that. I'll drop this commit and leave the cleaning up to others.

lpugin: Weird. It should not and it is not for me with xcode Version 13.3.1 (13E500a). Which version are you using?

rettinghaus: Yes, the same version I got. Maybe I did something wrong? Anyway this PR was about the comparison fix in the first place.

lpugin: OK, I'll have a look

PR #2831, opened 2022-04-29
Fix for slope in mixed beams

Monceber:
Fix for the horizontal beams that are supposed to be sloped:
image

After fix:
image


lpugin:
Can the example be added to the test-suite? Thanks

rettinghaus:
added in https://github.com/rism-digital/verovio.org/pull/70

PR #2845, opened 2022-05-05
Add option grouping in help manual

Monceber:
- separated MIDI options into new group
- rearranged some of the options to adhere to alphabetic order
- added grouping to the help manual output and added arguments to the `-h` option

Default output for the help manual looks as following:
```
Example usage:

 verovio [-s scale] [-r resource-path] [-o outfile] infile

Help manual categories: 
 -h base	Base short options
 -h general	Input and page configuration options
 -h layout	General layout options
 -h selectors	Element selectors and processing
 -h margins	Element margins
 -h midi	Midi options
 -h full	Print all help manual and exit
```

Help for specific categories can be printed with corresponding category name, or `full` for all possible options (previous behavior).

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -780,7 +780,7 @@ public:
     OptionDbl m_topMarginArtic;
     OptionDbl m_topMarginHarm;
     OptionDbl m_topMarginPgFooter;
-    
+
     /**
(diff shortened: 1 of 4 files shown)


lpugin:
Great!

PR #2847, opened 2022-05-06
Minor fix for the rest placement

Monceber:
- changed code to use method for checking relevant beam shortest duration, instead of using beam-wide value

Before:
image

After:
image


lpugin:
Please double-check rest-005

image

The example above should be added to the test-suite. Thanks

PR #2856, opened 2022-05-12
Adjust beam placement in regards to staff ends

Monceber:
- relaxed constraint for beams outside of the staff - beam will be adjusted only if it explicitly goes outisde of the staff

In attempt to avoid overlapping of beams with ledger lines they ended up being with too long stems:
image

This change should alleviate the problem a little bit.
image


lpugin:
Could you check beam-031 ? This looks weird

image

PR #2864, opened 2022-05-13
Improve dotted barlines for the single staves

Monceber:
Relates to #2839 

This introduces dotted barlines for single staves. Changes for multi-staff dotted barlines are yet to come.
Default rendering based on the barline width:
image

### DIFF
--- a/include/vrv/boundingbox.h
+++ b/include/vrv/boundingbox.h
@@ -341,22 +341,22 @@ public:
     /**
      * Check if the segmented line is empty
      */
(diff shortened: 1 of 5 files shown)


craigsapp:
Related, but for dashed barlines: I would expect dashed barlines at the ends of systems to be a little further to the right so that staff lines do not extend past the barline:

Screen Shot 2022-05-13 at 7 35 20 AM

lpugin:
@craigsapp I assume this is only because of the screenshot.

PR #2868, opened 2022-05-16
Fix for the stem length in cross-staff fTrem

Monceber:
Before:
image

After:
image

### DIFF
--- a/src/beam.cpp
+++ b/src/beam.cpp
@@ -207,7 +207,7 @@ void BeamSegment::CalcSetStemValues(Staff *staff, Doc *doc, BeamDrawingInterface
             }
             // handle cross-staff fTrem cases
             const auto [beams, beamsFloat] = beamInterface->GetFloatingBeamCount();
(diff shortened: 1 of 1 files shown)


src/beam.cpp
```diff
@@ -207,7 +207,7 @@ void BeamSegment::CalcSetStemValues(Staff *staff, Doc *doc, BeamDrawingInterface
             }
             // handle cross-staff fTrem cases
             const auto [beams, beamsFloat] = beamInterface->GetFloatingBeamCount();
-            if ((coord->GetStemDir() == STEMDIRECTION_down) && ((beams > 0) || (beamsFloat > 0))) {
+            if ((coord->m_stem->GetDrawingStemDir() == STEMDIRECTION_down) && ((beams > 0) || (beamsFloat > 0))) {
```
lpugin, line 210: It is not impossible to have rests within a mixed beam. You need to check that `m_stem` is set before accessing it.

lpugin: Sorry, I am noticing we are filtering by chords and notes above. But still, it would be best to check it

PR #2890, opened 2022-05-25
Font metadata update

rettinghaus:
This is a tiny update to Leipzig, where the `metadata.json` was outdated.


rettinghaus:
removed commit

PR #2900, opened 2022-05-30
Append transpose revision to header

Monceber:
Changed code to write `revisionDesc` to header on export when transposition is done. It contains timestamp and message about transposition action. For Interval/Semitone transposition message will state how many semitones up/own score has been transposed, for KeyTonic - message about transposing to closest/higher/lower key.
E.g.
`Transposed up 7 semitones by Verovio.`
or
`Transposed to closest G-flat by Verovio.`

### DIFF
--- a/include/vrv/iomei.h
+++ b/include/vrv/iomei.h
@@ -291,6 +291,11 @@ private:
     bool WriteDoc(Doc *doc);
 
     /**
+     * Write revisionDesc to the header
+     */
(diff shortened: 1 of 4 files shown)


lpugin:
Looks good for now. Eventually, we should probably have a `Header` class to regroup this kind of task not to overload MEIOutput. It could also be useful to add some simple helpers to retrieve some data from the header (e.g., a list of revisions) and regroup stuff such as `PgHead::GenerateFromMEIHeader`

PR #2907, opened 2022-06-02
Rename IsInBeam method to be more specific

Monceber:
- renamed IsInBeam and IsInFTrem to better convey their purpose - returning Beam object
- added method IsInBeam for purposes where Beam object is not required and only bool check is needed
- replaced previous usages of IsInBeam with GetAncestorBeam (where Beam is necessary) and removed check of IsInBeamSpan (since it's included within other method)

### DIFF
--- a/include/vrv/layerelement.h
+++ b/include/vrv/layerelement.h
@@ -121,16 +121,17 @@ public:
     /** Return true if the element is a note within a ligature */
     bool IsInLigature() const;
     /** Return the FTrem parten if the element is a note or a chord within a fTrem */
-    FTrem *IsInFTrem();
-    const FTrem *IsInFTrem() const;
+    FTrem *GetAncestorFTrem();
+    const FTrem *GetAncestorFTrem() const;
     /**
      * Return the beam parent if in beam
      * Look if the note or rest is in a beam.
      * Look for the first beam parent and check if the note is in its content list.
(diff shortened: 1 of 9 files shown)


include/vrv/layerelement.h
```diff
@@ -576,6 +581,9 @@ class LayerElement : public Object,
      * This also stores the negative values for identifying cross-staff
      */
     int m_alignmentLayerN;
+
+    // flag to indicate that layerElement belongs to the beamSpan
+    bool m_isInBeamspan;
```
lpugin, line 586: This needs to be initialized (I don't think it is) and reset in `LayerElement::ResetDrawing`

Monceber: This change is basically moving from **public** to **private** section, it's already existing value and is being initialized.
https://github.com/rism-digital/verovio/blob/e59394fffd1b48f3c68e72c1a1396be91fcd3930/src/layerelement.cpp#L134

Also I don't think `LayerElement::ResetDrawing` even exists. If you meant `LayerElement::ResetHorizontalAlignment`, then I don't think it's the best idea to reset it there - notes should not change whether they are in the beamspan or not based on alignment.

lpugin: Sorry, I meant `LayerElement::ResetData` (which used to be called `ResetDrawing`). And I can see it is actually reset here, so we are all good.

PR #2920, opened 2022-06-08
Clean-up handling keySig attributes

lpugin:
This PR removes KeySig::ClearKeyAccidAttribChildren because this method was problematic. `GetList` should not be used to modify the children of an object because it can (in theory) contain descendants that are not direct children of the object. Even though it was not the case with the KeySig attribute, I thought it would be better not to do it. With this PR, the clearing of the attribute objects is simply done every time they are updated.

### DIFF
--- a/include/vrv/keysig.h
+++ b/include/vrv/keysig.h
@@ -84,7 +84,6 @@ public:
      */
     ///@{
     bool HasNonAttribKeyAccidChildren() const;
-    void ClearKeyAccidAttribChildren();
     void GenerateKeyAccidAttribChildren();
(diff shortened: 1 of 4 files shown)


rettinghaus:
@lpugin Any idea what's happening here? https://github.com/rism-digital/verovio/runs/6791070333?check_suite_focus=true#step:9:647

lpugin:
The problem seems to have been introduced with https://github.com/rism-digital/verovio/pull/2864. It tries to draw a barline with no `@form` - at least this is what is happening with the score-011.mei test.

PR #2923, opened 2022-06-09
Proposal: move CONTRIBUTING to root directory

rettinghaus:
Currently the doc folder contains a hodgepodge of things and could probably be cleaned up. Especially since most of the actual documents there are now in the available in the Reference book.


lpugin:
This file should actually be remove

lpugin:
Then we should have a minimal with a link to https://book.verovio.org/contributing/guidelines.html

lpugin:
Closing for now

PR #2932, opened 2022-06-15
Skip layout for MIDI and timemap output

DavidBauer1984:
This PR introduces skipping the layout for midi and timemap output in the Verovio command line tool. This speeds up the total runtime of MIDI/timemap generation by a factor between 1,4 and 1,9 (depending on the size of the piece and the complexity of the layout).

### DIFF
--- a/include/vrv/toolkit.h
+++ b/include/vrv/toolkit.h
@@ -634,6 +634,13 @@ public:
     ///@{
 
     /**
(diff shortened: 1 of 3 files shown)


wergo:
Great PR! 

@DavidBauer1984, please excuse my question (if I have overseen something): how do you get the timemap from the command line? Thanks!

Issue #274, opened 2016-08-16
Creating xml:id values


craigsapp:
In the Humdrum Importer, every time I convert the score from Humdrum to MEI, the `xml:id`s are different.  It would be nice if the `xml:id`s could be produced the same way when the same data file (without any changes) is reconverted.  

(1) Is there a way, such as with a random seed, to produce the same `xml:id`s every time?  This would perhaps allow me to use the checksum for the file as a seed to randomly generated id numbers.

and/or

(2) Can I supply my own `xml:id` values?

An example verovio `xml:id`:  `note-000000189093144`.  If I can create my own xml ids, should the "note-" prefix (and similar for other classes) be maintained?  In other words, is this prefix for human readability, or does it have an additional function (such as MIDI playback)?

The C++ function names used to deal with xml ids imply that these are UUIDs (`GetUuid()`, `SetUuid()`), but technically they are not UUIDs  (https://en.wikipedia.org/wiki/Universally_unique_identifier).

I tracked down the id generator, which I see is a random-number generation using `rand()` from the C standard library:

https://github.com/rism-ch/verovio/blob/5637d0bd87dd0adde5b9210fdb48819234a381e3/src/object.cpp#L481-L489

But now I am wondering why they are different every time, since I cannot find a call to `srand()` which would initialize different sequences...

The ability to seed the random number generation would be useful, but also allowing user-supplied xml ids (or the part after the class name at least) would also be useful.

There are also C++ 11 ways to generate random numbers: http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution
The C++ ways might be able to fill in the leading zero in the current ids.

lpugin:
> (2) Can I supply my own xml:id values?

It think this is already be possible. You only need to call SetUuid in your importer.

> If I can create my own xml ids, should the "note-" prefix (and similar for other classes) be maintained?

No, the prefix has no meaning and is there only for readability. (I am not sure it actually helps.)

> The C++ function names used to deal with xml ids imply that these are UUIDs (GetUuid(), SetUuid()), but technically they are not UUIDs 

I am aware of this. They were actually UUIDs at some point, but I changed them to something shorter because it was making reading the MEI very difficult.

> But now I am wondering why they are different every time, since I cannot find a call to srand() which would initialize different sequences...

https://github.com/rism-ch/verovio/blob/master/tools/main.cpp#L150-L151
and
[...]

PR #2933, opened 2022-06-17
Use C++11 random library for ID generation

DavidBauer1984:
This PR replaces the C-style `rand()` used for ID generation by a generator from the C++11 random library. The main reasons for this change are thread safety (see [this post](https://stackoverflow.com/questions/6161322/using-stdlibs-rand-from-multiple-threads)) and higher quality random numbers (i.e. less likelihood of generating the same ID twice).

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -9,10 +9,10 @@
 #define __VRV_OBJECT_H__
 
 #include <cstdlib>
-#include <ctime>
 #include <functional>
 #include <iterator>
 #include <map>
+#include <random>
 #include <string>
 
 //----------------------------------------------------------------------------
@@ -1639,6 +1639,11 @@ private:
      * A static counter for uuid generation.
      */
     static thread_local unsigned long s_objectCounter;
+
+    /**
+     * Pseudo random number engine for ID generation
+     */
+    static thread_local std::mt19937 s_randomGenerator;
 };
 
 //----------------------------------------------------------------------------
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -94,7 +94,7 @@ std::string GetVersion();
  * Encode the integer value using the specified base (max is 62)
  * Base 36 uses 0-9 and a-z, base 62 also A-Z.
  */
-std::string BaseEncodeInt(int value, int base);
+std::string BaseEncodeInt(unsigned int value, unsigned int base);
 
 /**
  *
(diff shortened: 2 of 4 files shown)


craigsapp:
I am concerned about speed.  What are timings before and after using the C++ random library?

https://thompsonsed.co.uk/random-number-generators-for-c-performance-tested

https://medium.com/@odarbelaeze/how-competitive-are-c-standard-random-number-generators-f3de98d973f0

https://www.reddit.com/r/cpp_questions/comments/iueod0/is_stdrandom_faster_than_c_rand/

DavidBauer1984:
@craigsapp There should not be any perceivable difference in total runtime. Generating 100 million random numbers with Mersenne Twister takes 2.1 s on my machine. Generating a Beethoven piece (10 pages) takes 2.8 s. During that about 35000 objects are constructed. If we generate a random number for each object, then we are talking about `35000 / 10^8 * 2.8 s ~ 1 ms` spent for random number generation.

craigsapp:
> There should not be any perceivable difference in total runtime

`Should not be` and `is not` are not the same thing...  It would be useful to verify the expected behavior by comparing before and after the switch to the random library for rendering the same score.

Previous discussion in 2016 https://github.com/rism-digital/verovio/issues/274#issuecomment-240229054

DavidBauer1984:
Well I did compare it for the Beethoven and there was no perceivable difference in total runtime. Here is a test with another piece (2.8 MB, 23 pages) run with a time profiler. The total runtime is 3.574 s and it spends 2 ms for random number generation:
Performance
I agree with you that Mersenne is slower than `rand`, but _the time difference is neglectible w.r.t. total runtime of Verovio_. We even spend more time for calculating the base encoding (see line above).

lpugin:
@craigsapp, it seems to me that, if there is a difference, it will not be very significant. The change looks good to me.

craigsapp:
I am not complaining, but I want to make sure you keep your eye on efficiency (The Javascript language is not a multithreaded environment, unless WebWorkers are allowed to be multithreaded).   Also I just thought: are you testing in the javascript toolkit or in the native OS version?  I wonder if there are differences (but no need to bother doing that).

A thought related to this:  when an object is created, I am presuming that an ID is assigned to it?  An interesting idea would be to not assign an ID when an Object is created, and instead `Object::getID()` would automatically generate an ID if the ID is empty.   This would make the option to remove unused IDs probably faster since any object with no ID does not need to have it removed (after a search to see if it is used).  Also, I mostly do not use the random IDs, so calculating them is  90% wasteful for me.

PR #2937, opened 2022-06-20
WIP: JavaScript and npm package improvements

WolfgangDrescher:
See https://github.com/rism-digital/verovio/discussions/2815 for the discussion.

Changes:

* Added flag `-m` to `buildToolkit` to enable a build with `MODULARIZE=1`
* Add `$exportName` variable to pass an `EXPORT_NAME` to emscripten
* Removed `$NPM` variable in `buildToolkit`
* Added `buildNpmPackage` script to create the npm package
* Removed custom build option for `$wasmQ` in favour of `buildNpmPackage`
* Build and bundle JS files with rollup: `npm --prefix npm run prebundle -- -o "\$(pwd)/$BUILD_DIR/$FILENAME"`
* Added ESM export for the VerovioToolkit
* Added Verovio module with and without Humdrum as `MODULARIZE=1` emscripten module to the npm package
* Bump node version to v14.17.0 for full ESM support in `package.json`
* Added `exports`, `files` and `scripts` fields to `package.json`

To be discussed:

* Using ES6 JavaScript syntax will remove support for really old browsers (check relevant browser support here: [class](https://caniuse.com/?search=class), [proxy](https://caniuse.com/?search=proxy), [spread operator](https://caniuse.com/?search=spread%20operator))
* Naming of the new `-m` flag (Modularize emscripten build) is not good since there is already a `-M` (Clean existing makefile) and the similar naming can be confusing. Any idea for a better naming for this letter?
* Setting the default EXPORT_NAME to `VerovioModule` does not have any effect without `MODULARIZE=1`. Should I just set this for the new modularized builds, and leave it `Module` for the other builds?
* The new modularized build is exported with the name `createVerovioModule` as a factory. Naming is good I think and follows the emscripten recommendation in the docs.
* Since the humdrum module will be bundled in the npm package, should I remove the `npm-hum` directory?
* Since I don't know the build script for new releases I hope that I did not break anything in `buildToolkit`. This should still be tested.
* Filenames in `dist` are actually not so important. People can only import files that are specified in the `exports` field in `package.json`. The naming there should be chosen carefully. `verovio/next`, `verovio/wasm` and `verovio/wasm-hum` are not perfect yet.
* I could get rid of createEmscriptenProxy in `emscripten-proxy.js` and use `getToolkitFunction` directly, but it looks cleaner in the VerovioToolkit: `this.proxy.renderToSVG(this.ptr, pageNo, xmlDeclaration)` vs. `getToolkitFunction(this.VerovioModule, 'renderToSVG')(this.ptr, pageNo, xmlDeclaration)`
* I think we can remove `VerovioToolkit.instances` and the destructor in the `window.onunload` event. Also checkout this in my initial post of the discussion #2815
* Synchronous builds `verovio-toolkit.js`, `verovio-toolkit-light.js`, `verovio-toolkit-hum.js` should work without any breaking change. `verovio-toolkit-wasm.js`, as discussed, does currently need `verovio.module.onRuntimeInitialized` instead of `Module.onRuntimeInitialized`. Should I try to remove this breaking change?
* `buildToolkit` does currently create `build/verovio-toolkit-wasm-hum-modularize.js`, `build/verovio-toolkit-wasm-modularize.js`, but this is probably not needed in the current build and zip? Should we add a condition that prevents creating them when `$modularizeQ` is truthy?
* We should add `npm install` and eventually `rm -rf ./npm/node_modules` in `buildToolkit` before we execute `npm run ...`. Is there anything against it?
* npm must be available in the build step. Probably we just need to add `actions/setup-node` as a step into the ci_build.yml GitHub action and other workflows.
* Should we also add `verovio-toolkit-wasm-hum.js` to the release build artifacts and/or the development build step in ci_builds.yml?

Todo:

- [ ] Documentation in the Verovio Book


lpugin:
I am trying to build the wasm toolkit. I installed `rollup` with Homebrew. When I do
```bash
./buildToolkit -c -H -w
```
I get
```
[!] Error: Cannot find module '@rollup/plugin-commonjs'
Require stack:
- /Users/laurent/projects/verovio_develop/emscripten/npm/rollup.config.prebundle.js
- /opt/homebrew/Cellar/rollup/2.75.7/libexec/lib/node_modules/rollup/dist/shared/loadConfigFile.js
- /opt/homebrew/Cellar/rollup/2.75.7/libexec/lib/node_modules/rollup/dist/bin/rollup
Error: Cannot find module '@rollup/plugin-commonjs'
Require stack:
- /Users/laurent/projects/verovio_develop/emscripten/npm/rollup.config.prebundle.js
- /opt/homebrew/Cellar/rollup/2.75.7/libexec/lib/node_modules/rollup/dist/shared/loadConfigFile.js
- /opt/homebrew/Cellar/rollup/2.75.7/libexec/lib/node_modules/rollup/dist/bin/rollup
    at Module._resolveFilename (node:internal/modules/cjs/loader:946:15)
    at Module._load (node:internal/modules/cjs/loader:787:27)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/laurent/projects/verovio_develop/emscripten/npm/rollup.config.prebundle.js:5:16)
    at Module._compile (node:internal/modules/cjs/loader:1112:14)
    at require.extensions.<computed> [as .js] (/opt/homebrew/Cellar/rollup/2.75.7/libexec/lib/node_modules/rollup/dist/shared/loadConfigFile.js:612:20)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Module._load (node:internal/modules/cjs/loader:834:12)
    at Module.require (node:internal/modules/cjs/loader:1012:19)

gzip: can't stat: verovio-toolkit-wasm.js (verovio-toolkit-wasm.js): No such file or directory
```
Any ideas?

lpugin:
Ignore my previous comment, I got it working.

To your questions:

> * Using ES6 JavaScript syntax will remove support for really old browsers (check relevant browser support here: [class](https://caniuse.com/?search=class), [proxy](https://caniuse.com/?search=proxy), [spread operator](https://caniuse.com/?search=spread%20operator))
> 

This seems OK to me.

> * Naming of the new `-m` flag (Modularize emscripten build) is not good since there is already a `-M` (Clean existing makefile) and the similar naming can be confusing. Any idea for a better naming for this letter?
> 

It is OK for me to use `m`. We already have `c` and `C`, and as long as it is documented it is fine.

> * Setting the default EXPORT_NAME to `VerovioModule` does not have any effect without `MODULARIZE=1`. Should I just set this for the new modularized builds, and leave it `Module` for the other builds?

Yes, set it only for the modularized build.

> 
> * The new modularized build is exported with the name `createVerovioModule` as a factory. Naming is good I think and follows the emscripten recommendation in the docs.

OK

> 
> * Since the humdrum module will be bundled in the npm package, should I remove the `npm-hum` directory?

Yes, remove it

> 
> * Since I don't know the build script for new releases I hope that I did not break anything in `buildToolkit`. This should still be tested.
> 

I'll check that, but as fas as I can see it should be fine.

> * Filenames in `dist` are actually not so important. People can only import files that are specified in the `exports` field in `package.json`. The naming there should be chosen carefully. `verovio/next`, `verovio/wasm` and `verovio/wasm-hum` are not perfect yet.
> 
> * I could get rid of createEmscriptenProxy in `emscripten-proxy.js` and use `getToolkitFunction` directly, but it looks cleaner in the VerovioToolkit: `this.proxy.renderToSVG(this.ptr, pageNo, xmlDeclaration)` vs. `getToolkitFunction(this.VerovioModule, 'renderToSVG')(this.ptr, pageNo, xmlDeclaration)`
> 

Yes, this is good as it is.

> * I think we can remove `VerovioToolkit.instances` and the destructor in the `window.onunload` event. Also checkout this in my initial post of the discussion

Not sure about that. I will have a look.

> Synchronous builds verovio-toolkit.js, verovio-toolkit-light.js, verovio-toolkit-hum.js should work without any breaking change. verovio-toolkit-wasm.js, as discussed, does currently need verovio.module.onRuntimeInitialized instead of Module.onRuntimeInitialized. Should I try to remove this breaking change?

If you think you can maybe we can keep both, but only to have a deprecation warning and give users a bit more time to change it. Eventually we should keep only `verovio.module` because `Module` can yield conflicts anyway.

> buildToolkit does currently create build/verovio-toolkit-wasm-hum-modularize.js, build/verovio-toolkit-wasm-modularize.js, but this is probably not needed in the current build and zip? Should we add a condition that prevents creating them when $modularizeQ is truthy?

Yes, that would be better.

> We should add npm install and eventually rm -rf ./npm/node_modules in buildToolkit before we execute npm run .... Is there anything against it?
> npm must be available in the build step. Probably we just need to add actions/setup-node as a step into the ci_build.yml GitHub action and other workflows.

Maybe we can see how this work with GH action. Emscripten includes an npm installation, which is the one I am using.

> Should we also add verovio-toolkit-wasm-hum.js to the release build artifacts and/or the development build step in ci_builds.yml?

I would certainly not add it to the development build steps. I would also prefer not to have an additional one for the releases, so unless @craigsapp thinks we should have it, I would not.

lpugin:
Could you redo the  PR to `develop-npm` branch? That way we can test it further before merging it to develop. Thanks!

WolfgangDrescher:
The best way to install `rollup` is with npm. You can use `npm install -g rollup` to install in globally and without homwbrew. I had to use `@rollup/plugin-commonjs` as another dependency for the build process so rollup can interpret the emscripten CJS module. But I already included everything you need as `devDependencies`. Can you try this:

```sh
cd emscripten/npm
npm install
cd ..
./buildToolkit -c -H -w
```

To prevent using the cd commands I am using `npm --prefix npm install` (executes `npm install` inside the ./npm directory) from inside the emscripten directory in `./buildToolkit`. But I added and commited an additional npm install inside ./buildToolkit so this should not happen again and you should just need to call `./buildToolkit -c -H -w` now.

> If you think you can maybe we can keep both, but only to have a deprecation warning and give users a bit more time to change it. Eventually we should keep only verovio.module because Module can yield conflicts anyway.

Okay, I will try to find a ways for this. But I'm not sure if this is possible with rollup. Maybe with another rollup custom plugin, or by hacking `window.Module = Module` somewhere.

> > Setting the default EXPORT_NAME to VerovioModule does not have any effect without MODULARIZE=1. Should I just set this for the new modularized builds, and leave it Module for the other builds?

> Yes, set it only for the modularized build.

Ok, I removed this for `MODULARIZE=0` builds.

> Could you redo the PR to develop-npm branch? That way we can test it further before merging it to develop. Thanks!

I changed the base of this PR to `develop-npm`.

I hope my last changes did not break anything. I'm having problems with emscripten currently due to `homebrew updated` I executed yesterday, so I cannot test it myself right now, sorry.

emscripten/buildToolkit
```diff
@@ -220,9 +217,9 @@ $exports .= "'_malloc',";
 $exports .= "'_free'";
 $exports .= "]\"";
 
-my $extra_exports = "-s EXPORTED_RUNTIME_METHODS='[\"cwrap\"]' -s EXPORT_NAME=\"'$exportName'\"";
+my $extra_exports = "-s EXPORTED_RUNTIME_METHODS='[\"cwrap\"]'";
 
-my $modularize = "-s MODULARIZE=".($modularizeQ ? "1" : "0");
+my $modularize = "-s MODULARIZE=".($modularizeQ ? "1" : "0")." -s EXPORT_NAME=\"'createVerovioModule'\"";
```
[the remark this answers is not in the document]
WolfgangDrescher: Yes, I just realized this. Fixed in c6be81b

emscripten/buildToolkit
```diff
@@ -246,6 +246,7 @@ system($ccCmd);
 if ($? == 0) {
 	print " Done.\n";
 
+    `npm --prefix npm install`;
```
lpugin: Isn't this adding a `npm/node_modules` directory? and a `npm/package-lock.json` file? This is what I can now see after having run `buildNpmPackage`. I am not sure what will happen with these when I do an `npm publish` from there. Do you know if they will be published as well? I don't think we want that, do we?

lpugin: Good. Then they should be added to `.gitignore`

WolfgangDrescher: They should be ignored in `emscripten/npm/.gitignore`. Also `emscripten/npm/dist` should not be commited and is ignored in the same file.

WolfgangDrescher:
Really nice! Last things left before we can merge into `develop`:

* Naming for the `exports` of the npm package. I'm still searching for a better naming of `verovio/next`, `verovio/wasm` and `verovio/wasm-hum`.
* Implement `window.Module = Module` if you don't want the breaking change for the new `verovio.module` export.
* Figure out if `window.onunload` is really not needed as I would expect. I realized that I actually broke this currently with my changes (`this.proxy` is not present there). Where can I find what `vrvToolkit_destructor` is actually running in C++? But JS probably won't need it.
* Documentation. Is the Verovio Boook open sourced and open for contributions?
* Testing. I found a typo today 6822a1f. I did not test all of the toolkit methods and usually only need a few of them (`loadData`, `renderToSVG`, etc.). Maybe it could be worth trying to automate this as much as possible with a unit test framework (e.g. [vitest](https://vitest.dev/)).

lpugin:
> * Naming for the `exports` of the npm package. I'm still searching for a better naming of `verovio/next`, `verovio/wasm` and `verovio/wasm-hum`.

Yes, `verovio/next` is not great. What about `verovio/esm`? Would that make sense? `verovio/wasm` sounds ok, but maybe we can be more specify. `verovio/wasm-module`? or just `verovio/module`? 

I think we need to have names describing what things are or do rather than if they are old or new. Is it a module, a proxy, a toolkit factory, a wrapper? Just throwing  name ideas, here...

> 
> * Implement `window.Module = Module` if you don't want the breaking change for the new `verovio.module` export.
> 

Yes, your proposal sounded OK. Even if we cannot raise a warning, that would leave a bit of time for projects to change it.

> * Figure out if `window.onunload` is really not needed as I would expect. I realized that I actually broke this currently with my changes (`this.proxy` is not present there). Where can I find what `vrvToolkit_destructor` is actually running in C++? But JS probably won't need it.

As a C++ programmer I prefer to be always on the safe side regarding memory freeing. So unless 1) having it is problematic, or 2) showing that it is not strictly necessary, it can be kept because calling the destructor explicitly cannot be bad.

> 
> * Documentation. Is the Verovio Boook open sourced and open for contributions?

Yes, absolutely. At a first glance, it could go [there](https://book.verovio.org/installing-or-building-from-sources/javascript-and-webassembly.html). There is an `Edit this page` button at the top that takes you to the right place. Of course, can also simply fork and edit locally. For changes related to stuff that has not been released yet, the PR should be made to the develop branch. It will be merged when the release is made. 

> 
> * Testing. I found a typo today [6822a1f](https://github.com/rism-digital/verovio/commit/6822a1fcfa9605eeef0c152dd6740b689e58f3a4). I did not test all of the toolkit methods and usually only need a few of them (`loadData`, `renderToSVG`, etc.). Maybe it could be worth trying to automate this as much as possible with a unit test framework (e.g. [vitest](https://vitest.dev/)).

The more tests the better. However, this is a fairly limited list of functions and a good proof reading should do it, hopefully. If you have proposals for test setups, I would suggest to have a separate discussion.

PR #2940, opened 2022-06-27
Improve glyph extraction

rettinghaus:
This is an update to the font extraction process to take into account the default value for `horiz-adv-x`. Also some additional type hints were added. 
Font files have been regenerated with added previously missing `h-a-x` values.


fonts/extract-bounding-boxes.py
```diff
@@ -51,16 +51,18 @@ def read_svg_font_file(font_file_name: str):
...
     root = ET.fromstring(bytes(font_svg_content, encoding="utf-8"))
-    font_faces = get_svg_elements(root, "font-face")
+    font = root.find("svg:defs/svg:font", ns)
+    default_hax = font.get("horiz-adv-x", "0")
+    font_faces = get_svg_elements(font, "font-face")
     if len(font_faces) != 1:
         print(
             f"Error: the file {font_file_name} should have a unique font-face element!")
-        print(f"Please check that the svg has correct namespace: {SVG_NS}")
+        print(f"Please check that the svg has correct namespace: {ns}")
```
ahankinson, line 60: This will now print the full dictionary, rather than just the NS value. Is that intended?

rettinghaus: No. Thanks for the hint.

ahankinson (approved):

PR #2946, opened 2022-06-30
More JavaScript and npm package improvements 

WolfgangDrescher:
Changes since #2937:

* Added export of Module on global `window.Module` variable for backwards compatibility
* Fixed `VerovioToolkit.destroy()` and `window.unload` event
* Improved error message if no module is passed to the VerovioToolkit constructor
* Changed export name from `verovio/next` to `verovio/esm`
* Added documentation and ESM examples in `README.md`


lpugin:
Thanks. I would prefer to have the documentation only in the Reference book. I can still keep the README, though, but it should instead have a link to the chapter.

PR #2947, opened 2022-07-01
Vertical placement of mixed beams

DavidBauer1984:
This PR addresses the vertical positioning of mixed beams. We

1. Request staff space in order to enforce minimal stem lengths for mixed cross staff beams.
2. Improve the vertical beam placement for mixed beams by maximising the minimal stem length on both sides of the beam.

The change improves several examples in the test library, such as:

**beams-050**
beams-050

**cross-staff-007**
cross-staff-007

**cross-staff-009**
cross-staff-009

**cross-staff-014**
cross-staff-014

### DIFF
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -125,7 +125,7 @@ private:
     int CalcBeamSlopeStep(
         const Doc *doc, const Staff *staff, BeamDrawingInterface *beamInterface, int noteStep, bool &shortStep);
 
-    void CalcMixedBeamStem(const BeamDrawingInterface *beamInterface, int step);
+    void CalcMixedBeamPosition(const BeamDrawingInterface *beamInterface, int step);
 
     void CalcBeamPosition(const Doc *doc, const Staff *staff, BeamDrawingInterface *beamInterface, bool isHorizontal);
 
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -172,7 +172,7 @@ private:
 
     // Helper to check whether beam fits within certain bounds
     bool DoesBeamOverlap(
-        int staffTop, int topOffset, int staffBottom, int bottomOffset, bool isCrossStaff = false) const;
+        int staffTop, int topOffset, int staffBottom, int bottomOffset, int unit, bool isCrossStaff = false) const;
 
     // Helper to calculate the minimal stem length of above/below notes
     std::pair<int, int> GetMinimalStemLength(const BeamDrawingInterface *beamInterface, bool hasFrenchStyle) const;
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -171,14 +171,13 @@ private:
     void CalcSetValues();
 
     // Helper to check whether beam fits within certain bounds
-    bool DoesBeamOverlap(
-        int staffTop, int topOffset, int staffBottom, int bottomOffset, int unit, bool isCrossStaff = false) const;
+    bool DoesBeamOverlap(const BeamDrawingInterface *beamInterface, int topBorder, int bottomBorder, int unit) const;
 
     // Helper to calculate the vertical offset of the beam top/bottom w.r.t. the beam center
     std::pair<int, int> GetVerticalOffset(const BeamDrawingInterface *beamInterface) const;
 
     // Helper to calculate the minimal stem length of above/below notes
-    std::pair<int, int> GetMinimalStemLength(const BeamDrawingInterface *beamInterface, bool hasFrenchStyle) const;
+    std::pair<int, int> GetMinimalStemLength(const BeamDrawingInterface *beamInterface) const;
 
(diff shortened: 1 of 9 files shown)


lpugin:
Could you check beam-049?

image

The encoding is explicit about the stem directions, and the between placement should stay.

DavidBauer1984:
The question is how hard we want to enforce encoded stem directions when there is no space to properly draw a (single staff) mixed beam. We have two rules:

1. The mixed beam must be drawn inside the staff to prevent collisions with ledger lines.
2. The mixed beam must have minimal stem length at least **M** units. For beam chords we measure the distance from the beam to the closest note.

These rule existed before and are implemented in `BeamSegment::DoesBeamOverlap`. If the beam cannot be placed satisfying these rules, then the beam is flipped over and drawn above/below instead. Verovio displays a warning in this case. Regarding rule 2 the question arises: which value of **M** should we pick to make everybody happy?

A nice example to play around with this is the following:
<details>
<summary> Show MEI </summary>

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Beams with opposite stem directions</title>
         </titleStmt>
         <pubStmt>
            <respStmt>
               <persName role="encoder">Klaus Rettinghaus</persName>
               <corpName role="funder">Enote GmbH</corpName>
            </respStmt>
            <date isodate="2020">2020</date>
         </pubStmt>
         <seriesStmt>
         </seriesStmt>
         <notesStmt>
            <annot>No stem should be shorter than 5vu and the beams have to avoid top and bottom stave-lines.</annot>
         </notesStmt>
         <sourceDesc>
            <source>
               <bibl>Behind Bars, p. 26, example 1</bibl>
            </source>
         </sourceDesc>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="3.0.0" label="0">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" />
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure>
                     <staff n="1">
                        <layer n="1">
                           <beam>
                             <note dur="8" oct="5" pname="f" stem.dir="down" />
                             <note dur="8" oct="4" pname="a" stem.dir="up" />
                             <note dur="8" oct="5" pname="f" stem.dir="down" />
                           </beam>
                           <beam>
                             <note dur="8" oct="5" pname="f" stem.dir="down" />
                             <note dur="8" oct="4" pname="g" stem.dir="up" />
                             <note dur="8" oct="5" pname="f" stem.dir="down" />
                           </beam>
                           <beam>
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                              <note dur="8" oct="4" pname="f" stem.dir="up" />
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                           </beam>
                           <beam>
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                              <note dur="8" oct="4" pname="e" stem.dir="up" />
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                           </beam>
                           <beam>
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                              <note dur="8" oct="4" pname="d" stem.dir="up" />
                              <note dur="8" oct="5" pname="f" stem.dir="down" />
                           </beam>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```
</details>

**Before** this PR the rendering was:
15-Before
**M** was chosen quite small (I think M=1), but the beams were not placed vertically in an optimal way. So still the first three mixed beams flip over.

**After** this PR we have better vertical beam placement. So the rendering becomes:
- with M=4:
15-M4

- with M=3.5:
15-M3-5

- with M=3:
15-M3

The smaller we pick the minimal stem length M, less beams flip over, but the beams also look increasingly degenerated.

To make example Beam-049 look as before, we have to pick M=3 or less:
Beam-049

lpugin:
Yes, I agree that these are tricky cases. In 18th century engraving we can find many mixed beams that would not look appropriate according to later engraving practices. Typically mixed beam with a gap smaller than an octave and where the beam has to be slanted in order to fit.

At some point, it will fall down to the responsibility of the encoder to make sure stem directions represent something acceptable - or even possible, because encoding can actually be nonsense.

Maybe we can also have an option to indicate if we want single-staff mixed beams to be flipped at all? `--beam-mixed-preserve`, which would be `false` by default and meaning that Verovio can flip them. With the option enable, Verovio would produce whatever is encoded.

As for the flipping point (`--beam-mixed-stem-min` ?) I would suggest 3.5 as a default.

DavidBauer1984:
Options were added. The previous output
beam-049
for beam-049 can now be obtained by either setting `--beam-mixed-preserve` or `--beam-mixed-stem-min 3.0` (or smaller).

PR #2950, opened 2022-07-05
Add support for MEI basic and new option for moving score definition encoding

lpugin:
The PR add a `mei-basic` output format that performs some conversion to MEI basic when saving to MEI. The internal data remains untouched because the conversion is performed only in the writing process.

The conversion includes:
* Removing of the editorial markup. When there is some parallel segmentation in the data, only one (the visible) subtree is written to the ouput
* Moving `@accid` and `@artic` to `./accid` and `./artic`
* Moving `./clef`, `./keySig`, `./meterSig` within `scoreDef` and `staffDef` as corresponding attributes

At this stage, no check or removing of unsupported elements or attributes is performed

### Option to move score definition to staff

The PR also introduces a new `--move-score-definition-to-staff` option that moves all the clef, keySig, meterSigGrp, meterSig and mensur elements to the staffDef level when loading a file. This option can be used in conjunction with MEI basic output to have MEI files that follow the same encoding structure.

### DIFF
--- a/src/keysig.cpp
+++ b/src/keysig.cpp
@@ -238,6 +238,43 @@ int KeySig::GetFifthsInt() const
     return 0;
 }
 
+data_KEYSIGNATURE KeySig::ConvertToSig() const
+{
+    data_KEYSIGNATURE sig = std::make_pair(-1, ACCIDENTAL_WRITTEN_NONE);
+    const ListOfConstObjects &childList = this->GetList(this);
+    if (childList.size() > 1) {
+        const KeyAccid *firstKeyAccid = vrv_cast<const KeyAccid *>(childList.front());
+        assert(firstKeyAccid);
+        const data_ACCIDENTAL_WRITTEN accidType = firstKeyAccid->GetAccid();
+        if (accidType != ACCIDENTAL_WRITTEN_f && accidType != ACCIDENTAL_WRITTEN_s) {
+            LogWarning(
+                "KeySig content cannot be converted to @sig because this accidental type is not a flat or a sharp");
+            return sig;
+        }
+        bool isCommon = true;
+        int pos = 0;
+        for (auto &child : childList) {
+            const KeyAccid *keyAccid = vrv_cast<const KeyAccid *>(child);
+            assert(keyAccid);
+            if (accidType == ACCIDENTAL_WRITTEN_f && s_pnameForFlats[pos] != keyAccid->GetPname()) {
+                isCommon = false;
+                break;
+            }
+            else if (accidType == ACCIDENTAL_WRITTEN_s && s_pnameForSharps[pos] != keyAccid->GetPname()) {
+                isCommon = false;
+                break;
+            }
+            pos++;
+        }
+        if (!isCommon) {
+            LogWarning("KeySig content cannot be converted to @sig because the accidental series is not standard");
+            return sig;
+        }
+        sig = std::make_pair((int)childList.size(), accidType);
+    }
+    return sig;
+}
+
 //----------------------------------------------------------------------------
 // Static methods for KeySig
 //----------------------------------------------------------------------------
--- a/src/keysig.cpp
+++ b/src/keysig.cpp
@@ -243,7 +243,7 @@ data_KEYSIGNATURE KeySig::ConvertToSig() const
     data_KEYSIGNATURE sig = std::make_pair(-1, ACCIDENTAL_WRITTEN_NONE);
     const ListOfConstObjects &childList = this->GetList(this);
     if (childList.size() > 1) {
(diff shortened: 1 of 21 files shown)


src/options.cpp
```diff
@@ -1012,6 +1012,11 @@ Options::Options()
     m_mmOutput.Init(false);
     this->Register(&m_mmOutput, "mmOutput", &m_general);
 
+    m_moveScoreDefinitionToStaff.SetInfo("Move score definition to staff",
+        "Move score definition (clef, keySig, meterSig, etc.) from scoreDef to staffDef");
+    m_moveScoreDefinitionToStaff.Init(false);
+    this->Register(&m_moveScoreDefinitionToStaff, "moveScoreDefinitionToStaff", &m_general);
+
```
rettinghaus, line 1019: Why this without using it?

tools/main.cpp
```diff
@@ -447,10 +447,11 @@ int main(int argc, char **argv)
...
     }
 
-    if ((outformat != "svg") && (outformat != "mei") && (outformat != "midi") && (outformat != "timemap")
-        && (outformat != "humdrum") && (outformat != "hum") && (outformat != "pae") && (outformat != "pb-mei")) {
+    if ((outformat != "svg") && (outformat != "mei") && (outformat != "mei-basic") && (outformat != "midi")
+        && (outformat != "timemap") && (outformat != "humdrum") && (outformat != "hum") && (outformat != "pae")
+        && (outformat != "pb-mei")) {
         std::cerr << "Output format (" << outformat
-                  << ") can only be 'mei', 'pb-mei', 'svg', 'midi', 'humdrum' or 'pae'." << std::endl;
+                  << ") can only be 'mei', 'mei-basic', 'pb-mei', 'svg', 'midi', 'humdrum' or 'pae'." << std::endl;
```
rettinghaus, line 454: To me `basic` seems sufficient. Or change `pb-mei` into `mei-pb`?

src/keysig.cpp
```diff
@@ -238,6 +238,43 @@ int KeySig::GetFifthsInt() const
...
+            if (accidType == ACCIDENTAL_WRITTEN_f && s_pnameForFlats[pos] != keyAccid->GetPname()) {
+                isCommon = false;
+                break;
+            }
+            else if (accidType == ACCIDENTAL_WRITTEN_s && s_pnameForSharps[pos] != keyAccid->GetPname()) {
+                isCommon = false;
+                break;
+            }
+            pos++;
+        }
```
rettinghaus, line 268: You're missing key changes with encoded naturals. (Don't know if this is super important, but I've seen this coming from MusicXML.)

rettinghaus (commented):
Looks good. Some small comments.

src/options.cpp
lpugin: What do you mean? The option is false by default but used with `--move-score-definition-to-staff`

rettinghaus: Never mind. I missed the spot in iomei.

tools/main.cpp
lpugin: Changed to `mei-pb` and made `pb-mei` deprecated

src/toolkit.cpp
```diff
@@ -163,7 +163,11 @@ bool Toolkit::SetOutputTo(std::string const &outputTo)
     else if (outputTo == "mei-basic") {
         m_outputTo = MEI;
     }
+    else if (outputTo == "mei-pb") {
+        m_outputTo = MEI;
+    }
     else if (outputTo == "pb-mei") {
+        LogWarning("Output to 'pb-mei' is deprecaded, use 'mei-pb' instead.");
```
DavidBauer1984, line 170: (MINOR) Here and below: typo `deprecated`

src/keysig.cpp
rettinghaus: It comes from encoded key cancellation, so moving from _d minor_ to _d major_ you can find:

```xml
<keySig>
    <keyAccid accid="n" pname="b" />
    <keyAccid accid="s" pname="f" />
    <keyAccid accid="s" pname="c" />
 </keySig>
```

This would then require an `sig.showchange="true"`. 
We could check for correct order, or simply use any natural as flag.

rettinghaus (approved):
The logic isn't very easy to get, but should do the trick.

lpugin:
The logic is:
* Skip all naturals
* Use the first accid which is not a natural as the type for the key signature and count their number
* Stop when an accid which is not of the type of the key signature (or natural) is encountered because such a key signature cannot be represented as attribute (i.e., in MEI basic)
What is not clear?

lpugin:
I changed the variable initial assignment to ACCIDENTAL_WRITTEN_NONE but I am not sure it makes any difference in the logic.

ahankinson (approved):

PR #2952, opened 2022-07-05
Minor code maintenance 

rettinghaus:
This PR does some minor code maintenance like reducing scopes, formatting, and unification.

### DIFF
--- a/src/view_page.cpp
+++ b/src/view_page.cpp
@@ -1330,8 +1330,8 @@ void View::DrawLedgerLines(DeviceContext *dc, Staff *staff, const ArrayOfLedgerL
 
     if (below) {
         gClass = "below";
-        y -= ySpace * (staff->m_drawingLines - 1);
-        ySpace = -ySpace;
+        ySpace *= -1;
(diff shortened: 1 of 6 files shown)


bindings/qt/lib/src/toolkit.cpp
```diff
@@ -37,15 +37,14 @@ int Toolkit::adjustedDisplayHeightForPage(int page)
     }
     else {
         int averagePageHeight = 0;
-        int defaultPageHeight = 1000;
```
lpugin, line 40: I find it clearer to have it set separately and not within an instruction. But you can add `const` if you want.

src/view_page.cpp
```diff
@@ -1330,8 +1330,8 @@ void View::DrawLedgerLines(DeviceContext *dc, Staff *staff, const ArrayOfLedgerL
 
     if (below) {
         gClass = "below";
-        y -= ySpace * (staff->m_drawingLines - 1);
-        ySpace = -ySpace;
```
lpugin, line 1334: What is the improvement here?

PR #2955, opened 2022-07-06
Move Stem into separate file

Monceber:
Moved Stem out of elementpart.cpp file into separate one.
This makes code cleaner + easier to follow up with implementation of mensural stems.


src/stem.cpp
```diff
@@ -28,6 +28,8 @@ namespace vrv {
 // Stem
 //----------------------------------------------------------------------------
 
+static const ClassRegistrar<Note> s_factory("stem", STEM);
```
lpugin, line 31: Why do you want to have `stem` in the factory? This is not an MEI element

PR #2962, opened 2022-07-13
Minor code improvements

rettinghaus:


src/verse.cpp
```diff
@@ -249,11 +249,8 @@ int Verse::InitProcessingLists(FunctorParams *functorParams)
     return FUNCTOR_SIBLINGS;
 }
 
-int Verse::GenerateMIDI(FunctorParams *functorParams)
+int Verse::GenerateMIDI(FunctorParams *)
 {
-    GenerateMIDIParams *params = vrv_params_cast<GenerateMIDIParams *>(functorParams);
-    assert(params);
-
     Verse *previousVerse = vrv_cast<Verse *>(this->GetFirstAncestor(NOTE)->GetPrevious(this, VERSE));
```
lpugin, line 254: Looking at it, this is problematic. `this->GetFirstAncestor(NOTE)` need to be asserted / checked before calling `GetPrevious`

rettinghaus: Right. Also increases readability.

PR #2965, opened 2022-07-14
Update action and test time maps

rettinghaus:
This PR updates the GitHub Action to use Ubuntu 22.04 (which comes with Python 3.10.0). 

The test suite run now seeds the random generator to produce stable IDs and creates a timemap that can also be compared. For now it throws just an info if the timemap changes, but we can extend it if we want to add a deeper insight to the generated HTML.

### DIFF
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -6,9 +6,9 @@ import lxml.etree as etree
 import PIL.Image as Image
 import PIL.ImageChops as ImageChops
 import PIL.ImageOps as ImageOps
-import xmldiff.main as main
 from diffimg import diff as pngdiff
 from jsondiff import diff as jsondiff
+from xmldiff.main import diff_trees as xmldiff
 
 ns = {'svg': 'http://www.w3.org/2000/svg'}
 
@@ -201,7 +201,7 @@ if __name__ == "__main__":
                 root2.remove(e)
 
             # unused for now
-            diff = main.diff_trees(root1, root2)
+            diff = xmldiff(root1, root2)
             if (len(diff) > 0):
                 print(f'Node diff: {len(diff)}')
 
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -7,7 +7,7 @@ import PIL.Image as Image
 import PIL.ImageChops as ImageChops
 import PIL.ImageOps as ImageOps
 from diffimg import diff as pngdiff
-from jsondiff import diff as jsondiff
+from jsondiff import similarity as jsondiff
 from xmldiff.main import diff_trees as xmldiff
 
 ns = {'svg': 'http://www.w3.org/2000/svg'}
@@ -137,7 +137,7 @@ if __name__ == "__main__":
 
             timeMap1 = json.load(open(jsonFile1, 'r'))
             timeMap2 = json.load(open(jsonFile2, 'r'))
-            if jsondiff(timeMap1, timeMap2):
+            if jsondiff(timeMap1, timeMap2) != 1:
                 print(f'{name} produced a changed time map')
 
             diffValue = pngdiff(pngFile1, pngFile2, delete_diff_file=True)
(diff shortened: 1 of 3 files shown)


doc/test-suite.py
```diff
@@ -77,30 +75,39 @@
...
 
             tk.setOptions(json.dumps(options))
             tk.loadFile(inputFile)
+            # render to SVG
             svgString = tk.renderToSVG(1)
             svgString = svgString.replace(
                 "overflow=\"inherit\"", "overflow=\"visible\"")
             ET.ElementTree(ET.fromstring(svgString)).write(svgFile)
             cairosvg.svg2png(bytestring=svgString, scale=2, write_to=pngFile)
+            tk.resetOptions()
```
lpugin, line 110: Why resetting the options before rendering the timemap?

rettinghaus: Without reason.

doc/test-suite-diff.py
```diff
@@ -128,12 +128,19 @@ def end_webpage(html, body, htmlOutFile):
...
+            jsonFile2 = os.path.join(path_in2, item1, name + '.json')
             pngFileOut = os.path.join(path_out, item1, name + '.png')
             pngFile1Out = os.path.join(path_out, item1, name + '.after.png')
             pngFile2Out = os.path.join(path_out, item1, name + '.before.png')
             print(f'Comparing {name}')
 
-            diffValue = diffimg.diff(pngFile1, pngFile2, delete_diff_file=True)
+            timeMap1 = json.load(open(jsonFile1, 'r'))
+            timeMap2 = json.load(open(jsonFile2, 'r'))
+            if jsondiff(timeMap1, timeMap2):
```
ahankinson, line 140: Since you're not actually doing anything with the diff, you're only checking to see if they're different, you could probably get away with using the `similarity` function instead. This seems to do a bit less processing, and just returns a number indicating the similarity which is all you need. This also avoids implicitly checking the truthyness of a Dict result, which I generally avoid because it has some unexpected behaviours in some cases.

```
from jsondiff import similarity as jsonsimilar

... 

if jsonsimilar(t1, t2) < 1.0:
    print("...")
```

Dictionary truthiness:

```
>>> {} is False
False
>>> {} is True
False
>>> {} is None
False
>>> if {}: print("hello world")
>>> if not {}: print("hello world")
hello world
>>>

File: doc/test-suite-diff.py [hunk 3]
```diff
@@ -1,14 +1,14 @@
...
 
-import diffimg
 import lxml.etree as etree
 import PIL.Image as Image
 import PIL.ImageChops as ImageChops
 import PIL.ImageOps as ImageOps
-import xmldiff.main as main
+from diffimg import diff as pngdiff
+from jsondiff import diff as jsondiff
+from xmldiff.main import diff_trees as xmldiff
```
ahankinson, line 11: Good change! 👍 (who names an exported function `main`???)

doc/test-suite.py
```diff
@@ -77,30 +75,39 @@
...
             # filenames (input MEI/XML and output SVG)
             inputFile = os.path.join(path1, item1, item2)
-            print(inputFile)
+            options.update({"xmlIdSeed": int(hashlib.sha256(
+                inputFile.encode("utf-8")).hexdigest(), 16) % 10**9})
+            print(f'Rendering {item2}')
             name, ext = os.path.splitext(item2)
             svgFile = os.path.join(path2, item1, name + '.svg')
             pngFile = os.path.join(path2, item1, name + '.png')
+            timeMapFile = os.path.join(path2, item1, name + '.json')
```
ahankinson, line 88: Not a big deal, but I find using `f` strings a bit nicer than concatenating them... 

```suggestion
            timeMapFile = os.path.join(path2, item1, f"{name}.json")
```

rettinghaus: Plan for the next step is to switch to pathlib and clean up the whole section. So I'll leave this for now.

ahankinson (approved):

PR #2967, opened 2022-07-15
Support MEI input/output of Stem

DavidBauer1984:
This PR adds support for reading and writing `Stem` in MEI. To achieve this `Stem` now derives from `AttStemVis` and when constructing attribute stems we convert from `AttStems`, the base class of `Note/Chord`.

There is a [related PR on LibMEI](https://github.com/rism-digital/libmei/pull/17) that adjusts default values of stem length attributes.

Example:
```xml
<note dur="4" oct="4" pname="c" accid.ges="n" >
   <stem len="3.0" dir="down" pos="right"/>
</note>
```
Bildschirmfoto 2022-07-15 um 17 03 42

<details>
<summary>Show full MEI</summary>

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Stem length adjustment</title>
         </titleStmt>
         <pubStmt>
            <date>2017-05-17</date>
         </pubStmt>
         <notesStmt>
            <annot>The length of stems can be individually adjusted using the "stem.len" attribute.</annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="1.0.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure right="end" n="0">
                     <staff n="1">
                        <layer n="1">
                          <note dur="4" oct="4" pname="c" accid.ges="n" >
                            <stem len="3.0" dir="down" pos="right"/>
                          </note>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```
</details>

### DIFF
--- a/include/vrv/iomei.h
+++ b/include/vrv/iomei.h
@@ -401,6 +401,7 @@ private:
     void WriteProport(pugi::xml_node currentNode, Proport *proport);
     void WriteRest(pugi::xml_node currentNode, Rest *rest);
     void WriteSpace(pugi::xml_node currentNode, Space *space);
+    void WriteStem(pugi::xml_node currentNode, Stem *stem);
     void WriteSyllable(pugi::xml_node currentNode, Syllable *syllable);
     void WriteTabDurSym(pugi::xml_node currentNode, TabDurSym *tabDurSym);
     void WriteTabGrp(pugi::xml_node currentNode, TabGrp *tabGrp);
--- a/include/vrv/iomei.h
+++ b/include/vrv/iomei.h
@@ -700,6 +700,7 @@ private:
     bool ReadProport(Object *parent, pugi::xml_node proport);
     bool ReadRest(Object *parent, pugi::xml_node rest);
     bool ReadSpace(Object *parent, pugi::xml_node space);
+    bool ReadStem(Object *parent, pugi::xml_node stem);
     bool ReadSyl(Object *parent, pugi::xml_node syl);
     bool ReadSyllable(Object *parent, pugi::xml_node syllable);
     bool ReadTabDurSym(Object *parent, pugi::xml_node tabDurSym);
--- a/include/vrv/stem.h
+++ b/include/vrv/stem.h
@@ -9,8 +9,7 @@
 #define __VRV_STEM_H__
 
 #include "atts_cmn.h"
-#include "atts_shared.h"
-#include "atts_visual.h"
+#include "atts_mensural.h"
 #include "layerelement.h"
 
 namespace vrv {
@@ -22,9 +21,9 @@ class Flag;
 //----------------------------------------------------------------------------
 
 /**
- * This class models a stem as a layer element part and has not direct MEI equivlatent.
+ * This class models a stem as a layer element part and as MEI <stem> element.
  */
-class Stem : public LayerElement, public AttGraced, public AttStems, public AttStemsCmn {
+class Stem : public LayerElement, public AttGraced, public AttStemVis, public AttVisibility {
 public:
     /**
      * @name Constructors, destructors, reset and class name methods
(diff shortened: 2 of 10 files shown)


lpugin:
Because `<stem>` is part of the mensural MEI module, the encoding above is problematic - actually it is not valid with `mei-CMN`. Keeping `Stem` only as an internal layer element was the reason behind it. [Moving](https://github.com/rism-digital/verovio/pull/2955) it into distinct file is still fine, but this is now something quite different. Why not using `@stem.dir`, `@stem.pos` and `@stem.len`? Could you elaborate? I would suggest to bring the discussion to the MEI repository.

rettinghaus:
The reason to do it this way is to implement proper support for mensural stems like in this MEI: 

```xml
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Stem element</title>
         </titleStmt>
         <pubStmt>
            <date>2022-07-22</date>
         </pubStmt>
         <notesStmt>
            <annot>A mensural note may have two stems.</annot>
         </notesStmt>
      </fileDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" clef.shape="C" clef.line="3" notationtype="mensural" />
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure>
                     <staff n="1">
                        <layer n="1">
                          <note dur="minima" oct="4" pname="c" accid.ges="n" >
                            <stem dir="up" flag.form="curled"/>
                            <stem dir="down"/>
                          </note>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

Usually `[elementName].[attributeName]` attributes are a subgroup to their corresponding `[elementName]@[attributeName]` versions (e.g., `key.pname` equals `keySig@pname`). In Verovio we always resolve the former into the latter (e.g., `note@accid` -> `accid@accid`) for better handling and SVG output. Basically we do the same with stem already, so it seems natural to use the element's respective attributes. As `stem` is an element that came late to the party, `att.stems` is not really a subgroup of the attributes available on `stem`, so we need some way to translate them into each other. But from our point of view this is better than having `att.stems` _and_ `att.stem.vis` for stem. (Also we need `@flag.form` and `@flag.pos`.) 

There is already a [PR open](https://github.com/music-encoding/music-encoding/pull/981) to reduce the incompatibility between the attribute classes. (Still I'm not quite sure what should happen with `@stem.mod`.)

lpugin:
I was really pointing to the fact that example given is not valid CMN MEI:
```xml
<note dur="4" oct="4" pname="c" accid.ges="n" >
   <stem len="3.0" dir="down" pos="right"/>
</note>
```
Reading and writing stem elements should be limited to mensural notation. Otherwise that needs to be clarified in MEI.

(The handling of attributes as corresponding elements is still different here than what we do with accidentals because the conversion happens in `PrepareLayerElementParts` and not in MEIInput.)

rettinghaus:
How should this be limited to mensural notation? It is valid in `MEI-all`, and that is what we are supporting, correct? 

The following is also not valid in CMN (and nonsense anyway):
```xml
<chord dur="maxima">
    <note oct="4" pname="c" stem.dir="down" stem.pos="left" />
    <note oct="4" pname="e" stem.dir="up" stem.pos="right" />
</chord>
```

But Verovio makes something out of it:
image

The main question is when and how do we want to construct stems. Trying to do everything during load we would still need the `PrepareLayerElementParts` functor.

lpugin:
I do not find that the argument that, because Verovio already supports nonsense, it is OK for it to support more nonsense to be very constructive. You know very well that the stem element is part of the [MEI-mensural](https://music-encoding.org/guidelines/dev/elements/stem.html) module. You also know that using MEI-all is not necessary the best recommended practice - precisely because it can very quickly become nonsense. (no need to add an example of a `<ligature>` containing a `<chord>` here, we all know this is possible with MEI-all but does not make much sense musically.)

Do you think we should make Verovio more picky about these cases that mixes modules in a way that is it not really meant in MEI?

rettinghaus:
If you think the conversion should better be done in the `iomei` module we can move it there of course.

lpugin:
> If you think the conversion should better be done in the iomei module we can move it there of course.

Let's leave it like this for now.

lpugin:
@DavidBauer1984 any ideas why [this](https://ci.appveyor.com/project/LaurentPugin/verovio-r1t6l) is failing? Adding the forward declaration of [Stem](https://github.com/rism-digital/verovio/commit/4d90bc347a6565151904e7c5c17c02421bbfa43f) did not fix it.

lpugin:
I think we reached the stupid 127 limit of `else if` Microsoft sets https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1061?view=msvc-170
It is hard to believe they would put such a low limit... It is especially irritating to see "In any case, deeply nested code is difficult to understand and reason about".

DavidBauer1984:
Thanks for the discussion (I missed it due to summer vacation)!

@lpugin Regarding the 127 limit another option would be to use a switch-case on the class ID instead of (nested) if-else. This should be faster and allow more cases (the minimal number is still compiler dependent, see [this discussion](https://stackoverflow.com/questions/20355948/maximum-number-of-cases-that-can-be-addressed-using-switch-statement)). What do you think?

lpugin:
@DavidBauer1984 welcome back! 

I fixed it [here](https://github.com/rism-digital/verovio/commit/908f795231580249169640924f4938f291fe2068) by grouping LayerElement checks. I am fine as it is now and I do not think we will hit a compiler limit any time soon.

PR #2972, opened 2022-07-20
Replace C standard library headers

rettinghaus:
This is a small cosmetic change to prefer `cassert` and `cstdlib` over the C standard library headers.
A corresponding commit for libmei has been added to the [open PR](https://github.com/rism-digital/libmei/pull/16).


ahankinson:
Out of curiosity, what's the advantage?

rettinghaus:
There's no real advantage to us, except usage of C standard library headers was considered deprecated in C++17. Also they are simply a subset of the respective C++ headers. 

And there may be a bit more clarity in using namespaces: 

C++17 standard

rettinghaus:
The [C++20 standard](https://isocpp.org/files/papers/N4860.pdf) states:
> For compatibility with the C standard library, the C++ standard library provides the C headers enumerated in D.9, but their use is deprecated in C++.

ahankinson:
ok, thanks

lpugin:
I think that in fact C++20 has un-deprecated them

https://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0619r0.html#3.5

src/pugi/pugixml.cpp
```diff
@@ -19,7 +19,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include <assert.h>
```
lpugin, line 22: You should open a PR on pugixml and not change it here.

rettinghaus: Sorry. Actually there is one already: https://github.com/zeux/pugixml/pull/350/

PR #2973, opened 2022-07-21
Use VRV_UNSET

rettinghaus:
This PR switches to use VRV_UNSET for all integer attribute values as default by updating LibMEI. Corresponds to /rism-digital/libmei/pull/18.


rettinghaus:
Made some more tests and opened a [second PR for LibMEI](https://github.com/rism-digital/libmei/pull/19).
To me everything looks good now.

PR #2978, opened 2022-07-22
Fix beam overlaps with other elements

Monceber:
Fix for overlaps of beams with other elements within (rests, clefs, etc.)

Previous rendering:
image
image

New rendering:
image
image

Example is too long to be attached add it to body of PR :(


lpugin:
Could you add the second example? That one is short enough. Thanks!

PR #2985, opened 2022-07-26
Another small improvements for unisons

rettinghaus:


lpugin:
Could you give an example of what this is fixing? Thanks

rettinghaus:
It fixes `layer-007` (see penultimate measure), that was affected by the previous PR.
layer-007 before
layer-007 after

But it keeps `layer-014` intact, that obviously was broken before (see https://github.com/rism-digital/verovio/suites/7521899534/artifacts/309725072).

_Before_ the last PR we had cases like this:
image

_After_ these changes it looks more like expected:
image

PR #2991, opened 2022-08-08
Remove unused code

### LINKED ISSUE #2989: Never executed code
In the `editortoolkit_neume.cpp` there is a weird statement.
At this point `syllableFi` is always NULL, so the else block will never get executed.
https://github.com/rism-digital/verovio/blob/bb0c961574d0dbb5ce7253a82bec04fd32342e80/src/editortoolkit_neume.cpp#L1726-L1745

@JRegimbal could you please have a look there?

rettinghaus:
This PR removes some lines of unused code and makes some variables `const`. 
Also it removes the deprecated `codecvt` library and replaces it with the internal `UTF8to16` method instead.
Closes #2989

### DIFF
--- a/include/vrv/editortoolkit_neume.h
+++ b/include/vrv/editortoolkit_neume.h
@@ -43,7 +43,7 @@ public:
         std::vector<std::pair<std::string, std::string>> attributes);
     bool Merge(std::vector<std::string> elementIds);
     bool Set(std::string elementId, std::string attrType, std::string attrValue);
-    bool SetText(std::string elementId, std::string text);
+    bool SetText(std::string elementId, const std::string &text);
     bool SetClef(std::string elementId, std::string shape);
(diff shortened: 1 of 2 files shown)


lpugin:
Thanks. In the future, could you try to be a bit more specific in the commits comments? "minor code improvements" or "minor update" are not very helpful. Something like "Adding const and fix type" for [2886f61](https://github.com/rism-digital/verovio/pull/2991/commits/2886f612eecc6d915c51b04d2a22eee99fc51772) would have been about the same length and more informative. Thanks!

rettinghaus:
Good point. I tend to oversimplify things if functionality isn't changed.

lpugin:
Thanks. Yes, the problem is really when something breaks and we need to go through the commits to find it. Changing something that does not change functionality can still break, and having a hint when seeing only the commit message is useful.

PR #3006, opened 2022-08-26
Option for compound dynamics glyph usage

rettinghaus:
This PR adds another option to switch from using SMuFL's pre-defined compound dynamics glyphs to using single glyphs for rendering.

### DIFF
--- a/include/vrv/dynam.h
+++ b/include/vrv/dynam.h
@@ -77,7 +77,7 @@ public:
      * Return the SMuFL str for the dynamic symbol.
      * Call IsSymbolOnly first to check.
      */
-    std::wstring GetSymbolStr() const;
+    std::wstring GetSymbolStr(const bool singleGlyphs) const;
 
     /**
      * See FloatingObject::IsExtenderElement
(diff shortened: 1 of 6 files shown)


lpugin:
Could you add a couple of examples? Thanks

rettinghaus:
With `dynam-001.mei` from the test suite Verovio normally produces this: 

dynam-001-compound-glyphs

With the new option `--dynam-single-glyphs` it uses the corresponding single dynamic glyphs for f, m, p, r, s, and z. The result looks almost the same: 

dynam-001-single-glyphs

But in the SVG we now have single entities:

```xml
<g id="d1t8wmsc" class="dynam">
    <use xlink:href="#E524-1uoe3tk" x="6396" y="1803" height="720px" width="720px" />
    <use xlink:href="#E522-1uoe3tk" x="6535" y="1803" height="720px" width="720px" />
    <use xlink:href="#E525-1uoe3tk" x="6702" y="1803" height="720px" width="720px" />
 </g>
```

E.g., allowing coloring of each entity with CSS: 

dynam-001-single-glyphs

lpugin:
Great, thanks. I suspect this will not work with Bravura because of how the advance is set - unless this has changed since last time I looked at it.

rettinghaus:
Indeed, Bravura and Leland do not really work, especially the spacing in forte is bad:  
image
image

craigsapp:
You could have an advance correction applied to Bravura and Leland, and change that correction to zero when fixed in Bravura and Leland.  I thought a recent version of Leland fixed this problem in the font metrics?

lpugin:
Eventually I think Bravura should be fixed. If we correct it, it will remain unnoticed, so it might be counter-productive.

PR #3010, opened 2022-08-31
Fix end of measure clef import

### LINKED ISSUE #3009: Clef change at beginning of measure renders strangely
When introducing a clef change on the first beat of a measure, Verovio renders it in a previous measure instead.

**To Reproduce**
Minimal XML file (generated by music21) that demonstrates the behavior is [here](https://gist.github.com/KunstDerFuge/790cc85bd4e2f526100b0a12a2557c83).
When rendered in Verovio, the bass clef, which should appear at the beginning of measure 2, instead appears on the & of beat one in measure one.

image

**Expected behavior**
A bass clef should appear at the beginning of measure 2, as happens when the file is rendered with Musescore:

image

**Input data**
MusicXML for this example is available [in this gist](https://gist.github.com/KunstDerFuge/790cc85bd4e2f526100b0a12a2557c83).

**Verovio information**
 - Verovio 3.9.0
 - Javascript version

The problem can also be observed when uploading this file on the [Verovio website's MusicXML viewer](https://www.verovio.org/musicxml.html).

 - OS: Linux (Manjaro)
 - Firefox

**Additional info**
This behavior does not happen when the time signature is changed, e.g. to 6/4 (so that the new clef no longer falls on the first beat of a new measure), or when the new clef is moved to the next or previous eighth note.

rettinghaus:
closes #3009

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -218,7 +218,8 @@ void MusicXmlInput::ProcessClefChangeQueue(Section *section)
             else {
                 // For previous measure we need to make sure that clef is set at the end, so pass high duration value
                 // (since it won't matter there) and set measureNum to empty, since it doesn't matter as well
-                const int endDuration = 1024;
+                int endDuration = m_ppq;
+                for (int &num : m_meterCount) endDuration *= num;
                 musicxml::ClefChange previousClefChange(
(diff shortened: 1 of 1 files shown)


lpugin:
Thanks

PR #3011, opened 2022-09-01
Update Leland

rettinghaus:


fonts/generate_all.sh
```diff
@@ -3,22 +3,24 @@
 # Requires python
 # Requires svgpathtools and lxml python libraries (install with 'pip3 install %name%')
 
+PYTHON="python3.9"
+
```
ahankinson, line 7: Do you have an example where the `python3` binary doesn't work? If I have `python3.10` this will break my usage, but `python3` will still work.

rettinghaus: In my specific case my default points to Python 3.10, but I have all requirements for Verovio on 3.9. Pinning it here maybe is better than dealing with venv?
But we can default it to `python3` .

ahankinson: I use virtualenvs for everything; makes it easier to keep the different versions of things straight. But I think setting it to `python3` by default will keep it working in more instances. For example, on my machine:

```
$ python3.9
zsh: command not found: python3.9
```

PR #3013, opened 2022-09-02
Code maintenance

rettinghaus:
This PR brings some minor code changes to js files static code checkers were complaining about. 
It also updates the iOS headers by running the included shell script. 

No changes in behaviour are expected.

### DIFF
--- a/bindings/iOS/all.h
+++ b/bindings/iOS/all.h
@@ -41,16 +41,19 @@
 #import <VerovioFramework/hairpin.h>
 #import <VerovioFramework/staffgrp.h>
 #import <VerovioFramework/pitchinflection.h>
+#import <VerovioFramework/resources.h>
 #import <VerovioFramework/text.h>
 #import <VerovioFramework/space.h>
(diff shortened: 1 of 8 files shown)


lpugin:
Out of curiosity: what is the problem with single quotes?

craigsapp:
I would say nothing as they are interchangeable in javascript, but it does make C++ programmers happy :-)

rettinghaus:
It's just to make Codacy happy. Like javascript I don't care too much (although I also have a slight preference to doublequotes).

ahankinson:
If we're introducing these sorts of standards into the code base there should be some sort of shared configuration. Otherwise we're going to get lots of these little commits that just bring them into a "local" standard that is not shared by anyone else on the project.

lpugin:
Consistency can be a justification for such changes, but just personal preferences cannot.

In any case, it would be good if such changes could be documented with more information in the PR / commits together with some coding guidelines and configurations where ever they apply.

Issue #1533, opened 2020-07-06
Add pageWidth/HeightScaled parameters


rettinghaus:
The default value of `--unit` is 9, but it is unclear from the documentation/options, what this "9" means, i.e. what real-world units are meant. Internally it translates to 90 pixel for the output, which then is "scaled back" by 1 : 10 ratio in the svg between width/height and `viewBox`. Why has this way been chosen, and why are only integer value allowed/accepted for unit? 

Another problem is, that `--mm-output` doesn't translate the effective pixels correctly into mm, because px and mm are absolute values. 
Converting between units isn't just renaming them. If `--mm-output` is used, the `vu` isn't set to 9 pixels, nor to 9 millimeters, so it feels like just a random number controlling the staff height. 

The absolute values should be converted this way: 
in = inches (1in = 96px = 2.54cm)
px = pixels (1px = 1/96th of 1in)
pt = points (1pt = 1/72 of 1in)
pc = picas (1pc = 12 pt)

Or, to put it another way: using Verovio in a browser, independently of which unit is used for output, it should always look the same.

lpugin:
> Why has this way been chosen, and why are only integer value allowed/accepted for unit?

This used to be thought for printed output with a staff space of 1.8mm (unit 0.9mm). Accordingly the default page size is 2970 and 2100 which coincidentally matches A4 mm times 10 

`--mm-output` was added to have the conversion from SVG to PDF to produce the desired size.

lpugin:
To clarify the `--mm-output`: it does not mean 'converting' to mm, but instead output units in mm (actually 1/10mm). I agree we should clarify this. So do we need to clarify that default units are pixels.

(AFAIK, SVG does not make any assumptions on the dpi to be used for the conversion - CSS2 does, but not SVG. Some tools use 72dpi, some 90dpi, others 96dpi (e.g., browsers). This is notoriously problematic. So I think it is better to stay away from the idea that we can convert pixels to mm)

rettinghaus:
So, why not use `mm` as default? All the default values are pointing to a A4 format layout. Instead we then could use a `--output-width` option (including unit), which sets svg's `@width` and `@height` accordingly, e.g. `500pt`, `382px`, or `12em`. That would be easier than trying to figure out a suitable solution with `scale` and `page-width` and `unit` …

lpugin:
[...]

Issue #1055, opened 2019-05-10
Toolkit: Document units on `pageHeight` and `pageWidth`


earboxer:
What are the units on pageHeight and pageWidth?

`vrvToolkit.getAvailableOptions().groups['1-general'].options.pageHeight.description` should say more than "The page height"

The tutorial at https://www.verovio.org/tutorial.xhtml?id=topic01 shows

```
var zoom = 30;
    pageHeight = $(document).height() * 100 / zoom ;
    pageWidth = $(window).width() * 100 / zoom ;
    options = {
                pageHeight: pageHeight,
                pageWidth: pageWidth,
                scale: zoom,
                adjustPageHeight: true
            };
    vrvToolkit.setOptions(options);
```

Would it make sense/be possible to have a `pixelHeight` and `pixelWidth` option to set the SVG size without needing to be a product with the scale?

lpugin:
Units are indeed pixels with scale = 100. If we add another option then we will raise other issues. For example, which one of scale or pageHeight should be preserve in order to obtain the desire output.
If you have a good solution, please make a PR.

PR #3014, opened 2022-09-03
Add option to scale the output to the page size

lpugin:
When the `scale` parameter is set, Verovio scales the output page accordingly. This means that, when targetting a certain output size, one needs to calculate the page size according to the scale parameter. 

The `--scale-to-page-size` option makes it possible to have Verovio behaving the other way around. That is, always producing an output with the same size independently from the scale factor and instead scaling the content. `RedoLayout()` needs to be called once the scale value is changed.

For example, with the option `--scale-to-page-size` on and `--page-height 800` and `--page-width 1000`, we have the following output with a scale of `100`, `50` and `10` respectively:

image

---

image

---

image

---

This addresses requests raised before in #1533 and and #1055

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -427,6 +427,12 @@ void SvgDeviceContext::StartPage()
                                                         .c_str();
     }
 
+    // page rectangle - for debugging
+    // pugi::xml_node pageRect = m_currentNode.append_child("rect");
+    // pageRect.append_attribute("fill") = "pink";
+    // pageRect.append_attribute("height") = StringFormat("%d", this->GetHeight()* DEFINITION_FACTOR).c_str();
+    // pageRect.append_attribute("width") = StringFormat("%d", this->GetWidth() * DEFINITION_FACTOR).c_str();
+
     // a graphic for the origin
     m_currentNode = m_currentNode.append_child("g");
     m_svgNodeStack.push_back(m_currentNode);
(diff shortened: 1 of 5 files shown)


src/svgdevicecontext.cpp
```diff
@@ -427,19 +427,26 @@ void SvgDeviceContext::StartPage()
                                                         .c_str();
     }
 
+    // page rectangle - for debugging
```
ahankinson, line 430: Should this be removed?

src/svgdevicecontext.cpp
```diff
@@ -427,19 +427,26 @@ void SvgDeviceContext::StartPage()
...
     m_currentNode.append_attribute("transform")
         = StringFormat("translate(%d, %d)", (int)((double)m_originX), (int)((double)m_originY)).c_str();
 
-    // margin rectangle for debugging
-    // pugi::xml_node rect = m_currentNode.append_child("rect");
-    // rect.append_attribute("fill") = "pink";
-    // rect.append_attribute("height") = StringFormat("%d", this->GetHeight() * DEFINITION_FACTOR - 2 *
-    // m_originY).c_str(); rect.append_attribute("width") = StringFormat("%d", this->GetWidth() * DEFINITION_FACTOR - 2
-    // * m_originX).c_str();
+    // margin rectangle - for debugging
```
ahankinson, line 443: This too?

src/svgdevicecontext.cpp
lpugin: I left it for documentation. It is often useful for checking it

ahankinson:
Can I merge this? It looks good to me.

Issue #2651, opened 2022-02-16
Improve support for SMuFL glyphs within text


lpugin:
Verovio currently supports mixing text and SMuFL glyphs in the following situations:
* Lyrics for text elisions
* Tempo indication with notes
* Dynamics mixing text and dynamic glyphs (e.g., "sempre pp")
* Directives, for example for repeat signs

In such cases, the VerovioText font is used because we need the SMuFL glyphs to be treated as text in the SVG - and not as SVG shapes like elsewhere in the music rendering. VerovioText is a very small subset of Leipzig which is embedded in the SVG as woff.

There are a number of limitations:
* Only the Leipzig typeface is supported, and the music font option is ignored. As a result, music fonts can be mixed up when another font than Leipzig is selected, which is not optimal
* There might be some scaling difference between the original Leipzig and the set included in VerovioText. These are difficult to handle

### Proposal

It would be much better to have a `woff` being a subset for all music fonts supported by Verovio and not only for Leipzig. It would also be much better if these would be generated from the original music fonts and not maintained by hand.

What we would need is
* A list of glyph we would like to be available to be mixed up with text (elision, dynamics, repeats, notes for tempo and tuplets, etc.)
* Adjust the script to extract these glyphs from the music fonts and generate a woff for each music font (Leipzig, Bravura, Leland, etc.)
* Adjust Verovio to select the matching text font (for example, if Bravura is selected, we want to use the corresponding woff file)

### Open question

Because the woff file is embedded in the SVG, there is no need to have the font installed on a machine for it to display properly. We would maintain this. In some cases, however, it is necessary to have the font installed (e.g., when converting to PDF or PNG). For this, it is necessary to have the corresponding font available as TrueType (or similar) to be installable. We would need to think about the font name. 

VerovioText was a quick naming solution but would not be appropriate any more. Furthermore the resulting fonts would not exactly be a LeipzigText or a BravuraText, because SMuFL text fonts fulfil a different purpose. That means that we probably need to think about another font names, assuming that installing the complete music font would not work - even thought maybe that would work?

(Somehow related to https://github.com/music-encoding/music-encoding/pull/897 and https://github.com/music-encoding/music-encoding/pull/704)

lpugin:
See https://github.com/rism-digital/verovio/pull/3024

PR #3017, opened 2022-09-09
Replace the VerovioText font with the Leipzig font

lpugin:
This PR is a first step in improving support for SMuFL glyphs within text as explain [here](https://github.com/rism-digital/verovio/issues/2651)

All the glyphs previously defined in VerovioText are now available as standard music glyphs. They are part of Leipzig (5.2.77). 

Some additional glyphs have been added because the glyphs previously used in VerovioText were not exactly correct:
* accidentals for figured bass (previously using music accidentals in VerovioText)
* metronom mark notes (previously using the whole notes in VerovioText)

With this, we can now drop VerovioText, including the text bounding box resource we had to calculate the spacing. All the spacing is now calculated using the music font resource.

The PR also changes the way triggering the music font works. Previously, we would do
```xml
<rend fontname="VerovioText">&#xe048;</rend>
```
Verovio now expect a `@fontfam` with `"smufl"`. That is:
```xml
<rend fontfam="smufl">&#xe048;</rend>
```

The MusicXML and Humdrum have been [changed](https://github.com/rism-digital/verovio/commit/b7d1bcc94c29c9ff17988facf99aaf7e06c6a57f) accordingly. Verovio will still loads `rend` with `@fontname="VerovioText"` but will display a deprecation warning. Eventually we will need to adjust the examples in the test suite.

### Web font

We still use a web font embedded in the SVG. However, with a few differences: 
* The font is no longer VerovioText but Leipzig.
* We use Web Font version 2, which has a slightly better compression (30%) - the XML to be included for VerovioText used to be 32kb, with Leipzig in woff2 is now 52kb, which we can still very reasonably have included in the SVG.
* The font is now defined at the end of the SVG. It seems to work fine, but we should keep this change in mind if this is an issue with some SVG renders.

### Expected changes

This PR should not trigger changes except for some small alignment / position differences because the font system is different. One difference thought will be that we need to install `Leipzig` when running the test suite. This means that the CI test suite evaluation for the PR will probably not be very helpful. However, I evaluated it locally and everything looks fine.

### Next steps

The next step will be to allow for other fonts to be used as well. This requires essentially generating a Web Font 2 for each font we support and that we can then include in the SVG in the same way.

Another thing that will follow is the support for mixed content with `<symbol>`. This will remove the current limitation that we cannot have `<symbol>` within a `<dir>` that also includes text. This will be done by making Symbol a TextElement class.


samuelbradshaw:
> * The font is now defined at the end of the SVG. It seems to work fine, but we should keep this change in mind if this is an issue with some SVG renders.

I'm not very familiar with how fonts are currently being defined, but does this mean that if the SVG is split into multiple pages, the font will be defined multiple times?

lpugin:
> I'm not very familiar with how fonts are currently being defined, but does this mean that if there are multiple SVGs rendered (multiple pages in the score), the font will be defined multiple times?

Yes, that is correct. However, nothing has changed on that side. I do not think that has ever been an issue, but maybe I am wrong. Should it be the case, then we can always add an option not to include the font within the SVG and instead let the use set the URL of the font.

samuelbradshaw:
OK. If the result is the same as before, it should be fine for my use cases (I haven't had any issues because of fonts previously). Thanks!

rettinghaus:
I played around with this a bit, and mostly things seems to be fine. Biggest concern is that we loose coda/segno symbols when importing MusicXML. We could switch to those glyphs from the SMuFL range, but many users would probably still try to use unicode code points in similar cases? 
I added those glyphs to the respective code points but still couldn't get them to be shown. Please have a look.

lpugin:
> Biggest concern is that we loose coda/segno symbols when importing MusicXML. We could switch to those glyphs from the SMuFL range, but many users would probably still try to use unicode code points in similar cases?

I would suggest to do the same what is done for accidentals in figured bass. That is to map standard unicode points to SMuFL corresponding ones when rendering. That way we do not need to have duplicated glyphs in the font.

https://github.com/rism-digital/verovio/blob/7188cb1396dd1e7948a319b8f0e953ac41e110c7/src/view_text.cpp#L154-L171

(We also map SMuFL standard accidentals to SMuFL figured bass accidentals because this was previously supported, but this encoding practice is not to be recommended)

We could also adjust the MusicXML importer, but if this is happening at rendering anyway it would not be necessary.

lpugin:
@rettinghaus are you willing to support
```xml
<dir>𝄌</dir>
```
or
```xml
<dir><symbol>𝄌</symbol></dir>
```
or both?

rettinghaus:
For segno and coda we need `<dir><rend>𝄌</rend></dir>`, otherwise it will be impossible to center the glyphs.

lpugin:
Yes, and you probably also want to change the font style because `dir` are rendered in italic by default. 

For now we used to support:
```xml
 <dir><rend fontname="VerovioText" fontstyle="normal">𝄌</rend></dir>
```
That would still work because it would be converted to
```xml
 <dir><rend fontfam="smufl" fontstyle="normal">𝄌</rend></dir>
```
But that does not really make sense because these glyphs are not part of SMuFL. I could leave with it, though. 

Ideally it should be just:
```xml
<dir><rend fontstyle="normal">𝄌</rend></dir>
```
We could try to convert all unicode music glyph to SMuFL equivalent as I suggested. It would be straight forward if the `rend` contains only music glyphs. If they are mixed with text it would be more complicated, even though not impossible. I am not sure it is worth doing it, though. Any thoughts?

lpugin:
(Ideally I would prefer to have something like:
```xml
<dir func="coda"/>
```
which could even be empty by default and have some text only if needed. But it is a whole different story, and we had that discussion [already](https://github.com/music-encoding/music-encoding/issues/712) but it never went anywhere...)

lpugin:
Unicode glyphs within `<rend>` with `@fontfam="smufl"` now render properly

image

```xml
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title>Directives with repeat symbols</title>
            </titleStmt>
            <pubStmt>
                <date>2022-02-15</date>
            </pubStmt>
            <notesStmt>
                <annot>The following SMuFL symbols can be used within directives.</annot>
            </notesStmt>
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application version="3.9.0" label="2">
                    <name>Verovio</name>
                </application>
            </appInfo>
        </encodingDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef>
                        <staffGrp>
                        </staffGrp>
                    </scoreDef>
                    <section>
                        <measure>
                            <staff n="1">
                                <layer n="1">
                                    <note dur="4" oct="4" pname="c" />
                                    <note dur="4" oct="4" pname="c" />
                                    <note dur="4" oct="4" pname="c" />
                                    <note dur="4" oct="4" pname="c" />
                                </layer>
                            </staff>
                            <dir place="above" staff="1" tstamp="1"><rend fontfam="smufl" fontstyle="normal">&#x1d109;</rend></dir>
                            <dir place="above" staff="1" tstamp="2"><rend fontfam="smufl" fontstyle="normal">&#x1d10a;</rend></dir>
                            <dir place="above" staff="1" tstamp="3"><rend fontfam="smufl" fontstyle="normal">&#x1d10b;</rend></dir>
                            <dir place="above" staff="1" tstamp="4"><rend fontfam="smufl" fontstyle="normal">&#x1d10c;</rend></dir>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

rettinghaus (approved):
Looks good to me.

PR #3027, opened 2022-09-20
Change std::wstring to std::u32string

lpugin:
The use of `std::wstring` (and `wchar_t`) is problematic with music unicode characters on Windows because `wchar_t` on Windows are limited to 16 bits.

This PR switch to `std::u32string` and `char32_t`. A couple of methods using `std::wstringstream` had to be changed because these do not work as it used to with `std::u32string`s.


Monceber (approved):

rettinghaus (approved):

include/vrv/view.h
```diff
@@ -546,19 +546,19 @@ class View {
...
         DeviceContext *dc, Point bezier[4], int thickness, int staffSize, int penWidth, int penStyle = AxSOLID);
     void DrawPartFilledRectangle(DeviceContext *dc, int x1, int y1, int x2, int y2, int fillSection);
-    void DrawTextString(DeviceContext *dc, std::wstring str, TextDrawingParams &params);
-    void DrawDirString(DeviceContext *dc, std::wstring str, TextDrawingParams &params);
-    void DrawDynamString(DeviceContext *dc, std::wstring str, TextDrawingParams &params, Rend *rend);
-    void DrawHarmString(DeviceContext *dc, std::wstring str, TextDrawingParams &params);
-    void DrawSmuflLine(DeviceContext *dc, Point orig, int length, int staffSize, bool dimin, wchar_t fill,
-        wchar_t start = 0, wchar_t end = 0);
-    void DrawSmuflString(DeviceContext *dc, int x, int y, std::wstring s, data_HORIZONTALALIGNMENT alignment,
+    void DrawTextString(DeviceContext *dc, std::u32string str, TextDrawingParams &params);
```
DavidBauer1984, line 553: Here and the following: Pass strings as const reference whenever possible.

src/harm.cpp
```diff
@@ -80,19 +80,19 @@ bool Harm::GetRootPitch(TransPitch &pitch, unsigned int &pos) const
...
     if (!textObject) return false;
-    std::wstring text = textObject->GetText();
+    std::u32string text = textObject->GetText();
 
     if (text.length() > pos && text.at(pos) >= 'A' && text.at(pos) <= 'G') {
         int pname = (text.at(pos) - 'C' + 7) % 7;
         int accid = 0;
         for (pos++; pos < text.length(); pos++) {
-            if (text.at(pos) == L'𝄫')
+            if (text.at(pos) == U'\U0001D12B')
```
DavidBauer1984, line 89: Can we define those unicode literals together in one place and use constants (with proper names) instead of cryptic literals?

include/vrv/view.h
lpugin: Yes, but here that is not possible. I changed it because I had to drop the `std::stringstream` and tokenize the string 'by hand'. This means modifying it. 
https://github.com/rism-digital/verovio/blob/3f15f1f9034edd4ced47d34cc2bc9aaea94fc652/src/view_text.cpp#L291
If we pass it as reference we need to make a copy, which would not make it more efficient.

src/harm.cpp
lpugin: This is what I did first but had to [change](https://github.com/rism-digital/verovio/commit/9bdecc24f4dc091e4b9380869cf23595f04d373e) it because MS Visual Studio does not like it:
https://ci.appveyor.com/project/LaurentPugin/verovio-r1t6l/builds/44834728

We will add them as comment for improving readability.

lpugin: https://github.com/rism-digital/verovio/pull/3027/commits/1174fe2567162e3e325e6af9fdb07c74087ef768

include/vrv/view.h
DavidBauer1984: Yes, this is in `DrawLyricString` and I agree that passing by value is better there. For the other functions:

- `DrawTextString`: `str` is not modified and hence should be passed as const ref.
- `DrawDirString`: `str` is not modified, but copied at the first opportunity into `adjustedStr` which then is modified. Here we should keep passing `str` by value and remove the local variable `adjustedStr`.
-  `DrawDynamString`:  `str` is not modified and hence should be passed as const ref.
-  `DrawHarmString`:  `str` is not modified and hence should be passed as const ref.

craigsapp: In such cases I pass by const reference, and then copy the string inside of the function.  In such cases there will be no loss in efficiency since there has to be a copy made somewhere, so I am doing it explicitly rather than implicitly.   This makes the function declaration look nicer and you don not have to wonder if the const reference was not added by accident.

lpugin: Yes, this is probably clearer. I'll change it.

lpugin:
@craigsapp I am going to merge this. I double checked the changes in `iohumdrum.cpp` and it should be fine.

include/vrv/view.h
DavidBauer1984: > In such cases there will be no loss in efficiency since there has to be a copy made somewhere

Just want to remark that this is not true if an rvalue is passed into the function. If the string is passed by value, then the string will be moved and not copied in this case which is more efficient. It’s like copying a pointer instead of a whole container.
However, let’s leave it as above and passing everything by const ref should be good enough, I’m happy with that.

PR #3028, opened 2022-09-21
Refactor font generation

ahankinson:
This PR refactors the font generation process. Several Python scripts are combined into a new utility, `generate.py` which currently handles the following tasks:

 - generates the `smufl.h` file containing the supported glyphs
 - extracts the supported glyphs from an SVG font
 - creates a CSS `@font-face` definition with a base64-encoded woff2 font embedded
 - converts fonts to SVG or WOFF2

The last two require a `fontforge` installation to run. 

`lxml` has been deprecated since the attribute ordering problem was fixed in Python 3.8.

`svgpathtools` are still required, but since it's so heavy (requiring numpy and scipy, which requires LAPACK/BLAS and a fortran compiler) suggestions for alternate solutions are welcome!

The `supported.xsl` file has been changed to `supported.xml`, and the XSLT functions removed. It was not being used as an XSLT anyway, so it was decided to change this. 

From conversations with @lpugin the generate.py script does *not* currently handle text fonts, unlike previous versions. This can be changed if desired.

---

`generate.py` uses the built-in Python `argparse` module, which provides some additional niceties for the command-line utility. 

`python3 generate.py --help` should provide help about the script itself. The individual functions are implemented as sub-commands, so `python3 generate.py css --help` will provide help specific to the `css` sub-command.

---

How to upgrade: 

The structure of the fonts directory has changed significantly. Each font now has its own directory, and the generate.py script is built to work against this. If you add new fonts, please follow the `Fontname/Fontname.[ttf|otf]` standard. You can specify a custom directory for this with the `--source` parameter on the `extract` command, but then the structure should still be `/your/custom/fonts/Fontname/Fontname.[ttf|otf]` if the command is `--source /your/custom/fonts/`

Likewise, the SMuFL metadata for each font has been moved to the font directory. It should follow the `fontname_metadata.json` standard. (`fontname` should be lower-case here, e.g., `bravura_metadata.json`)

Where previously you might have called: `python3 generate-smufl-code.py`, you can now call `python3 generate.py smufl`. 

Where previously you might have called: `python3 extract-bounding-boxes.py Bravura` you can now call `python3 generate.py extract Bravura`

Where you might have called `./generate_woff2.sh` you can now call `python3 generate.py woff2 Bravura Bravura/Bravura.otf`

The `./generate_all.sh` script has been updated with the new functions.

---

All of the data has been re-generated with this script, as a test to ensure that it produces no significant differences. As a reviewer, please check that you do not see any signficant differences. The most changes seem to be an additional space added between the last attribute and the self-closing end-tag, as a result of moving to the built-in ElementTree writer.


ahankinson:
NB: After review with @lpugin a few changes to WOFF and SVG generation are incoming.

lpugin (approved):

ahankinson:
@rettinghaus it's probably a good idea to have a look at this, as we think it might have an impact on your workflows!

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+    glyph file containing only the supported SMuFL glyphs.
+
+    :param opts: A set of options from the argument parser sub-command.
+    :return: True if successful, False otherwise.
+    """
+    source_dir: str = opts.source
+    fontname: str = opts.fontname
+    data_dir: str = opts.data
+
+    source_pth: Path = Path(source_dir)
```
rettinghaus, line 134: Is there a specific reason for abbreviating "path" to "pth"? I think one additional character wouldn't harm.

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+
+def extract_fonts(opts: Namespace) -> bool:
+    """
+    Takes a font file and extracts the necessary files for Verovio. Generates a glyph file for each glyph, and an SVG
+    glyph file containing only the supported SMuFL glyphs.
+
+    :param opts: A set of options from the argument parser sub-command.
+    :return: True if successful, False otherwise.
+    """
+    source_dir: str = opts.source
```
rettinghaus, line 130: Why use intermediate variables? Use directly `source_pth: Path = Path(opts.source)` and build subsequent paths with `source_pth / fontname`, etc.

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+        log.error(f"Could not read {str(font_pth)}. Does it exist?")
+        return False
+
+    if not os.access(output_pth, os.W_OK):
+        log.error(f"Could not write to {str(output_pth)}. Check permissions.")
+        return False
+
+    if not glyph_file_pth.is_dir():
+        log.debug("Creating %s", glyph_file_pth)
+        os.makedirs(glyph_file_pth, exist_ok=True)
```
rettinghaus, line 162: Why not use `Path.mkdir`?

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+            fmt_css: str = FONTFACE_WRAPPER.format(
+                fontname=fontname, b64encoding=b64_encoding.decode()
+            )
+            css_content.write(fmt_css)
+
+    if opts.keep_intermediates:
+        intermediate_pth = Path(font_data_pth, "tmp")
+        log.debug("Keeping intermediate files in %s", intermediate_pth.resolve())
+        os.makedirs(intermediate_pth, exist_ok=True)
+        tmp_svg_font.replace(Path(intermediate_pth, f"{fontname}_subset.svg"))
```
rettinghaus, line 259: Again, `intermediate_pth` is as path object, so just use `(intermediate_pth / f"{fontname}_subset.svg")`

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+    SVG version of a given font.
+
+    Base64-encodes the resulting WOFF2 font, and embeds it in a @font-face declaration.
+
+    :param opts: A set of options from the argument parser sub-command.
+    :return: True if successful, False otherwise.
+    """
+    fontname: str = opts.fontname
+    source_dir: str = opts.source
+    source_pth: Path = Path(source_dir)
```
rettinghaus, line 202: See above

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+
+    if not font_data_pth.is_dir() or not os.access(font_data_pth, os.R_OK):
+        log.error(f"Could not read font information from {str(font_data_pth)}")
+        return False
+
+    if not os.access(metadata_pth, os.R_OK):
+        log.error(f"Could not read {str(metadata_pth)}. Does it exist?")
+        return False
+
+    if not os.access(font_pth, os.R_OK):
```
rettinghaus, line 152: Why having all these if statements for checking access. I think it would be more elegant to catch a `PermissionError` in `__read_svg_font_file` and log the error there.

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+
+def generate_smufl(opts: Namespace) -> bool:
+    """
+    Generates the `smufl.h` file for Verovio.
+
+    :param opts: A set of options from the argument parser sub-command.
+    :return: True if successful, False otherwise.
+    """
+    supported_glyphs_fname: str = opts.supported
+    header_out_dir: str = opts.header_out
```
rettinghaus, line 88: See above

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+
+    svg_data = __read_svg_font_file(str(font_pth))
+    if not svg_data:
+        log.error("Could not extract SVG data from %s", font_pth.resolve())
+        return False
+
+    family, units, hax, glyphs = svg_data
+
+    log.debug(
+        "SVG Data extracted: Family: %s, Units: %s, H-A-X: %s, Number of glyphs: %s",
```
rettinghaus, line 172: Why avoid an f-string here?

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+    font-weight: normal;
+    font-style: normal;
+}}"""
+
+
+B64_FONT_LICENSE = """
+This font is licensed under the SIL Open Font License, http://scripts.sil.org/OFL, with 
+Reserved Font Name "{fontname}".
+
+See https://github.com/rism-digital/verovio/blob/develop/fonts/README.md for more information.
```
rettinghaus, line 71: What does the "blob" part in this URL?

rettinghaus (changes requested):
Looks mostly fine. Added some ideas and questions.

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+
+SMUFL_HEADER = """/////////////////////////////////////////////////////////////////////////////
+// Name:        smufl.h
+// Author:      Laurent Pugin
+// Created:     2014-2022
+// Copyright (c) Authors and others. All rights reserved.
+/////////////////////////////////////////////////////////////////////////////
+
+/////////////////////////////////////////////////////////////////////////////
+// NOTE: this file was generated by the ./fonts/generate smufl script
```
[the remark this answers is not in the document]
rettinghaus: Forget this comment.

src/svgdevicecontext.cpp
```diff
@@ -108,13 +108,22 @@ void SvgDeviceContext::Commit(bool xml_declaration)
...
+        std::ifstream cssFontFile(cssFontPath);
+        if (!cssFontFile.is_open()) {
+            LogWarning("The CSS font for '%s' could not be loaded and will not be embedded in the SVG");
+        }
+        else {
+            std::stringstream cssFontStream;
+            cssFontStream << cssFontFile.rdbuf();
+            pugi::xml_node css = m_svgNode.append_child("style");
+            css.append_attribute("type") = "text/css";
+            css.append_child(pugi::node_pcdata).set_value(cssFontStream.str().c_str());
```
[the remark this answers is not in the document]
lpugin: This is a different issue and the current PR does not make any changes in that regard. The font remains included only if there are some music glyphs mixed with text.

fonts/generate.py
ahankinson: In the `log` module, strings using the `%s` statements are lazily evaluated, while `f""` statements are not. So if your log level is `info` then percent-formatted strings do not actually get interpolated in `debug` statements.

See: https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

fonts/generate.py
ahankinson: That's what GitHub URLs look like if you link to a non-branch-specific (e.g., whatever the default branch is) file.

fonts/generate.py
ahankinson: Because there are several different failure modes being checked. `PermissionError` only checks the permission, while os.R_OK checks both permission and existence.

fonts/generate.py
ahankinson: It was an attempt to avoid [a Python anti-pattern](https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html), while at the same time trying to make sure I'm keeping all the different variations of that data separate and well-named, and also trying to keep some semblance of the old code so that I could easily find my way back to it while I was porting it over. (The XML processors, for example, don't take `Path` objects, so I have to use the string variant on them...)

fonts/generate.py
ahankinson: Fixed! Thanks.

fonts/generate.py
ahankinson: Fixed! Thanks.

fonts/generate.py
ahankinson: I just can't bring myself to use a division operator on a string. Sorry! I know it's supported, but I'm not a fan. 😄

fonts/generate.py
ahankinson: Fixed, thanks!

fonts/generate.py
ahankinson: Fixed, thanks!

fonts/generate.py
ahankinson: Pylint will flag this, which is how I learned about it.

fonts/generate.py
```diff
@@ -0,0 +1,657 @@
...
+                a_element.set("x", str(round(anc[0], 2)))
+                a_element.set("y", str(round(anc[1], 2)))
+
+    tree: Et.ElementTree = Et.ElementTree(root)
+    Et.indent(tree)
+    log.debug("Writing SVG file %s", output.resolve())
+    tree.write(str(output), encoding="UTF-8", xml_declaration=True)
+
+
+def __get_alternate_glyphs(glyphs: dict, metadata: dict) -> dict:
```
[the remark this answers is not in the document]
rettinghaus: Forget it, I mixed things up again. However, I'll make a mental note to specify alternates directly in Leipzig so they'll get exported in the JSON.

rettinghaus (approved):

Issue #3005, opened 2022-08-26
Proposal for Functor Classes


DavidBauer1984:
### Motivation
Verovio currently implements functors as virtual methods of the Object class and stores the functor state in struct-like parameter objects (consisting only of public data members plus constructor for initialization). This approach has the following drawbacks:

1. The functor implementation is spread over several element classes. In order to understand what a certain functor does, one must search and inspect the code in several classes. Reversely, element classes are not self-contained and are crowded with functor code. An extreme example is the _Measure class_ which consists of 1700 LOC with 1000 LOC functor implementation.

2. There exists no mechanism for functors to share common functionality. For example, it would make sense to split the current transpose functor into three functors Transpose, TransposeMdiv and TransposeToSoundingPitch which share the core transposition functionality on KeySig, Note and Rest. Similarly, it would make sense to split the MEI export into several functors for page based, score based, filtered score based and MEI basic.

3. The functor state is currently not properly encapsulated. Usually parameters serve as input, output or intermediate values. However, there is no restriction for setting and getting parameter values.

4. The functor state must be passed as argument into the functor implementation and casted afterwards. Even though this does not have a negative impact on runtime (the casts are static), it adds a certain amount of boilerplate code.

5. There is no natural place to implement functor specific helper functions. This either leads to very long functor implementations (i.e. `LayerElement::CalcAlignmentPitchPos` with 260 LOC) or helper methods within an element class which are very functor specific (i.e. `Note::GenerateGraceNoteMIDI`).

### Proposal
These drawbacks could be overcome by implementing functors as classes and using a [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern) to call them. The basic idea is to merge the functor implementation and state into one class for each functor.
There would be an interface class defining virtual methods for each element (i.e. `VisitorInterface::VisitMeasure(Measure *measure)`, `VisitorInterface::VisitNote(Note *note)`, `VisitorInterface::VisitRest(Rest *rest)`, … ) from which all functors are derived.
To transform a functor into a class, the following steps must be done:

1. Create a new class for the functor (say `CastOffPagesFunctor`)
2. Use the content of the param object (`CastOffPagesParams`) to define the member variables (`m_contentPage, m_doc, m_currentPage, …`)
[...]

PR #3032, opened 2022-09-27
Functor classes - getting started

DavidBauer1984:
This PR implements the basic framework for functor classes (see #3005 ) and moves the first functors, namely most of the search functors and `CalcDots` as well as `CalcLedgerLines`.

- We have const functors (class `ConstFunctor`) and mutable functors (class `MutableFunctor`). The latter are allowed to make changes on elements that they visit. Both are derived from `FunctorBase` which stores basic properties such as traversal direction, the current functor code or filters. They are derived from different interfaces (`ConstFunctorInterface` and `FunctorInterface`) which contain the virtual methods for element visitation. For convenience, we have the classes `DocFunctor` and `DocConstFunctor` which additionally store a Doc pointer.
- For now we implement visitation for `Object`, `Doc`, all layer elements and all container elements. This will be further extended in the future.
- The search functors demonstrate the implementation and usage of simple functor classes.
- `CalcDots` demonstrates the implementation of an advanced functor with private helper function.
- `CalcLedgerLines` demonstrates the implementation of an advanced functor that also uses the end interface (formerly end functor). The call of functors now becomes quite neat: 
```cpp
CalcLedgerLines calcLedgerLines(doc);
this->Process(calcLedgerLines);
```
- Changing the code structure already becomes apparent in `note.cpp` which is reduced by ~100 LOC.

This PR should not trigger any changes in performance or rendering.


lpugin:
Thanks! I changed based to `develop-functors`.

I am planning to have a release with 3.12 this week. We can start merging this after that.

DavidBauer1984:
@lpugin Release build size (from Xcode) is now 17.855 MB compared to 17.620 MB before.

src/object.cpp
```diff
@@ -1104,6 +1104,140 @@ void Object::Process(Functor *functor, FunctorParams *functorParams, Functor *en
...
+                if (this->FiltersApply(filters, *iter)) {
+                    (*iter)->Process(functor, deepness);
+                }
+            }
+        }
+    }
+
+    if (functor.ImplementsEndInterface() && !skipFirst) {
+        FunctorCode code = this->VisitEnd(functor);
+        functor.SetCode(code);
```
lpugin, line 1160: Why not setting the `code` directly in `Visit` or `VisitEnd` methods? Then you would do just
```cpp
this->VisitEnd(functor);
```

DavidBauer1984: We would have to set the code in each override of `Visit(End)`, now `Accept(End)`. So we would trade setting it twice in `Process` against setting it 250 times in `Accept`.

DavidBauer1984:
@lpugin Is there anything left to do here?

Issue #2974, opened 2022-07-21
Support for <graphic> element


wergo:
At the moment (3.11.), Verovio does not read or write the `<graphic>` element within `<surface>`, although the other elements inside facsimile group (`<facsimile>`, `<zone>`) are supported. Thus, this element gets lost upon MEI input/output. See discussion here #2859.

It would be great to add support for an element <graphic> to the Verovio codebase (creating graphic.cpp and graphic.h) and add support for input and output of it.

Issue #2859, opened 2022-05-12
Facsimile is not written to MEI 


rettinghaus:
MEI files loose their `facsimile` element (and all children) when they are loaded and saved back.

Minimal input file: 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Facsimile</title>
         </titleStmt>
         <pubStmt>
            <date isodate="2022-05-12">2022-05-12</date>
         </pubStmt>
         <seriesStmt>
            <title>Verovio test suite</title>
         </seriesStmt>
         <notesStmt>
            <annot />
         </notesStmt>
      </fileDesc>
   </meiHead>
   <music>
      <facsimile>
         <surface>
            <graphic />
         </surface>
      </facsimile>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" />
                  </staffGrp>
               </scoreDef>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

wergo:
Verovio writes `<facsimile>` and `<zone>` elements, but still looses `<graphic>` elements upon im-/exporting MEI (using Verovio 3.10.0 that contains #2889).

Example above extended by <zone> elements: 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Facsimile</title>
         </titleStmt>
         <pubStmt>
            <date isodate="2022-05-12">2022-05-12</date>
         </pubStmt>
[...]

PR #3033, opened 2022-09-27
Develop graphic element

wergo:
Adds element `<graphic>` to Verovio, to support import and export of it; addresses #2974 and #2859.

In atts_shared.cpp, the attribute class `ATT_POINTING` (both for setters and getters) had `@target` and `@targettype` incorrectly parsed as `xlink:target` and `xlink:targettype` which was corrected in this PR (see https://music-encoding.org/guidelines/v4/attribute-classes/att.pointing.html). 

Open question: `@width` and `@height` in `<graphic>` currently have `vu` as default units (will be added upon output at the moment), which seems strange for mostly pixel units. See discussion here: https://github.com/music-encoding/music-encoding/issues/603. Hope that this will solved soon with an extra optional `@unit` attribute and both `@width` and `@height` as pure floating values.


rettinghaus:
Looks good, but LibMEI shouldn't be touched manually. I spotted the problem there, fixed it in https://github.com/rism-digital/libmei/pull/21 and regenerated LibMEI in #3034. 
After that is merged, please rebase and revert the changes in `libmei/atts_shared.cpp`.

wergo:
Thanks a lot @rettinghaus! I was not aware about touching libmei when modifying that code. It is reverted and #3034 merged now, so this PR should only include the graphic-related things.

rettinghaus (approved):

lpugin (approved):

wergo:
Thanks for merging! 

Let's fix the width/height vu issue, once MEI is clarified (https://github.com/music-encoding/music-encoding/issues/603).

PR #3034, opened 2022-09-27
Fix wrong namespaces in LibMEI

rettinghaus:
Another update to LibMEI with [fixed namespace propagation](https://github.com/rism-digital/libmei/pull/21/commits/ed1248f767ced2a02a1ba181972ad8381b4da1c2).

### DIFF
--- a/libmei/attclasses.h
+++ b/libmei/attclasses.h
@@ -296,6 +296,6 @@ enum AttClassId {
     ATT_CLASS_max
 };
 
-} // vrv namespace
+} // namespace vrv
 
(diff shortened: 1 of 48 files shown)


lpugin:
Why are you changing this?

rettinghaus:
All other (clang-formatted) files end with `// namespace vrv`, so why should LibMEI be different? 
But the main changes are in `atts_shared` by fixing the wrong namespace prefixes.

lpugin:
I was looking at the namespace issue. I can now see it is corrupted and had to be fixed. Thanks

PR #3036, opened 2022-09-28
Fixed: Add a command for checking a font

### LINKED ISSUE #3035: Check the status of a font
It would be useful to have some feedback from the font generation script about the status of the font in terms of glyphs available. More precisely, what we need to know is the list glyphs that are selected in Verovio but not available in the font.

ahankinson:
The `check [fontname]` command will check a font's full SVG file against the list of supported glyphs, and will report the code of any glyph that is supported by Verovio, but that is not in that font.

Optionally, it can also show the list of glyphs in the font that are not supported by Verovio by passing the `--show-unsupported` flag.

Fixes #3035

### DIFF
--- a/fonts/generate.py
+++ b/fonts/generate.py
@@ -350,7 +350,7 @@ def __combine_alternates_and_supported(opts) -> dict:
         metadata: dict = json.load(jsonfile)
 
     if not os.access(metadata_pth, os.R_OK):
-        log.warning("The metadata file could be read at %s", metadata_pth)
+        log.warning("The metadata file could not be read at %s", metadata_pth)
         return supported_glyphs
 
     alternate_glyphs: dict = __get_alternate_glyphs(supported_glyphs, metadata)
--- a/fonts/generate.py
+++ b/fonts/generate.py
@@ -172,7 +172,6 @@ def extract_fonts(opts: Namespace) -> bool:
     )
 
     supported_glyphs: dict = __combine_alternates_and_supported(opts)
-    metadata_pth: Path = Path(font_data_pth, f"{fontname.lower()}_metadata.json")
 
     with open(metadata_pth, "r") as jfile:
         metadata: dict = json.load(jfile)
--- a/fonts/generate.py
+++ b/fonts/generate.py
@@ -1,5 +1,4 @@
 import base64
-import datetime
 import json
 import logging
 import os
@@ -8,7 +7,7 @@ import subprocess
 import sys
 import tempfile
 import xml.etree.ElementTree as Et
-from argparse import ArgumentParser, Namespace
+from argparse import ArgumentParser, Namespace, RawTextHelpFormatter
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


fonts/generate.py
```diff
@@ -83,21 +82,20 @@ def generate_smufl(opts: Namespace) -> bool:
...
+        log.error("Could not find or read %s file.", supported_pth)
         return False
 
     if not os.access(header_out_pth, os.W_OK):
         log.error("Could not write to %s.", header_out_pth)
         return False
 
     log.debug("SMuFL header will be stored in %s", header_out_pth)
-    supported_glyphs: dict = __get_supported_glyph_codes(supported_glyphs_fname)
+    supported_glyphs: dict = __get_supported_glyph_codes(supported_pth)
```
rettinghaus, line 98: same

rettinghaus (approved):

fonts/generate.py
```diff
@@ -83,21 +82,20 @@ def generate_smufl(opts: Namespace) -> bool:
...
     """
-    supported_glyphs_fname: str = opts.supported
-    supported_pth: Path = Path(supported_glyphs_fname)
+    supported_pth: Path = Path(opts.supported)
     header_out_pth: Path = Path(opts.header_out)
     header_file_pth: Path = Path(header_out_pth, "smufl.h")
 
     if not supported_pth.is_file() or not os.access(supported_pth, os.R_OK):
-        log.error("Could not find or read %s file.", supported_glyphs_fname)
+        log.error("Could not find or read %s file.", supported_pth)
```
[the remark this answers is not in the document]
ahankinson: I want the full path as provided so that users can see where it's looking, since we do some implicit path creation based on font names, so having the full path is useful for checking that the file actually exists. 

`.name` only shows the last component of the path, whereas casting a `Path` to a string (or just printing it) will print the full path.

```
from pathlib import Path

p = Path("/foo/bar/baz.txt")
p.name  #=> 'baz.txt'
str(p)     #=> '/foo/bar/baz.txt'
f"{p}"     #=> '/foo/bar/baz.txt'
```

fonts/generate.py
ahankinson: This would actually cause a problem with `.name` since the full (relative) path is needed, and `.name` would only provide the filename.

fonts/generate.py
```diff
@@ -313,7 +349,11 @@ def __combine_alternates_and_supported(opts) -> dict:
     with open(metadata_pth, "r") as jsonfile:
         metadata: dict = json.load(jsonfile)
 
-    alternate_glyphs = __get_alternate_glyphs(supported_glyphs, metadata)
+    if not os.access(metadata_pth, os.R_OK):
+        log.warning("No alternate glyphs file could be read at %s", metadata_pth)
```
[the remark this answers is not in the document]
ahankinson: Fixed, thanks

fonts/generate.py
```diff
@@ -313,7 +349,11 @@ def __combine_alternates_and_supported(opts) -> dict:
     with open(metadata_pth, "r") as jsonfile:
         metadata: dict = json.load(jsonfile)
 
-    alternate_glyphs = __get_alternate_glyphs(supported_glyphs, metadata)
+    if not os.access(metadata_pth, os.R_OK):
+        log.warning("The metadata file could be read at %s", metadata_pth)
```
lpugin, line 353:
```suggestion
        log.warning("The metadata file could not be read at %s", metadata_pth)
```

Issue #3037, opened 2022-09-28
ToolKit.loadZipDataBuffer() stopped working: 'Module not defined'


wergo:
Starting from release 3.11, the toolkit method `loadZipDataBuffer()` to load compressed musicXml files (.mxl) fails to work. The error message is `"Module is not defined."` This worked fine until including version 3.10 (as can be tested by dragging an .mxl file on https://mei-friend.mdw.ac.at/)

Please see below a command line example with a valid ArrayBuffer and a byteLength (and with the `inputFrom` option set to `xml`):

<img width="900" alt="grafik" src="https://user-images.githubusercontent.com/45632600/192733697-c2571b90-ae3b-47b8-b740-e0b88f3d9473.png">

Same situation with ToolKit 3.10 works just fine:

<img width="650" alt="Bildschirmfoto 2022-09-28 um 10 39 41" src="https://user-images.githubusercontent.com/45632600/192732037-4c356dfb-baaf-4a5a-96fb-c356a5e34966.png">

It seems that with the big changes for Version 3.11 this method broke (see #2937) as it still refers to a variable `Module` that apparently does not exist. @WolfgangDrescher, @rettinghaus, any ideas?

MXL test file: https://iwk.mdw.ac.at/goebl/mei/BeetAnGeSample.mxl

lpugin:
It seems that the problem is with `https://www.verovio.org/javascript/latest/verovio-toolkit-hum.js`. It works fine for me with `https://www.verovio.org/javascript/latest/verovio-toolkit-wasm.js`. Could you confirm?

WolfgangDrescher:
My first thought is that this probably has something to do with `verovio-toolkit-hum.js` because it is a synchronous build compared to the asynchronous `verovio-toolkit-wasm.js` file. I never used `loadZipDataBuffer` before. How can I test this? Do I need an ArrayBuffer for this in any case?

```js
import verovio from 'verovio';

document.addEventListener('DOMContentLoaded', (event) => {
    verovio.module.onRuntimeInitialized = () => {
        const tk = new verovio.toolkit();
        fetch('https://raw.githubusercontent.com/music-encoding/sample-encodings/main/MEI_4.0/Music/Encoding_alternatives/Mozart_Veilchen/Das_Veilchen_0Parameters.mei').then(response => {
            console.log({response});
            return response.arrayBuffer();
        }).then(arrayBuffer => {
            console.log({arrayBuffer});
[...]

PR #3041, opened 2022-09-29
Fix access to global Module variable in loadZipDataBuffer

WolfgangDrescher:
Fixes https://github.com/rism-digital/verovio/issues/3037.

Introduces access to the VerovioModule with a new `module`getter in VerovioToolkit:

```js
export class VerovioToolkit {
    // ...
    get module() {
        return this.VerovioModule;
    }
}

const tk = new VerovioToolkit();
console.log(tk.module);
```

I still get an error when running `loadZipDataBuffer` as described in https://github.com/rism-digital/verovio/issues/3037#issuecomment-1260655896. But this seems to be unrelated to the issue and @wergo confirmed that this fixes the problem.

Bildschirmfoto 2022-09-28 um 11 39 22

### DIFF
--- a/emscripten/npm/src/verovio-toolkit.js
+++ b/emscripten/npm/src/verovio-toolkit.js
@@ -14,6 +14,10 @@ export class VerovioToolkit {
         VerovioToolkit.instances.push(this);
     }
 
+    get module() {
+        return this.VerovioModule;
+    }
+
     destroy() {
         VerovioToolkit.instances.splice(VerovioToolkit.instances.findIndex(i => i.ptr === this.ptr), 1);
         console.debug("Deleting toolkit instance");
@@ -119,10 +123,10 @@ export class VerovioToolkit {
         }
         var dataArray = new Uint8Array(data);
         var dataSize = dataArray.length * dataArray.BYTES_PER_ELEMENT;
-        var dataPtr = Module._malloc(dataSize);
-        Module.HEAPU8.set(dataArray, dataPtr);
+        var dataPtr = this.module._malloc(dataSize);
+        this.module.HEAPU8.set(dataArray, dataPtr);
         var res = this.proxy.loadZipDataBuffer(this.ptr, dataPtr, dataSize);
-        Module._free(dataPtr);
+        this.module._free(dataPtr);
         return res;
     }
 


ahankinson:
@WolfgangDrescher if the code that you posted in https://github.com/rism-digital/verovio/issues/3037#issuecomment-1260655896 is causing you problems, it's probably because you're loading an MEI file, which is not ZIP-compressed. You would need to try it with an `.mxl` file -- a couple URLs have been posted.

(Verovio should probably gracefully handle the case where the data being uncompressed isn't actually compressed, but that's likely a separate issue.)

wergo:
Well, the solution that worked on my machine is this #3042. I have not tested this one.

emscripten/npm/src/verovio-toolkit.js
```diff
@@ -119,10 +123,10 @@ export class VerovioToolkit {
         }
         var dataArray = new Uint8Array(data);
         var dataSize = dataArray.length * dataArray.BYTES_PER_ELEMENT;
-        var dataPtr = Module._malloc(dataSize);
-        Module.HEAPU8.set(dataArray, dataPtr);
+        var dataPtr = this.module._malloc(dataSize);
```
ahankinson, line 126: can't this just be `this.VerovioModule` directly? (also, are you missing the method call parens `()`?)

lpugin:
@WolfgangDrescher do we really need the extra function, or is #3042 OK?

emscripten/npm/src/verovio-toolkit.js
WolfgangDrescher: See my comment below. My syntax is correct. Class getters are accessed as properties in JavaScript and do not need to be called: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

WolfgangDrescher:
> it's probably because you're loading an MEI file, which is not ZIP-compressed. You would need to try it with an `.mxl` file

Thank you for this clarification @ahankinson.

wergo:
I hade the chance to test this PR now and it works great on my code, @WolfgangDrescher. Thanks a lot. I am certainly fine with any variant of the solution.

lpugin:
Using and accessing `VerovioModule` seems to be the clearest and simplest option, so I merged #3042. Thanks for looking into this.

PR #3045, opened 2022-09-30
WIP: Recognize CMAKE_INSTALL_PREFIX

ahankinson:
By setting an absolute path to the install directory in the CMakeLists.txt file, the cmake build was ignoring the CMAKE_INSTALL_PREFIX.  By changing it to a relative path the CMAKE_INSTALL_PREFIX seems to work as expected.

These changes make it possible to adapt the installation of Verovio to different OS filesystem layouts. (e.g., homebrew on Mac uses /opt/homebrew instead of /usr/local now). 

In addition, the path to the resources directory was changed to work with CMAKE_INSTALL_PREFIX. This necessitated a change in the source code, since the value was previously hard-coded. (Please review this bit -- my C++ `#defines` are pretty bad!)

This was tested and seemed to work, but would benefit from wider testing.

```
$ mkdir -p /tmp/verovio
$ cd /path/to/verovio/tools
$ cmake ../cmake -DCMAKE_INSTALL_PREFIX=/tmp/verovio  (also possible: cmake ../cmake --install-prefix=/tmp/verovio)
$ make -j 8
$ make install
```

The executable should now be in `/tmp/verovio/bin` and the resources in `/tmp/verovio/share`.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -19,11 +19,6 @@
 #include "att.h"
 #include "vrv.h"
 
-#ifdef RESOURCE_DIR
-#define VRV_RESOURCE_DIR RESOURCE_DIR
-#else
-#define VRV_RESOURCE_DIR "/usr/local/share/verovio"
-#endif
 
 namespace vrv {
 
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -19,7 +19,6 @@
 #include "att.h"
 #include "vrv.h"
 
-
 namespace vrv {
 
 const std::map<int, std::string> Option::s_breaks = { { BREAKS_none, "none" }, { BREAKS_auto, "auto" },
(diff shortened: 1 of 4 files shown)


src/options.cpp
```diff
@@ -19,6 +19,12 @@
 #include "att.h"
 #include "vrv.h"
 
+#ifdef RESOURCE_DIR
+#define VRV_RESOURCE_DIR RESOURCE_DIR
+#else
+#define VRV_RESOURCE_DIR "/usr/local/share/verovio"
+#endif
```
lpugin, line 26: This should go in `vrvdef.h`.

ahankinson:
I just tested the Python and emscripten builds, and they seem unchanged. So i think I'll mark this as ready.

ahankinson:
The XCode build seems to work ok too.

PR #3054, opened 2022-10-05
fix midi for lyrics on chords

rettinghaus:
small fix for midi output with lyrics in chords

### DIFF
--- a/src/verse.cpp
+++ b/src/verse.cpp
@@ -251,9 +251,10 @@ int Verse::InitProcessingLists(FunctorParams *functorParams)
 
 int Verse::GenerateMIDI(FunctorParams *)
 {
-    Note *note = vrv_cast<Note *>(this->GetFirstAncestor(NOTE));
-    assert(note);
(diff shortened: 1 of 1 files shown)


src/verse.cpp
```diff
@@ -251,9 +251,10 @@ int Verse::InitProcessingLists(FunctorParams *functorParams)
 
 int Verse::GenerateMIDI(FunctorParams *)
 {
-    Note *note = vrv_cast<Note *>(this->GetFirstAncestor(NOTE));
-    assert(note);
-    Verse *previousVerse = vrv_cast<Verse *>(note->GetPrevious(this, VERSE));
+    LayerElement *parent = vrv_cast<Note *>(this->GetFirstAncestor(NOTE));
+    if (!parent) parent = vrv_cast<Chord *>(this->GetFirstAncestor(CHORD));
+
```
lpugin, line 256: You still want to assert parent before accessing it

PR #3060, opened 2022-10-06
add git_commit only in dev

### LINKED ISSUE #3059: No git commit available when building from zipped sources
When downloading a release from the release pages, GitHub strips the `.git` directory out. Then, when building from these sources cmake gives a warning: 

`fatal: not a git repository (or any of the parent directories): .git`

This results in a version string of: 

```
$ verovio --version
Verovio 3.12.1-[undefined]
```

To fix this, either the git revision check should be skipped in the make process if it's not a git repo, and the code adjusted (the `GIT_COMMIT` variable), or the version printing string should be modified to omit the trailing revision if the `GIT_COMMIT` is set to `[undefined]` -- or some combination of the two.

rettinghaus:
Closes #3059

### DIFF
--- a/src/vrv.cpp
+++ b/src/vrv.cpp
@@ -264,8 +264,8 @@ std::string GetFilename(std::string &fullpath)
 std::string GetVersion()
 {
     std::string dev;
-    if (VERSION_DEV) dev = "-dev";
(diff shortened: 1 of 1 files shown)


ahankinson:
It looks OK to me! Is it a problem with building release code from the latest git checkout, though? Probably a question for @lpugin

lpugin:
Implemented differently with #3062

PR #3069, opened 2022-10-11
Fix repeated duration with dot in PAE output

carlolic:
The generated PAE was not as expected in case of notes with dotted duration. In particular, the dotted duration was output only for the first note and not for all notes before duration change. Example: `4.AAA` -> `4.AAA` instead of `4.A4.A4.A` (in case of no dots, the output was as expected: `4AAA` -> `4A4A4A`).
I removed the if-statement checking whether duration or number of dots has changed since, as far as I understand, we want to output duration after each note. The code worked for non-dotted duration only because in this case `interface->GetDots()` is not assigned, whereas `m_currentDots` is 0 or -1.

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -539,34 +539,32 @@ void PAEOutput::WriteDur(DurationInterface *interface)
 {
     assert(interface);
 
-    if ((interface->GetDur() != m_currentDur) || (interface->GetDots() != m_currentDots)) {
-        m_currentDur = interface->GetDur();
-        m_currentDots = (interface->HasDots()) ? interface->GetDots() : 0;
-        std::string dur;
-        switch (m_currentDur) {
-            case (DURATION_long): dur = "0"; break;
-            case (DURATION_breve): dur = "9"; break;
-            case (DURATION_1): dur = "1"; break;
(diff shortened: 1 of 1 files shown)


lpugin:
We certainly need to fix it but maybe not like that. Looking at the code it seems that the intended output was not to have the duration for each note. However, the problem was that because the dots would be unset with un-dotted notes, the comparison with the dot current value would always be false. This is the reason why we have value for each note without dots.

You need to fix it by taking into account that `GetDots` returns `VRV_UNSET` when there is no dots. So you want to check with `HasDots` first.

carlolic:
Hi @lpugin, I'm not sure I understand. Do we want to have the duration of each note or not? If yes, I think I should remove the if-statement, since we wouldn't care if duration or number of dots has changed.

lpugin:
>  Do we want to have the duration of each note or not?

No, otherwise there would be no `if` statement, as you just said. So having the value for every note was a bug resulting from a comparison with VRV_UNSET in all cases where notes have no dots.

carlolic:
OK! I assumed the normalized PAE should have repeated duration, but it is just the other way around. I'm closing this PR and will open a new one.

PR #3070, opened 2022-10-11
Minor improvements to script

rettinghaus:
This PR clarifies that we're actually create a bounding-box file and not an svg.

### DIFF
--- a/fonts/generate.py
+++ b/fonts/generate.py
@@ -118,7 +118,7 @@ def generate_smufl(opts: Namespace) -> bool:
 def extract_fonts(opts: Namespace) -> bool:
     """
     Takes a font file and extracts the necessary files for Verovio.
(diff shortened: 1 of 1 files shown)


fonts/generate.py
```diff
@@ -118,7 +118,7 @@ def generate_smufl(opts: Namespace) -> bool:
 def extract_fonts(opts: Namespace) -> bool:
     """
     Takes a font file and extracts the necessary files for Verovio.
-    Generates a glyph file for each glyph,and an SVG glyph file
+    Generates a glyph file for each glyph, and an SVG glyph file
```
ahankinson, line 121: This comma is not needed.

craigsapp: Since there is no verb after the `and`, it is not an independent clause, so no comma is needed (or allowed by English teachers).

PR #3073, opened 2022-10-12
Bug fix: repeated duration in PAE output

carlolic:
Note duration should be shown in the PAE output only if the duration has changed.
Bug description: in case of duration without dots, duration is repeated for each note, even if it has not changed. 
Example: `4A4CE` -> `4A4C4E` but it should be `4ACE`

### DIFF
--- a/src/iopae.cpp
+++ b/src/iopae.cpp
@@ -538,7 +538,7 @@ void PAEOutput::WriteTupletEnd(Tuplet *tuplet)
 void PAEOutput::WriteDur(DurationInterface *interface)
 {
     assert(interface);
- 
+
     const int ndots = (interface->HasDots()) ? interface->GetDots() : 0;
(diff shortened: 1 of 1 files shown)


src/iopae.cpp
```diff
@@ -539,9 +539,10 @@ void PAEOutput::WriteDur(DurationInterface *interface)
 {
     assert(interface);
 
-    if ((interface->GetDur() != m_currentDur) || (interface->GetDots() != m_currentDots)) {
+    int m_dots = (interface->HasDots()) ? interface->GetDots() : 0;
```
lpugin, line 542: local variables are not prefixed with `m_`

rettinghaus: (And in cases like these we usually make them const.)

rettinghaus (approved):

PR #3077, opened 2022-10-13
Refactor JSON options in Python

lpugin:
This branch refactors the way parameters passed to methods and values returned in JSON format are handled.

Instead of passing and returning them as string they are now Python dictionaries. This means we can do
```python
Python 3.9.13 (main, May 24 2022, 21:13:51) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import verovio
>>> tk = verovio.toolkit(False)
>>> tk.getDefaultOptions().get('scale')
100
>>> tk.getDefaultOptions().get('mmOutput')
False
>>> tk.setOptions({'mmOutput': True})
True
>>> tk.getOptions().get('mmOutput')
True
>>> 
```


lpugin:
We would need all methods having a JSON string parameter or returning a JSON string to be adjusted in the same way. Help welcome!

This will break compatibility with previous Python version of the toolkit, and we cannot really deprecate them before enabling them.

lpugin:
An additional change with the PR is now that `getOptions(bool)` is no longer available. Instead we have
* `getOptions()` for the current values
* `getDefaultOptions()` for the default values

There is a deprecation warning but only for the JS version

rettinghaus (approved):
Looks great. Thanks!

PR #3087, opened 2022-10-21
Fix Xcode build with humdrum scheme

DavidBauer1984:
This PR properly adds `graphic.h/cpp` to the Verovio Xcode project.


rettinghaus:
Seems to address the same as #3085

ahankinson:
Yes, I will merge #3085 and close this one.

ahankinson:
They're basically the same except for ` settings = {ATTRIBUTES = (Public, ); }; };` -- is this necessary?

DavidBauer1984:
> They're basically the same except for ` settings = {ATTRIBUTES = (Public, ); }; };` -- is this necessary?

No idea, but we use this setting for all other header files. So I usually include it, to keep things consistent.

rettinghaus:
I just noticed that the header file is in the wrong place. I'll open another PR in some minutes.

Issue #3122, opened 2022-11-15
Add log level controls 


ahankinson:
Currently Verovio can enable or disable logging, but there are no other fine-grained controls for silencing specific log levels. So you cannot currently silence "Info" messages and only see "error" messages.

There are some global functions currently available for logging. These would need to be modified by including an enum to configure the logging. There should be four levels: error, warning, info, and debug. Each level includes the previous, so if you set "warning" you would get "warning" and "error"; if you choose "info" you would get warning, error, and info, etc.

In the JS toolkit we would need to expose a new global function to configure this behaviour. The built-in logging is already available in the C++ code (it just needs to be modified to support levels), so it needs to be exported in the various emscripten proxies, etc. Maybe @WolfgangDrescher has an idea of how this could work?

Related to this, I'm going to be submitting a PR to make a minor change to the JS toolkit to remove the `console.debug` statements there for the constructor / destructor. This should prevent the toolkit from emitting debug messages in production code.

lpugin:
For the JS toolkit I added the functions to the C wrapper and exposed them in the emscripten build. They still needs to be made accessible from the module
https://github.com/rism-digital/verovio/commits/develop-js-logging
@WolfgangDrescher could you have a look? Thanks

WolfgangDrescher:
> The built-in logging is already available in the C++ code (it just needs to be modified to support levels), so it needs to be exported in the various emscripten proxies, etc.

Are you speaking about the proxies in `emscripten/npm/src/emscripten-proxy.js`? lpugin already added a method there.
Will it be enough to make this function accessible via `VerovioToolkit`? Or is any kind of subscriber needed that listens to the logs of the module and then prints them in the console? Or will the emscripten module print them to the console automatically when `enableLog` is activated?

How would this logger be activated? E.g. like this?

```js
const LOG_LEVEL = 'debug';
verovio.module.onRuntimeInitialized = () => {
    const tk = new verovio.toolkit();
    tk.enableLog(LOG_LEVEL);
}
```

ahankinson:
I think it might be something like:

```
const LOG_LEVEL = 'debug';
verovio.module.onRuntimeInitialized = () => {
    verovio.enableLog(LOG_LEVEL);
    const tk = new verovio.toolkit();
}
```

But maybe @lpugin can confirm?

lpugin:
[...]

PR #3125, opened 2022-11-17
Develop js logging

WolfgangDrescher:
Implements https://github.com/rism-digital/verovio/issues/3122.

Setting `VerovioModule` as second parameter and a wrapper function with the default value of `DefaultVerovioModule` in `prebundle-module.js` allows us to use `verovio.enableLog(verovio.logLevel.error)` without the need of passing the module as an argument. But at the same time we can use this method in ESM:

verovio-toolkit-wasm.js:

```html
<script src="verovio-toolkit-wasm.js"></script>
<script>
verovio.module.onRuntimeInitialized = () => {
    verovio.enableLog(verovio.logLevel.error);
}
</script>
```

ESM:

```js
import createVerovioModule from 'verovio/wasm';
import { logLevel, enableLog } from 'verovio/esm';

createVerovioModule().then(VerovioModule => {
    enableLog(logLevel.error, VerovioModule);
});
```

I'm not exposing `enableLog` in the VerovioToolkit because this should enabled per module one not per toolkit instance, is this correct? The downside of this is that we have to create a second EmscriptenProxy and call the constructor to get the pointer:

```js
export function enableLog(level = logLevel.debug, VerovioModule) {
    const proxy = createEmscriptenProxy(VerovioModule);
    const ptr = proxy.constructor();
    proxy.enableLog(ptr, level);
}
```

Could this be prevented by adding it to the toolkit and then calling is per toolkit instance and not per module?

Currently there is no difference when I test this. It does not matter what log level I pass or if I call `enableLog` at all. Is this because it is not yet implemented on the C++ side or is there something wrong on my side?

I set `logLevel.debug` as default value for the parameter so you can use `enableLog()` to print all logs.

### DIFF
--- a/emscripten/npm/src/verovio-logger.js
+++ b/emscripten/npm/src/verovio-logger.js
@@ -8,7 +8,7 @@ export const logLevel = {
     error: 4,
 };
 
-export function enableLog(level, VerovioModule) {
+export function enableLog(level = logLevel.debug, VerovioModule) {
     const proxy = createEmscriptenProxy(VerovioModule);
     const ptr = proxy.constructor();
     proxy.enableLog(ptr, level);
--- a/emscripten/npm/src/verovio-logger.js
+++ b/emscripten/npm/src/verovio-logger.js
@@ -1,0 +1,15 @@
+import { createEmscriptenProxy } from "./emscripten-proxy.js";
+
+export const logLevel = {
+    off: 0,
+    debug: 1,
+    info: 2,
+    warning: 3,
+    error: 4,
+};
+
+export function enableLog(level, VerovioModule) {
+    const proxy = createEmscriptenProxy(VerovioModule);
+    const ptr = proxy.constructor();
+    proxy.enableLog(ptr, level);
+}
--- a/emscripten/npm/src/index.js
+++ b/emscripten/npm/src/index.js
@@ -1,5 +1,8 @@
 import { VerovioToolkit } from "./verovio-toolkit.js";
+import { logLevel, enableLog } from "./verovio-logger.js";
 
 export {
     VerovioToolkit,
+    logLevel,
+    enableLog,
 };
--- a/emscripten/npm/src/prebundle-module.js
+++ b/emscripten/npm/src/prebundle-module.js
@@ -1,10 +1,15 @@
 import DefaultVerovioModule from "../../build/verovio.js";
 import { VerovioToolkit } from "./verovio-toolkit.js";
+import { logLevel, enableLog } from "./verovio-logger.js";
 
 class VerovioToolkitDefaultModule extends VerovioToolkit {
     constructor(VerovioModule = DefaultVerovioModule) {
         super(VerovioModule);
     }
+}
+
+function enableLogDefaultModule(level, VerovioModule = DefaultVerovioModule) {
+    enableLog(level, VerovioModule);
 }
 
 // Assign Module to window to prevent breaking changes.
@@ -16,4 +21,6 @@ if (typeof window !== "undefined") {
 export default {
     module: DefaultVerovioModule,
     toolkit: VerovioToolkitDefaultModule,
+    logLevel,
+    enableLog: enableLogDefaultModule,
 };
--- a/emscripten/npm/src/prebundle-module.js
+++ b/emscripten/npm/src/prebundle-module.js
@@ -8,7 +8,7 @@ class VerovioToolkitDefaultModule extends VerovioToolkit {
     }
 }
 
-function enableLogDefaultModule(level, VerovioModule = DefaultVerovioModule) {
+function enableLogDefaultModule(level = logLevel.debug, VerovioModule = DefaultVerovioModule) {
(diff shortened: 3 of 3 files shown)


lpugin:
Thanks. In between I had try to move the function directly in the pre-bundle step. See [here](https://github.com/rism-digital/verovio/commit/76b9a9d42cfeee42a141848a763108df4933d57c)

That works for non ESM uses, and it looks more correct not to have the function in the mapping. Could you see if this approach can also work for ESM?

WolfgangDrescher:
> Thanks. In between I had try to move the function directly in the pre-bundle step. See here

Good input, I refactored this pull request with your solution.

> That works for non ESM uses, and it looks more correct not to have the function in the mapping. Could you see if this approach can also work for ESM?

No, for ESM we will need a wrapper function that in any case. There is no global module available. You have to import it manually in any case and therefore you need to pass it to the logging wrapper function to enable it.

I think it's best to keep the wrapper function with the module as second parameter and just provide a default value for `verovio-toolkit-wasm.js`.

Please ping me when the implementation of the new logging system is ready. Currently I cannot test this. ESM is still untested by me, but looks good so far.

emscripten/npm/src/verovio-logger.js
```diff
@@ -0,0 +1,15 @@
+export const logLevel = {
```
lpugin, line 1: With this [PR](https://github.com/rism-digital/verovio/pull/3132), the resutling calls will be:

C++:
```cpp
vrv::EnableLog(vrv::LOG_ERROR);
```
Python
```python
import verovio
verovio.enableLog(verovio.LOG_ERROR)
```

If we want to be inline with this and be able to do

JS
```js
verovio.enableLog(verovio.LOG_ERROR);
```
I assume we should do
```
export const LOG_OFF = 0;
export const LOG_ERROR = 1;
export const LOG_WARNING = 2;
export const LOG_INFO = 3;
export const LOG_DEBUG = 4;
```
And then (in `./src/index.js` and './prebundle-module.js`)
```
export default {
    LOG_OFF,
    LOG_ERROR,
    ...
}
```
It is not great to have to set each value explicitly, but I don't think we have another option.

WolfgangDrescher: I would prefer the wrapping `logLevel` object. But of course I will refactor this accordingly.

What are the correct `int` values for these constants? I'm not a C++ developer so im not sure about this line: https://github.com/rism-digital/verovio/blob/57a7a4b778e275eace0d5d7a6eb9f6f33c11c679/include/vrv/toolkitdef.h#L32

```js
export const LOG_OFF = 0;
export const LOG_ERROR = 1;
export const LOG_WARNING = 2;
export const LOG_INFO = 3;
export const LOG_DEBUG = 4;
```

or what @ahankinson suggested in https://github.com/rism-digital/verovio/issues/3122#issuecomment-1315814132

```js
export const LOG_OFF = 0;
export const LOG_DEBUG = 1;
export const LOG_INFO = 2;
export const LOG_WARNING = 3;
export const LOG_ERROR = 4;
```

WolfgangDrescher: And `EnableLogToBuffer` will still accept a `bool` and not the same `LogLevel` enum as `EnableLog`. Correct?

https://github.com/rism-digital/verovio/blob/57a7a4b778e275eace0d5d7a6eb9f6f33c11c679/src/vrv.cpp#L194-L197

WolfgangDrescher: Okay, I pushed to change this (untested).

emscripten/npm/src/verovio-logger.js
```diff
@@ -1,15 +1,13 @@
...
-    error: 4,
-};
+export const LOG_OFF = 0;
+export const LOG_ERROR = 1;
+export const LOG_WARNING = 2;
+export const LOG_INFO = 3;
+export const LOG_DEBUG = 4;
 
-export function enableLog(level = logLevel.debug, VerovioModule) {
+export function enableLog(level = LOG_DEBUG, VerovioModule) {
```
lpugin, line 7: Do we need a default here? It would be better not to have one. If we need set it to `LOG_WARNING`

WolfgangDrescher: > I set logLevel.debug as default value for the parameter so you can use enableLog() to print all logs.

We don't need this. But it could be a handy shortcut. Should I remove it to make it consistent with C++ and Python?

I could throw an error instead if no level is passed. Since logging by the emscripten module will be disabled until you set `enableLog` this could be a good idea.

lpugin: Yes, remove it. We can throw an error if you think this is appropriate. (Logging is enabled by default)

WolfgangDrescher: Okay, changed and pushed it. I will have a look if we need to throw there once this is testable.

lpugin (approved):
Thanks!

PR #3126, opened 2022-11-17
Regenerate fonts

DavidBauer1984:
After #3124 fonts should be regenerated. Otherwise editorial dynamics get lost:

| Before | After |
| ------ | ----- |
| Before | After |


rettinghaus:
I thought that I did, but probably I forgot to push the css files. Sorry.

PR #3133, opened 2022-11-22
Improve page rescaling

Monceber:
When `--scale` and `--scale-to-page-size` are used together, sometimes resulting width/height are not equal to the original values.
This PR addresses that problem. Additionally, when `px` are used for size, it's rounded up to make sure that we have size in exact number of pixels.

### DIFF
--- a/include/vrv/devicecontext.h
+++ b/include/vrv/devicecontext.h
@@ -113,11 +113,17 @@ public:
         m_userScaleX = scaleX;
         m_userScaleY = scaleY;
     }
+    void SetBaseSize(int width, int height)
+    {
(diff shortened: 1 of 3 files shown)


include/vrv/devicecontext.h
```diff
@@ -331,6 +337,10 @@ class DeviceContext {
     int m_width;
     int m_height;
 
+    /** stores base width and height of the device context before application of scale */
+    int m_baseWidth = 0;
+    int m_baseHeight = 0;
```
lpugin, line 342: Initialize in constructor (for consistency)

PR #3139, opened 2022-11-23
Fix font generation

rettinghaus:
This PR improves the font generation script to enable starting with a clean output. 

Checking if it is possible to write to the bounding box file before creating the directory doesn't make sense. With that it is not possible to remove this file and start the generation new. (Also there is no check in place for the css file.)

Also `exist_ok` doesn't make much sense because we should never get to the line if it exists. Instead, using `parents=True` makes sure you can delete the whole `data` folder.

### DIFF
--- a/fonts/generate.py
+++ b/fonts/generate.py
@@ -148,13 +148,9 @@ def extract_fonts(opts: Namespace) -> bool:
         log.error("Could not read %s. Does it exist?", font_pth)
         return False
 
-    if not os.access(output_pth, os.W_OK):
-        log.error("Could not write to %s. Check permissions.", output_pth)
-        return False
-
     if not glyph_file_pth.is_dir():
         log.debug("Creating %s", glyph_file_pth)
-        glyph_file_pth.mkdir(exist_ok=True)
+        glyph_file_pth.mkdir(parents=True)
 
     svg_data = __read_svg_font_file(str(font_pth))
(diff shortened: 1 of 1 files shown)


fonts/generate.py
```diff
@@ -148,13 +148,9 @@ def extract_fonts(opts: Namespace) -> bool:
         log.error("Could not read %s. Does it exist?", font_pth)
         return False
 
-    if not os.access(output_pth, os.W_OK):
```
ahankinson, line 151: I think it's actually supposed to be checking if `data_pth` exists. So it's incorrect to check `output_pth`, but we still should check if `data_pth` exists (which I think we don't do... )

ahankinson:
I'm not sure about:

> makes sure you can delete the whole data folder.

The `data` folder is shipped with Verovio -- you should be able to regenerate some things in it, but I don't think you would ever delete it?

fonts/generate.py
rettinghaus: Then `os.W_OK` is also wrong?

ahankinson: We need to be able to write to the `data_pth`, don't we?

lpugin:
There is no reason for the script to delete `data` entirely because some of its content is not generated by it.

ahankinson (approved):

Issue #3121, opened 2022-11-14
Conversion from Humdrum to MEI is producing meterSig@unit=0 sometimes


gregchapman-dev:
**Describe the problem**
A Humdrum file with e.g. "*M2/0" as a time signature is (sometimes; I think only if there is also a *met present) turned into an MEI file with a `<meterSig>` with `@count=2` and `@unit=0`.  This should be `@count=4` and `@unit=1`.

The following code is the only place (of several) I could find that has this bug:

```
            // But always need to provide @meter.unit since timestamps
            // are in reference to it (can't add meter.count since
            // this will also print a time signature.
            // Count now allowed (suppress time signature display with @form="invis").
            unit = stoi(matches[2]);
            MeterSig *vrvmetersig = getMeterSig(element);
            vrvmetersig->SetForm(METERFORM_invis);
            vrvmetersig->SetCount({ { std::stoi(matches[1]) }, MeterCountSign::None });
            vrvmetersig->SetUnit(unit);
```

... and should be modified as follows to match the other places ...

```
            // But always need to provide @meter.unit since timestamps
            // are in reference to it (can't add meter.count since
            // this will also print a time signature.
            // Count now allowed (suppress time signature display with @form="invis").
            unit = stoi(matches[2]);
            if (unit == 0) {
                count *= 2;
                unit = 1;
            }
            MeterSig *vrvmetersig = getMeterSig(element);
            vrvmetersig->SetForm(METERFORM_invis);
            vrvmetersig->SetCount({ { count }, MeterCountSign::None });
            vrvmetersig->SetUnit(unit);
```

craigsapp:
I will fix the case you point out.

Background:

A time signature of "2/0" means a measure of two breves (double whole notes).

`meterSig@unit` is currently an integer, which causes problems representing breves as the beat in time signatures.   Ideally the text `breve` and `long` would be added to `meterSig@unit`  (which seems best and is similar to [note@dur](https://music-encoding.org/guidelines/v4/data-types/data.duration.cmn.html)), or as a fraction 0.5, or as in Humdrum as "0" ("00" for `long`, which would not work as an integer).

Currently I convert time signatures such as "2/0" (two breve beats per measure) into "4/1" (four semibreve (whole note) beats per measure), which is mathematically correct but not semantically correct.

gregchapman-dev:
There's a fix for this in PR #3150.

craigsapp:
Test case demonstrating the problem:

```tsv
**kern
[...]

PR #3152, opened 2022-12-07
One more place requires the "timeSig with zero denominator" fix.

gregchapman-dev:
This fixes issue #3121 more completely.


gregchapman-dev:
Closed because I didn't mean it to be merged directly to develop.  See PR 3153 instead, which merges to humdrum-develop.

PR #3164, opened 2022-12-15
Humdrum import: Handle clef changes in the middle of a rest

gregchapman-dev:
Clef change in the middle of a space was already handled.  In the case of a (visible) rest, we need to keep the visual duration of the original rest, set it's gestural duration to the pre-clef duration, and add a space after the clef containing the remainder of the duration (i.e. the post-clef duration).


gregchapman-dev:
Shoot, I did it again.  Ignore this, use the next PR I write instead.

Issue #3158, opened 2022-12-09
Handle Python path objects


rettinghaus:
Instead of having filenames and paths as string in Python it is more common to handle these as path objects. Currently Verovio only handles strings, so one has to cast the path object to string. 

It would be good if Verovio could accept path objects (and simply casts them to strings):

```python
from pathlib import Path
import verovio

file = Path("path-to-mei-file")
tk = verovio.toolkit()
tk.loadFile(file)
```

lpugin:
From the experience we have with the json conversion in the Python binding, there is an overhead when adjusting the binding and I think we should modify it as little as possible. So I am not convinced this would make a big difference for the user, at least not big enough to justify the hurdles.

PR #3169, opened 2022-12-20
Improve python toolkit

rettinghaus:
This PR brings some improvements and fixes to the Python toolkit.

#### Improvements:
* the discouraged use of `pkg_resources ` is replaced with [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources)
* the overloaded functions have now docstrings and type hints, presenting more information on functions:

```python
>>> tk.getTimesForElement.__doc__
'Return a dictionary with the following key values for a given note.'
>>> tk.getTimesForElement.__annotations__
{'xml_id': <class 'str'>, 'return': <class 'dict'>}
``` 

* toolkit logging has been unified for retiring an empty object
* `renderToTimemapFile` and `saveFile` cast `filename` to str, allowing to pass path objects here (see #3158)
* functions that are not overloaded get [auto-generated docstrings](https://www.swig.org/Doc4.1/Python.html#Python_nn67)

```python
>>> tk.getNotatedIdForElement.__doc__
'getNotatedIdForElement(toolkit self, std::string const & xmlId) -> std::string'
``` 

#### Fixes:
* `getMIDIValuesForElement` led to a crash, because instead of a string/json an int was returned
*  wrongly overloaded functions were removed (there is no `vrv::Toolkit::RenderToMIDI(const std::string &)`)

**NB**: calling `editInfo` directly after loading a file leads to a crash.


.github/workflows/python-ci-wheel.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install twine
+          python -m pip install importlib_resources twine
```
ahankinson, line 70: The latest version that we currently "support" is 3.7, and `importlib.resources` seems to be available there?

bindings/python/verovio.i
```diff
@@ -27,142 +29,140 @@
 // Toolkit::Edit
 %feature("shadow") vrv::Toolkit::Edit( const std::string & ) %{
 def edit(toolkit, editor_action):
+    """Edit the MEI data."""
```
ahankinson, line 32: Do these need to be hand-edited? If the doc for the C++ is updated, will it be updated here too, or are there two places?

bindings/python/verovio.i
```diff
@@ -27,142 +29,140 @@
...
+    """Return a vector of ID strings of all elements (the notated and the expanded) for a given element."""
     return json.loads($action(toolkit, xml_id))
 %}
 
 // Toolkit::GetMEI
 %feature("shadow") vrv::Toolkit::GetMEI(const std::string & = "") %{
-def getMEI(toolkit, options = None):
-    if options == None:
-        options = {}
+def getMEI(toolkit, options: dict = {}) -> str:
```
ahankinson, line 87: This is incorrect. Mutable structures should not be used as default arguments.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/toolkit.cpp
```diff
@@ -839,7 +839,7 @@ std::string Toolkit::GetMEI(const std::string &jsonOptions)
 
     if (this->GetPageCount() == 0) {
         LogWarning("No data loaded");
-        return "";
+        return {};
```
ahankinson, line 842: doesn't this method return a string?

src/toolkit.cpp
```diff
@@ -1165,7 +1165,7 @@ std::string Toolkit::GetElementAttr(const std::string &xmlId)
     }
     // If not found at all
     if (!element) {
-        LogInfo("Element with id '%s' could not be found", xmlId.c_str());
+        LogInfo("Element '%s' not found", xmlId.c_str());
```
ahankinson, line 1168: The original message seems to be a bit clearer?

src/toolkit.cpp
```diff
@@ -1704,7 +1705,7 @@ std::string Toolkit::GetTimesForElement(const std::string &xmlId)
     jsonxx::Object o;
 
     if (!element) {
-        LogWarning("Element '%s' not found", xmlId.c_str());
+        LogInfo("Element '%s' not found", xmlId.c_str());
```
ahankinson, line 1708: Why was this log level changed?

.github/workflows/python-ci-wheel.yml
ahankinson: Oh, sorry -- I see that it's installing it in the CI...

bindings/python/verovio.i
```diff
@@ -27,142 +29,140 @@
...
-    return json.loads($action(toolkit, xml_id))
-%}
-
-// Toolkit::GetNotatedIdForElement
-%feature("shadow") vrv::Toolkit::GetNotatedIdForElement(const std::string &) %{
-def getNotatedIdForElement(toolkit, xml_id):
+def getMIDIValuesForElement(toolkit, xml_id: str) -> dict:
+    """Return MIDI values of the element with the ID (xml:id)."""
     return json.loads($action(toolkit, xml_id))
 %}
```
lpugin, line 97: Why is this one removed?

rettinghaus: It returns a single id, not a dict.

src/toolkit.cpp
rettinghaus: It does.

https://stackoverflow.com/questions/26587110/returning-an-empty-string-efficient-way-in-c

src/toolkit.cpp
rettinghaus: Why should this message be different than all the others?

src/toolkit.cpp
ahankinson: ok!

bindings/python/__init__.py
```diff
@@ -1,5 +1,8 @@
...
+try:
+    from importlib.resources import files
+except ImportError:
+    from importlib_resources import files
 
 from .verovio import *
 
-verovio.setDefaultResourcePath(resource_filename('verovio', 'data'))
\ No newline at end of file
+verovio.setDefaultResourcePath(str(files("verovio") / "data"))
```
ahankinson, line 8: Urghle... overloaded division operators as path separators. That's the smelliest part of Python IMO.

src/toolkit.cpp
ahankinson: Why can't it be different? Unless you're filtering on that specific string, does it matter if they're different?

src/toolkit.cpp
rettinghaus: Because it returns an empty json. To me this looks awkward. When an unknown id produces a warning I'd expect that nothing is returned. But then we need to handle it differently.

bindings/python/verovio.i
rettinghaus: Using equality operators here hurts even more. But I'll change it, thanks.

bindings/python/verovio.i
rettinghaus: Unfortunately I don't see another way as copying them here manually.

.github/workflows/python-ci-wheel.yml
rettinghaus: The files package is available from 3.9 onwards: https://docs.python.org/3.9/library/importlib.html#module-importlib.resources

src/toolkit.cpp
lpugin: Since there were quite a few issues with strings returned and the JavaScript binding I would prefer not to change this. Thanks.

rettinghaus: Do we need to generate a c_string or can we use the faster `return std::string();`?

lpugin: The speed difference is insignificant. Just leave `return "";`

src/toolkit.cpp
lpugin: Since you made it a warning line 1642 it would probably make sense to keep it consistent and to leave it a warning. Furthermore, since the default log level is now `warning`, then it is probably wise to have a message logged by default when the element is not found.

bindings/python/verovio.i
ahankinson: You’re correct, though — it should be `is None` instead of `== None`

lpugin: > it should be is None instead of == None

That was me... ;-)

bindings/python/verovio.i
rettinghaus: Turns out swig is able to use doxygen for generating docstring (although only for functions that are not overloaded): https://www.swig.org/Doc4.0/Doxygen.html#Doxygen

Replaced the deprecated `-py3` option with `-doxygen` option in 937bb8f3b551658e1b72a325db5e269bd9c5ebba

rettinghaus:
```console
>>> print(tk.renderToSVG.__doc__)

        Render a page to SVG.

        :type pageNo: int, optional
        :param pageNo: The page to render (1-based)
        :type xmlDeclaration: boolean, optional
        :param xmlDeclaration: True for including the xml declaration in the SVG output
        :rtype: string
        :return: The SVG page as a string
        
```

lpugin: Not sure what to do with the duplicated documentation of the overloaded methods. Maybe we can just point to the reference book? I made sure the ones there are all coming from the source code.

ahankinson: From what I understand the docstrings in the SWIG file can be removed, and SWIG will automatically add them to the Python methods. Or maybe I'm wrong?

ahankinson: Does this also need to happen here: 

https://github.com/rism-digital/verovio/blob/develop/setup.py#L121

rettinghaus: Well, I tried removing them from the swig file, but that will just create empty docstrings. Maybe we'll come up with a better solution later.

lpugin (changes requested):
Could you remove the cast to string for the filenames? As I commented https://github.com/rism-digital/verovio/issues/3158 I do not think this is useful.

As it stand we are missing all the methods that are file access related that are currently not overloaded. Having just two of them but not the others does not make sense, and I do not think we want more methods overloaded because that means loosing more of the documentation we have through doxygen.

bindings/python/verovio.i
ahankinson: So... 

It seems that the docstrings are missing from the shadowed Python methods because those override it. However, this is fixable. 

Rather than, for example:

```
%feature("shadow") vrv::Toolkit::GetAvailableOptions() const %{
def getAvailableOptions(toolkit) -> dict:
     return json.loads($action(toolkit))
%}
```

This can be changed to:

```
%feature("pythonappend") vrv::Toolkit::GetAvailableOptions() const %{
    val = json.loads(val)
%}
```

It seems that the auto-generated Python proxy method sets the value of `$action` to the `val` variable, so you can intercept that before the return and modify it.

This generates a method that looks like this:

```
    def getAvailableOptions(self):
        r"""
        Return all available options grouped by category

        For each option, returns the type, the default value, and the minimum and maximum value (when available)

        :rtype: string
        :return: A stringified JSON object
        """
        val = _verovio_module_3_11_0.toolkit_getAvailableOptions(self)

        val = json.loads(val)


        return val
```

I'm not super happy about the value of `val` being reset, but _c'est la vie_ I guess.

ahankinson: For the shadowed methods that have stuff happening before the C++ call, there is also `%pythonprepend`, which does the same except the arguments are held in an `args[]` list.

ahankinson: Slight correction -- not held in the `args[]` list but set to a variable named after the C++ variable name. So:

```
%feature("pythonprepend") vrv::Toolkit::SetOptions(const std::string &) %{
    jsonOptions = json.dumps(jsonOptions)
%}
```

generates:

```
    def setOptions(self, jsonOptions):
        r"""
        Set option values

        The name of each option to be set is to be given as JSON key.

        :type jsonOptions: string
        :param jsonOptions: A stringified JSON objects with the output options
        :rtype: boolean
        :return: True if the options were successfully set
        """

        jsonOptions = json.dumps(jsonOptions)


        return _verovio_module_3_11_0.toolkit_setOptions(self, jsonOptions)
```

lpugin: I guess I went of the %feature("shadow") because in some methods we need to do adjustments before **and** after.

lpugin: Yes, I remember reading that.

If we can do both then we should give a try. @rettinghaus are you willing to try this?

rettinghaus: I'm on it. 

Any specific reason `GetDescriptiveFeatures` takes `options` instead of `jsonOptions`?

lpugin: No, It can be renamed.

ahankinson:
I think this line will also need to be updated:

https://github.com/rism-digital/verovio/blob/develop/cmake/CMakeLists.txt#L218

.github/workflows/python-ci-wheel.yml
rettinghaus: btw, pypi.org still says "Python >=3.5". Should we update this?

rettinghaus:
This should do the trick. 

As we have to do a check anyway I also added the possibility to still pass a string, so you can do now:
`tk.renderToTimemapFile(filename)`
`tk.renderToTimemapFile(filename, {options})`
`tk.renderToTimemapFile(filename, "{options}")`

And with the new build option you get:

```terminal
>>> print(tk.renderToTimemapFile.__doc__)

        Render a document to timemap and save it to the file.

        Remarks: nojs

        :type filename: string
        :param filename: The output filename
        :type jsonOptions: string, optional
        :param jsonOptions: A stringified JSON objects with the timemap options
        :rtype: boolean
        :return: True if the file was successfully written
        
```

ahankinson:
Thanks @rettinghaus -- @lpugin and I reviewed your changes this morning. 

The ability to pass a stringified JSON object in Python can be removed; if we're doing this for the ability to pass `dict` objects directly it seems like a step in the opposite direction to also allow strings.

There is, of course, another issue here though: The docstrings generated from the C++ are wrong when parsed out to Python:

```
:type jsonOptions: string, optional
:param jsonOptions: A stringified JSON objects with the timemap options
```

These should probably be something like:

```
:type jsonOptions: dict, optional
:param jsonOptions: A dictionary with the timemap options
```

I'm going to do a bit more SWIG digging to see if there's any way we can fix this...

ahankinson:
OK, so after spending most of yesterday on this issue, here's what I've found. 

 - Using the prepend / append method gives us docstrings, yes; but those docstrings are based on the C++ code, not the Python, so they're wrong.
 - You can add [Python-specific docs](https://www.swig.org/Doc4.1/Doxygen.html#Doxygen_ignore) in the C++ docstrings, but there's no way to if / else it -- you can't say "use X for Python, and Y for everything else."
 -  The doxygen markup in the C++ docstrings is automatically expanded by SWIG, so `@param foo: A great foo parameter` expands to `:type foo: string, :param foo: A great foo parameter`. This means we don't have any way of controlling or overriding the `:type` expansion to say "In Python, `foo` is a dict" since it gets it from the C++ method signature.
 - This means that we either have to a) deal with auto-generated but wrong docstrings using prepend / append, or, using the "shadow" method have correct type signatures but no docstrings. (Assuming we don't want to put the docstrings in two places, which @lpugin is against). 
 - The presence of a default value for a parameter in an argument list in C++ affects the translation of arguments in Python. `std::string RenderToTimemap(const std::string &jsonOptions = "");` produces `def renderToTimemap(self, *args)`, while `std::string RenderData(const std::string &data, const std::string &jsonOptions);` produces `def renderData(data, jsonOptions)`.
 - This also means that using the prepend / append method, we don't get named parameters in the Python function signature, since everything is collapsed to the `*args` array. @lpugin isn't really a fan of this, nor the extra `*args` array parsing needed to convert the incoming arguments.
 - I looked, but could not find, a way to auto-include the C++ docstrings using the "shadow" method. And, given that they would be wrong anyway, I'm not sure that would be useful.

So after all that, and chatting with @lpugin, it would probably better to go back to using the "shadow" method and not have the docstrings in the Python interface, but properly annotating the arguments and return types.

lpugin (commented):
Seeing that `%feature("pythonprepend")` does not have named parameters when there is more than one of them, I find it not an improvement from the `%shadow` syntax. Having to do parameter array manipulations makes it hard to read and error prone.

I would suggest to drop the refactoring in `verovio.i` and we can merge the other changes already. What do you think?

bindings/python/verovio.i
```diff
@@ -1,6 +1,6 @@
 /////////////////////////////////////////////////////////////////////////////
 // Name:        verovio.i
-// Author:      Laurent Pugin
+// Author:      Laurent Pugin and Klaus Rettinghaus
```
lpugin, line 3: The author of a file is creator of it. We are not going to start adding the names of everybody making changes to the files in the repository.

bindings/python/verovio.i
```diff
@@ -19,144 +19,163 @@
...
 // Toolkit::SaveFile
 %feature("shadow") vrv::Toolkit::SaveFile(const std::string &, const std::string & = "") %{
-def saveFile(toolkit, filename, options = None):
-    if options == None:
+def saveFile(toolkit, filename: str, options: Optional[dict] = None) -> None:
+    """Get the MEI and save it to the file."""
+    if options is None:
         options = {}
-    return $action(toolkit, filename, json.dumps(options))
+    return json.loads($action(toolkit, filename, json.dumps(options)))
```
lpugin, line 158: This is wrong. `SaveFile` returns a `bool`

bindings/python/verovio.i
```diff
@@ -19,144 +19,163 @@
...
 // Toolkit::Edit
 %feature("shadow") vrv::Toolkit::Edit( const std::string & ) %{
 def edit(toolkit, editor_action):
+    """Edit the MEI data."""
     return $action(toolkit, json.dumps(editor_action))
 %}
 
 // Toolkit::EditInfo
 %feature("shadow") vrv::Toolkit::EditInfo() %{
 def editInfo(toolkit):
```
lpugin, line 39: The return type annotation is missing

bindings/python/verovio.i
```diff
@@ -19,144 +19,163 @@
...
-def setOptions(toolkit, json_options):
+def setOptions(toolkit, json_options: dict) -> bool:
+    """Set option values."""
     return $action(toolkit, json.dumps(json_options))
 %}
 
 // Toolkit::ValidatePAE
 %feature("shadow") vrv::Toolkit::ValidatePAE(const std::string &) %{
-def validatePAE(toolkit, data):
+def validatePAE(toolkit, data: str) -> dict:
```
ahankinson, line 177:
```suggestion
def validatePAE(toolkit, data: Union[str, dict]) -> dict:
```

This will allow both a string or a Python dictionary as input.

bindings/python/verovio.i
```diff
@@ -19,144 +19,163 @@
...
+    if options is None:
         options = {}
     return json.loads($action(toolkit, json.dumps(options)))
 %}
 
 // Toolkit::RenderToTimemapFile
 %feature("shadow") vrv::Toolkit::RenderToTimemapFile(const std::string &, const std::string & = "") %{
-def renderToTimemapFile(toolkit, filename, options = None):
-    if options == None:
+def renderToTimemapFile(toolkit, filename: str, options: Optional[dict] = None) -> None:
```
ahankinson, line 145:
```suggestion
def renderToTimemapFile(toolkit, filename: str, options: Optional[dict] = None) -> bool:
```

ahankinson (approved):

Issue #3168, opened 2022-12-20
'verovio-module.js' does not provide an export named 'default'


sonovice:
**Describe the problem**
When using the official NPM package, this code works without any problems:

Shell:
```bash
npm install verovio
```

JS:
```javascript
import createVerovioModule from 'verovio/wasm';
import { VerovioToolkit } from 'verovio/esm';

const mei_string = "..."; // some valid MEI data

createVerovioModule().then(VerovioModule => {
    const vrv = new VerovioToolkit(VerovioModule);
    vrv.loadData(mei_string);
    const data = vrv.renderToSVG(1);
    console.log(data);
});
```

However, if I build the package locally using the `emscripten/buildNpmPackage` script and try to use it, it fails with the following error:
```
Uncaught SyntaxError: The requested module '/@fs/home/simon/Repositories/verovio/emscripten/npm/dist/verovio-module.js' does not provide an export named 'default'
```

**To Reproduce**
```bash
mkdir -p ~/verovio
git clone https://github.com/rism-digital/verovio ~/verovio
cd verovio/emscripten
./buildNpmPackage

cd ~/app # go to JS source code
npm install ~/verovio/emscripten/npm
npm run dev # depends on the project, of course
```

**Environment information (as appropriate)**
 - OS: Linux x64 with Kernel v6.0.6
 - Emscripten: v3.1.28
 - Browser: All

WolfgangDrescher:
Are you using `"type": "module"` in your `package.json`? There was no full support for asynchronous ES6 modules in emscripten until v3.1.27. This is why the current build of `verovio-module.js` is actually a CommonJS module. ESM can import CommonJS modules if they just expose a default export (which is the case for verovio). If you have `"type": "module"` in your `package.json`, then `.js` files will be treated as ESM ("`.mjs`"). You can force CommonJS modules by using `.cjs` as file extension:

https://stackoverflow.com/a/57492606/9009012

> Node.js will treat `.cjs` files as CommonJS modules and `.mjs` files as ECMAScript modules. It will treat `.js` files as whatever the default module system for the project is (which is CommonJS unless package.json says `"type": "module",`).

Recently there was some progress and https://github.com/emscripten-core/emscripten/issues/11792 was implemented in https://github.com/emscripten-core/emscripten/pull/17915.

[...]

PR #3174, opened 2022-12-21
Build modularized JS toolkit with support for `-s EXPORT_ES6`

WolfgangDrescher:
Related to https://github.com/rism-digital/verovio/issues/3168.

Recently there was some progress in emscripten to support ECMAScript modules and https://github.com/emscripten-core/emscripten/issues/11792 was implemented in https://github.com/emscripten-core/emscripten/pull/17915. This means that since `em++` v3.1.27 we can use the `EXPORT_ES6` flag to generate an ESM build of the emscripten module instead of a CommonJS module (that can be imported with ESM).

Is there anything I can add to force usage of em++ v3.1.27?

I did some tests an even without 3bca155 everything did work just fine in my project. But adding `mjs` as file extension will probably be the safer method to ensure the verovio-module will be used with ESM. For other usage you can still use `verovio-toolkit-wasm.js` with CommonJS due to the UMD format.

Probably this won't fix all cases when using verovio with bundlers (see this comment https://github.com/emscripten-core/emscripten/issues/11792#issuecomment-1319507031) but it should still improve ESM support.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -193,6 +193,9 @@ if (BUILD_AS_LIBRARY)
 ###########
 
 elseif (BUILD_AS_WASM)
+    if (EMSCRIPTEN_VERSION VERSION_LESS "3.1.27")
+        message(FATAL_ERROR "You need em++ 3.1.27 or greater")
+    endif()
     message(STATUS "***** Building Verovio as WASM BC *****")
     add_library(verovio STATIC ../tools/c_wrapper.cpp ${all_SRC})
 
--- a/emscripten/buildNpmPackage
+++ b/emscripten/buildNpmPackage
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e -o pipefail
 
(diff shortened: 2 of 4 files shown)


lpugin:
Thanks! 
 
> Is there anything I can add to force usage of em++ v3.1.27?

Could you try to add this to the cmake file (in `elseif (BUILD_AS_WASM)`:
```
    if (EMSCRIPTEN_VERSION VERSION_LESS "3.1.27")
        message(FATAL_ERROR "You need em++ 3.1.27 or greater")
    endif()
```

WolfgangDrescher:
Thanks, I added your snippet into CMakeLists.txt.

Additionally I improved error handling in `buildToolkit` (die when `system($cmakeCmd)` fails) and `buildNpmPackage` (script will stop at the first fatal error). Without this `buildToolkit` would still compile and link something but display red messages and `buildNpmPackage` would still build a corrupted toolkit and bundle something with `npm`.

PR #3178, opened 2022-12-23
Fix memory leak

valeriyvan:


lpugin (changes requested):
No, there is no memory leak here. The `fermata` is added to the measure in `Measure::ConvertMarkupAnalyticalEnd`.

I agree it is not intuitive. It could be better documented or the name of the method could be changed to make it clearer. Suggestion welcome.

PR #3184, opened 2022-12-28
add deg attribute to note

rettinghaus:
This adds the ability to store data from a `**deg` spine in the `@deg` attribute of a note. 

See https://music-encoding.org/guidelines/dev/data-types/data.SCALEDEGREE.html

However, when there are multiple `**deg` spines I'm not sure what should be done.

### DIFF
--- a/include/vrv/note.h
+++ b/include/vrv/note.h
@@ -13,6 +13,7 @@
 //----------------------------------------------------------------------------
 
 #include "accid.h"
+#include "atts_analytical.h"
 #include "atts_externalsymbols.h"
 #include "atts_frettab.h"
 #include "atts_mensural.h"
@@ -54,6 +55,7 @@ class Note : public LayerElement,
              public AttCue,
              public AttExtSym,
              public AttGraced,
+             public AttHarmonicFunction,
              public AttMidiVelocity,
              public AttNoteGesTab,
              public AttNoteHeads,
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -2668,6 +2668,7 @@ void MEIOutput::WriteNote(pugi::xml_node currentNode, Note *note)
     note->WriteCue(currentNode);
     note->WriteExtSym(currentNode);
     note->WriteGraced(currentNode);
+    note->WriteHarmonicFunction(currentNode);
     note->WriteMidiVelocity(currentNode);
     note->WriteNoteGesTab(currentNode);
     note->WriteNoteHeads(currentNode);
@@ -6620,6 +6621,7 @@ bool MEIInput::ReadNote(Object *parent, pugi::xml_node note)
     vrvNote->ReadCue(note);
     vrvNote->ReadExtSym(note);
     vrvNote->ReadGraced(note);
+    vrvNote->ReadHarmonicFunction(note);
     vrvNote->ReadMidiVelocity(note);
     vrvNote->ReadNoteGesTab(note);
     vrvNote->ReadNoteHeads(note);
--- a/src/note.cpp
+++ b/src/note.cpp
@@ -61,6 +61,7 @@ Note::Note()
     , AttCue()
     , AttExtSym()
     , AttGraced()
+    , AttHarmonicFunction()
     , AttMidiVelocity()
     , AttNoteGesTab()
     , AttNoteHeads()
@@ -78,6 +79,7 @@ Note::Note()
     this->RegisterAttClass(ATT_CUE);
     this->RegisterAttClass(ATT_EXTSYM);
     this->RegisterAttClass(ATT_GRACED);
+    this->RegisterAttClass(ATT_HARMONICFUNCTION);
     this->RegisterAttClass(ATT_NOTEGESTAB);
     this->RegisterAttClass(ATT_NOTEHEADS);
     this->RegisterAttClass(ATT_NOTEVISMENSURAL);
@@ -104,6 +106,7 @@ void Note::Reset()
     this->ResetCue();
     this->ResetExtSym();
     this->ResetGraced();
+    this->ResetHarmonicFunction();
     this->ResetNoteGesTab();
     this->ResetNoteHeads();
     this->ResetNoteVisMensural();


craigsapp:
Regarding the regular expression for `**deg` data on the page:
https://music-encoding.org/guidelines/dev/data-types/data.SCALEDEGREE.html

It needs a slight modification.  Here is the original:

```
<rng:param name="pattern">(\^|v)?[1-7](\+|\-)?</rng:param>
```

The chromatic alterations `+` and `-` indicate a semitone adjustment, but it is possible to have an adjustment of more than one semitone, so this regular expression is better:

```
<rng:param name="pattern">(\^|v)?[1-7](\+*|-*)?</rng:param>
```
A backslash should not come before the `-`, since it is not a metacharater (only inside `[]`).   In this regular expression, 0 or more `+`/`-` are allowed rather than 0 or 1.

For example `7++` in C major would be B-double-sharp, and `7--` would be B-double-flat.    These 2-semitone adjustments are rare in any case, with `++` being exceedingly rare.  There could be triple sharps/flats, but those would be mostly theoretical.  If you want to limit to two semitones alteration, then one of these regular expressions should work:

```
<rng:param name="pattern">(\^|v)?[1-7](\+{,2}|-{,2})?</rng:param>
```

or if the `{}` operator is not allowed (it should be since it is in the "extended" regular expressions such as `+`, `()` and `|` &mdash; Another way of expressing the limitation to +/- two semitones could be:

```
<rng:param name="pattern">(\^|v)?[1-7](\+\+?|--?)?</rng:param>
```

The last sentence of the description:

> The amount of chromatic alternation is not indicated.

Is not correct: The number of `+` or `-` indicate the chromatic alteration.

Also note that `alternation` is a typo for `alteration`.


The `**deg` documentation is ambiguous on this point.   There was a loss of the primary documentation files around 2005, and there is some information missing from the page:

Screen Shot 2022-12-28 at 12 37 46 PM

Notice `--` is mentioned, but `++` would be a reasonable interpretation of two semitones as well.  There is also `r` for rests, but that is not very useful (since `<rest>` cannot have a scale degree, and would obviously be `r`).



---------------------

There is a complication that I will probably address soon in the `**deg` representation.   The scale degree alterations are relative to the pitch classes of the key.   For major keys there is no problem, but for minor keys the `**deg` standard is the harmonic minor (with a raised 7ths from the natural minor).    I am thinking of allowing an `n` after the `7` to indicate that it is in reference to the natural minor rather than the harmonic minor.   In that case the unlimited semitone alteration version would be:

```
<rng:param name="pattern">(\^|v)?([1-6]|7n?)(\+*|\-*)?</rng:param>
```

In C minor: `7` is B-natural, `7n` is B-flat, `7-` is B-flat, and `7n-` is B-double-flat.   The `n` can be thought of as adding an extra flat to the harmonic minor scale.  `7n+` would be B-natural (equivalent to `7-+`) and `7n++` would be B-sharp.    This will allow notation of the alterations in melodic minor, but still keep the harmonic minor as the default scale degrees so there is no confusion about what the chromatic alteration of the 7th scale degree in minor keys is.

craigsapp:
The pull request looks good to me.  How would you programmatically set `note@deg`?

I am presuming:

```cpp
Note *note = new Note();
std::string data = "^4+";
note->SetDeg(data);
```

Or perhaps inputting a `u32string`?

```cpp
Note *note = new Note();
std::u32string data = U"^4+";
note->SetDeg(data);
```

craigsapp:
Regarding multiple degrees for a single note:  This is probably not something to be concerned with until someone complains about it.   Usually there will be only a singe scale degree interpretations for a note, which is relative to the active key of the music.   There can be cases as found in the discussion https://github.com/rism-digital/verovio/discussions/3166 where there are multiple scale degree interpretations for multiple keys.   This is possible in music in particular during modulations where the notes can be considered in two keys at the same time (and will therefore have different scale degree functions for each of the keys).

craigsapp (approved):

rettinghaus:
A `std::string` should be sufficient here. I'll make a corresponding PR to fix the datatype in MEI. 

Note that this PR is for the Humdrum branch to accompany the ongoing process there.

rettinghaus:
btw the sentence 

> The amount of chromatic alteration is not indicated

(except for the typo) comes from here: https://www.humdrum.org/Humdrum/representations/deg.rep.html

lpugin:
The PR is a good start but does not render anything. What would be good is to convert `@deg` into `<harm>` in `Object::ConvertMarkupAnalytical`

It would probably be good to take into account the fact that memory management can currently be misleading and that the code can be clarified. See PR #3178

craigsapp:
> The PR is a good start but does not render anything. What would be good is to convert `@deg` into `<harm>`

The purpose of `note@deg` is for analysis, and it should not be rendered.   

If someone wants to display it, then they should write their own converter in XSLT to do so.   The problem is that encoding `@deg` data in `<harm>` content tends to add a lot of formatting to display in a specific style (see discussion https://github.com/rism-digital/verovio/discussions/3166 where there are lots of styling options being discussed).   The resulting `<harm>` contents then becomes difficult to use for analytic processing.

In the Humdrum-to-MEI converter, the `@deg` information is being converted into `<harm>`, and ideally this conversion should also have `harm/startid` that points to the source note that includes the original `@deg` data (if there is a known mapping to a note).  Alternatively or in addition, it might be nice to add `harm@deg` to MEI.

rettinghaus:
Thinking of this a bit more it makes sense (to me) that the amount of chromatic alteration is not shown. But I will not argue about that with the master of Humdrum. 😉

craigsapp:
> (except for the typo) comes from here: https://www.humdrum.org/Humdrum/representations/deg.rep.html

I fixed it.  That page is a copy of the documentation before it was accidentally deleted around 2005 in an os migration (not by me or on my computer).   The documentation is being renovated from plain HTML to jekyll/markdown, so the newest documentation is at the link in the MEI documentation (but that link is now older than the mirrored copy ;-)

https://www.humdrum.org/rep
https://www.humdrum.org/rep/deg

> Thinking of this a bit more it makes sense (to me) that the amount of chromatic alteration is not shown. But I will not argue about that with the master of Humdrum. 😉

It will depend on what you want to analyze.  If you want to measure the amount of alterations, you would check for `/[+-]/` and this would not matter if `+` or `++` were in the token.  If you wanted to search for extreme cases, then you could use `/(\+\+|--)/`.  Not encoding `++` means you could not do that search.

I checked the `deg` command from the Humdrum Toolkit.  It only outputs `-` or `+` regardless of the size of the alteration:

```tsv
**kern	**kern
*C:	*a:
4c	4c
4d	4d
4e	4e
*^	*^
4f#	4f	4f#	4f
*v	*v	*	*
*	*v	*v
4g##	4g##
4a#	4a#
4b--	4b--
4b-	4b-
4b	4b
4cc 4ee 4g	4cc 4ee 4g
*-	*-
```

deg output:

```tsv
**deg	**deg
*C:	*a:
1	3
^2	^4
^3	^5
*^	*^
^4+	^4	^6+	^6
*v	*v	*	*
*	*v	*v
^5+	^7+
^6+	^1+
v7-	v2-
^7-	^2-
^7	^2
^1 ^3 v5	^3 ^5 v7-
*-	*-
```

It does not interleave the data with the input score, which is a limitation that should be expanded on in the future (otherwise it is not easy to see the scale degrees via music notation).

Creating a census of the degrees (and keeping track of melodic approaches and basic alterations):

```bash
deg test.krn  | serialize -S | ridx -H | tr " " "\n" | sortcount
```

```
3	^3
2	^4
2	^6+
2	^2
2	^5
2	v7-
1	^2-
1	^1
1	3
1	^5+
1	v2-
1	1
1	^7+
1	^7-
1	^1+
1	^6
1	v5
1	^4+
1	^7
```

The most common scale degree is the third approached from below.   Here is the same analysis for J.S. Bach's WTC 1 Fugue 1:

```bash
humcat h://wtc/wtc1f01.krn  | deg | serialize -S | ridx -H | tr " " "\n" | sortcount
```

```
67	^6
61	r
56	v2
54	^2
54	^3
50	v3
49	^5
49	v1
47	^1
47	v5
46	v6
43	v7
35	v4
35	^7
31	^4
24	5
16	2
14	v4+
14	^4+
9	v5+
8	1
8	^7-
6	4
5	^5+
5	v7-
5	6
5	3
5	v1+
3	7
2	^1+
```

Where the most common note is the sixth scale degree approached from below (the data only has an initial key designation, and does not consider modulations).

Here is an analysis of just scale degrees, ignoring their melodic approach:

```bash
 humcat h://wtc/wtc1f01.krn  | deg -a | serialize -S | ridx -H | tr " " "\n" | sortcount
```

```
126	2
120	5
118	6
109	3
104	1
81	7
72	4
61	r
28	4+
14	5+
13	7-
7	1+
```

the 7th and 4th are used noticeably less than the (so approaching a pentatonic scale somewhat).   The `4+` is due to a leading tone to G (so there is likely a modulation to G major in the music, or at least secondary dominants/tonicizations to G).  And less so to A (minor) due to `5+`.

It is not possible to filter out repeated notes in the current labelling done by the `deg` command, since it does not mark the melodic approach of initial notes nor notes following rests as well as repeated notes.  (it would be good to add an option to allow/disallow analyzing melodic approaches across rests, with an additional requirement for a cutoff duration for rests when analyzing melodic approach -- such as allowing continuation across eighth notes, but not a whole measure rest).   And there are complexities when dealing with chords going to single note or chords with different numbers of notes when calculating melodic approaches...

PR #3198, opened 2023-01-11
Fix extend import for MusicXML figure 

rettinghaus:
This fixes a small issue with creating empty `harm` elements for extender stops, like:

```xml
<figured-bass>
   <figure>
      <extend type="stop" />
   </figure>
</figured-bass>
```

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2480,8 +2480,8 @@ void MusicXmlInput::ReadMusicXmlFigures(pugi::xml_node node, Measure *measure, c
     if (!figures.empty()) {
         Harm *harm = new Harm();
         Fb *fb = new Fb();
-        for (auto iter = figures.begin(); iter != figures.end(); ++iter) {
-            fb->AddChild(*iter);
+        for (auto &fig : figures) {
+            fb->AddChild(fig);
         }
         harm->AddChild(fb);
         harm->SetTstamp((double)(m_durTotal + m_durFb) * (double)m_meterUnit / (double)(4 * m_ppq) + 1.0);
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -2450,35 +2450,40 @@ void MusicXmlInput::ReadMusicXmlFigures(pugi::xml_node node, Measure *measure, c
...
-        harm->SetTstamp((double)(m_durTotal + m_durFb) * (double)m_meterUnit / (double)(4 * m_ppq) + 1.0);
-        m_durFb += node.child("duration").text().as_int();
-        m_controlElements.push_back({ measureNum, harm });
-        m_harmStack.push_back(harm);
+        Text *text = new Text();
+        text->SetText(UTF8to32(textStr));
+        f->AddChild(text);
+        fb->AddChild(f);
     }
+    if (!fb->GetFirst()) return;
```
lpugin, line 2481: This creates a memory leak

rettinghaus: Then why isn't this:
https://github.com/rism-digital/verovio/blob/b5325efb83d770710610538fd498d600a64623d2/src/view_control.cpp#L2173

lpugin: I am not sure why you are pointing to that and the purpose of the comparison.

You allocate a `Harm`on the heap line 2455, and then a `Fb`, and then exit the function without doing anything with them. This creates a memory leak.

rettinghaus: Oh, sorry! I thought you meant `->GetFirst`. You're right of cause.

src/iomusxml.cpp
```diff
@@ -2476,14 +2475,21 @@ void MusicXmlInput::ReadMusicXmlFigures(pugi::xml_node node, Measure *measure, c
         Text *text = new Text();
         text->SetText(UTF8to32(textStr));
         f->AddChild(text);
-        fb->AddChild(f);
+        figures.push_back(f);
+    }
+    if (!figures.empty()) {
```
lpugin, line 2480: I think it would be a bit clearer here to return if the vector is empty instead of having an `if` block if not empty – since nothing is happening after it anyway

src/iomusxml.cpp
```diff
@@ -2476,14 +2475,21 @@ void MusicXmlInput::ReadMusicXmlFigures(pugi::xml_node node, Measure *measure, c
...
+        Harm *harm = new Harm();
+        Fb *fb = new Fb();
+        for (auto iter = figures.begin(); iter != figures.end(); ++iter) {
```
lpugin, line 2483: You can use a C++11 range loop here which is more compact

lpugin (commented):
Thanks. The memory leak is gone!

src/iomusxml.cpp
rettinghaus: sure

PR #3202, opened 2023-01-12
Use pen width for bounding boxes

DavidBauer1984:
This PR unifies the handling of pen width for bounding boxes and adds consideration of pen width for polygons and polylines.

As a consequence, in some cases the bounding boxes of endings now slightly horizontally overlap. We instruct the layout that this is OK. See Ending-002 in the test suite.

### DIFF
--- a/include/vrv/bboxdevicecontext.h
+++ b/include/vrv/bboxdevicecontext.h
@@ -159,6 +159,11 @@ private:
      */
     void ResetGraphicRotation();
 
+    /**
+     * Get the overlap due to pen width on the left/right
+     */
+    std::pair<int, int> GetPenWidthOverlap() const;
+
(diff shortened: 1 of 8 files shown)


DavidBauer1984:
There is a regression in `Ending-002`. Investigating ...

lpugin:
If I understand correctly, the endings where pushing themselves up 
image
I think it would be better simply to shorten their length and avoid they touch instead of allowing for some overlap. I would visually be better too in my opinion

DavidBauer1984:
Yes, if we don't allow for overlap, then the endings are pushing themselves up. I'll see what I can do to horizontally separate them. Anyway, this PR has taken a funny transition from bounding boxes to endings :)

PR #3204, opened 2023-01-12
Humdrum: Support for negative figured bass numbers

WolfgangDrescher:
In preparation for the upcoming feature for automatically generated figured bass numbers in https://github.com/craigsapp/humlib/pull/53 I am adding functionality for negative numbers to iohumdrum.cpp.

For figured bass this is probably not important as I have never seen negative figured bass numbers. But since this can also create an *Intervallsatz* I think it's a good idea a enable support for negative numbers.

@craigsapp and I have been discussing about the signifier. I chose `m` for `minus` since it probably won't be necessary to add a plus sign for positive numbers in future.

Negative numbers can be encoded by adding the `m` signifier:

```tsv
**kern	**kern	**fb
*clefG2	*clefG2	*
4cc	4f	m5
4cc	4a#	#m3
*-	*-	*-
```

Examples:

Bildschirm­foto 2023-01-12 um 20 27 52

Bildschirm­foto 2023-01-12 um 20 32 55

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -8660,17 +8660,10 @@ std::u32string HumdrumInput::convertFBNumber(const std::string &input, hum::HTp 
         accidental = 0;
     }
 
-    if (input.find("v") != std::string::npos) {
-        // display minus sign for negative numbers if "v" signifier is present
+    if (input.find("~") != std::string::npos) {
+        // display minus sign for negative numbers if "~" signifier is present
         output += '-';
     }
-
-    /*
-    if (input.find("^") != std::string::npos) {
-        // display plus sign for positive numbers if "^" signifier is present
-        output += '+';
-    }
-    */
 
     // accidental in front of number unless an "r" is present:
     if ((!slash) && (input.find("r") == std::string::npos) && (!reverse)) {
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -8660,8 +8660,8 @@ std::u32string HumdrumInput::convertFBNumber(const std::string &input, hum::HTp 
         accidental = 0;
     }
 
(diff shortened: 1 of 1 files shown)


WolfgangDrescher:
I just realized that `m` actually seems to be already reserved for minor intervals: https://doc.verovio.humdrum.org/humdrum/figured_bass/#encoding-chromatic-interval.

This does not seem to be conflicting yet with `iohumdrum.cpp` and could still be used. But probably it would be better to take something different.

`-` is already reserved for flats and `n` (=negative) is already reserved for natural. Is there a better option?

craigsapp:
> I just realized that m actually seems to be already reserved for minor intervals: https://doc.verovio.humdrum.org/humdrum/figured_bass/#encoding-chromatic-interval.

> This does not seem to be conflicting yet with iohumdrum.cpp and could still be used. But probably it would be better to take something different.

> - is already reserved for flats and n (=negative) is already reserved for natural. Is there a better option?

I was thinking the same just now reading the posts above. I think the best is to use `v` for negative numbers, where `v` is similar to the downward melodic approach in `**deg` data.   `^` can be reserved if above/positive is ever needed.

WolfgangDrescher:
> I was thinking the same just now reading the posts above. I think the best is to use v for negative numbers, where v is similar to the downward melodic approach in **deg data. ^ can be reserved if above/positive is ever needed.

Personally I think that `v` and `^` are bit confusing as they look like the melodic approach signifier of `deg` or similar. It looks like there is some kind of directional information. But since I have no better suggestion I updated the PR accordingly.

WolfgangDrescher:
This sound good to me. But I'll leave it up to you to define the standards 😊️. For me it's fine to merge like this. If you think different I can revert 84d8b27.

craigsapp:
The `~` looks good, so we will go with that.

PR #3205, opened 2023-01-13
Change gliss to default to wavy

rettinghaus:
This PR is a small change for glissandos to default to wavy lines if no `@lform` is specified.

### DIFF
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -2048,18 +2048,6 @@ void View::DrawGliss(DeviceContext *dc, Gliss *gliss, int x1, int x2, Staff *sta
     }
 
     switch (gliss->GetLform()) {
-        case LINEFORM_wavy: {
-            const int length = static_cast<int>(hypot(x2 - x1, y2 - y1));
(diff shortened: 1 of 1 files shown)


lpugin:
I would personally expect a glissando to be a straight line by default. So does Gould for the "Conventional glissando notation" (See Behind Bars p. 265). What is the basis of your proposal?

rettinghaus:
Usually straight lines indicate a portamento. But it's not super important. If you'd prefer to leave it as is, I'm also fine with that.

lpugin:
I am not against it. Let's see if there are some other opinions on this.

lpugin:
Maybe this is something that can be clarified or improved in MEI?

lpugin:
Since there is no other opinion we can leave it as it is. It can still be changed but following a discussion / clarification in MEI.

craigsapp:
In piano music the gliss lines can be seen both as straight or wavy lines, depending on the repertory and publisher.  So I don't have a preference either way, and wavy can be the default to differentiate them from portamento (pianos can't do portamento :-)

Issue #3218, opened 2023-01-19
Expose skipLayoutOnLoad() in toolkit


wergo:
With PR #2932, @DavidBauer1984 introduced a nice flag to skip time-intensive layout processing when merely requesting MIDI or timemap through the commend line. 

It seems that this feature cannot be accessed from the toolkit. Am I missing something here? 

It would be great to expose it to the toolkit too.

rettinghaus:
It's available in the Python toolkit, but nowhere documented. Perhaps we should also move it out from the _nodoc_ group?

DavidBauer1984:
When using Verovio as command-line program we don't need an extra option. We already have full information and know when to skip the layout (i.e. depending on the output format).

When using it as toolkit, we need some extra information. When calling `Load` after construction of the toolkit, the toolkit doesn't know what the user intends to do with it, i.e. if layout is required for later commands such as rendering. So in my view there are two alternatives:

1. We create an option. Then it would be neat, if the value was calculated from the output format on the command line program while it should be switched off per default on the toolkit. This feels a bit awkward compared to other options.

2. Make it available via toolkit API.

I refrained from making the functionality available in the toolkit, since skipping the layout can lead to problems, if not used properly. It should be activated only, if the toolkit is not used for any rendering purposes. If the user loads the MEI without layout and then calls `Render`, this probably leads to a crash. We can leave this hazard to the user or must improve error handling.

wergo:
> It's available in the Python toolkit, but nowhere documented. Perhaps we should also move it out from the _nodoc_ group?

@rettinghaus Where do I see that this method exposed for the Python toolkit? I cannot find it in `/bindings/python/verovio.i`.

wergo:
@DavidBauer1984 Thanks a lot for your reply. 

Independently whether it is an option or exposed as part of the API, if data is loaded to the toolkit with the rendering-skipped option and renderToSvg() methods are called afterwards, the toolkit should go and rerender the layout then. What do you think?

rettinghaus:
@wergo swig provides all public methods in the toolkit. In `verovio.i` we simply redefine a set of methods to be more Pythonic.

wergo:
@rettinghaus Ah, I see. Thanks a lot.

DavidBauer1984:
> Independently whether it is an option or exposed as part of the API, if data is loaded to the toolkit with the rendering-skipped option and the `renderToSvg()` method is called afterwards, the toolkit should go and re-render the layout then.

[...]

PR #3228, opened 2023-01-26
Add skipLayoutOnLoad to js toolkit

rettinghaus:
This PR is an attempt to add `skipLayoutOnLoad` to the JavaScript toolkit and to the Doxygen documentation.


lpugin:
As I said it, having this very specific function exposed to the toolkit is over the top. As I commented, we simply need to improve the documentation and explain that setting `breaks` to `none` is the most efficient way to retrieve the timemap.

rettinghaus:
ok, so you also don't want to hav it in the documentation?

lpugin:
Yes, I do, in the `renderToTimemap` method (see https://github.com/rism-digital/verovio/issues/3218#issuecomment-1404723764). I will add it.

PR #3229, opened 2023-01-26
Unify iterators

rettinghaus:
This PR tries to unify iterators by not declaring counters before and trying to avoid casting when simply iterate by `size()` or `length()`.

### DIFF
--- a/src/bboxdevicecontext.cpp
+++ b/src/bboxdevicecontext.cpp
@@ -216,7 +216,7 @@ void BBoxDeviceContext::DrawPolygon(int n, Point points[], int xOffset, int yOff
     int y1 = points[0].y + yOffset;
     int y2 = y1;
 
-    for (int i = 0; i < n; i++) {
+    for (int i = 0; i < n; ++i) {
         x1 = std::min(x1, points[i].x + xOffset);
         x2 = std::max(x2, points[i].x + xOffset);
         y1 = std::min(y1, points[i].y + yOffset);
@@ -362,7 +362,7 @@ void BBoxDeviceContext::DrawMusicText(const std::u32string &text, int x, int y, 
     char32_t smuflGlyph = 0;
     if (setSmuflGlyph && (text.length() == 1)) smuflGlyph = text.at(0);
 
-    for (unsigned int i = 0; i < text.length(); i++) {
+    for (size_t i = 0; i < text.length(); ++i) {
         char32_t c = text.at(i);
         const Glyph *glyph = resources->GetGlyph(c);
         if (!glyph) {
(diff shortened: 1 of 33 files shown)


lpugin:
I like casting ;-). Didn't we have this discussion already? Keeping `int` everywhere is more unified than have `int` in some cases and `size_t` in others.

rettinghaus:
Well, we didn't have exactly _this_ discussion ;-). 

In some cases there was `unsigned int` used, in some other places already `size_t`. Just changed the cases where we don't do anything else than counting up, whenever the value is passed nothing was changed.

lpugin:
I added a clarification here https://github.com/rism-digital/verovio-reference-book/commit/6916eafe5071bbe56b6d489760534a017a56ac2e

Please adjust accordingly. Thanks!

rettinghaus:
Done.
NB: There are still some `size_t` left where `std::string::npos` is used.

craigsapp:
> In some cases there was unsigned int used, in some other places already size_t. Just changed the cases where we don't do anything else than counting up, whenever the value is passed nothing was changed.

They are similar but not the same (so you have to be careful), since different computers/OSes can have different sizes for the two types:

https://en.cppreference.com/w/c/language/arithmetic_types#Data_models

`std::string::npos` is `size_t` which could be 64-bits while `unsigned int` could be 32-bits at the same time. 

https://en.cppreference.com/w/c/types/size_t#:~:text=typedef%20%2F*implementation%2Ddefined*%2F,is%20not%20less%20than%2016.

> size_t is commonly used for array indexing and loop counting. Programs that use other types, such as unsigned int, for array indexing may fail on, e.g. 64-bit systems when the index exceeds [UINT_MAX](https://en.cppreference.com/w/c/types/limits) or if it relies on 32-bit modular arithmetic.

So you should not compare a `size_t` value to an `unsigned int` in the most general case since they could have different storage sizes, and in particular `std::string::npos` on a 64-bit computer cannot be compared to a 32-bit unsigned int without an error.

Doing so is OK, as long as you are not using values above the size that fits in 32-bits.   In other words, as long as a vector has less than 4 billion values, using unsigned int is OK (and less than 2 billion when using int).  But if you have a vector with more than 4 billion entries, you have to use the correct type, which is `size_t`.

There should be no problems caused when changing `unsigned int` to `size_t` since `size_t` is always larger or equal to `unsigned int`'s storage space.

> NB: There are still some size_t left where std::string::npos is used.

That is good since `size_t` is always larger or equal to `unsigned int` storage space, and is required with dealing with `std::string::npos`.

> Keeping int everywhere is more unified than have int in some cases and size_t in others.

There is `int` and there is `unsigned int`.  Comparing `int` to `size_t` values is obviously stupid, while comparing `unsigned int` to `size_t` values is subtly stupid.  I like int to avoid cases where an error in subtraction would otherwise generate a very large index into an array, so I like doing the obviously stupid thing.

The correct data type for `std::vector::operator[]` is `size_t`, so it might be better to avoid `unsigned int` in favor of `size_t`:

https://en.cppreference.com/w/cpp/types/size_t

> std::size_t can store the maximum size of a theoretically possible object of any type (including array). A type whose size cannot be represented by std::size_t is ill-formed. On many platforms (an exception is systems with segmented addressing) std::size_t can safely store the value of any non-member pointer, in which case it is synonymous with [std::uintptr_t](https://en.cppreference.com/w/cpp/types/integer).
>
>std::size_t is commonly used for array indexing and loop counting. Programs that use other types, such as unsigned int, for array indexing may fail on, e.g. 64-bit systems when the index exceeds [UINT_MAX](https://en.cppreference.com/w/cpp/types/climits) or if it relies on 32-bit modular arithmetic.
>
>When indexing C++ containers, such as [std::string](https://en.cppreference.com/w/cpp/string/basic_string), [std::vector](https://en.cppreference.com/w/cpp/container/vector), etc, the appropriate type is the member typedef size_type provided by such containers. It is usually defined as a synonym for std::size_t.

PR #3230, opened 2023-01-26
Remove SkipLayoutOnLoad and use --breaks instead

lpugin:
* Since the method / option was used only in the command-line, we can replace it with `--breaks none`

### DIFF
--- a/include/vrv/toolkit.h
+++ b/include/vrv/toolkit.h
@@ -626,13 +626,6 @@ public:
     ///@{
 
     /**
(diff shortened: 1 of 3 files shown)


rettinghaus:
Well, I used it in Python, but fine.

lpugin:
Yes, people who use it in Python need to change `.skipLayoutOnLoad()` to `.setOptions({ "breaks": "none"})`  - or add this option when they set some.

rettinghaus (approved):

PR #3232, opened 2023-01-26
Change option keyword

### LINKED ISSUE #3231: ResetXmlIdSeed
The toolkit references the` --xml-id-checksum` option, yet this is not available in the toolkit. 

And why is this needed at all when you can simply `setOptions({"xmlIdSeed": value})`?

rettinghaus:
I don't know, but this makes it clearer to me. Closes #3231


lpugin:
When options are referred to in the documentation we normally use the command line spelling (e.g., see `--spacing-staff` [here](https://book.verovio.org/advanced-topics/layout-options.html)). I would suggest to leave it the same here for consistency.

PR #3234, opened 2023-01-27
Various adjustments to the code style (no functional changes)

lpugin:
* Use c++11 range loop when possible
* Use explicit types instead of auto for facilitating code understanding
* Revert changes yielding warnings with `size_t` change to `int`

### DIFF
--- a/src/chord.cpp
+++ b/src/chord.cpp
@@ -445,8 +445,8 @@ bool Chord::IsVisible() const
     // if the chord doesn't have it, see if all the children are invisible
     const ListOfConstObjects &notes = this->GetList(this);
 
-    for (const Object *iter : notes) {
-        const Note *note = vrv_cast<const Note *>(iter);
+    for (const Object *object : notes) {
+        const Note *note = vrv_cast<const Note *>(object);
         assert(note);
         if (!note->HasVisible() || note->GetVisible() == BOOLEAN_true) {
             return true;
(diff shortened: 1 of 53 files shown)


src/vrv.cpp
```diff
@@ -180,10 +180,8 @@ LogLevel StrToLogLevel(const std::string &level)
 
 bool LogBufferContains(const std::string &s)
 {
-    std::vector<std::string>::iterator iter = logBuffer.begin();
-    while (iter != logBuffer.end()) {
-        if ((*iter) == s) return true;
-        ++iter;
+    for (const std::string &logStr : logBuffer) {
+        if (logStr == s) return true;
```
DavidBauer1984, line 184: Looks much cleaner. Even nicer would be to use `std::any_of` here. But maybe that's something for another refactoring chapter.

DavidBauer1984 (approved):
Looks good as far as I can see. I've added some small comments with optional improvements.

src/vrv.cpp
lpugin: Yes, I'll leave it like this.

src/chord.cpp
```diff
@@ -515,7 +515,7 @@ int Chord::AdjustOverlappingLayers(const Doc *doc, const std::vector<LayerElemen
     int actualElementsInUnison = 0;
 
     // process each note of the chord separately, storing locations in the set
-    for (auto iter : notes) {
+    for (Object *iter : notes) {
```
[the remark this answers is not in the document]
lpugin: Good point

PR #3249, opened 2023-02-03
Code style updates

rettinghaus:
* prefer C++11 `using` keyword for alias declaration
* use C++11 braced-init-list
* replace deprecated headers

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -95,6 +95,7 @@
       "familyName": "Regimbal"
     },
     {
+      "@id": "https://orcid.org/0000-0001-5523-2085",
       "@type": "Person",
       "givenName": "Simon",
       "familyName": "Waloschek"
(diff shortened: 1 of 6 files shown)


lpugin:
Thanks. Beyond the preference, why would you change `typedef` to `using`? What difference does it make? As far as I know, `using` is useful when doing templates, but we do not do that.

rettinghaus:
It's really just about readability.

ahankinson:
I can't actually see anything in PEP 257 that would necessitate those changes? Would you mind pointing out the line or sentence?

rettinghaus:
> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

https://peps.python.org/pep-0257/#one-line-docstrings

rettinghaus:
rebased, staying with `typedef` for the time being.

PR #3267, opened 2023-02-13
Improve style of generated libmei

rettinghaus:
This brings some small changes for the generated LibMEI to be better in line with our clang-format style.


libmei/dist/attclasses.h
```diff
@@ -297,6 +297,6 @@ enum AttClassId {
     ATT_CLASS_max
 };
 
-} // namespace
+} // namespace vrv
```
ahankinson, line 300: I think this was removed intentionally by @lpugin, due to the namespace option that we can now pass in.

lpugin: Yes. We can remove ` // namespace` if you want

lpugin (changes requested):
Revert changes on namespace end

libmei/dist/attclasses.h
rettinghaus: It's generated (and also put there automatically by clang-format).

lpugin: I know. But we do not run clang format on the libMEI code - for the obvious reason that (most of) the code is generated.

rettinghaus: But we *could* run clang-format now. ;-) 
It doesn't hurt, and personally I think it's at least as useful as the  unneeded endif comment.

lpugin (approved):
OK, leave it like that

libmei/dist/attclasses.h
lpugin: > But we could run clang-format now. ;-)

No, because of line-length dependant formatting. If you run clang-format on the `libmei/dist` folder, you will see things like that because the code is generated: 
```
-    data_ACCIDENTAL_GESTURAL_extended StrToAccidentalGesturalExtended(const std::string &value, bool logWarning = true) const;
+    data_ACCIDENTAL_GESTURAL_extended StrToAccidentalGesturalExtended(
+        const std::string &value, bool logWarning = true) const;
```
Of course, we could add `// clang-format off` everywhere, or even worse, make libmei re-implement clang-format and check for the line length.

Alternatively, we could force users to run clang-format after libmei, but you seem to agree that this is not a good idea since you adjusted the libmei output to align with it.

> It doesn't hurt, and personally I think it's at least as useful as the unneeded endif comment.

Since when are comments needed  ? This 
```
} // namespace
```
is nearly as good as 
```
} // namespace vrv
```
but clearly much better than
```
}
```
So I disagree an the unneeded comment is as needed as the namespace name

PR #3269, opened 2023-02-14
Reduce memory and executable size

DavidBauer1984:
This PR reduces memory consumption of Verovio by 10-15% through optimisations in the generated LibMEI code. 
Here are the sizes of selected types in bytes:

| Class | Before | After | Reduction |
| ----- | ------ | ----- | --------- |
| Accid | 1136 | 984 | 13.4% |
| Measure | 3216 | 2984 | 7.2% |
| Note | 1792 | 1480 | 17.4% |

The reduction is mainly achieved by generating the attributes non-polymorphic with protected constructor and destructor. To still be able to use them outside the class hierarchy (as required i.e. in `iomei.cpp`), we also generate instantiable versions named `ExtAtt*`.

An interesting side effect of this is that the size of the executable is also reduced:

| Before | After | Reduction |
| ------ | ----- | ------- |
| 15.924 MB | 14.930 MB | 6.2% |

Runtime is not effected (neither positive nor negative). This PR addresses ideas discussed before in https://github.com/rism-digital/verovio/discussions/3065 .

### DIFF
--- a/libmei/addons/att.h
+++ b/libmei/addons/att.h
@@ -23,24 +23,24 @@ namespace vrv {
 
 /**
  * This is the base class for all MEI att classes.
- * It is not an abstract class but it should not be instanciated directly.
+ * It cannot be instanciated outside the Att class hierarchy.
  * The att classes are generated with the libmei parser for Verovio.
  */
-class Att : public AttConverter {
-public:
+class Att : public AttConverterBase {
+protected:
     /** @name Constructors and destructor */
     ///@{
-    Att();
-    virtual ~Att();
+    Att() = default;
+    ~Att() = default;
     ///@}
 
+public:
     static data_ACCIDENTAL_WRITTEN AccidentalGesturalToWritten(data_ACCIDENTAL_GESTURAL accid);
     static data_ACCIDENTAL_GESTURAL AccidentalWrittenToGestural(data_ACCIDENTAL_WRITTEN accid);
 
     static data_STAFFREL StaffrelBasicToStaffrel(data_STAFFREL_basic staffrelBasic);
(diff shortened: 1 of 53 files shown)


libmei/addons/att.h
```diff
@@ -23,24 +23,24 @@ namespace vrv {
 
 /**
  * This is the base class for all MEI att classes.
- * It is not an abstract class but it should not be instanciated directly.
+ * It cannot be instanciated outside the Att class hierarchy.
```
ahankinson, line 26:
```suggestion
 * It cannot be instantiated outside the Att class hierarchy.
```

lpugin (commented):
An afterthought: I am not sure the `ExtAtt` name for the derived classes is very explanatory. What about `AttCleffingLogInst` instead of `ExtAttCleffingLog` ? Or `InstCleffingLog` or `InstAttCleffingLog`? Any other ideas?

DavidBauer1984:
Yes, I was not happy with it either. I did not want to mix prefix and suffix, since there could actually be an attribute ending on `Ext` or `Inst` which would trigger more confusion. But if you prefer, we can change the prefix `ExtAtt` to `InstAtt`. Or maybe someone has an even better idea.

lpugin:
It is very minor, but I think that `Inst` would be a better prefix. I'll make the change. Thank you for the actual improvement!

rettinghaus:
I'm just thinking out loud if it would make sense to make it configurable in LibMEI if a class should be instantiable?

lpugin:
I checked and it would make a difference of about 20K in the JS build so I don't think it is significant enough

PR #3270, opened 2023-02-15
Generator script improvements

DavidBauer1984:
This PR improves the usage of the LibMEI generator script a bit.
- The configuration is now a required argument. Before the script was crashing when leaving it out.
- Running the script from another directory now works, i.e.
`python3 tools/parseschema2.py -c tools/config.yml mei/develop/mei-verovio_compiled.odd`

### DIFF
--- a/libmei/tools/cpp.py
+++ b/libmei/tools/cpp.py
@@ -1214,7 +1214,7 @@ public:
 
 
 def create_basic_validator(configure: dict, outdir: Path):
-    basic_path = Path(configure["basic_odd"])
+    basic_path = Path(configure["config_dir"]) / Path(configure["basic_odd"])
     with basic_path.open("r") as basic_schema:
         bschema = MeiSchema(basic_schema, resolve_elements=True)
 
@@ -1268,7 +1268,8 @@ def create(schema, configure: dict) -> bool:
     global DATATYPES
     lg.debug("Begin Verovio C++ Output ...")
 
-    datatypes: Path = Path(configure["datatypes"])
+    config_dir: Path = Path(configure["config_dir"])
+    datatypes: Path = config_dir / Path(configure["datatypes"])
     if not datatypes.exists():
         lg.error("Datatypes file could not be found")
         return False
@@ -1276,13 +1277,13 @@ def create(schema, configure: dict) -> bool:
     DATATYPES = yaml.safe_load(datatypes.open("r"))
 
     ns: str = configure["namespace"]
-    outdir: Path = Path(configure["output_dir"])
+    outdir: Path = config_dir / Path(configure["output_dir"])
 
     # if we output element classes, then we do not output the attmodule class
     if configure["elements"]:
         create_element_classes(ns, schema, outdir)
         if addons := configure.get("addons_dir"):
-            copy_addons(ns, Path(addons), outdir)
+            copy_addons(ns, config_dir / Path(addons), outdir)
         else:
             lg.warning("Element output was configured, but no addons directory was provided.")
     else:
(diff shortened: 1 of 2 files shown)


libmei/tools/cpp.py
```diff
@@ -1214,7 +1214,7 @@ class MEIBasic {{
 
 
 def create_basic_validator(configure: dict, outdir: Path):
-    basic_path = Path(configure["basic_odd"])
+    basic_path = Path(configure["config_dir"]) / Path(configure["basic_odd"])
```
ahankinson, line 1217: It's a style thing, but I really really dislike the division operator overrides for path segments.

libmei/tools/parseschema2.py
```diff
@@ -56,7 +56,7 @@ def main(configure: dict) -> bool:
     p = ArgumentParser(usage='%(prog)s [-c config path] [compiled odd path]')
 
     p.add_argument("compiled", help="A compiled ODD file", type=Path)
-    p.add_argument("-c", "--config", help="Path to a config file", type=Path)
+    p.add_argument("-c", "--config", help="Path to a config file", required=True, type=Path)
```
lpugin, line 59: The https://github.com/rism-digital/verovio/blob/develop/libmei/README.md needs to be adjusted accordingly

ahankinson:
Are you assuming that the config file *directory* is the same one that has the addons, etc? 

The intention of the `-c` argument was to provide a path to a config file, but the config file already contains paths to directories for addons, etc. So you should be changing those in the config file, instead of appending the config file's parent directory to their paths.

libmei/tools/cpp.py
DavidBauer1984: Should I use `joinpath` or what exactly is the best alternative?

ahankinson: In this specific case, something like:

`Path(configure["config_dir"], configure["basic_odd"])`

would do the same thing.

libmei/tools/parseschema2.py
DavidBauer1984: We can probably also make the config argument optional and use a default path when it is left out. How would you prefer? Anyway the script should not crash.

ahankinson: I did have a default path in there before, but I think it got missed when I did a refactoring. I can add that back in.

DavidBauer1984:
There are two issues here:
1. When running the script without the config argument, it crashes:
Crash
2. When running the script from the parent directory, it fails since the relative paths in the config file are not properly resolved:
Parent_Dir

I stumbled upon these when trying to run the generator for the first time. From a usability perspective, it would be nice to get those fixed at some point. If you want, you can close this PR and fix it later.

ahankinson:
I'll fix it today.

DavidBauer1984:
Great, thank you!

rettinghaus:
It would also be good to have a `requirements.txt` here or at least mention the use of `pyyaml` and `lxml` in the `README`.

ahankinson:
> It would also be good to have a requirements.txt here or at least mention the use of pyyaml and lxml in the README.

Yeah, +1 to that. The problem was that we actually have two Python projects in Verovio -- the 'fonts' setup, and now 'libmei'. It might make sense to combine them BUT the fonts setup depends on svgpathtools, which depends on scipy, which requires a FORTRAN compiler with BLAS and LAPACK, and it's a pain to install. So I didn't want to make the libmei stuff dependent on that.

Talking with @lpugin about this, he's fine with adding another separate set of Python project files to the libmei directory, so I'll do that when I fix the config file paths.

I've been using `pyproject.toml` files, so I would prefer to keep doing that.

PR #3273, opened 2023-02-15
Fixed config deps

ahankinson:
- Adds python project files for installing dependencies
 - Moves the config and datatypes files outside of the Python module
 - Updates the default config to work from the libmei directory
 - Updates README with more instructions and info
 - Adds a default parameter if the `-c` option is not specified
 - Removes an outdated comment (whoops)

Refs #3270


ahankinson:
Tagging @DavidBauer1984 as well.

rettinghaus (approved):
Looks fine to me.

rettinghaus:
Maybe also update the README in fonts? The `export LDFLAGS="-L/opt/homebrew/opt/lapack/lib"` line appears twice.

lpugin (approved):

libmei/tools/parseschema2.py
```diff
@@ -56,7 +56,7 @@ def main(configure: dict) -> bool:
     p = ArgumentParser(usage='%(prog)s [-c config path] [compiled odd path]')
 
     p.add_argument("compiled", help="A compiled ODD file", type=Path)
-    p.add_argument("-c", "--config", help="Path to a config file", type=Path)
+    p.add_argument("-c", "--config", default="./config.yml", help="Path to a config file", type=Path)
```
DavidBauer1984, line 59: Now the default config is not working when running the script from the tools directory:
FromTools
When running it from `libmei` with default config, it works. I would still prefer to have more flexibility here.

Apart from this, it looks fine to me.

ahankinson: So you would like the paths to resolve based on the parent directory of the `config.yml` file, rather than the `cwd` from where you are running the script?

ahankinson:
> Maybe also update the README in fonts? The export LDFLAGS="-L/opt/homebrew/opt/lapack/lib" line appears twice.

Yeah; although newer versions of scipy seem to not need this now. But I will fix it.

libmei/tools/parseschema2.py
DavidBauer1984: I think it would be nice to resolve the relative paths into reasonable absolute paths, i.e. the config default path should point to the location where the config file actually sits, independent of where the script is called from. Also relative paths in the config file should be considered relative to the config file location. I think using the `cwd` to resolve relative paths is not a good idea, in principle the `cwd` could be anything.

It's a nice-to-have, feel free to merge this PR without this.

PR #3275, opened 2023-02-15
Fix page head regression

DavidBauer1984:
This PR fixes a regression introduced in a recent refactoring.

| Before | After |
| ------ | ----- |
| Before | After |


craigsapp:
Thanks!  I have recently been getting a similar problem with headers when creating PDFs which is likely the same problem (but I hadn't tried to track down the source yet).

PR #3279, opened 2023-02-18
Fix line breaks [skip-ci]

rettinghaus:
This small PR changes line breaks from CRLF to LF unifying the code files, reducing size by 15549 bytes.

Maybe we want to document that in the reference book and / or add a a `.gitattributes` file?


lpugin:
* Leave the library files untouched
* Leave view_neume.cpp untouch too for now because  I need to merge another branch first

rettinghaus:
pugixml provides a release with unix line endings: https://pugixml.org/

lpugin:
I am OK to change to unix line endings when there is a new release of pugixml. I agree to it is not great to have different line endings in the repository, but I find it even more annoying that changing the line endings yields a entire file diff. This makes it difficult to track changes. Luckily, different line endings do not cause major problems.

PR #3281, opened 2023-02-20
try to calculate missing divisions

### LINKED ISSUE #3263: MusicXML: Absent divisions attribute causes errors in visual spacing and timemap output
**Describe the problem**

Major notation programs don't always include `<divisions>` information correctly in the `<measure><attributes>...` block.
Most of the time, this absence of the division information passes silently, not causing any rendering problems: The visual output looks fine, timemap output is correct. Obviously, Verovio infers the division information needed to handle the spacing correctly.

However, if there is for example a dynamic mark in the middle or end of the bar, the infered division calculation breaks, resulting in 
- incorrect placement of the dynamic mark
- incorrect "extra spacing" in the end of the bar
- incorrect timemap output 

incorrect

**To Reproduce**
1. Load the attached musicXml file
2. Render page X
3. See errors as described above

**Expected behavior**

The fact that Verovio handles absent divisions info without problem as long as there isn't a dynamic mark or anything like that, makes it hard to spot why the layout and the timemap output isn't correct.

The best case would of course be that Verovio could infer the divisions information needed, and make the calculations correctly based on that. The second best would be a clear console output telling that measure>attribute>divisions information is missing and that the necessary calculations for layout and timemap output couldn't be performed correctly.

The current behaviour, with Verovio "silently" rendering incorrectly and outputting incorrect timemap info, is unfortunate.

**Verovio information**
 - 3.15.0-dev

```xml
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 4.0 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="4.0">
    <movement-title>Missing divisions problem</movement-title>
    <part-list>
        <score-part id="P1">
        </score-part>
    </part-list>

    <part id="P1">
        <measure number="1">
            <attributes>
                <key>
                    <fifths>0</fifths>
                </key>
                <clef>
                    <sign>G</sign>
                    <line>2</line>
                </clef>
                <time>
                    <beats>4</beats>
                    <beat-type>4</beat-type>
                </time>
                <!-- <divisions>2</divisions> -->
            </attributes>

            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <duration>8</duration>
                <voice>1</voice>
                <type>whole</type>
                <stem>down</stem>
            </note>
        </measure>

        <measure number="2">
            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <voice>1</voice>
                <type>quarter</type>
                <duration>2</duration>
                <stem>up</stem>
            </note>

            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <voice>1</voice>
                <type>quarter</type>
                <duration>2</duration>
                <stem>up</stem>
            </note>

            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <voice>1</voice>
                <type>quarter</type>
                <duration>2</duration>
                <stem>up</stem>
            </note>

            <direction>
                <direction-type>
                    <dynamics>
                        <mf/>
                    </dynamics>
                </direction-type>
            </direction>

            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <voice>1</voice>
                <type>quarter</type>
                <duration>2</duration>
                <stem>up</stem>
            </note>

        </measure>

        <measure number="3">
            <note>
                <pitch>
                    <step>G</step>
                    <octave>4</octave>
                </pitch>
                <duration>8</duration>
                <voice>1</voice>
                <type>whole</type>
                <stem>down</stem>
            </note>
        </measure>

    </part>

</score-partwise>


```

rettinghaus:
closes #3263

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -467,7 +467,7 @@ private:
     /* measure repeats */
     bool m_slash = false;
     /* MIDI ticks */
-    int m_ppq = 0;
+    int m_ppq = -1;
     /* measure time */
     int m_durTotal = 0;
     /* measure time */
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -467,7 +467,7 @@ private:
     /* measure repeats */
     bool m_slash = false;
     /* MIDI ticks */
-    int m_ppq = 1;
+    int m_ppq = 0;
     /* measure time */
     int m_durTotal = 0;
     /* measure time */
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2681,7 +2681,7 @@ void MusicXmlInput::ReadMusicXmlNote(
     const int duration = node.child("duration").text().as_int();
     const int noteStaffNum = node.child("staff").text().as_int();
     const pugi::xml_node rest = node.child("rest");
-    if (m_ppq == 0 && (duration && !typeStr.empty())) {
+    if (m_ppq < 0 && (duration && !typeStr.empty())) {
         // if divisions are missing, try to calculate
         m_ppq = (double)duration * pow(2, ConvertTypeToDur(typeStr) - 2) / 4;
     }
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2681,7 +2681,7 @@ void MusicXmlInput::ReadMusicXmlNote(
     const int duration = node.child("duration").text().as_int();
     const int noteStaffNum = node.child("staff").text().as_int();
     const pugi::xml_node rest = node.child("rest");
-    if (m_ppq < 0 && (duration && !typeStr.empty())) {
+    if (m_ppq < 0 && duration && !typeStr.empty()) {
         // if divisions are missing, try to calculate
         m_ppq = (double)duration * pow(2, ConvertTypeToDur(typeStr) - 2) / 4;
     }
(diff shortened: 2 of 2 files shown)


include/vrv/iomusxml.h
```diff
@@ -467,7 +467,7 @@ class MusicXmlInput : public Input {
     /* measure repeats */
     bool m_slash = false;
     /* MIDI ticks */
-    int m_ppq = 1;
+    int m_ppq = 0;
```
ahankinson, line 470: can `m_ppq` ever be the denominator? If so, then this could lead to divide-by-zero.

Such as here? https://github.com/rism-digital/verovio/blob/6587876e2a7b336a85e2472c7d3bc54776ce2e9a/src/iomusxml.cpp#L528

ahankinson: Maybe `-1` is better if you want to indicate that it's unset?

rettinghaus: Good point!

ahankinson:
It looks ok to me, but @lpugin is on vacation this week, so I won't merge it until he gets back and has a look too.

src/iomusxml.cpp
```diff
@@ -2681,6 +2681,11 @@ void MusicXmlInput::ReadMusicXmlNote(
     const int duration = node.child("duration").text().as_int();
     const int noteStaffNum = node.child("staff").text().as_int();
     const pugi::xml_node rest = node.child("rest");
+    if (m_ppq < 0 && (duration && !typeStr.empty())) {
```
lpugin, line 2684: `()` inside are not necessary

rettinghaus: Well spotted!

PR #3282, opened 2023-02-20
Update functor branch to current develop

DavidBauer1984:
This PR merges _develop_ into _develop-functors_ and resolves various conflicts.


lpugin:
Any thing in particular I should look at?

DavidBauer1984:
Not really. Maybe run some examples or tests, if you have something. Behaviour should be the same as on _develop_ . 

The more functors are moved into classes, the more conflicts arise. So we need to merge on a regular basis.

PR #3288, opened 2023-02-23
Ensure unique xml IDs

DavidBauer1984:
So far Verovio generated xml IDs from random 32 bit numbers. Interestingly, even with good pseudo random generators, there is a fair chance to obtain duplicates pretty soon. The probability of having at least one duplicate already hits 1% after 9300 generated IDs:
birthday-prob
This is known as [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem). It is not just theory, duplicated IDs were found by Klaus in larger MEI files.

This PR drops the randomness in the ID generation and uses a hash function to calculate deterministic IDs. This ensures uniqueness, since the [hash function used is bijective](https://stackoverflow.com/questions/664014/what-integer-hash-function-are-good-that-accepts-an-integer-hash-key). Thus we can have short and unique xml IDs without the need to switch to longer 128 bit UUIDs.

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -651,6 +651,8 @@ public:
 
     static std::string GenerateRandID();
 
+    static uint32_t Hash(uint32_t number, bool reverse = false);
+
     static bool sortByUlx(Object *a, Object *b);
 
     /**
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -12,7 +12,6 @@
 #include <functional>
 #include <iterator>
 #include <map>
-#include <random>
 #include <string>
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


ahankinson:
A couple questions:

 - We use the CRC hash of the input as the seed for generating unique, but consistent, IDs. Will that be the same? (`--xml-id-checksum`). 
 - What do the new hash IDs look like?

DavidBauer1984:
- The CRC hash of the input is used to determine a seed specific to the input file. For the same seed (stemming from the same input) the generated IDs will be the same. Each seed corresponds to a sequence of IDs and the mapping _Seed_ -> _ID sequence_ is deterministic.
- For similar seeds (i.e. 123 and 124) the generated IDs will look very different. The trick is to apply the hash function once to the seed (L1204 in object.cpp).
- The new hash IDs _look_ the same as before, i.e.:
Look

PR #3317, opened 2023-03-09
Remove exceptions when setting options

lpugin:
Instead of catching `std::stoi` and `std::stod` exception, we preliminary check the content of the string with a regex.

### DIFF
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -57,6 +57,11 @@ LogLevel StrToLogLevel(const std::string &level);
 bool AreEqual(double dFirstVal, double dSecondVal);
 
 /**
+ * Utility for check the string is a number
+ */
+bool IsNumber(const std::string value);
+
+/**
  * Extract the ID from any URI
  */
 std::string ExtractIDFragment(std::string refID);
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -57,9 +57,9 @@ LogLevel StrToLogLevel(const std::string &level);
 bool AreEqual(double dFirstVal, double dSecondVal);
 
 /**
- * Utility for check the string is a number
+ * Utility to check if the string is a number
  */
-bool IsNumber(const std::string value);
+bool IsNumber(const std::string &value);
 
 /**
  * Extract the ID from any URI
--- a/include/vrv/vrv.h
+++ b/include/vrv/vrv.h
@@ -57,9 +57,14 @@ LogLevel StrToLogLevel(const std::string &level);
 bool AreEqual(double dFirstVal, double dSecondVal);
 
 /**
- * Utility to check if the string is a number
+ * Utility to check if the string is a valid integer for std::stoi
  */
-bool IsNumber(const std::string &value);
+bool IsValidInteger(const std::string &value);
+
+/**
+ * Utility to check if the string is valid double for std::stod
+ */
+bool IsValidDouble(const std::string &value);
 
 /**
  * Extract the ID from any URI
--- a/src/vrv.cpp
+++ b/src/vrv.cpp
@@ -243,7 +243,7 @@ bool IsValidInteger(const std::string &value)
 bool IsValidDouble(const std::string &value)
 {
     // Accept "1.0" " 1.0 " ".0"  "1." "+1.0" "-1.0"
-    std::regex re("^\\s*[+-]?(\\d+\\.?\\d*|\\.?\\d+)\\s*$");
+    std::regex re("^\\s*[+-]?(?:\\d+\\.?\\d*|\\.\\d+)\\s*$");
     return std::regex_match(value, re);
 }
 
--- a/src/vrv.cpp
+++ b/src/vrv.cpp
@@ -235,12 +235,14 @@ bool AreEqual(double dFirstVal, double dSecondVal)
 
 bool IsValidInteger(const std::string &value)
 {
+    // Accept "1" " 1 " "+1" "-1" "1." "1.0"
     std::regex re("^\\s*[+-]?\\d+\\.?\\d*\\s*$");
     return std::regex_match(value, re);
 }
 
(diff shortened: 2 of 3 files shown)


include/vrv/vrv.h
```diff
@@ -56,6 +56,11 @@ LogLevel StrToLogLevel(const std::string &level);
  */
 bool AreEqual(double dFirstVal, double dSecondVal);
 
+/**
+ * Utility for check the string is a number
+ */
+bool IsNumber(const std::string value);
```
DavidBauer1984, line 62: Please pass as const reference. Also the comment above is a bit bumpy.

DavidBauer1984 (approved):
Looks good!

src/vrv.cpp
```diff
@@ -232,6 +233,12 @@ bool AreEqual(double dFirstVal, double dSecondVal)
     return std::fabs(dFirstVal - dSecondVal) < 1E-3;
 }
 
+bool IsNumber(const std::string &value)
+{
+    std::regex re("^\\s*[+-]?(\\d+\\.?|\\.?\\d+)\\d*\\s*$");
```
ahankinson, line 238: Tested a slight modification:

`^\s*[+-]?\d*?\.?\d+\s*$`

Advantages:
 - Removes the capturing groups `()`, since we don't need to use the result
 - Removes the `|` clause and replaces it with an optional leading digit

Test cases (Matches):

```
100
100.0
0.99
-19
-19.9
 -19.9 
-.5
-99999999
```

Does not match:

```
foo
foo.bar
0foo
0foo.bar
foo.0bar
foo.bar0
```

lpugin: I think that would fail on `1.`

lpugin: And if you do `\d*\.?\d*` then is passed on `.` which is not great, and why we need the `|`. But we can probably do
`\d+\.?|\.?\d+`

lpugin:
We actually need to distinct regex for integers and doubles

src/vrv.cpp
craigsapp: > Removes the capturing groups (), since we don't need to use the result

If you don't want to create a capture group with `(xxx)`, then you can use `(?:xxx)`, which is called a "non-capture group".  So the original re would be:

`std::regex re("^\\s*[+-]?(?:\\d+\\.?|\\.?\\d+)\\d*\\s*$");`

removing the capture group (which should result in faster processing since it would not need to be saved in the search results).

> `^\s*[+-]?\d*\.?\d+\s*$`
> -0000.

Do you want to exclude cases where there are digits before the decimal point but not afterwards?  I always avoid writing numbers like that, but it seems like a legal number that might surprise the general user if it is invalid.

-------------------------

Also, it might be useful to allow the following cases by removing the `\s*$` part of the regular expression:

```
0foo
0foo.bar
0000000.foo
```

In other words, if there is a number at the start of the string and some text afterwards, would you like to keep the number, and ignore the trailing text?  That seems more graceful.

And I presume that cases such as `foo` for a number parameter would  result in the default value remaining (i.e., `foo` is ignored), and there would be a warning message generated about bad parameter contents.  If `0foo` is allowed, then perhaps also a warning message could be generated complaining that "foo" is present, but also saying that the "0" will be used for the numeric parameter (but that warning might not be needed).

Also note in C++ you can express numbers with exponent notation:

```
double num = 1.23e5; 
double num = 1.23E5; 
```

which are two ways of expressing the number: 123000.0.   I don't see any need for that, but maybe some software-generate parameters might do that for very small numbers.

Here is an example C++ program:

```c++
#include <iostream>
using namespace std;
int main(void) {
   cout << 0.0000000000000000000123 << endl;
	return 0;
}
```

which will output:

`1.23e-20`

The `strtod()` function knows how to parse such strings:

```cpp
#include <iostream>
#include <stdlib.h>
using namespace std;
int main(void) {
	double value = strtod("1.23e-20", NULL);
	cout << value << endl;
	return 0;
}
```

which outputs `1.23e-20`.

Also note that `strtod()` ignores any trailing non-numeric string contents:

```cpp
#include <iostream>
#include <stdlib.h>
using namespace std;
int main(void) {
	double value = strtod("1.23e-20SOMETEXT", NULL);
	cout << value << endl;
	return 0;
}
```

The output for this example program will also be `1.23e-20`.

Also note that `strtod()` will automatically ignore leading spaces:

```cpp
#include <iostream>
#include <stdlib.h>
using namespace std;
int main(void) {
	double value = strtod("   	1.23e-20SOMETEXT", NULL);
	cout << value << endl;
	return 0;
}
```

Also note that if you give `strtod()` text with no leading numeric characters, then it will assign a value of 0 to the string:

```cpp
#include <iostream>
#include <stdlib.h>
using namespace std;
int main(void) {
	double value = strtod("SOMETEXT123", NULL);
	cout << value << endl;
	return 0;
}
```

In this case the program outputs `0`.





-------------------------------------

Somewhat related:  When happens if a numeric parameter is out of range?  It would be useful to set a parameter value to the max or min allowed value, and then given a warning message saying that the parameter was trimmed to the max or min value.

--------------------------------------

Tangential:

`std::regex re("^\\s*[+-]?(?:\\d+\\.?|\\.?\\d+)\\d*\\s*$");`

You might want to consider a raw string in this case:

`std::regex re(R"(^\s*[+-]?(?:\d+\.?|\.?\d+)\d*\s*$)");`

I tried to use raw strings several years ago in the iohumdrum.cpp code, but clang-format would always put a space after the `R` that I had to always fix afterwards.  Perhaps with more recent versions of clang-format that bug is fixed?

Note that raw strings are useful to quote JSON data in C++ code:

```
string jsonData = R"json({
    "foo": "bar"
})json";
```

Where the `json` arbitrary label is added to prevent needing to escape `)"` in the contents.

ahankinson:
That sounds good to me

craigsapp:
That sounds good (@craigsapp)

PR #3342, opened 2023-03-22
Humdrum: Hide rests with extra duration with `yy` signifier

WolfgangDrescher:
A small code review would be good for this one, as I do not really know what `embedQstampInClass()`, `appendElement(elements, pointers, irest)` and `convertRhythm(irest, trest)` does. But I took it from a few lines above and it seems to work.

<details>
  <summary>Click for MEI for the demo score</summary>

```tsv
**kern	**text	**kern	**text	**kern	**text
*clefC3	*	*clefC2	*	*clefG2	*
*k[]	*	*k[]	*	*k[]	*
*M2/1	*	*M2/1	*	*M2/1	*
*met(C|)	*	*met(C|)	*	*met(C|)	*
=1	=1	=1	=1	=1	=1
1r	.	00ryy	.	1dd	Se-
1g	Se-	.	.	2dd	-lig
.	.	.	.	2dd	zu
=2	=2	=2	=2	=2	=2
2g	-lig	.	.	1b	prei-
2g	zu	.	.	.	.
2e	prei-	.	.	2cc	-sen
2f	-sen	.	.	2a	ist
=3	=3	=3	=3	=3	=3
1e	ist	0r	.	2g	der
.	.	.	.	2a	.
1d	der	.	.	4b	.
.	.	.	.	4cc	.
.	.	.	.	8b	.
.	.	.	.	8a	.
.	.	.	.	4b	.
=4	=4	=4	=4	=4	=4
1c	mann/	1g	Se-	2ee	mann/
.	.	.	.	2ee	se-
2r	.	2g	-lig	2dd	-lig
[2c	ist	2g	zu	2ee	zu
=5	=5	=5	=5	=5	=5
*-	*-	*-	*-	*-	*-
```
</details>

Before:

Bildschirm­foto 2023-03-22 um 21 49 46

After:

Bildschirm­foto 2023-03-22 um 21 49 27

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -11244,20 +11244,21 @@ bool HumdrumInput::fillContentsOfLayer(int track, int startline, int endline, in
                     // (so not an mRest), and update the visual duration
                     // of the rest because there will be invisible rests
                     // added in later measure(s).
-					if (trest->find("yy") != std::string::npos) {
-						Space *irest = new Space();
-						if (m_doc->GetOptions()->m_humType.GetValue()) {
-							embedQstampInClass(irest, trest, *trest);
-						}
-						setLocationId(irest, trest);
-						appendElement(elements, pointers, irest);
-						convertRhythm(irest, trest);
-					} else {
-						Rest *rest = new Rest;
-						setLocationId(rest, trest);
-						appendElement(layer, rest);
-						convertRest(rest, trest, -1, staffindex);
-					}
+                    if (trest->find("yy") != std::string::npos) {
+                        Space *irest = new Space();
+                        if (m_doc->GetOptions()->m_humType.GetValue()) {
+                            embedQstampInClass(irest, trest, *trest);
+                        }
+                        setLocationId(irest, trest);
+                        appendElement(elements, pointers, irest);
+                        convertRhythm(irest, trest);
+                    }
+                    else {
+                        Rest *rest = new Rest;
+                        setLocationId(rest, trest);
+                        appendElement(layer, rest);
+                        convertRest(rest, trest, -1, staffindex);
+                    }
                 }
                 else {
                     std::cerr << "Strange error for adding rest " << trest << std::endl;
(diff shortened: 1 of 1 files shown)


craigsapp:
> A small code review would be good for this one, as I do not really know what `embedQstampInClass()`, `appendElement(elements, pointers, irest)` and `convertRhythm(irest, trest)` does. 

`embedQstampInClass()` Is used to embed a timestamp within note and rest elements in the SVG output of Verovio.   This is in turn used to allow harmonic and melodic navigation when click on a note in VHV.  There is an animated demo of this:

https://doc.verovio.humdrum.org/graphic/pitch/#graphical-notation-navigation

Try click on a note in a score.  the press the up/down/left/right arrow keys to navigate in the score melodically and harmonically.  If any note/rest in the score does not have qstamp, then the arrow navigation will probably get stuck (in which case `embedQstampInClass()` needs to be applied to that corresponding MEI element.

Here is an example inspection of a note element in the SVG output from verovio (when `--humtype` option is set):

```xml
<g id="note-L20F2" class="note qon-1_2 qoff-1 pname-c acc-n oct-4 b40c-2 b12c-0 ">
```

The class `qon-1_2` means that the notes starts 1/2 of a quarter note from the start of the score.  And `qoff-1` means that the note stops at 1 quarter note duration from the start of the score (so the note is a half note).

----------------------

`appendElement(elements, pointers, irest)` 

This is (very) important in order to inset a newly created element into the MEI xml hierarchy.  The elements/pointers is a vector that stores the current element into which musical content is being added.   This can be the staff, measure layer, beam, etc.  So in this case the `irest` is being inserted into the MEI-like data structure in the current element container that is active (which in this case is mostly likely a `<layer>` (but could be a `<beam>` or other things that could contain a rest (technically `<space>`).  If you don't append a newly created element, then it would be easy to see its lack in the data or the rendering (although invisible rests are invisible in the SVG output, especially at the end of a measure).

----------------

`convertRhythm(irest, trest)` It (seems) to convert  a `**kern` (`**recip`) rhythm into MEI parameters.

PR #3356, opened 2023-03-30
Fix warnings in humdrum code

rettinghaus:


craigsapp:
As this is an external library, these changes will be overwritten the next time I update these files.  So they should go to `craigsapp/humlib` issues instead.   I have made a note of them and can update the library files.  The implicit `int` type conversions are in the `deg` tool which is relatively recent (from December).  I don't know why I haven't noticed these warnings before.  `humlib.cpp` and `humlib.h` are combined source code files for easy copying into verovio, so fixing those files directly on humlib will also result in automatic overwriting (I should probably but a notice at the tops of these files about that).

lpugin:
OK. I added a reminder for @craigsapp https://github.com/craigsapp/humlib/issues/75

craigsapp:
Fixed in commit https://github.com/craigsapp/humlib/commit/9cdbe2793f8a1da1db11e55b6dddaeaea34ecf81

Strange that there were no int/unsigned int warnings when I compiled in `Apple clang version 13.1.6 (clang-1316.0.21.2.5)` as well as `gcc version 12.2.1 20221121 (Red Hat 12.2.1-4) (GCC)`.

I will submit the updated code as a PR.

PR #3360, opened 2023-04-03
Expose expansion map in Javascript toolkit

wergo:
This PR exposes the expansion map in the Javascript toolkit with a new method `renderToExpansionMap()` and `renderToExpansionMapFile()`. `renderToExpansionMap()` returns a JSON object of the expansion map (see PR #1152). 

Exposing the expansion map makes it easier to access the relation between elements in the different expanded repetitions (for expample of a MIDI file) while using the toolkit for other tasks.


lpugin:
Could you give a minimal example and the corresponding output? Thanks

wergo:
A minimal MEI file is this here: https://raw.githubusercontent.com/wergo/test-encodings/main/Repetition_Study.mei

`tk.setOptions({ expand: 'expansion_default'});`
`let expansionMap = tk.renderToExpansionMap();` 

Beginning of `expansionMap` (for each key, it shows all repetitions, the first always being the original notated): 
Bildschirmfoto 2023-04-03 um 16 45 14

wergo:
There is the method `getNotatedIdForElement(std::string xmlId)` that returns the id of the notated/original element. And there is `getExpansionIdsForElement(std::string xmlId)` that does the opposite: returns all element ids that a given element may be expanded to. 

These methods work, if the toolkit still has the data loaded that you are asking it about. However, if you request a MIDI file (and the corresponding expansionMap), the toolkit instance may already have different data loaded (such as in speed mode in mei-friend), while you still want to get some info from the expansionMap. That's why I think, exposing it to the user is usefull.

lpugin:
Retrieving an expansionmap that do not correspond the the data loaded seems to be a dangerous idea to me. This probably works for you now because of an inconsistency in Verovio, namely that that the expansion is reset only when loading a data file but not when loading a data string.
https://github.com/rism-digital/verovio/blob/71862699bc0d19606b039e6f4b2c523fc6a14ce6/src/toolkit.cpp#L314

This should be changed and the expansion map should also be reset when loading a data string. We can still add the extension map export implemented in the PR if you think it would still be useful after that - I think it would.

wergo:
`Retrieving an expansionmap that does not correspond the the data loaded seems to be a dangerous idea to me.`
This is certainly not my intention. The expansion map should always be cleared whenever new data is loaded.

(My use case was referring to retrieving a  MIDI rendering and the corresponding expansion map, and then use the toolkit for other purposes.)

This is something I forgot three years ago when creating the expansion map code. 

I have added the expansion resetting in  `Toolkit::LoadData(const std::string &data)`. Could you please check whether it is at the right position? Thanks a lot.

lpugin:
> I have added the expansion resetting in `Toolkit::LoadData(const std::string &data)`. Could you please check whether it is at the right position? Thanks a lot.

Yes, it is the right place. You should also remove it from `Toolkit::LoadFile` since that one call `LoadData`.

wergo:
Thanks a lot! Done in 4dbb863.

emscripten/npm/src/emscripten-proxy.js
```diff
@@ -102,6 +102,9 @@ function getToolkitFunction(VerovioModule, method) {
     // char *renderData(Toolkit *ic, const char *data, const char *options)
     mapping.renderData = VerovioModule.cwrap("vrvToolkit_renderData", "string", ["number", "string", "string"]);
 
+    // char *renderToExpansionMap(Toolkit *ic)
```
lpugin, line 105: We always have  a `["number"]` parameter even for function without any. The number is the pointer to the toolkit. See `getVersion()`

lpugin (commented):
You also need to add both functions to `./bindings/python/verovio.i`

wergo:
Thanks for this thorough review. I am not overseeing the entire Verovio architecture. 

I have copied/adapted the code for the Python bindings, but have not tested it (12dd996). 

In 6d80c62, I try to handle an empty expansion map so that the JSON parsing does not generate an error. Should we do this also for the Python scripts?

lpugin:
> In https://github.com/rism-digital/verovio/commit/6d80c629a48c1e8aae8d59083088020d36ec1c69, I try to handle an empty expansion map so that the JSON parsing does not generate an error. Should we do this also for the Python scripts?

That's a good point. Maybe we can just make sure that `RenderToExpansionmap` always returns "{}" ? This is should probably be change for other similar methods, such as `RenderToTimemap`

wergo:
That's better. Done in 955be89 for `RenderToExpansionmap()` and `RenderToTimemap()`. The other related method such as `getExpansionIdsForElement()` work fine with empty returns as far as I can oversee.

PR #3366, opened 2023-04-06
Improve header from MusicXML import

rettinghaus:
This refines the generation of `xml:id` values for the `meiHead` from MusicXML import to be in line with all other ids. Also it no suppresses those ids now when the `--remove-ids` option is being used.

Also it simplifies some remaining `pugi::node_pcdata` by using `text().set()`.

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -310,7 +310,7 @@ void Doc::GenerateMEIHeader(bool meiBasic)
         // projectDesc
         pugi::xml_node projectDesc = encodingDesc.append_child("projectDesc");
         pugi::xml_node p1 = projectDesc.append_child("p");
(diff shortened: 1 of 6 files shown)


src/doc.cpp
```diff
@@ -310,7 +310,7 @@ void Doc::GenerateMEIHeader(bool meiBasic)
         // projectDesc
         pugi::xml_node projectDesc = encodingDesc.append_child("projectDesc");
         pugi::xml_node p1 = projectDesc.append_child("p");
-        p1.text().set(StringFormat("MEI encoded with Verovio").c_str());
+        p1.text().set(StringFormat("MEI encoded with Verovio.").c_str());
```
lpugin, line 313: This is not a sentence, which is why there is no full stop.

rettinghaus: I does look like one to me.

PR #3371, opened 2023-04-11
Follow barline xRel

### LINKED ISSUE #2763: fermata + spacingLinear interaction
When the `spacingLinear` option is small, strange things happen with the horizontal placement of the fermata in the following example (the fermata should remain centered on the final barline):

linearSpacing = 0.10:

Screen Shot 2022-04-03 at 10 31 55 PM

linearSpacing = 0.08:

Screen Shot 2022-04-03 at 10 32 03 PM

linearSpacing = 0.06:

Screen Shot 2022-04-03 at 10 32 10 PM

linearSpacing = 0.04:

Screen Shot 2022-04-03 at 10 32 20 PM

linearSpacing = 0.02:

Screen Shot 2022-04-03 at 10 32 32 PM

linearSpacing = 0.00:

Screen Shot 2022-04-03 at 10 32 43 PM

<details>
<summary>
Click to view MEI data for above examples
</summary>

```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-04-03T22:32:49" version="3.10.0-dev-4a1d799">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m8opjql">
    <score xml:id="sbq1yuu">
     <scoreDef xml:id="s5ksdo4" midi.bpm="400.000000">
      <staffGrp xml:id="sb7gj0k">
       <staffDef xml:id="sbdrx1k" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="F" line="4" />
        <keySig xml:id="keysig-L3F1" pname="c" mode="major" sig="0" />
        <meterSig xml:id="metersig-L5F1" count="2" sym="cut" unit="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
       <staff xml:id="s4jt7vl" n="1">
         <note xml:id="note-L10F1" type="placed" dur="2" oct="3" pname="f" stem.dir="up" accid.ges="n" />
         <note xml:id="note-L11F1" type="placed" dur="2" oct="3" pname="e" stem.dir="up" accid.ges="n" />
        </layer>
        <layer xml:id="layer-L2F1N2" n="2">
         <note xml:id="note-L10F2" dur="1" oct="2" pname="g" accid.ges="n" />
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L12" n="2">
       <staff xml:id="staff-L12F1N1" n="1">
        <layer xml:id="layer-L12F1N1" n="1">
         <note xml:id="note-L13F1" dur="1" oct="3" pname="g" accid.ges="n" />
        </layer>
        <layer xml:id="layer-L12F2N2" n="2">
         <chord xml:id="chord-L13F2" dur="1">
          <note xml:id="note-L13F2S1" oct="3" pname="d" accid.ges="n" />
          <note xml:id="note-L13F2S2" oct="2" pname="g" accid.ges="n" />
         </chord>
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L15" right="end" n="2">
       <staff xml:id="staff-L15F1N1" n="1">
        <layer xml:id="layer-L15F1N1" n="1">
         <note xml:id="note-L16F1" dur="1" oct="3" pname="c" accid.ges="n" />
        </layer>
        <layer xml:id="layer-L15F2N2" n="2">
         <chord xml:id="chord-L16F2" dur="1">
          <note xml:id="note-L16F2S1" oct="2" pname="g" accid.ges="n" />
          <note xml:id="note-L16F2S2" oct="2" pname="e" accid.ges="n" />
          <note xml:id="note-L16F2S3" oct="2" pname="c" accid.ges="n" />
         </chord>
        </layer>
       </staff>
       <fermata xml:id="fermata-L18F1N1" staff="1" tstamp="3.000000" place="above" />
       <fermata xml:id="fermata-L18F1N2" staff="1" tstamp="3.000000" place="below" />
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

Monceber:
closes #2763 
- changed code to make sure that timestamp-only alignments with the same time as barline follow its xRel
image

### DIFF
--- a/src/adjustxposfunctor.cpp
+++ b/src/adjustxposfunctor.cpp
@@ -246,6 +246,15 @@ FunctorCode AdjustXPosFunctor::VisitMeasure(Measure *measure)
 
         m_measureTieEndpoints = measure->GetInternalTieEndpoints();
         measure->m_measureAligner.Process(*this);
+
+        // In some cases with low enough linear spacing xRel of barline might shift due to overlaps with previous
+        // alignments. Make sure that timestamp-only alignments that have the same time as barline follow its xRel. 
+        const Alignment *rightBar = measure->m_measureAligner.GetRightBarLineAlignment();
+        Alignment *rightBarDefault
+            = measure->m_measureAligner.GetAlignmentAtTime(rightBar->GetTime(), ALIGNMENT_DEFAULT);
+        if (rightBarDefault->HasTimestampOnly() && (rightBarDefault->GetXRel() != rightBar->GetXRel())) {
+            rightBarDefault->SetXRel(rightBar->GetXRel());
+        }
     }
 
     this->SetFilters(previousFilters);
--- a/src/adjustxposfunctor.cpp
+++ b/src/adjustxposfunctor.cpp
@@ -248,7 +248,7 @@ FunctorCode AdjustXPosFunctor::VisitMeasure(Measure *measure)
         measure->m_measureAligner.Process(*this);
 
         // In some cases with low enough linear spacing xRel of barline might shift due to overlaps with previous
-        // alignments. Make sure that timestamp-only alignments that have the same time as barline follow its xRel. 
+        // alignments. Make sure that timestamp-only alignments that have the same time as barline follow its xRel.
         const Alignment *rightBar = measure->m_measureAligner.GetRightBarLineAlignment();
         Alignment *rightBarDefault
             = measure->m_measureAligner.GetAlignmentAtTime(rightBar->GetTime(), ALIGNMENT_DEFAULT);


src/adjustxposfunctor.cpp
```diff
@@ -246,6 +246,15 @@ FunctorCode AdjustXPosFunctor::VisitMeasure(Measure *measure)
 
         m_measureTieEndpoints = measure->GetInternalTieEndpoints();
         measure->m_measureAligner.Process(*this);
+
+        // In some cases with low enough linear spacing xRel of barline might shift due to overlaps with previous
```
lpugin, line 250: Do you understand why the barline shifts? The fix works, but I am not sure I understand its logic. Isn't there somewhere a place to fix the issue more at its root?

Monceber: Yes. The issue here is that positioning of timestamp-only alignments is done by placing them proportionally to their time. However as this happens before any of the overlaps can be processed, this can lead to situations where timestamp alignment initially matches that of non-timestamp (in current case barline) but after overlaps are calculated actual position of the timstamp remains unchanged (since the element itself does not have any overlaps). This does not generally happen unless linear spacing is low enough, like in current example.
To illustrate it further, here is the same example (without the fix), but there is no overlap happening (since only stem is present):
image
And this is example with original problem:
image
As you can see, position of timestamps remain unchanged.

The problem here is that barline and fermata don't have the same alignments, so the shift for one doesn't translate into shift for another. Alternatively this could be fixed by merging them, but I'm not sure whether them being in BARLINE alignment instead of DEFAULT could lead to some issues.

lpugin: Thank you for the explanation. I am still looking into it.

lpugin:
The underlying issue here is having no duration spacing knowing that timestamps take no space. This is what you get with `--spacing-linear 0.0` and the following encoding

<details>


```xml
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-04-03T22:32:49" version="3.10.0-dev-4a1d799">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m8opjql">
    <score xml:id="sbq1yuu">
     <scoreDef xml:id="s5ksdo4" midi.bpm="400.000000">
      <staffGrp xml:id="sb7gj0k">
       <staffDef xml:id="sbdrx1k" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="F" line="4" />
        <keySig xml:id="keysig-L3F1" pname="c" mode="major" sig="0" />
        <meterSig xml:id="metersig-L5F1" count="2" sym="cut" unit="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L15" right="end" n="2">
       <staff xml:id="staff-L15F1N1" n="1">
        <layer xml:id="layer-L15F1N1" n="1">
        </layer>
       </staff>
       <fermata xml:id="fermata-L18F1N1" staff="1" tstamp="0.50000" place="above" />
       <fermata xml:id="fermata-L18F1N2" staff="1" tstamp="0.000000" place="below" />
       <fermata xml:id="fermata-L18F1N3" staff="1" tstamp="1.50000" place="above" />
       <fermata xml:id="fermata-L18F1N4" staff="1" tstamp="3.000000" place="below" />
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

image

This is with `0.1`

image

And this with `0.05`

image

This is `0.05` and disabling the minimum measure width

image

I am not sure what should happen when the value is `0.0`. Having all the timestamps at the same position is correct but does not make sense at the same time. Any thoughts?

Monceber:
Well, the value `0.0` is not the only problematic one.

Currently the code for horizontal space is as following:
```c++
return pow(intervalTime, spacingNonLinear) * spacingLinear * 10.0;
```
Linear spacing is the easiest that can set this to 0, but considering that return type is `int`, effectively any value <1 is 0. So with sufficiently low non-linear and linear option values the same results could be achieved.

In the end it all comes down to question whether we want to allow such rendering. I would say - yes, considering there are two dedicated options for this. However, adding a warning/error message if resulting spacing is or close to 0 would be a good indication.

lpugin:
@Monceber please have a look at #3389

lpugin:
Fix differently

Issue #3374, opened 2023-04-15
Implement keySig@color and keyAccid@color


craigsapp:
MEI 5 (development) adds `keySig@color` and `keyAccid@color`, which would be useful to implement in verovio.

Test example for `keySig`:

Target rendering:

<img width="620" alt="Screenshot 2023-04-14 at 4 54 35 PM" src="https://user-images.githubusercontent.com/3487289/232171986-3d065dbb-c703-4a19-8006-452d9812220f.png">

<details>
<summary> Click to view test MEI data for above rendering </summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2023-04-14T16:46:03" version="3.16.0-dev-9880298">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m71d0d">
    <score xml:id="smk0gne">
     <scoreDef xml:id="svqzwzz">
      <staffGrp xml:id="syrd2ub">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5" trans.diat="-1" trans.semi="-2">
        <label xml:id="label-L2F1">Clarinet in B<rend xml:id="r1d1nk76" fontfam="smufl"></rend>
        </label>
        <clef xml:id="clef-L4F1" shape="G" line="2" />
        <keySig xml:id="keysig-L6F1" color="hotpink" type="visual-key-signature" sig="2f" />
        <meterSig xml:id="metersig-L7F1" count="3" unit="4" />
        <instrDef xml:id="inmq77c" midi.instrnum="71" midi.instrname="Clarinet" />
       </staffDef>
      </staffGrp>
     </scoreDef>
[...]

PR #3376, opened 2023-04-15
Add keySig color support

rettinghaus:
addresses #3374

image

NB: Currently we use Verovio custom graphics for `keyAccid`, so coloring there is not supported.

Additionally this removes (empty) SVG Bounding Boxes from milestone elements.

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -2484,7 +2484,7 @@ void MEIOutput::WriteKeySig(pugi::xml_node currentNode, KeySig *keySig)
     // Only write att values if representing an attribute or in MEI basic
     if (!this->IsTreeObject(keySig)) {
         InstKeySigDefaultAnl attKeySigDefaultAnl;
-        // Broken in MEI 4.0.2 - waiting for a fix
+        // Broken in MEI 5.0.0-dev - waiting for a fix
         // attKeySigDefaultAnl.SetKeyAccid(keySig->GetAccid());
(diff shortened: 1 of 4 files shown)


craigsapp:
Thanks! That was fast...

I don't need `keyAccid` coloring myself (but just being complete in the issue).  So that can be implemented some time in the future when someone needs it.  Bracket enclosures for keyAccid will be of more use&mdash;I haven't needed them yet, but I am guessing that is already possible).

So issue #3374 can be closed when the PR is merged.

lpugin (changes requested):
Please revert the change on the milestone bounding box.

src/iomei.cpp
```diff
@@ -2484,7 +2484,7 @@ void MEIOutput::WriteKeySig(pugi::xml_node currentNode, KeySig *keySig)
     // Only write att values if representing an attribute or in MEI basic
     if (!this->IsTreeObject(keySig)) {
         InstKeySigDefaultAnl attKeySigDefaultAnl;
-        // Broken in MEI 4.0.2 - waiting for a fix
```
lpugin, line 2487: Why are you changing this? We are still waiting for a fix, aren't we?

rettinghaus: As far as I know there is now version 4.0.2?

rettinghaus:
I rebased it and dropped the commit. However, I will complain about this in the latest issue.

PR #3388, opened 2023-04-21
Add support for tone clusters

rettinghaus:
This is the first (simple) implementation for `chord@cluster`.

image

chord-009

NB: Examples will be added to the test suite. Full midi support will be implemented at a later stage.


include/vrv/view.h
```diff
@@ -348,6 +348,7 @@ class View {
     ///@{
     void DrawAcciaccaturaSlash(DeviceContext *dc, Stem *stem, Staff *staff);
     void DrawClefEnclosing(DeviceContext *dc, Clef *clef, Staff *staff, char32_t glyph, int x, int y);
+    void DrawCluster(DeviceContext *dc, Chord *chord, Layer *layer, Staff *staff, Measure *measure);
```
lpugin, line 351: It would be better name `DrawChordCluster`

PR #3391, opened 2023-04-21
update fonts for divLine & rephrase member name

yinanazhou:


rettinghaus (changes requested):
LibMEI files shall not be clang formatted.

rettinghaus:
LibMEI should be updated before this PR. See #3392.

rettinghaus:
@yinanazhou You write "update fonts" but this PR doesn't update the fonts, so this will crash on attempt to render this. Also the introduced glyphs here are not available in Leipzig font. Please add them there first.

lpugin:
> LibMEI should be updated before this PR. See #3392.

This has already been done https://github.com/rism-digital/verovio/commit/873020f54d4002e44548b2fde918ac3b77e4ac4b

rettinghaus:
Didn't see that this is not supposed to go into develop.

lpugin (changes requested):
* Remove the `emscripten/emsdk` submodule
* Move `View::DrawDivLine` implementation from `view_element.cpp` to `view_neumes.cpp`

We can then see for the fonts

src/toolkit.cpp
```diff
@@ -1010,7 +1010,7 @@ std::string Toolkit::GetAvailableOptions() const
         }
 
         grp << "options" << opts;
-        grps << optionGrp->GetId() << grp;
```
lpugin, line 1013: This does not look right. Could you double check?

libmei/dist/atttypes.h
```diff
@@ -2515,4 +2515,4 @@ enum whitespace_XMLSPACE : int8_t {
 
 } // namespace vrv
 
-#endif // __LIBMEI_ATT_TYPES_H__
```
lpugin, line 2518: Discard this change too

Issue #2545, opened 2021-12-17
Saving memory & attribute composition


DavidBauer1984:
**Saving memory**
Verovio is getting quite memory hungry. The following table shows memory peaks for load+layout+rendering:

| MEI size | Memory peak | Ratio |
| -------- | -------------- | ----- |
| 226 KB | 21,9 MB | 96,9 |
| 1,1 MB | 54,1 MB | 49,2 |

A lot of the memory is consumed by the object tree, so let's have a look at some object sizes:

| Class | Size in Bytes |
| ----- | ---- |
| Note | 1768 |
| LayerElement | 816 |
| LinkingInterface | 296 |

This is just the fixed object size (i.e the result of the `sizeof` operator), the actual memory consumed by these objects will be larger due to container members like std::string or std::vector. But it is already interesting to see that a sixth of _each_ Note object consists of the linking interface which sits inside each LayerElement and ControlElement. And is hardly ever used.

In Verovio we have this unwritten law that we always store attributes by inheritance. So instead of storing attributes which are actually encoded, we store _all_ attributes that could be used (from a rendering point of view). This has some neat advantages:
- Attribute getter and setter are automatically available.
- Availability of attributes can be checked via `dynamic_cast`

But it also has the disadvantage that it unnecessarily eats up memory. 

So the principal idea would be to allow for attributes also to be stored via composition instead of inheritance. Imagine that we rename the current LinkingInterface into LinkingInterfaceImpl and create a new LinkingInterface class which stores a pointer to LinkingInterfaceImpl. This pointer would be NULL as long as no linking is encoded, reducing the size of LinkingInterface to 8 bytes. So this alone would save almost 300 bytes on most of the layer and control elements. There still is a catch: how would an element gain access to an attribute stored inside LinkingInterfaceImpl? `dynamic_cast` would not work anymore. Even worse, the whole attribute detection mechanism that is currently used in Verovio and LibMEI would fail.

**Current attribute detection**
During element construction we currently register all attribute and interface ids and store them in a std::vector in Object (see `m_attClasses` and `m_interfaces`). We then have functions `bool Object::HasAttClass(AttClassId attClassId)` and `bool HasInterface(InterfaceId interfaceId)` which check the existence of an attribute or interface by searching the id in the corresponding vector. Finally, we typically use the following pattern to retrieve an attribute if it exists:

```C++
if (element->HasAttClass(ATT_NAME)) {
        AttClass *att = dynamic_cast<AttClass *>(element);
[...]

PR #3400, opened 2023-04-28
Add rendring for al niente on hairpins

rettinghaus:
This adds rendering for al niente circles to hairpins.

hairpin-al-niente

(Example from Gould, will be added to test suite.)

### DIFF
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -627,7 +627,7 @@ void View::DrawHairpin(
 
     const int cap = (style == AxDOT) ? AxCAP_ROUND : AxCAP_SQUARE;
 
-    dc->SetPen(m_currentColour, hairpinThickness, style, 0, 0, cap, AxJOIN_BEVEL);
+    dc->SetPen(m_currentColour, hairpinThickness, style, 0, 0, cap, AxJOIN_MITER);
 
     if ((startY == 0) && !niente) {
         Point p[3];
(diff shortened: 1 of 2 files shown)


src/view_control.cpp
```diff
@@ -533,40 +533,43 @@ void View::DrawHairpin(
     FloatingPositioner *leftLink = hairpin->GetCorrespFloatingPositioner(hairpin->GetLeftLink());
     FloatingPositioner *rightLink = hairpin->GetCorrespFloatingPositioner(hairpin->GetRightLink());
 
+    const int unit = m_doc->GetDrawingUnit(staff->m_drawingStaffSize);
+    const hairpinLog_FORM form = hairpin->GetForm();
+    const bool niente = (hairpin->GetNiente() == BOOLEAN_true);
```
lpugin, line 538: It usually better to check if an attribute is not default before getting its value. At least that would be necessary if one day we want to have [this](https://github.com/rism-digital/verovio/issues/2545#issuecomment-997668899) type of optimisation

lpugin: I also meant it for `const hairpinLog_FORM form = hairpin->GetForm();`

rettinghaus: We check for this in line 528.

lpugin: Correct

Issue #2530, opened 2021-12-03
Tablature: French tab bass slashes ugly


TimCrawford:
French tab: Bass course slashes are too thick and take up too much space 

![Basses_French](https://user-images.githubusercontent.com/2544314/144642511-aa6163d8-d9fc-41c6-a58c-c2b7776afc7b.jpg)

lpugin:
SMuFL has the slash directly combined with the fret number https://w3c.github.io/smufl/latest/tables/french-and-english-renaissance-lute-tablature.html

It seemed to me it would be simpler to have a single glyph for the slash and combine it appropriately. However, if `/a`, `//a` and `///a` (i.e., U+EBCE, U+EBCF and U+EBD0) are the only possible combinations we can still go that way, though.

TimCrawford:
Laurent,

We need the notes and slashes to be separate: we could have ////a and even /////a ! (I could show you original sources for these.)

Though these are ‘logically’ replaceable by bass-number-glyphs ‘4’ and ‘5’, respectively, we may need to encode the original appearance. (I have not, I think, ever seen the equivalent of //////a - i.e. corresponding to ‘6’, where, as far as I know, the bass-number was always used.)

So your ‘simpler’ preference is preferable to me, too.

A further consideration is that the order of the letter and slashes in the bass-note can vary; in MSS, usually it’s like the encoding: ‘///a’, but in at least one prominent print (Thomas Mace, Musick’s Monument, 1676) and in fact several MSS, it’s like ‘a///‘ . In this case, however, there is so little ambiguity that we can possibly ignore this, at least for now.

On the other hand, there is yet another complication, where one or two printed sources give special notation for playing the two strings of a bass course as two notes (at the fundamental and at the higher octave, always in this order) separately; this is done in one way (Mouton, c1680) by using a smaller typeface for the second note and in another way (Radolt, 1701) using a capital letter for the bass note, lower-case for the second. In both cases the two notes are connected visually by a slur (or tie, rather). (I should say that both of these sources are engraved rather than typeset.)
Again, I could provide examples easily enough.

BTW, separating the letter and the slashes would simplify the SMUFL table’s ‘Recommended stylistic alternates’ which provide different Unicodes for several combinations. (The underline and strike-through ones are needed, too, I’m afraid!)

Tim

On 17 Jan 2022, at 15:43, Laurent Pugin ***@***.******@***.***>> wrote:



[...]

PR #3401, opened 2023-05-02
Use SMuFL glyphs for open 7th .. 10th courses.  Partial fix of issue #2530

paul-bayleaf:
Partial fix of  Tablature: French tab bass slashes ugly #2530 

After this fix, no ugly slashes on open diapasons.  However, there will still be ugly slashes on stopped diapasons.

Screenshot_20230501_174651

Please re-generate fonts after merge, thanks.


lpugin:
Thanks. I will re-generate the fonts

PR #3408, opened 2023-05-05
Add support for diamond shape enclosing

rettinghaus:
image

### DIFF
--- a/include/vrv/rend.h
+++ b/include/vrv/rend.h
@@ -59,6 +59,11 @@ public:
      */
     bool IsSupportedChild(Object *object) override;
 
+    /**
+     * Check if rend has an enclosing.
+     */
+    bool HasEnclosure() const;
+
     //----------//
     // Functors //
     //----------//
@@ -75,9 +80,6 @@ public:
 
 private:
     //
-public:
-    //
-private:
 };
 
 } // namespace vrv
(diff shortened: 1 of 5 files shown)


src/view_text.cpp
```diff
@@ -447,14 +447,15 @@ void View::DrawRend(DeviceContext *dc, Rend *rend, TextDrawingParams &params)
...
-    if (rend->HasContentBB()) {
-        if (rend->HasRend()) {
-            params.m_enclosedRend.push_back(rend);
-            params.m_x = rend->GetContentRight() + m_doc->GetDrawingUnit(100);
-            params.m_explicitPosition = true;
-            params.m_enclose = rend->GetRend();
-        }
+    const bool isEnclosed = (rend->HasRend())
+        ? (((rend->GetRend() - TEXTRENDITION_box) | (TEXTRENDITION_tbox - rend->GetRend())) >= 0)
+        : false;
```
lpugin, line 452: This is cryptic. What's the logic behind it

lpugin: Since we are talking about four values I think the simplest would be to have a helper function (on `Rend`?) to check it.

include/vrv/rend.h
```diff
@@ -74,9 +80,6 @@ class Rend : public TextElement,
 
 private:
     //
-public:
-    //
-private:
```
lpugin, line 79: These are placeholders for member variables. No need to remove them

PR #3411, opened 2023-05-08
Functor classes for MIDI and transposition

### LINKED ISSUE #3005: Proposal for Functor Classes

DavidBauer1984:
This PR moves the remaining functors (for MIDI and transpose) into functor classes and cleans up the previous functor approach. This completes the functor refactoring. Closes #3005


DavidBauer1984:
@lpugin Thanks for supporting this refactoring. I'm quite happy about the new structure and hopefully it will help Verovio to grow further.

PR #3412, opened 2023-05-08
Fix a few bugs in latest develop-humdrum branch

gregchapman-dev:
1. Fix typo when setting turn@form (every turn's form was set to lower).
2. Fix crash in LayerElement::CalcLayerOverlap when sameDirElement is true, but stemInterface is NULL.
3. Fix crash in Layer::GetAtPos when this->GetFirst() returns NULL, by moving the NULL check a bit earlier.

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -25534,7 +25534,7 @@ void HumdrumInput::addTurn(hum::HTp token, const string &tok, int noteIndex)
     std::string noteid = getLocationId("note", token, noteIndex);
     turn->SetStartid("#" + noteid);
 
-    turn->SetForm(invertedQ ? turnLog_FORM_lower : turnLog_FORM_lower);
+    turn->SetForm(invertedQ ? turnLog_FORM_lower : turnLog_FORM_upper);
 
     if (m_signifiers.above) {
         if (turnend < (int)token->size() - 1) {
--- a/src/layer.cpp
+++ b/src/layer.cpp
@@ -197,13 +197,14 @@ LayerElement *Layer::GetAtPos(int x)
 const LayerElement *Layer::GetAtPos(int x) const
 {
     const Object *first = this->GetFirst();
-    if (!first || !first->IsLayerElement()) return NULL;
+    if (!first) return NULL;
 
     if (first->IsEditorialElement()) {
         IsEditorialElementComparison cmp;
         cmp.ReverseComparison();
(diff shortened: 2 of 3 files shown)


gregchapman-dev:
OK, it's actually two fixes.

src/iohumdrum.cpp
```diff
@@ -25534,7 +25534,7 @@ void HumdrumInput::addTurn(hum::HTp token, const string &tok, int noteIndex)
     std::string noteid = getLocationId("note", token, noteIndex);
     turn->SetStartid("#" + noteid);
 
-    turn->SetForm(invertedQ ? turnLog_FORM_lower : turnLog_FORM_lower);
+    turn->SetForm(invertedQ ? turnLog_FORM_lower : turnLog_FORM_upper);
```
craigsapp, line 25537: Thanks!  I was wondering a few weeks ago why the turns were upside down (but I never bothered to check).

craigsapp:
> Fix typo when setting turn@form (every turn's form was set to lower).
Fix crash in LayerElement::CalcLayerOverlap when sameDirElement is true, but stemInterface is NULL.
Fix crash in Layer::GetAtPos when this->GetFirst() returns NULL, by moving the NULL check a bit earlier.

Only the first commit should be PRed to the develop humdrum branch.  The other two should go to the `develop` branch.  @lpugin can either say that as well or he can approve merging into the `develop-humdrum` branch and then I will do a PR to the `develop` branch.

craigsapp (approved):

lpugin (changes requested):
Yes, please make a distinct PR to develop for the changes in layer.cpp and layerelement.cpp

gregchapman-dev:
New PR for the layer.cpp and layerelement.cpp fixes is PR #3413.

gregchapman-dev:
OK, I have made the same layer.cpp fix tweak here now that I made on develop, for consistency.  Do you need me to remove the two NULL checks from this PR?  Or can I leave them in place, so humdrum-develop can benefit from them as well?

craigsapp:
> Do you need me to remove the two NULL checks from this PR? Or can I leave them in place, so humdrum-develop can benefit from them as well, prior to any merge-back from develop?

The best thing is to remove them from `develop-humdrum`, and then I will pull `develop` into `develop-humdrum` (so they will end up there from your PR to `develop`).

lpugin (approved):

Issue #975, opened 2019-01-03
Unnecessary trailing zeros in timestamps


craigsapp:
When converting into MEI data from another data format (MusicXML, Humdrum, or I presume P&E), the timestamps are quite ugly:

From MusicXML importer:
```xml
       <harm xml:id="harm-0000001676348108" staff="1" tstamp="1.000000">Am</harm>
       <harm xml:id="harm-0000001651477441" staff="1" tstamp="7.750000">Dm</harm>
       <harm xml:id="harm-0000002053616025" staff="1" tstamp="14.500000">G</harm>
```

And via the Humdrum importer

```xml
         <harm xml:id="harm-L12F5" staff="1" tstamp="1.000000">Am</harm>
         <harm xml:id="harm-L13F5" staff="1" tstamp="4.000000">Dm</harm>
         <harm xml:id="harm-L14F5" staff="1" tstamp="7.000000">G</harm>
```

If someone is using verovio to generate SVG images, it will not matter since there is no MEI output in that case.  But when using MEI to convert from other formats, these extra zeros make the data harder to read and pad the filesize with unecessary characters.

A `@tstamp` of `1.000000` would look better as `1` instead.  The cause of the extra zeros is due to the use of sprintf rather than stringstream.  Is there a reason for this @lpugin?  It would be nicer to not have the trailing zeros by using stringstream instead.

I can make this fix myself, if you like 😃

lpugin:
> I can make this fix myself, if you like

Please change it if you want. However, I would suggest `1.0` instead of `1`

craigsapp:
> Please change it if you want. However, I would suggest 1.0 instead of 1

Any particular reason? (such as some validator expects `@tstamp` to contain a decimal point?)

Current:

```
         <harm xml:id="harm-L12F5" staff="1" tstamp="1.000000">Am</harm>
         <harm xml:id="harm-L13F5" staff="1" tstamp="4.000000">Dm</harm>
         <harm xml:id="harm-L14F5" staff="1" tstamp="7.000000">G</harm>
```

With one decimal digit minimum:

```
         <harm xml:id="harm-L12F5" staff="1" tstamp="1.0">Am</harm>
         <harm xml:id="harm-L13F5" staff="1" tstamp="4.0">Dm</harm>
         <harm xml:id="harm-L14F5" staff="1" tstamp="7.0">G</harm>
```

Without any fractional digits (if all are zero):


```
[...]

PR #3420, opened 2023-05-17
Unify spelling and shrink size of SVG

rettinghaus:
This PR unifies the spelling of color in the code and makes the SVG a bit smaller by dropping trailing zeros.

### DIFF
--- a/include/vrv/bboxdevicecontext.h
+++ b/include/vrv/bboxdevicecontext.h
@@ -49,11 +49,11 @@ public:
      * @name Setters
      */
     ///@{
-    void SetBackground(int colour, int style = AxSOLID) override;
+    void SetBackground(int color, int style = AxSOLID) override;
     void SetBackgroundImage(void *image, double opacity = 1.0) override{};
     void SetBackgroundMode(int mode) override;
-    void SetTextForeground(int colour) override;
-    void SetTextBackground(int colour) override;
+    void SetTextForeground(int color) override;
+    void SetTextBackground(int color) override;
     void SetLogicalOrigin(int x, int y) override;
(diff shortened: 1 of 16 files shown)


ahankinson:
I think `%g` can emit scientific notation as a format… is this valid input in the places where you’ve switched it?

https://stackoverflow.com/questions/827499/string-format-with-g-in-c

craigsapp:
`%g` might be interesting for this issue (which I was too lazy to implement). 

https://github.com/rism-digital/verovio/issues/975

 Scientific notation avoidance would be important as @ahankinson mentions.

rettinghaus:
It's valid. Most of these values are be between 0 and 1. If something weird should happen with scaling we're safe anyway, because [SVG supports scientific notation](https://www.w3.org/TR/SVG11/types.html#DataTypeNumber).  

[arpeggios_in_chords.svg.zip](https://github.com/rism-digital/verovio/files/11515860/arpeggios_in_chords.svg.zip)

rettinghaus:
Ok, I reverted the cases where there could be a potential risk. Still, where we only have values of 0 or 1, I don't see the point in writing `0.000000` and `1.000000`.

lpugin:
>  don't see the point in writing `0.000000` and `1.000000`.

I don't think it is great. But since most of the users do not look at the SVG code, I don't think it is a big deal. Machines do not care.

rettinghaus:
You don't like it, I don't like it, @craigsapp doesn't like it. And, as you said, machines do not care. ;-)

lpugin:
Since we agree that machines do not care, could you revert the other ones? Thanks!

PR #3422, opened 2023-05-17
fix JSON "\uXXXX" parsing

alpha0010:
The included [JSON++ library](https://github.com/hjiang/jsonxx) is archived, so I am submitting the patch here.  Is there somewhere else I should instead target?

This applies utf-8 encoding to unicode hex codepoint escape sequences, avoiding invalid symbols/crashes for non-ASCII codepoints.

I do not think this issue directly impacts normal Verovio usage.  We discovered the issue while utilizing the embedded JSON library in some of our own code.

### DIFF
--- a/src/json/jsonxx.cc
+++ b/src/json/jsonxx.cc
@@ -116,28 +116,51 @@ bool parse_string(std::istream& input, String& value) {
                     break;
                 case 'u': {
                         int i;
-                        std::stringstream ss;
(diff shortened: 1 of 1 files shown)


paul-bayleaf:
For completeness, consider also implementing decoding UTF-16 surrogate pairs, which JSON uses to encode characters outside the Basic Multilingual Plane.
https://www.rfc-editor.org/rfc/rfc8259 para 7. Strings

PR #3424, opened 2023-05-18
Implement German lute tablature

paul-bayleaf:
Implement German lute tablature.

Limitations:
Only implements Hans Newsidler's notation for the 6th course.
Had to substitute for some missing German lute tablature glyphs from elsewhere in the SMuFL fonts.
Notes are aligned to the top of the (invisible) staff.  Not yet implemented bottom or custom alignment.

If merged please regenerate the fonts.
LN132 tab-0 tab mei


src/view_beam.cpp
```diff
@@ -284,7 +284,7 @@ void View::DrawBeamSegment(
     int durRef = DUR_8;
     int durRef2 = DUR_16;
 
-    if (staff->IsTabLuteFrench() || staff->IsTabLuteItalian()) {
+    if (staff->IsTabLuteFrench() || staff->IsTabLuteItalian() || staff->IsTabLuteGerman()) {
```
rettinghaus, line 287: Again sort alphabetically.

src/note.cpp
```diff
@@ -266,12 +266,14 @@ const TabGrp *Note::IsTabGrpNote() const
     return vrv_cast<const TabGrp *>(this->GetFirstAncestor(TABGRP, MAX_TABGRP_DEPTH));
 }
 
-std::u32string Note::GetTabFretString(data_NOTATIONTYPE notationType) const
+std::u32string Note::GetTabFretString(data_NOTATIONTYPE notationType, bool &overline) const
```
rettinghaus, line 269: What is the advantage to hand `overline` over from here instead of getting it in `ViewTab` directly?

src/view_tab.cpp
```diff
@@ -144,9 +145,32 @@ void View::DrawTabNote(DeviceContext *dc, LayerElement *element, Layer *layer, S
...
+        // TODO is this the correct way to add an overline?
+        if (overline && !fret.empty()) {
+            const int lineWidth = m_doc->GetDrawingStaffLineWidth(staff->m_drawingStaffSize);
+            const int width = m_doc->GetGlyphWidth(fret.at(0), glyphSize, drawingCueSize);
+            const int x1 = x - width / 2;
+            const int x2 = x + (fret.size() - 1) * width + width * 8 / 10; // trim right hand overhang on last character
+            const int y1 = y + m_doc->GetGlyphHeight(fret.at(0), glyphSize, drawingCueSize) + lineWidth;
+            const int y2 = y1;
+
+            dc->SetPen(m_currentColour, ToDeviceContextX(lineWidth), AxSOLID);
```
rettinghaus, line 166: This looks weird. Why do you use `ToDeviceContextX` here?

src/view_tab.cpp
```diff
@@ -144,9 +145,32 @@ void View::DrawTabNote(DeviceContext *dc, LayerElement *element, Layer *layer, S
...
+        }
 
         dc->SetFont(m_doc->GetDrawingSmuflFont(glyphSize, false));
         this->DrawSmuflString(dc, x, y, fret, HORIZONTALALIGNMENT_center, glyphSize);
+
+        // Add overline if required
+        // TODO is this the correct way to add an overline?
+        if (overline && !fret.empty()) {
+            const int lineWidth = m_doc->GetDrawingStaffLineWidth(staff->m_drawingStaffSize);
+            const int width = m_doc->GetGlyphWidth(fret.at(0), glyphSize, drawingCueSize);
```
rettinghaus, line 160: Instead of checking only the first glyph you should try `dc->GetSmuflTextExtent` (e.g. see `DrawOctave`).

src/view_tab.cpp
```diff
@@ -144,9 +145,32 @@ void View::DrawTabNote(DeviceContext *dc, LayerElement *element, Layer *layer, S
...
+            const int x1 = x - width / 2;
+            const int x2 = x + (fret.size() - 1) * width + width * 8 / 10; // trim right hand overhang on last character
```
rettinghaus, line 162: After looking at this too long I understood what this means. ;-) Something like `x + extend.m_width - lineWidth` would be much more elegant.

rettinghaus (changes requested):
Thanks! I left some comments.

rettinghaus:
Could you point me please to the missing `et` and `con` glyphs in digitized prints?

src/note.cpp
paul-bayleaf: There are several different notations for the 6th, 7th ... courses that I have not yet implemented.  Some of these use overlines.  As the overline is part of the character it seemed better to put the logic in note.cpp where the characters are selected, rather than in view_tab.cpp where they are drawn.    The implementation of these other notations requires some means in MEI of specifying which to use.

src/view_tab.cpp
paul-bayleaf: Here I was copying the line drawing code from view_page.cpp DrawStaffLines - without having a full understanding of the drawing primitives.  Please advise, thanks.

src/beam.cpp
```diff
@@ -602,8 +602,8 @@ void BeamSegment::CalcBeamInit(
         beamInterface->m_beamWidthBlack /= 2;
         beamInterface->m_beamWidthWhite /= 2;
 
-        // Adjust it further for tab.lute.french and tab.lute.italian
-        if (staff->IsTabLuteFrench() || staff->IsTabLuteItalian()) {
+        // Adjust it further for tab.lute.french, tab.lute.italian and tab.lute.german
+        if (staff->IsTabLuteFrench() || staff->IsTabLuteItalian() || staff->IsTabLuteGerman()) {
```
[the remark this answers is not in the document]
paul-bayleaf: Will do

src/note.cpp
```diff
@@ -335,6 +337,66 @@ std::u32string Note::GetTabFretString(data_NOTATIONTYPE notationType) const
         }
         return fretStr;
     }
+    else if (notationType == NOTATIONTYPE_tab_lute_german) {
+        std::u32string fretStr;
+        const int fret = this->GetTabFret();
+        const int course = this->GetTabCourse();
+
+        // SMuFL has glyphs for German lute tablature following Hans Newsider's notation
```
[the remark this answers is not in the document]
paul-bayleaf: Will fix.

src/view_beam.cpp
paul-bayleaf: Will do

src/view_tab.cpp
paul-bayleaf: OK

src/view_tab.cpp
paul-bayleaf: OK

src/view_tab.cpp
rettinghaus: This should be just `lineWidth`. Maybe not base it on `staffLineWidth` but another value like `lyricLineThickness`? See what fits best and add just a factor for convenience.

paul-bayleaf:
Example of et and con.  From Hans Newsidler's Ein Newgeordent Kunstlich Lautenbuch, Lamora Isaac.  p1v.-p3
https://www.lutemusic.org/facsimiles/NewsidlerH/Ein_Newgeordent_Kunstlich_Lautenbuch/v.1_1536/p2v.png

Con is very common in Newsidler's music, et is very rare - I had to hunt to find one.  Con looks like a 9, et looks like a 7 with a bar and a tail.

Here is a blowup of measures 50,51,52.  Con is in each of these measures, top row 2nd note.
Et appears in the measure 51 1st chord, 2nd row.
et_con

rettinghaus (approved):

include/vrv/vrvdef.h
```diff
@@ -672,6 +672,7 @@ enum GraphicID { PRIMARY = 0, SPANNING, SYMBOLREF };
 //----------------------------------------------------------------------------
 
 #define TABLATURE_STAFF_RATIO 1.75
+#define GERMAN_TAB_STAFF_RATIO 2.2
```
lpugin, line 675: Could you explain why you need a distinct ratio for German tablature?

lpugin (commented):
Sorry for the late feedback. I have only one question.

include/vrv/vrvdef.h
paul-bayleaf: With the smaller ratio I found that the Gothic characters overlapped each other and were very cramped, the Gothic characters are quite bulky and some are very tall, especially the k.

craigsapp:
Here is the vocal score version of La Bernardina: https://verovio.humdrum.org/?file=jrp/Jos/Jos2721-La_Bernardina.krn

rettinghaus:
@craigsapp Actually that's completely different problem.

craigsapp:
Is that something that could be fixed within verovio? (i.e., where would the font metrics for Chinese characters be acquired?)  and is verovio otherwise guessing at the width of the characters, or perhaps the width of the characters wrong in the font?

lpugin:
Now that the glyphs have been added https://github.com/rism-digital/leipzig/pull/29 I will merge it. The fonts still need to be re-generated, though.

PR #3426, opened 2023-05-20
Fix check that was rejecting single character turns such as "4aS"

gregchapman-dev:
Beethoven piano sonata01-2.krn has two of these.


gregchapman-dev:
Crap, did it wrong _again_.

PR #3433, opened 2023-05-25
Clean project groups and MeterSigGrp base class

DavidBauer1984:
This PR moves Pb/Sb to system elements and MNum to control elements in the Xcode project. It further derives `MeterSigGrp` from `LayerElement`.


DavidBauer1984:
CI needs to rerun.

PR #3427, opened 2023-05-20
iohumdrum.cpp rejects valid single-character turns (e.g. "4aS").

gregchapman-dev:
Beethoven piano sonata01-2.krn has two of these.

PR #3436, opened 2023-05-30
Humdrum: Improve element visibility for suppressed tokens

WolfgangDrescher:
Demo data:

```tsv
**kern	**kern
=7	=7
*^	*^
2B;yyy	2GG;yy	2g;yy	2d;yyy
=:|!	=:|!	=:|!	=:|!
*	*	*	*
4dyy	4GGyy	4byy	[4gyy
=8	=8	=8	=8
4dyy	4GGyy	4byy	8gL]yy
.	.	.	8f#Jyy
4cyy	4AAyy	4ccyy	8eLyy
.	.	.	8f#Jyy
8BLyy	4BByy	4ddyy	[4gyy
8AJyy	.	.	.
=9	=9	=9	=9
8BLyy	4.BByy	4.ddyy	8gL]yy
8cJyy	.	.	8aJyy
4dyy	.	.	8gLyy
.	8AAyy	8ccyy	8f#Jyy
4dyy	4GGyy	4byy	4gyy
=10	=10	=10	=10
2d;yyy	2D;yy	2a;yy	2f#;yyy
4Byy	[4Eyy	4gyy	4eyy
=	=	=	=
*v	*v	*	*
*	*v	*v
*-	*-
```

Before:

Bildschirmfoto 2023-05-30 um 21 59 48

After:

Bildschirmfoto 2023-05-30 um 22 01 27

## Questions:

* @craigsapp Should I implement this for `HumdrumInput::insertGBeam` as well?
* I'm not sure if this could also be relevant for `HumdrumInput::insertBeamSpan`
* `beam->SetType("invisible");` does not seem to have any effect for the element visibility and there is no `beam::SetVisible` member for beams. I fixed this for now with `beam->SetColor("transparent")` but I'm sure there must be a better solution. Any suggestions?

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -20228,6 +20228,11 @@ Beam *HumdrumInput::insertBeam(
     std::vector<std::string> &elements, std::vector<void *> &pointers, const humaux::HumdrumBeamAndTuplet &tg)
 {
     Beam *beam = new Beam();
+    if (tg.token->find("yy") != std::string::npos) {
+        // Ignore beam when token is suppressed with yy signifier
+        beam->SetType("invisible");
+        beam->SetColor("transparent");
+    }
     appendElement(elements, pointers, beam);
     elements.push_back("beam");
     pointers.push_back((void *)beam);
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -20248,6 +20248,11 @@ Beam *HumdrumInput::insertGBeam(
     std::vector<std::string> &elements, std::vector<void *> &pointers, const humaux::HumdrumBeamAndTuplet &tg)
 {
     Beam *gbeam = new Beam();
+    if (tg.token->find("yy") != std::string::npos) {
+        // Ignore beam when token is suppressed with yy signifier
+        gbeam->SetType("invisible");
+        gbeam->SetColor("transparent");
+    }
     appendElement(elements, pointers, gbeam);
     elements.push_back("gbeam");
     pointers.push_back((void *)gbeam);
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -20249,7 +20249,7 @@ Beam *HumdrumInput::insertGBeam(
 {
     Beam *gbeam = new Beam();
     if (tg.token->find("yy") != std::string::npos) {
-        // Ignore beam when token is suppressed with yy signifier
+        // Ignore grace note beam when token is suppressed with yy signifier
         gbeam->SetType("invisible");
         gbeam->SetColor("transparent");
     }
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -20241,7 +20241,7 @@ Beam *HumdrumInput::insertBeam(
 
 //////////////////////////////
 //
-// HumdrumInput::insertGBeam --
+// HumdrumInput::insertGBeam -- insert beam for grace notes
 //
 
 Beam *HumdrumInput::insertGBeam(
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -26249,6 +26249,10 @@ void HumdrumInput::addTrill(Object *linked, hum::HTp token)
 
 void HumdrumInput::processTieStart(Note *note, hum::HTp token, const std::string &tstring, int subindex)
 {
+    // Ignore tie when token is suppressed with yy signifier
+    if (token->find("yy") != std::string::npos) {
+        return;
+    }
     if (token->isMensLike()) {
         return;
     }


src/iohumdrum.cpp
```diff
@@ -20228,6 +20228,11 @@ Beam *HumdrumInput::insertBeam(
     std::vector<std::string> &elements, std::vector<void *> &pointers, const humaux::HumdrumBeamAndTuplet &tg)
 {
     Beam *beam = new Beam();
+    if (tg.token->find("yy") != std::string::npos) {
+        // Ignore beam when token is suppressed with yy signifier
+        beam->SetType("invisible");
+        beam->SetColor("transparent");
```
WolfgangDrescher, line 20234: Any ideas how to improve this?

craigsapp: I think this is the best way.   This is equivalent to `visibility: none` in CSS: space will be allocated to the beams but they will not be shown.   This should be sufficient as long as you don't mind the extra vertical space between staves that this may create.

craigsapp:
> `beam->SetType("invisible");` does not seem to have any effect for the element visibility and there is no `beam::SetVisible` member for beams. 

the `@type` attribute is a more informal system, so it is useful to include for analytic purposes.   And at least in general it is not used by verovio for rendering (that is usually done by `@form` when doing some sort of rendering in verovio).  The `@type` parameter is converted into a `@class` attribute in the SVG output from verovio.  So you could add a verovio option to the Humdrum file to cause the elements with a `transparent ` class label to be colored transparent:

```tsv
!!!verovio: svgCss: g.invisible { fill: transparent; stroke: transparent; }
```

Or perhaps equivalently:

```tsv
!!!verovio: svgCss: g.invisible { visibility: none; }
```


I notice that ledger lines are still visible when notes are set to transparent.  Is there also a way to control the color of ledger lines @lpugin?  Maybe by adding a verovio option such as `--color-ledger-lines-as-note` might work (should also work for chords).


Screenshot 2023-05-30 at 6 02 16 PM

<details>
<summary>Click to view MEI data for above example</summary>

```
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2023-05-30T18:03:36" version="3.16.0-dev-b62cd9f-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mjwj3tb">
    <score xml:id="s1eybwzi">
     <scoreDef xml:id="s1edwjv1">
      <staffGrp xml:id="si7c6ug">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5">
        <clef xml:id="c1t15tkf" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="staff-L1F1" n="1">
         <note xml:id="note-L3F1" dur="4" oct="4" pname="c" accid.ges="n" />
         <note xml:id="note-L4F1" dur="4" oct="4" pname="d" accid.ges="n" />
         <note xml:id="note-L5F1" dur="4" oct="4" pname="e" accid.ges="n" />
         <note xml:id="note-L6F1" dur="4" oct="4" pname="f" accid.ges="n" />
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L7">
       <staff xml:id="staff-L7F1N1" n="1">
        <layer xml:id="layer-L7F1N1" n="1">
         <note xml:id="note-L9F1" dur="4" oct="4" pname="c" color="transparent" accid.ges="n" />
         <note xml:id="note-L10F1" dur="4" oct="4" pname="d" color="transparent" accid.ges="n" />
         <note xml:id="note-L11F1" dur="4" oct="4" pname="e" color="transparent" accid.ges="n" />
         <note xml:id="note-L12F1" dur="4" oct="4" pname="g" color="transparent" accid.ges="n" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

I think it would be good to to create `*invis` and `*Xinvis` to be mapped to `*color:transparent` and `*color:black`.   Then `yy` would not need to be added to every note that needs to be invisible (this could also be used for `**dynam` and `**text`).




> @craigsapp Should I implement this for HumdrumInput::insertGBeam as well?

Yes, that would be good (otherwise there will be unattached grace beams like there were with regular beams).

If the purpose of making the music invisible is for students to composite the missing music, then there is a question of security since the notes are still present in the MEI/SVG data and they might eventually figure that out...  In that case maybe add `*blank` and `*Xblank` which will convert notes into invisible rests.

lpugin:
@craigsapp could you review this and the [other one](https://github.com/rism-digital/verovio/pull/3427)? Thanks

WolfgangDrescher:
Commit 47616c3 fixes the same bug for grace notes

Before:

Bildschirmfoto 2023-06-19 um 11 43 55

After:

Bildschirmfoto 2023-06-19 um 11 44 09

Data:

```
**kern
16bqLyy
16ddqJyy
1cc
*-
```

WolfgangDrescher:
Having something like `*blank` does seem useful. But there is no urgent need for me right now. Same for `*invis` as it is currently easier to add `yy` to all data tokens with `shed` than adding an additional tandem interpretation in the browser. 

(Maybe we should add a new filter that allows adding new tandem interpretations.)

craigsapp (approved):

PR #3446, opened 2023-06-07
Fix MusicXML diminuendo <wedge> spread handling

paxbun:
The [MusicXML standard](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/wedge/) specifies that the `spread` attribute at the start of a crescendo and the end of a diminuendo is ignored. However, the current code only reads `spread` at the end of a `<wedge>` pair, which means it handles the spread of diminuendos wrongly.

This fix changes the behavior of `MusicXmlInput` so it sets the opening of a hairpin at the end for diminuendos or the start for crescendos. The followings show the result of this fix. [Test MusicXML files](https://github.com/rism-digital/verovio/files/11671536/Test.MusicXML.files.zip) are made by modifying `<wedge>` elements in `BeetAnGeSample.musicxml` from the example set on [musicxml.com](https://www.musicxml.com/music-in-musicxml/example-set/).

* `test-narrow-wedge.musicxml` (Made by replacing `spread="15"` with `spread="5"`)

| Before fix | After fix |
| --- | --- |
| test-narrow-wedge before | test-narrow-wedge after |

* `test-spread-0-at-dim-stop.musicxml` (Made by adding `spread="0"` to the end `<wedge>` of diminuendos)

| Before fix | After fix |
| --- | --- |
| test-spread-0-at-dim-stop before | test-spread-0-at-dim-stop after |

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2229,10 +2229,12 @@ void MusicXmlInput::ReadMusicXmlDirection(
                     if (wedge->node().attribute("niente")) {
                         iter->first->SetNiente(ConvertWordToBool(wedge->node().attribute("niente").as_string()));
                     }
-                    if (wedge->node().attribute("spread")) {
-                        data_MEASUREMENTSIGNED opening;
-                        opening.SetVu(wedge->node().attribute("spread").as_double() / 5);
-                        iter->first->SetOpening(opening);
+                    if (iter->first->GetForm() == hairpinLog_FORM_cres) {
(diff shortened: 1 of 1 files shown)


paxbun:
BTW, I've also just found that `niente` must be handled similarly (ignored at the end of a crescendo and the start of a diminuendo, which is the opposite of `spread`), but I've never seen this kind of notation, and I don't have any MusicXML file containing this. I don't know if making a similar change for this is okay.

rettinghaus (approved):

PR #3448, opened 2023-06-09
Remove doubled attributes in SVG

rettinghaus:
This PR removes attributes that are applied twice in SVG (parent and child).

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -853,8 +853,8 @@ void SvgDeviceContext::StartText(int x, int y, data_HORIZONTALALIGNMENT alignmen
 
     m_currentNode = m_currentNode.append_child("text");
     m_svgNodeStack.push_back(m_currentNode);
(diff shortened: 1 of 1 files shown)


src/svgdevicecontext.cpp
```diff
@@ -931,8 +931,8 @@ void SvgDeviceContext::DrawText(
         svgText.replace(svgText.size() - 1, 1, "\xC2\xA0");
     }
 
-    std::string currentFaceName
-        = (m_currentNode.attribute("font-family")) ? m_currentNode.attribute("font-family").value() : "";
+    pugi::xpath_node fontNode = m_currentNode.select_node("ancestor::*[@font-family][1]");
+    std::string currentFaceName = (fontNode) ? fontNode.node().attribute("font-family").value() : "";
```
rettinghaus, line 935: Usually it's not the currentNode to look at, because we have nested `tspan` it is at least its parent. With this change it will properly look-up if the `font-family` is already specified.

PR #3454, opened 2023-06-16
Humdrum: Fix key labels for harms

WolfgangDrescher:
Before:

Bildschirmfoto 2023-06-16 um 12 03 12

After:

Bildschirmfoto 2023-06-16 um 12 02 38

Demo data:

```
**kern	**deg
*	*arr
*	*circ
=	=
*C-:	*C-:
4c	1+
*C:	*C:
4c	1
*C#:	*C#:
4c	1-
=	=
*D-:	*D-:
4c	7
*D:	*D:
4c	7-
*D#:	*D#:
4c	7--
=	=
*E-:	*E-:
4c	6
*E:	*E:
4c	6-
*E#:	*E#:
4c	6--
=	=
*F-:	*F-:
4c	5+
*F:	*F:
4c	5
*F#:	*F#:
4c	5-
=	=
*G-:	*G-:
4c	4+
*G:	*G:
4c	4
*G#:	*G#:
4c	4-
=	=
*A-:	*A-:
4c	3
*A:	*A:
4c	3-
*A#:	*A#:
4c	3--
=	=
*B-:	*B-:
4c	2
*B:	*B:
4c	2-
*B#:	*B#:
4c	2--
=	=
*-	*-

```

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -8752,8 +8752,8 @@ void HumdrumInput::addHarmLabel(
     std::string tempChar;
     for (int i = 0; i < (int)label.size(); i++) {
         switch (label[i]) {
-            case '#': output += output += U"\u266f"; break; // unicode sharp
-            case '-': output += output += U"\u266d"; break; // unicode flat
+            case '#': output += U"\u266f"; break; // unicode sharp
+            case '-': output += U"\u266d"; break; // unicode flat
             default: tempChar = label[i]; output += UTF8to32(tempChar);
         }
     }


craigsapp:
Thanks 😉

craigsapp:
You can change the default styling for the key label if you like.  I made it red because it was hard to see as plain text.   Red makes the text stand out, but people will wonder why it is red.  Maybe making it bold is better.  When there is an enclosure around the next text, adding more space will be good to...

WolfgangDrescher:
I would prefer a bold font. Typically I would draw a box around the key label when I analyze a score by hand. But this does currently not look good because it seems like verovio does not automatically calculate the needed space for the boxes:

Bildschirmfoto 2023-06-19 um 11 25 53

What would be the best syntax in humdrum to allow dynamically setting the style? Perhaps something like this?

```
!LO:HM:r=box:w=bold
*C:
```

---

Also note that `rand->SetRend(TEXTRENDITION_circle);` currently does render ellipses:

Bildschirmfoto 2023-06-19 um 11 23 24

rettinghaus:
@WolfgangDrescher See https://github.com/music-encoding/music-encoding/pull/1100

craigsapp:
Automatic styling of key labels is removed in commit https://github.com/rism-digital/verovio/commit/b21091bcdc7c7def65f884189e293e3f42a8c577

Example:

Screenshot 2023-06-19 at 8 22 14 AM

```tsv
**kern	**harm
*C:	*C:
=1	=1
1c	I
=	=
*-	*-
```

Now the color and font-family can be controlled with the verovio `svgCss` option.  Here is an example to color the text `darkred` and in `sans-serif` font:

Screenshot 2023-06-19 at 8 19 38 AM

```tsv
**kern	**harm
*C:	*C:
=1	=1
1c	I
=	=
*-	*-
!!!verovio: svgCss g.key-label { fill:darkred; font-family:sans-serif; }
```

`fill` is the CSS parameter to color text in SVG images.

craigsapp:
Preferably `<dir>` and `<harm>` horizontal collision avoidance would take into account enclosures:

Screenshot 2023-06-19 at 8 33 14 AM

The minimum spacing between adjacent enclosures or text should be the same as that for pure text (about 1 diatonic step or MEI virtual units).

You could add a verovio issue about that enhancement.

WolfgangDrescher:
> I cannot change the color or the font-family, so perhaps the CSS values will not override explicit styling in the SVG (I tried adding `!important1 but that did not work.

`rend2->SetFontweight(FONTWEIGHT_bold);` did work for me, in case we want to hardwire it.

> What does HM stand for?

`<harm>`. But your suggestion with `LO:KL` is fine for me.

> Automatic styling of key labels is removed in commit b21091b

Could/should we add a tandem interpretation for this?

Issue #3457, opened 2023-06-20
scoreDef@midi.bpm being ignored


craigsapp:
Some recent change in the last few weeks is causing `scoreDef@midi.bpm` to be ignored (It was noticed after updating my compiled version of verovio last weekend after not doing that for several weeks).  This seems to be a bug, but perhaps the method of encoding tempos for MIDI files has changed related to MEI version updates?

I would suspect this PR being involved: https://github.com/rism-digital/verovio/pull/3411

Test example:

<img width="765" alt="Screenshot 2023-06-20 at 7 39 12 AM" src="https://github.com/humdrum-tools/verovio-humdrum-viewer/assets/3487289/d3bdb853-87ed-4d52-bae3-60a0aaf82463">

<details>
<summary>
Click to view MEI data for above notation
</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2023-06-20T07:32:38" version="3.16.0-dev-b21091b">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m1gmwh72">
    <score xml:id="sitao8z">
     <scoreDef xml:id="sd3b2z8" midi.bpm="20.000000">
      <staffGrp xml:id="swlpe0b" bar.thru="true" symbol="brace">
       <staffDef xml:id="staffdef-L1F2" n="1" lines="5">
        <clef xml:id="c3fdo1y" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F2" count="4" unit="4" />
       </staffDef>
       <staffDef xml:id="staffdef-L1F1" n="2" lines="5">
[...]

PR #3463, opened 2023-06-22
Warn about changes in time map creation

rettinghaus:
This is a follow-up on #3457. We missed the regression, because the GH Action test didn't reported this. This changes the print statements so that warning are properly thrown in the Action (and you even don't have to download the artifact).

image

### DIFF
--- a/doc/test-suite.py
+++ b/doc/test-suite.py
@@ -1,12 +1,11 @@
 # This script it expected to be run from ./bindings/python
 import argparse
-import hashlib
 import json
 import os
 import sys
 import xml.etree.ElementTree as ET
 
-import cairosvg
+from cairosvg import svg2png
 
 # Add path for toolkit built in-place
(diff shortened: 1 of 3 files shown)


doc/test-suite.py
```diff
@@ -55,32 +54,31 @@
...
-            if shortlist and not(os.path.join(item1, item2) in shortlist):
+            if shortlist and not (os.path.join(item1, item2) in shortlist):
                 continue
 
             # reset the options
             options = testOptions.copy()
 
             # filenames (input MEI/XML and output SVG)
             inputFile = os.path.join(path1, item1, item2)
-            options.update({"xmlIdSeed": int(hashlib.sha256(
```
lpugin, line 82: What is the improvement here?

rettinghaus: Using the built-in function instead of mimicking it in Python.

PR #3464, opened 2023-06-22
RISM - DDMAL Merge #5 - Invisible Bbox

yinanazhou:
This PR merges changes for invisible bounding boxes. 

- Add empty syl handling for neume notation (Issue: https://github.com/DDMAL/Neon/issues/1016#issuecomment-1578292074)

This is the last PR for RISM - DDMAL merge🎉

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -3916,6 +3916,25 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
             m_doc->ConvertMarkupDoc(!m_doc->GetOptions()->m_preserveAnalyticalMarkup.GetValue());
         }
 
+        // Add child to empty syl in facsimile mode for neume notation
+        if (success && (m_doc->GetType() == Facs) && (m_doc->m_notationType == NOTATIONTYPE_neume)) {
+            ListOfObjects syls = m_doc->FindAllDescendantsByType(SYL);
+            ListOfObjects::iterator iter = syls.begin();
+            while (iter != syls.end()) {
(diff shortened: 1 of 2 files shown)


src/view_page.cpp
```diff
@@ -1745,6 +1745,15 @@ void View::DrawTextChildren(DeviceContext *dc, Object *parent, TextDrawingParams
...
 
+    if ((!parent->GetChildCount()) && (m_doc->GetType() == Facs) && (m_doc->m_notationType == NOTATIONTYPE_neume)) {
+        // add child to empty syl in facsimile mode for neume notation
+        Text *text = new Text();
+        std::u32string str = U"";
+
+        text->SetText(str);
+        parent->AddChild(text);
+    }
+
```
lpugin, line 1756: Adding some content to the data when rendering the page is very weird. Can't this be done when loading the file? It would be slightly better.

src/iomei.cpp
```diff
@@ -3916,6 +3916,25 @@ bool MEIInput::ReadDoc(pugi::xml_node root)
...
+
+                    Text *text = new Text();
+                    std::u32string str = U"";
+
+                    text->SetText(str);
+                    syl->AddChild(text);
+                }
+                ++iter;
+            }
+        }
```
lpugin, line 3936: Can't you do it in `MeiInput::ReadSyl`? That would be a better place and you would not need to make a search in the whole document.

yinanazhou: Oh right. Thanks for pointing out!

PR #3475, opened 2023-07-04
Update ios

rettinghaus:
NB: This is missing in the last release, but I'm not sure if it justifies a bugfix release.


lpugin:
I think the release is all good. See :
https://github.com/rism-digital/verovio/blob/6dc700daa6b5ab3c89bcedc0d751da6ea237a489/bindings/iOS/all.h#L142
https://github.com/rism-digital/verovio/blob/6dc700daa6b5ab3c89bcedc0d751da6ea237a489/bindings/iOS/all.h#L174-L176

PR #3486, opened 2023-07-12
Use PPU factor directly

DavidBauer1984:
This PR slightly simplifies the `ApplyPPUFactorFunctor` by storing and using the factor directly (instead of the page).

### DIFF
--- a/include/vrv/miscfunctor.h
+++ b/include/vrv/miscfunctor.h
@@ -52,8 +52,8 @@ private:
 public:
     //
 private:
-    // The current page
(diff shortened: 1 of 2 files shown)


lpugin:
The PPU factor is meant to be adjustable by page when using the page-based approach. I don't expect accessing the value from the page to have a significant impact. Is this really a problem for you?

DavidBauer1984:
This was just about code simplification, I don't expect any difference in runtime as well. It's not a big deal, we can close this PR if you prefer the previous version.

PR #3493, opened 2023-07-28
Update libmei and adjust code

rettinghaus:


lpugin:
Thanks! Where is the change in `midi/Binasc.cpp` coming from?

rettinghaus:
This comes from mixing up branches and improper commits, sorry. Should be better now.

PR #3495, opened 2023-08-07
MusicXML: When computing vgrp, consider relative-y in addition to default-y.

martinellimarco:
This simple PR fixes an issue related to overlapping elements during the computation of `vgrp`.

For example, this piece of MusicXML, generated with MuseScore, assigns the same `default-y` value to both dynamics, which leads to their rendering on top of each other.

```xml
      <direction placement="below">
        <direction-type>
          <dynamics default-x="50.82" default-y="-40.00" relative-x="-12.52" relative-y="-51.19">
            <sfz/>
          </dynamics>
        </direction-type>
        <sound dynamics="124.44"/>
      </direction>
      <direction placement="below">
        <direction-type>
          <dynamics default-x="30.10" default-y="-40.00" relative-x="-10.25" relative-y="-22.72">
            <pp/>
          </dynamics>
        </direction-type>
```

In the example, both `vgrp` values were computed as 40, resulting in the following:

broken

With the applied patch, `relative-y` is taken into consideration, resulting in two distinct `vgrp` values, leading to the following outcome:

fixed

I have also confirmed that the same issue arises in other instances where `vgrp` is utilized. The identical solution has been implemented there as well.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -2103,7 +2103,7 @@ void MusicXmlInput::ReadMusicXmlDirection(
     if (containsWords && !containsTempo && !containsDynamics) {
         pugi::xpath_node_set words = node.select_nodes("direction-type/*[self::words or self::coda or self::segno]");
         defaultY = words.first().node().attribute("default-y").as_int();
-        defaultY = (defaultY*10) + words.first().node().attribute("relative-y").as_int();
+        defaultY = (defaultY * 10) + words.first().node().attribute("relative-y").as_int();
         std::string wordStr = words.first().node().text().as_string();
         if (wordStr.rfind("cresc", 0) == 0 || wordStr.rfind("dim", 0) == 0 || wordStr.rfind("decresc", 0) == 0) {
             containsDynamics = true;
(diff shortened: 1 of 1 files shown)


rettinghaus:
Thanks. I'm a little worried that mixed dynamics (glyphs and hairpins) will not be aligned anymore after this change. The (difference in) `relative-y` should only be taken into account if there are multiple instances of the same type.

martinellimarco:
Oh, I see your point. I will make some tests and change the code accordingly.

Is there some reference I can use to catch edge cases? I'm new to both MusicXML and verovio.

lpugin:
@martinellimarco could you:
* Sign the CLA (https://book.verovio.org/introduction/licensing.html)
* Run the clang-formatting (https://book.verovio.org/contributing/guidelines.html#formatting)

Thanks!

martinellimarco:
@lpugin

- I've signed the CLA and sent it via email.
- I've run clang-format and pushed the updated code.

@rettinghaus

I've created a simple script to compare the output of patched and unpatched Verovio using ImageMagick's compare function. This allows me to visually identify any differences in the output.

I've conducted tests on approximately 150 MusicXML scores, all exported from MuseScore. Additionally, I've used synthetic examples with complex and unrealistic scenarios to push the limits of the testing. I've also tested all the example MusicXML files on musicxml.com.

So far, I haven't noticed any differences in the majority of scores, except for cases where there were previously overlapping elements that are now being corrected.

I identified a pattern where new overlaps occurred when the sum of `default-y` and `relative-y` matched the `default-y` value of another element. This is why I'm adjusting the `default-y` by a factor of 10. It's possible this adjustment could be made even larger as a precaution, though currently, I haven't encountered any instances where this led to a different outcome.

I plan to expand my testing to include MusicXML files generated by other software as well. I have a friend who uses Finale and has offered to assist me with this.

Despite my efforts, I still feel like I might be missing some testing scenarios. What are your thoughts? How can I further enhance my testing approach?

rettinghaus (approved):
@martinellimarco Thanks. I've tried to put together some test cases myself and your approach seems to do the trick.

martinellimarco:
I've tested it with Finale, and the results are the same. The only differences between the patched and unpatched software are that previously overlapping elements are now separated. While I haven't had the opportunity to test it with other software, I believe it's working.

Issue #3473, opened 2023-07-03
Missing glyph in develop_tab branch


lpugin:
There is a missing glyph ornamentLeftVerticalStrokeWithCross used in the develop_tab branch.

https://github.com/rism-digital/verovio/blob/4812281fe7c96c37d5962fef14970887b1bbfe4c/src/note.cpp#L364

It should be added to Leipzig first

lpugin:
Tagging @paul-bayleaf

rettinghaus:
This is a replacement glyph for the slashed number one: 

![](https://user-images.githubusercontent.com/7693447/251417797-2615acfb-6ed5-4adf-8afc-b8fd3c85daf1.png)

Maybe it would make sense to use primitives for this (until SMuFL provides a proper glyph)?

craigsapp:
Related: _Scaramella fa la galla_ by Loyset Compere

https://verovio.humdrum.org/?file=https://raw.githubusercontent.com/josquin-research-project/Com/master/Com3052-Scaramella_fa_la_galla.krn&v=evenNoteSpacing

<img width="907" alt="Screenshot 2023-07-11 at 11 46 13 PM" src="https://github.com/rism-digital/verovio/assets/3487289/fc6600f9-bda7-46de-b47c-119d4efad03b">

Josquin's _Scaramella_:

https://verovio.humdrum.org/?file=https://raw.githubusercontent.com/josquin-research-project/Jos/master/Jos2830-Scaramella.krn&v=evenNoteSpacing

<img width="765" alt="Screenshot 2023-07-11 at 11 49 31 PM" src="https://github.com/rism-digital/verovio/assets/3487289/2af71170-e179-4e68-bf23-eec4c4dcee3d">

rettinghaus:
@craigsapp how's that related?

craigsapp:
They have the same name 😉    The Compere Bassus first two measures are the same rhythm as the tablature (but I cannot read the tablature pitches so I don't know about that).  The Josquin one has similar pitch contour as Compere but in a triple meter.

paul-bayleaf:
I'm preparing a pull request for the E-LAUTE project for vertical positioning of notes and rhythm signs and for @notationsubtype.   I'll replace SMUFL_E595_ornamentLeftVerticalStrokeWithCross with a struck through  digit 1 at the same time, so SMUFL_E595 won't be needed.

paul-bayleaf:
[...]

PR #3512, opened 2023-09-06
Develop tab glt positioning

paul-bayleaf:
As per E-laute project implements positioning for German lute tablature using loc and valign.  Implements notationsubtype for different variants of the 6th course - although there is much more to do here.   Fix due to change in GetListSize interface.   Fix to use explicit strikethrough of digit one for open 6th course etc instead of SMUFL_E595_ornamentLeftVerticalStrokeWithCross.


paul-bayleaf:
Fixes Issue #3473

lpugin:
Thanks. Do we have a corresponding PR that has been made (and merged?) to the tablature-ig customization? We need to make sure the what comes into Verovio is in sync with that. Pinging @reinierdevalk @musicog @DILewis

paul-bayleaf:
Thanks.   

I've updated my pull request for the MEI 4.0.1 schema customizations for the German lute tablature work.  I note that it also works for MEI 5.0.  I use it to verify the MEI output of my luteconv program.
https://github.com/music-encoding/tablature-ig/pull/26
But this has not been merged.

If you are referring to 
https://github.com/music-encoding/music-encoding/tree/tablature-ig
then that was last updated on 2021-11-16.  I've not touched this.

lpugin:
Great, thanks. @reinierdevalk or @DILewis: could you look at updating the [tag](https://github.com/music-encoding/music-encoding/tree/tablature-ig) to MEI 5.0 ?

lpugin:
I have not updated https://github.com/music-encoding/music-encoding/tree/tablature-ig and brought it inline with https://github.com/music-encoding/music-encoding/tree/develop.

The proposed changed should be added there. The two modifications adding `att.verticalAlign` to `note` and `staffDef` should be brought up for discussion since these are not tablature-specific elements.

src/note.cpp
```diff
@@ -350,23 +352,38 @@ std::u32string Note::GetTabFretString(data_NOTATIONTYPE notationType, bool &over
...
+        // parts of the SMuFL collection.  Overlines and strike throughs are drawn separately.
 
-        if (course == 6 && fret >= 0 && fret <= 13) {
-            if (fret == 0) {
-                fretStr = SMUFL_E595_ornamentLeftVerticalStrokeWithCross; // substitute for 1 with oblique stroke
-            }
-            else {
+        if (course >= 6 && fret >= 0 && fret <= 13) {
+            // TODO more GLT subtypes needed, is this subtype too specific?
+            if (notationSubtype == "judenkunig_1523") {
```
lpugin, line 368: Looking again at this PR, this is also problematic. We should not have in Verovio hardcoded values like that. For the glyphs missing I would suggest using a symbol table. This is already supported by Verovio (see the test suite examples for symbols). Pinging @musicog.

musicog: As decided at the MEI Development Workshop at mdw, we will be defaulting to Newsidler unless an alternate symbol is defined (via `@glyphnum` or `@altsym` on `<note>`, or via a `<symbolTable>`)

rettinghaus: Just for reference: https://github.com/w3c/smufl/issues/274

paul-bayleaf: Does this mean that @notationsubtype is no longer to be used for the >= 6th course variants?   I.e. there are no "predefined variant sets"?

If so I can remove the @notationsubtype code and just keep the Newsidler variant.   This pull request would then  just implement the vertical positioning  with @loc and @valign and fix Issue https://github.com/rism-digital/verovio/issues/3473.   Would that be sufficient to have this pull request accepted?  Or are more changes required?

I would like to deal with implementing @glyphnum and @altsym etc in a separate pull request - first I need to study how to do this.

Thanks

lpugin: @paul-bayleaf yes please remove the `@notationsubtype`-dependent code and I can then merge the PR.

The `@valign` on `staffDef`will be renamed at some point, but the code can easily be adjusted once the renaming is done. Conceptually nothing will change so I am OK to merge it.

lpugin (approved):

Issue #3521, opened 2023-09-28
JavaScript toolkit crashing


lpugin:
**Describe the problem**
it seems that the JavaScript version is crashing somehow randomly with the develop version. I cannot reproduce the problem locally but loading the file below in the verovio editor does crash.

**To Reproduce**
Load the file below in the [Verovio editor](https://editor.verovio.org/)

<details>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title />
         </titleStmt>
         <pubStmt>
            <date isodate="2023-09-28-21:03:59" />
         </pubStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application xml:id="verovio" version="4.1.0-dev-ae7c305">
               <name>Verovio (4.1.0-dev-ae7c305)</name>
            </application>
         </appInfo>
         <projectDesc>
            <p>MEI encoded with Verovio</p>
         </projectDesc>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv xml:id="m19nj5e9" n="1" type="variation">
            <score xml:id="soqsi69">
               <scoreDef xml:id="sjft1g2" keysig="1s" meter.count="3" meter.unit="4">
                  <staffGrp xml:id="smlovzj">
                     <staffDef xml:id="s13jbbey" n="1" lines="5" clef.shape="G" clef.line="2" />
                     <staffDef xml:id="shbsl03" n="2" lines="5" clef.shape="F" clef.line="4" />
                  </staffGrp>
               </scoreDef>
               <section xml:id="ssm35er" n="1">
                  <measure xml:id="m2rcrlx" n="1">
                     <staff xml:id="s36inxr" n="1">
                        <layer xml:id="la56g6u" n="1">
                           <note xml:id="n0111-1" dur="4" oct="5" pname="g" />
                           <note xml:id="n0111-2" dur="4" oct="5" pname="g" />
                           <beam xml:id="b1kzq82m">
[...]

PR #3522, opened 2023-09-28
Revert "Replace handling of current scoreDef in Doc"

lpugin:
Reverts rism-digital/verovio#3518

@DavidBauer1984 I am reverting this to narrow down the #3521 issue. Maybe it is unrelated to this PR, which could mean this is a WebAssembly issue.


DavidBauer1984:
OK let me know if I can help somehow or whenever you found any problems with the scoreDef PR.

DavidBauer1984:
@lpugin I guess I see the problem: the visible scores are not cleaned up during document reset. When doing two successive loads with the toolkit, `m_visibleScores` still contains scores from the first load. The list is only lazily updated when autogenerating headers and footers, and at this point accesses a destroyed score from the first load.

To fix this, it is probably enough to add `m_visibleScores.clear()` in `Doc::Reset`. I can do some more testing and open another PR in the next days.

lpugin:
Yes, it does look like `m_visibleScores` is holding invalid data. Since I reverted the PR you can take your time to fix it.

PR #3526, opened 2023-10-02
update podspec

rettinghaus:
I made some updates to the podspec file, but now I'm stuck with this error:

`humlib.h:45:10: fatal error: 'algorithm' file not found`

### DIFF
--- a/Verovio.podspec
+++ b/Verovio.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name           = 'Verovio'
-  s.version        = '4.0.1'
+  s.version        = '4.1.0-dev'
   s.license        = { :type => 'LGPL', :file => 'COPYING' }
   s.homepage       = 'https://www.verovio.org/index.xhtml'
   s.authors        = { 'Contributors List' => 'https://github.com/rism-digital/verovio/graphs/contributors' }
--- a/Verovio.podspec
+++ b/Verovio.podspec
@@ -1,21 +1,21 @@
 Pod::Spec.new do |s|
   s.name           = 'Verovio'
-  s.version        = '4.1.0-dev'
-  s.license        = { :type => 'LGPL' }
+  s.version        = '4.0.1'
+  s.license        = { :type => 'LGPL', :file => 'COPYING' }
   s.homepage       = 'https://www.verovio.org/index.xhtml'
   s.authors        = { 'Contributors List' => 'https://github.com/rism-digital/verovio/graphs/contributors' }
   s.summary        = 'Verovio'
   s.source         = { :git => 'https://github.com/rism-digital/verovio.git', :tag => 'version-' + s.version.to_s }
-  s.swift_versions = ['3.0', '4.0', '4.1', '4.2', '5.0','5.1']
+  s.swift_versions = ['3.0', '4.0', '4.1', '4.2', '5.0', '5.1', '5.2', '5.3', '5.4', '5.5', '5.6', '5.7']
   s.source_files        = 'src/**/*.{h,cpp,cc}',
-                          'include/{crc,hum,json,midi,pugi,utf8,vrv,zip}/*.{h,hpp}',
-                          'libmei/{dist,addons}/*.{h,cpp}'
+                        'include/{crc,hum,json,midi,pugi,utf8,vrv,zip}/*.{h,hpp}',
+                        'libmei/{dist,addons}/*.{h,cpp}'
   s.public_header_files = 'src/**/*.{h}',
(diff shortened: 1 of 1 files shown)


craigsapp:
> `humlib.h:45:10: fatal error: 'algorithm' file not found`

Line 45:

```cpp
#include <algorithm>
```

This is a standard C++ header file:

https://cplusplus.com/reference/algorithm

Do you know what compiler/version is being used?

rettinghaus:
LLVM clang version 15.0.0 on macOS 14.

rettinghaus:
It's strange, because we have it also in other headers, but no complaints on them.

craigsapp:
Yes, that is strange.  Perhaps `humlib.h` is being treated as a C include file somehow?

There are two C header files loaded just before `<algorithm>`:

```cpp

#include <stdarg.h>
#include <string.h>

#include <algorithm>
#include <cctype>
#include <chrono>
#include <cmath>
#include <cstring>
#include <ctime>
#include <fstream>
#include <functional>
#include <iomanip>
#include <iostream>
#include <list>
#include <locale>
#include <map>
#include <regex>
#include <set>
#include <sstream>
#include <string>
#include <utility>
#include <vector>
```

Perhaps converting 

```cpp
#include <stdarg.h>
#include <string.h>
```

will fix the problem.   Mixing C and C++ header files should not be a problem, but it would be good to try.

In C++, you can include C headers like `<stdarg.h>` and `<string.h>` using the #include directive as you've shown. C++ is designed to be compatible with C, so you can use C headers in C++ code without any issues.

Here's how you can include them in C++ code to use only C++ headers:

```cpp
#include <cstdarg> // for <stdarg.h>
#include <cstring>  // for <string.h>
```

And see if the problem with `#include <algorithm>` is then fixed.

rettinghaus:
Funny, because now it's 

`humlib.h:42:10: fatal error: 'cstdarg' file not found`

Verovio.podspec
```diff
@@ -1,21 +1,21 @@
 Pod::Spec.new do |s|
   s.name           = 'Verovio'
-  s.version        = '4.1.0-dev'
-  s.license        = { :type => 'LGPL' }
+  s.version        = '4.0.1'
```
lpugin, line 3: keep `4.1.0-dev`

lpugin:
> `humlib.h:45:10: fatal error: 'algorithm' file not found`

When are you having this error? Building what?

Verovio.podspec
rettinghaus: No, that will not work. The pod linter checks out the repo and builds it locally, and there is no `4.1.0-dev` tag.

craigsapp:
> Funny, because now it's

`humlib.h:42:10: fatal error: 'cstdarg' file not found`

Funny, but predictable yet still strange: it is complaining about the first C++ header file, but skipped over it when it was formatted as a C header include.   So somehow it seems that humlib.h is treated as C source code.

I have clang 14.0.3 on MacOS 13.4.1 (Ventura).   Maybe I should risk updating to macOS 14 Sonoma which was released last week (https://en.wikipedia.org/wiki/MacOS_Sonoma).

On Linux, I have clang 16.0.6 (but uses gcc 13.2.1).  I am getting only one warning compiling verovio, but it is a new one:

```
/user/c/craig/git-cloud/verovio/src/iopae.cpp: In member function ‘void vrv::PAEInput::ParseHeader(jsonxx::Object&)’:
/user/c/craig/git-cloud/verovio/src/iopae.cpp:3231:29: warning: loop variable ‘key’ of type ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’} binds to a temporary constructed from type ‘const char* const’ [-Wrange-loop-construct]
 3231 |     for (const std::string &key : { "scoring", "key_mode", "role", "voice_intrument" }) {
      |                             ^~~
/user/c/craig/git-cloud/verovio/src/iopae.cpp:3231:29: note: use non-reference type ‘const std::string’ {aka ‘const std::__cxx11::basic_string<char>’} to make the copy explicit or ‘const char* const&’ to prevent copying
```

rettinghaus:
It's not connected to macOS, I hit this particular thing before.

Verovio.podspec
lpugin: No, it will be push from `master` with `4.1.0` then and there will be a corresponding.

lpugin:
It builds without problem and no warning for me with xcode 15.0 on macos 13.5.2

lpugin:
And seems to build fine here https://github.com/rism-digital/verovio/actions/runs/6378048545/job/17307890306

lpugin (changes requested):
Could you change back the version? Then I can merge it and update the pod after the next release.

lpugin (approved):

PR #3555, opened 2023-12-05
More metadata fixes

gregchapman-dev:
Biggest change is a reworking of Humdrum date parsing to handle complex date ranges such as "!!!CDT:1750^1751-~1802/6" (which means composer birthdate is between 1750 and 1751, and composer death date is approximately June 1802).  Several other fixes as well.  Verovio and converter21 now produce exactly the same MEI metadata when converting Humdrum to MEI (tested on a few thousand Humdrum scores from various organizations).

I believe my Humdrum to MEI metadata work is now complete.

### DIFF
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -509,12 +509,12 @@ public:
      * A copy of the back tree stored as pugi::xml_document
      */
     pugi::xml_document m_back;
-    
+
     /**
      * The music@decls value
      */
     std::string m_musicDecls;
-    
+
     /** The current page height */
     int m_drawingPageHeight;
     /** The current page width */
--- a/include/vrv/doc.h
+++ b/include/vrv/doc.h
@@ -509,7 +509,12 @@ public:
      * A copy of the back tree stored as pugi::xml_document
      */
     pugi::xml_document m_back;
-
+    
+    /**
+     * The music@decls value
+     */
+    std::string m_musicDecls;
+    
     /** The current page height */
     int m_drawingPageHeight;
     /** The current page width */
--- a/include/vrv/iohumdrum.h
+++ b/include/vrv/iohumdrum.h
@@ -466,8 +466,9 @@ struct DateWithErrors {
 };
 
 struct DateConstruct {
-    // constructType can be any of "" (invalid), "DateSingle" (one date), "DateRelative" (one date, qualifier="before" or "after"),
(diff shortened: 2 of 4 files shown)


gregchapman-dev:
@lpugin please review my iomei.cpp/doc.h changes.  @craigsapp please review the iohumdrum.{h, cpp} changes.

gregchapman-dev:
One last change to emitted EDTF (in MADS data) from a review by Perry.

lpugin:
Could you post a sample Humdrum and MEI file? Thanks

include/vrv/doc.h
```diff
@@ -510,6 +510,11 @@ class Doc : public Object {
      */
     pugi::xml_document m_back;
 
+    /**
+     * The music@decls value
+     */
+    std::string m_musicDecls;
```
lpugin, line 516: this should be stored as a `pugi::xml_document`

lpugin (changes requested):

include/vrv/doc.h
gregchapman-dev: @lpugin That's easy to do. Question: it's a bit more than iomei.cpp actually needs; should iomei ignore any sub-elements (since there are already doc->m_back and doc->m_front), and thus only take attributes from doc->m_music?  Or should any sub-elements of doc->m_music be merged with doc->m_back and doc->m_front and place in the output `<music>` element?

gregchapman-dev:
Here is an example Humdrum file and the resulting output.mei file (both with .txt extensions, so GitHub will take them).
[Agnus_00.krn.txt](https://github.com/rism-digital/verovio/files/13619944/Agnus_00.krn.txt)
[output.mei.txt](https://github.com/rism-digital/verovio/files/13619949/output.mei.txt)

include/vrv/doc.h
gregchapman-dev: Alternatively, I could get rid of doc->m_back and doc->m_front, replacing them with doc->m_music (which would then contain any back/front matter).  But that would be a backward-incompatible change, requiring other changes throughout verovio, and might break other folks' code that is based on verovio code.

lpugin: Sorry, ignore my previous comment. I thought you were dealing with a sub-tree, not just an attribute.

lpugin (approved):

craigsapp (approved):

PR #3578, opened 2024-01-24
Upgrading to C++20

lpugin:
No major changes beside:
* Updating make files
* Moving the `LedgerLine` definition up (before `Staff) because of one compiler complaining about it
* Ajust the CI (remove unsupported OS / compiler combinations and add new ones)

If upgrading to C++20 is a problem, please speak up. It is not absolutely necessary to do this now, but at the same time there seems to be no reason no to do it.


lpugin:
Thanks for checking this. This is the status of the building tests on GH
https://github.com/rism-digital/verovio/actions/runs/7583048214

They all look fine. There is however one problem on Windows https://ci.appveyor.com/project/LaurentPugin/verovio-r1t6l/builds/49024120

PR #3580, opened 2024-01-24
Fix resource path long option

igneus:
both `verovio -h base` and [online documentation](https://book.verovio.org/installing-or-building-from-sources/command-line.html#building-on-macos-or-linux) advertise `--resource-path` as long option for `-r`, but it doesn't work and the actual long option is `--resources`. This is due to the options being defined on two places (currently out of sync), one definition being used to print help, the other to actually parse the command line arguments.

This PR consists of two commits. The first one is the minimal change required to make the `--resource-path` option work as promised by the documentation. The latter is a refactoring which eliminates (most of) the code duplication that was at the root of the issue.

### DIFF
--- a/tools/main.cpp
+++ b/tools/main.cpp
@@ -72,6 +72,8 @@ bool optionExists(const std::string &option, int argc, char **argv, std::string 
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


lpugin:
Please fix the formatting with clang-format https://book.verovio.org/contributing/guidelines.html#formatting

igneus:
Done.

lpugin:
Thanks!

Issue #3575, opened 2024-01-22
Implement fontname attr support on clef and meterSig elements 


fernandoherreradelasheras:
It would be great to have support for using a different fontname for some of the symbols in the score.

My main use case for this is to include, as an indication, the first or two first measures of the score with the original keys, clefs and meters along the original initial notes so the reader can know what changes has been done in the transcription. This is a usual practice when editing early baroque music that involves meter adaptations (like for example the ternary CZ in the iberian baroque). For example: 
![Informational measures on a baroque score](https://github.com/rism-digital/verovio/assets/20201639/4fd40ff3-1671-4e35-8a9b-ed27b1aa02f8)

I have been trying to achieve this with MEI+verovio using different approaches and none of them worked out well:
* editorial content: cannot layout in the same line
* using mensural notation for one section and then switching to CMN: not supported and also the mensural symbols doesn't fit perfectly with the intended look. 
* rendering those measures from a separate score as an incipit and the importing it as a graphic: hard to get it looking nice and not responsibe
* add those specific bits as svgs symbolDef in the symbolTable and then use altsym="#xxx" to render a custom graphic: this was problematic because of the symbols rendered as svg are supported mostly for floating elements and getting anything in the staff with the proper size and spacing was too complicated. (At leas from playing with this code I got a cuple of bug fixed on the bounding boxes that will sumit later)

So at this point my only viable option was via custom font, but maybe I am missing something and there could be other options. This would look like this using MEI fontname attribute:
```
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Custom fonts</title>
         </titleStmt>
         <pubStmt>
            <date isodate="2024-01-21"/>
         </pubStmt>
         <seriesStmt>
            <title>Verovio test suite</title>
         </seriesStmt>
         <notesStmt>
            <annot>Verovio supports fontname attribute in clef and meterSig elements</annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="4.2.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef keysig="2s">
                  <meterSig glyph.num="U+E080" count="3" unit="2" fontname="Patriarca"/>
                  <staffGrp>
                     <staffDef n="1" lines="5">
[...]

PR #3582, opened 2024-01-26
Support multiple fonts adding an option to add extra fonts.  Add fontname support for clef and meter symbols. 

fernandoherreradelasheras:
See issue #3575

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -686,6 +686,7 @@ public:
     OptionDbl m_extenderLineMinSpace;
     OptionDbl m_fingeringScale;
     OptionString m_font;
+    OptionArray m_addCustomFont;
     OptionDbl m_graceFactor;
     OptionBool m_graceRhythmAlign;
     OptionBool m_graceRightAlign;
--- a/include/vrv/clef.h
+++ b/include/vrv/clef.h
@@ -35,6 +35,7 @@ class Clef : public LayerElement,
              public AttOctave,
              public AttOctaveDisplacement,
              public AttStaffIdent,
+             public AttTypography,
              public AttVisibility {
 public:
     /**
--- a/include/vrv/metersig.h
+++ b/include/vrv/metersig.h
@@ -8,6 +8,7 @@
 #ifndef __VRV_METERSIG_H__
 #define __VRV_METERSIG_H__
 
+#include "atts_externalsymbols.h"
 #include "atts_shared.h"
 #include "atts_visual.h"
 #include "layerelement.h"
@@ -25,8 +26,10 @@ class ScoreDefInterface;
  */
 class MeterSig : public LayerElement,
                  public AttEnclosingChars,
+                 public AttExtSymNames,
                  public AttMeterSigLog,
                  public AttMeterSigVis,
+                 public AttTypography,
                  public AttVisibility {
 public:
     /**
--- a/include/vrv/resources.h
+++ b/include/vrv/resources.h
@@ -56,12 +56,13 @@ public:
      */
     ///@{
     /** Init the SMufL music and text fonts */
-    bool InitFonts();
+    bool InitFonts(const std::vector<std::string> &extraFonts, const std::string &defaultFont);
     /** Init the text font (bounding boxes and ASCII only) */
     bool InitTextFont(const std::string &fontName, const StyleAttributes &style);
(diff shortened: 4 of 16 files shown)


lpugin:
Thanks. Could you make your font `Patriarca` available somewhere?

lpugin:
You need to run clang-format. See https://book.verovio.org/contributing/guidelines.html#formatting

fernandoherreradelasheras:
> Thanks. Could you make your font `Patriarca` available somewhere?

Sure!

Patriarca font is described in [^1]. It is based on Capitan font from another project (described in [^2]). The version I am using was included in the source code for the MuRET project [^3], however, the public repository with MuRET code [^4] was removed from github some months ago. There is still a public repository with one of the libraries developed for MuRET that has a copy of Patriarca font:

https://github.com/davidrizo/im3/tree/master/imcore/src/main/resources/fonts/patriarca

The font files don't have any copyright notice so I am not sure about the license for them (MuRET code itself was licensed using a mix of Apache and GPLv3).



[^1]: https://grfia.dlsi.ua.es/repositori/grfia/pubs/403/1953-8089-1-PB.pdf
[^2]: https://anuariomusical.revistas.csic.es/index.php/anuariomusical/article/view/221/221
[^3]: https://muret.dlsi.ua.es/muret/
[^4]: https://github.com/davidrizo/muret

ahankinson:
Maybe @davidrizo can clarify the license question?

fernandoherreradelasheras:
> You need to run clang-format. See https://book.verovio.org/contributing/guidelines.html#formatting

Done!

include/vrv/options.h
```diff
@@ -686,6 +686,7 @@ class Options {
     OptionDbl m_extenderLineMinSpace;
     OptionDbl m_fingeringScale;
     OptionString m_font;
+    OptionArray m_addCustomFont;
```
lpugin, line 689: Rename to m_fontCustom

fernandoherreradelasheras:
> Could you comment on what is hacky and patchy?

 When loading a font :
`LoadFont(const std::string &fontName, bool withFallback = true, bool buildNameTable = false);`
these semantics are confusing. We explicitly indicate in the call that we want to build the global name table from it and that we don't want to use a fallback font for any missing glyphs. However which font acts as fallback is implicitly indicated by the order of loading. Maybe we could add a separate call `SetFallbackFont` or add a `isFallbackFont` parameter to `LoadFont` .

About the withFallback parameter I am not sure, because it is not exposed via any option. Without being exposed it's just a policy . What should be this? verovio distributed fonts are expected to have all glyphs and they are safe to use without fallback? then any user added font (included if set as default) should be assumed that need fallback? Is there a use case when people might want to not render a all missing glyphs from they selected font?

Then the data structures I used for keeping several font maps are neither the most optimal or the most straightforward to follow: a map with file path as key and (glyph, id) struct as value, where id is calculated on insertion based on the number of glyphs loaded before for this same code + another map to keep that count. This was motivated because I thought maintaining svg output from different versions of verovio for the same input as similar as possible across different versions of verovio was a desirable goal (and this implementation was mapped around trying to keep the same format of the ids as SmuflCode-VariableSuffix for all the non-new cases). But then I though I was overthinking this and maybe that goal doesn't make that much sense in this case (with the variable suffix a diff between two runs of the same version is already not useful and only maybe some xslt filters may rely on this format).

Finally, code using different fonts does:
```
Resources &resources = m_doc->GetResourcesForModification();
resources.SetCurrentFont(customFont);
```
and maybe we should move the current font somewhere else because we are not really modifying the resources themselves just choosing one of the for the next operation.


Those were my main concerns about this implementation plus, of course, any other thing I might have missed.

davidrizo:
> Maybe @davidrizo can clarify the license question?

Hi @fernandoherreradelasheras, @ahankinson 

Currently, Patriarca is being used inside MuRET to render the agnostic encoding. 

We have not yet finished the font to make it compatible with Verovio. There are always more urgents things to do... We have done some experiments with it, but elements such as the ligature are not working well yet, and some glyphs need to be corrected. Our plan is to have it ready for the summer, with the new version of MuRET.

It will be made publicly available, I don't know which license yet, it's possible that we'll use SIL Open Font License.

@fernandoherreradelasheras You can contact me by mail to know more details if you need.

PR #3583, opened 2024-01-27
base options: also argument requirement governed by the Option object

igneus:
In #3580 I was removing, as far as possible, code duplication regarding definition of the base options. But I left the `required_argument` / `no_argument` specification in `main.cpp`, as I wasn't confident that my understanding of the context is sufficient not to accidentally break something. Having read a bit more, I propose one more refactoring to get rid of this last bit of duplicate definition. Let the `Option` objects specify also if they require a command line argument or not.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -122,6 +122,7 @@ public:
     void SetShortOption(char shortOption, bool isCmdOnly);
     char GetShortOption() const { return m_shortOption; }
     bool IsCmdOnly() const { return m_isCmdOnly; }
+    bool IsArgumentRequired() const { return true; }
 
     /**
      * Return a JSON object for the option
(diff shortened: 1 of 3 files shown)


lpugin:
There was a problem with your [previous PR](https://github.com/rism-digital/verovio/pull/3580). The `const char*` in `option` is a dangling pointer because it relies on the string returned by `FromCamelCase`. I'll fix it after having merged this.

igneus:
I'm sorry for that.

Given that my attempt to reduce duplicate definitions resulted only in code more convoluted and harder to understand, but with almost the same amount of hardcoded duplicate data, at this point I would probably revert
762f6a99be8bc99c9a99bfc1c025d7574df2e22a 985137d264f2d765aa3fdad9eddafcd35970ff5a c9ddec064646fc76db831348d2c8500164610412 43cbfdead83b5bd707cad410f460ab4551e38da5 (i.e. my refactoring attempts foiled by poor memory management + your subsequent cleanup). Before there were duplicate definitions of the base options, but the code was (after the little fix in 175f0f49673ee3e0ac33bf41da4e46bc587b127d) correct and easy to read.

lpugin:
No worries. I agree it is now not easier to read, but at least the option names are sure to be right. We can leave it like this. Thanks for contributing!

PR #3584, opened 2024-01-29
Update actions to remove deprecation warning

rettinghaus:


lpugin:
I had to revert it because it broke artifact upload

Issue #3589, opened 2024-02-05
De-emphasise dividing lines when rendering German Lute Tablature


musicog:
This issue concerns the rendering of German Lute Tablature (GLT) as implemented by @paul-bayleaf and currently available in branch `develop-tab`.

![image](https://github.com/rism-digital/verovio/assets/5521800/d26c74c0-4ed7-408b-ac0f-d3bc9a365269)

The  image above displays the current rendering produced for an example GLT encoding as of PR #3581. It's wonderful to see this rendering in Verovio! However, having run the output by @Kateryna333 (Kateryna Schöning, P.I., E-LAUTE), I would like to forward her suggestion to de-emphasise the dividing lines between systems, by: 

1. slightly lowering the lines to ensure no contact between the note glyphs and the dividers, and 
2. making the dividing lines lighter in colour so as not to dominate the image

This would allow the lines to better serve in their intended role as reading aids, without becoming distracting. 

It's possible that 1. is font-dependent -- the glyphs are sufficiently separated from the lines when the Bravura font is specified:

![image](https://github.com/rism-digital/verovio/assets/5521800/edbfac79-0eb4-4d5a-98d5-f8ffe3a573ba)

PR #3590, opened 2024-02-06
Fix issue #3589 GLT de-emphasise dividing lines

paul-bayleaf:
Fix issue #3589 Slightly move dividing lines down and reduce line thickness.  (Also added a trivial but unconnected fix for when @lines.visible="true" for GLT).

Leipzig font before this pull request
before-Leipzig
 
Leipzig font after this pull request, slight lower and thinner dividing lines 
after-Leipzig

Bravura font before this pull request
before-Bravura

Bravura font after this pull request, slight lower and thinner dividing lines 
after-Bravura


lpugin (approved):

musicog (approved):
Thanks for getting around to this so quickly! Looks good to me.

Issue #3605, opened 2024-02-24
notes with colored="true" and dur="2" get rendered in without the note head color inversion (CMN)


fernandoherreradelasheras:
in CMN colored="true" is reflected by inverting the note head fill / void. This works fine for all note durations but for dur="2" that remains as a normal half note (and it should look like a quarter note).

See the following MEI:
```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp symbol="bracket">
                     <staffDef n="1" lines="5">
                        <clef shape="G" line="2" />
                        <meterSig count="2" unit="8" />
                     </staffDef>
                     <staffDef n="2" lines="5">
                        <clef shape="G" line="2" />
                        <meterSig count="2" unit="8" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <note oct="5" pname="e" dur="1" />
                           <note oct="5" pname="e" dur="2" />
                           <note oct="5" pname="e" dur="4" />
                           <note oct="5" pname="e" dur="8" />
                           <note oct="5" pname="e" dur="16" />
                        </layer>
                     </staff>
                     <staff n="2">
                        <layer n="1">
                           <note oct="5" pname="e" dur="1" colored="true" />
                           <note oct="5" pname="e" dur="2" colored="true" />
                           <note oct="5" pname="e" dur="4" colored="true" />
                           <note oct="5" pname="e" dur="8" colored="true" />
                           <note oct="5" pname="e" dur="16" colored="true" />
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

currently renders to:

![note-012](https://github.com/rism-digital/verovio/assets/20201639/aaf3463c-2db4-404a-b8ea-04c57ce844ff)

and it should be:

[...]

PR #3606, opened 2024-02-24
Fix issue #3605: draw colored notes and dur="2" with empty note head

fernandoherreradelasheras:


src/view_element.cpp
```diff
@@ -1488,16 +1488,22 @@ void View::DrawNote(DeviceContext *dc, LayerElement *element, Layer *layer, Staf
...
         else {
             // Whole notes
             char32_t fontNo;
             if (note->GetColored() == BOOLEAN_true) {
-                fontNo = (drawingDur == DUR_1) ? SMUFL_E0FA_noteheadWholeFilled : SMUFL_E0A3_noteheadHalf;
+                if (DUR_1 == drawingDur) {
+                    fontNo = SMUFL_E0FA_noteheadWholeFilled;
+                }
+                else if (DUR_2 == drawingDur) {
+                    fontNo = SMUFL_E0A4_noteheadBlack;
```
rettinghaus, line 1502: Shouldn't that then be `noteheadHalfFilled`?

fernandoherreradelasheras: uhmm, why the half?

rettinghaus: Because it's a filled half note. In _Leipzig_ the glyph is basically the same as the quarter head, but there are other fonts where they differ. I've added the missing glyph in https://github.com/rism-digital/leipzig/pull/32

lpugin (changes requested):
Please change to `noteheadHalfFilled`

fernandoherreradelasheras:
> Please change to `noteheadHalfFilled`

Sorry for the delay. Done now and force pushed and update commint

rettinghaus (approved):

lpugin (approved):

PR #3607, opened 2024-02-24
fonts update with filled half note

rettinghaus:


lpugin:
Do you know what has changed in `E0FA` in Leipzig?

rettinghaus:
Yes, I took `E0A2` as basis, which was constructed simpler.

PR #3611, opened 2024-02-28
Update vrv.cpp

ammatwain:
To ensure correct code formatting for floating-point values, two lines of code have been added to the "StringFormat" method, temporarily resetting the compiler's std::locale configurations. This prevents formatting issues stemming from specific OS or development environment settings, ensuring consistent and accurate numerical representation in the C++ source code during program execution. (At least, they do on my machine; I hope it's the same on yours... :) )

### DIFF
--- a/src/vrv.cpp
+++ b/src/vrv.cpp
@@ -211,12 +211,14 @@ void EnableLogToBuffer(bool value)
 
 std::string StringFormat(const char *fmt, ...)
 {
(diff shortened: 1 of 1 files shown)


craigsapp:
That sounds good.  Can you give an example of what happens in your locale that causes a problem?

Exponentiation of small numbers can be a problem, I imagine (where `0.00000001` might be output as `1e-8`).

craigsapp:
I found the initial issue for this PR :-)

PR #3617, opened 2024-03-06
Added ClearCoords call when editing a beam

Matthew-w56:
An error was occurring when a beam was deleted from, because the beam's element coordinate references were not being updated.  This fixes that problem

### DIFF
--- a/src/editortoolkit_cmn.cpp
+++ b/src/editortoolkit_cmn.cpp
@@ -213,7 +213,7 @@ bool EditorToolkitCMN::Delete(std::string &elementId)
 {
     Object *element = this->GetElement(elementId);
     if (!element) return false;
-
+	
     if (element->Is(NOTE)) {
         return this->DeleteNote(vrv_cast<Note *>(element));
     }
@@ -247,15 +247,15 @@ bool EditorToolkitCMN::KeyDown(std::string &elementId, int key, bool shiftKey, b
 
     // For elements whose y-position corresponds to a certain pitch
     if (element->HasInterface(INTERFACE_PITCH)) {
-        PitchInterface *interface = element->GetPitchInterface();
-        assert(interface);
+        PitchInterface* pitch_interface = element->GetPitchInterface();
+        assert(pitch_interface);
         int step;
         switch (key) {
             case KEY_UP: step = 1; break;
             case KEY_DOWN: step = -1; break;
             default: step = 0;
         }
-        interface->AdjustPitchByOffset(step);
+        pitch_interface->AdjustPitchByOffset(step);
         return true;
     }
     return false;
@@ -301,11 +301,11 @@ bool EditorToolkitCMN::Insert(std::string &elementType, std::string const &start
     }
 
     assert(element);
-    TimeSpanningInterface *interface = element->GetTimeSpanningInterface();
-    assert(interface);
+    TimeSpanningInterface *timespan_interface = element->GetTimeSpanningInterface();
+    assert(timespan_interface);
     measure->AddChild(element);
-    interface->SetStartid("#" + startid);
-    interface->SetEndid("#" + endid);
+    timespan_interface->SetStartid("#" + startid);
+    timespan_interface->SetEndid("#" + endid);
 
     m_chainedId = element->GetID();
     m_editInfo.import("uuid", element->GetID());
@@ -512,7 +512,7 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
 
     Chord *chord = note->IsChordTone();
     Beam *beam = note->GetAncestorBeam();
-
+	
     if (chord) {
         if (chord->HasEditorialContent()) {
             LogInfo("Deleting a note in a chord that has editorial content is not possible");
@@ -560,6 +560,7 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
         }
     }
     else if (beam) {
+		// If the beam has exactly 2 notes (take apart and leave a single note and a rest)
         if ((int)beam->m_beamSegment.GetElementCoordRefs()->size() == 2) {
             bool insertBefore = true;
             LayerElement *otherElement = beam->m_beamSegment.GetElementCoordRefs()->back()->m_element;
@@ -584,7 +585,8 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
             m_chainedId = rest->GetID();
             return true;
         }
-        if (beam->IsFirstIn(note)) {
+		// If the beam has more than 2 and this is first
+        else if (beam->IsFirstIn(note)) {
             Rest *rest = new Rest();
             rest->DurationInterface::operator=(*note);
             Object *parent = beam->GetParent();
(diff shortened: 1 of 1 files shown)


rettinghaus:
Please clang-format the code.

Matthew-w56:
I added so little and don't see a place where my code isn't in conformity with the standards, so you'll have to let me know what you don't like about it

src/editortoolkit_cmn.cpp
```diff
@@ -213,7 +213,7 @@ bool EditorToolkitCMN::Delete(std::string &elementId)
 {
     Object *element = this->GetElement(elementId);
     if (!element) return false;
-
+
```
ahankinson, line 216: Empty whitespace.

src/editortoolkit_cmn.cpp
```diff
@@ -247,15 +247,15 @@ bool EditorToolkitCMN::KeyDown(std::string &elementId, int key, bool shiftKey, b
 
     // For elements whose y-position corresponds to a certain pitch
     if (element->HasInterface(INTERFACE_PITCH)) {
-        PitchInterface *interface = element->GetPitchInterface();
-        assert(interface);
+        PitchInterface* pitch_interface = element->GetPitchInterface();
```
ahankinson, line 250: I think you've just renamed this, without changing what it does?

src/editortoolkit_cmn.cpp
```diff
@@ -512,7 +512,7 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
 
     Chord *chord = note->IsChordTone();
     Beam *beam = note->GetAncestorBeam();
-
+
```
ahankinson, line 515: Additional whitespace.

src/editortoolkit_cmn.cpp
```diff
@@ -560,6 +560,7 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
         }
     }
     else if (beam) {
+		// If the beam has exactly 2 notes (take apart and leave a single note and a rest)
```
ahankinson, line 563: Indentation doesn't match.

src/editortoolkit_cmn.cpp
```diff
@@ -584,16 +585,17 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
             m_chainedId = rest->GetID();
             return true;
         }
-        if (beam->IsFirstIn(note)) {
+		// If the beam has more than 2 and this is first
```
ahankinson, line 588: Indentation doesn't match.

src/editortoolkit_cmn.cpp
```diff
@@ -602,18 +604,26 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
             parent->InsertAfter(beam, rest);
             beam->DeleteChild(note);
             m_chainedId = rest->GetID();
-            return true;
         }
+		// If the beam has more than 2 and this in the middle
         else {
-            Rest *rest = new Rest();
+			Rest *rest = new Rest();
```
ahankinson, line 610: Perhaps you have your editor set to add tabs? This indentation is way off.

src/editortoolkit_cmn.cpp
```diff
@@ -602,18 +604,26 @@ bool EditorToolkitCMN::DeleteNote(Note *note)
...
             rest->DurationInterface::operator=(*note);
             beam->ReplaceChild(note, rest);
             delete note;
             m_chainedId = rest->GetID();
-            return true;
         }
+		
+		// All but the first IF statement branches lead here
+
```
ahankinson, line 618: All this indentation is way off too.

ahankinson:
I've approved the clang format action so you can see what doesn't pass. See: https://github.com/rism-digital/verovio/actions/runs/8178137301/job/22363152445?pr=3617

I've also had  a go at manually looking through the style, and added some comments.

src/editortoolkit_cmn.cpp
Matthew-w56: Yup.  The keyword 'interface' is the subject of a macro that expands it to "__STRUCT__", or at least my IDE was warning me of one.  So I renamed it to both be less vague, and avoid that overlap.

ahankinson: You may need to configure your IDE to not do that, as the `*interface` name is used in a lot of places in Verovio.

src/editortoolkit_cmn.cpp
```diff
@@ -213,7 +213,6 @@ bool EditorToolkitCMN::Delete(std::string &elementId)
 {
     Object *element = this->GetElement(elementId);
     if (!element) return false;
-
```
ahankinson, line 216: It's generally better if you don't change the existing formatting unless there's a reason. Removing the blank line here is innocent enough, but it adds additional unnecssary changes to the diff.

Matthew-w56: Personally I liked that space there.  I just received guff about having the gap there

ahankinson:
It may also be helpful to review the Coding Guidelines, which tells you how to install and run clang-format locally so you can fix the errors before committing.

https://book.verovio.org/contributing/guidelines.html

src/editortoolkit_cmn.cpp
ahankinson: "guff" from  me? 

I was commenting on the additional spaces (trailing / hanging spaces) that had been added to the line, not to the blank line itself. Sorry if I wasn't clear about that.

Matthew-w56:
Yeah, I'm not sure how to set up clang on Windows, and the instructions I can find involve a long, manual build process with other libraries.  I don't have the time to be doing that

ahankinson:
You could try an online version. I haven't tried it, but it seems this one allows you to upload a format file, and then paste your code in. 

Just pay attention to the version of clang-format used by Verovio (15) because sometimes things change between releases.

Matthew-w56:
Thanks for the point to that tool.  This next push should be the last

src/editortoolkit_cmn.cpp
```diff
@@ -247,15 +246,15 @@ bool EditorToolkitCMN::KeyDown(std::string &elementId, int key, bool shiftKey, b
 
     // For elements whose y-position corresponds to a certain pitch
     if (element->HasInterface(INTERFACE_PITCH)) {
-        PitchInterface *interface = element->GetPitchInterface();
-        assert(interface);
```
lpugin, line 251: You should not have to change this. Please revert

Matthew-w56: I will revert that change.  Though I am quite frustrated that the only time I get a directed response from you is telling me I did something wrong, when I've asked you how to do things in this library in 3 different ways over the last week.

Matthew-w56:
Okay I reverted that name change, it should be good to go

src/editortoolkit_cmn.cpp
lpugin: Sorry about that. This is a part of the code that hasn't been touched for years. One reason is that improving it needs proper thinking about how to refactor it. Guiding you would require quite some time that I do not have right at the moment. I understand you would expect clear and elaborated answers straight away but I am sorry this is not always possible. Nonetheless, the PR looks fine and I am happy to merge it.

Issue #3622, opened 2024-03-11
Lyrics attached to short duration notes are inserted at wrong tick in midi files


fernandoherreradelasheras:
When rendering this file to MIDI:

```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" meter.count="4" meter.unit="4">
                        <clef shape="G" line="2" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="1">
                     <staff n="1">
                        <layer n="1">
                           <note dur="8" oct="4" pname="b">
                              <verse n="1">
                                 <syl con="d" wordpos="i">Pri</syl>
                              </verse>
                           </note>
                           <note dur="8" oct="5" pname="c">
                              <verse n="1">
                                 <syl wordpos="t">mum</syl>
                              </verse>
                           </note>
                           <note dur="8" oct="4" pname="b">
                              <verse n="1">
                                 <syl wordpos="s">et</syl>
                              </verse>
                           </note>
                           <note dur="8" oct="4" pname="a">
                              <verse n="1">
                                 <syl wordpos="s">per</syl>
                              </verse>
                           </note>
                           <note dur="8" oct="4" pname="g">
                              <verse n="1">
                                 <syl con="d" wordpos="i">Sem</syl>
                              </verse>
                              <accid accid="s" />
                           </note>
                           <note dur="8" oct="4" pname="a">
                              <verse n="1">
                                 <syl wordpos="t">per</syl>
                              </verse>
                           </note>
                           <rest dur="4"/>
                        </layer>
                     </staff>
                  </measure>
                  <measure n="2">
                     <staff n="1">
                        <layer n="1">
                           <meterSig count="4" unit="2"/>
                           <note dur="4" oct="4" pname="b">
                              <verse n="1">
                                 <syl con="d" wordpos="i">Pri</syl>
                              </verse>
[...]

PR #3623, opened 2024-03-11
Fix issue #3622: wrong lyrics position in midi output for <syl> in notes <= quarter

fernandoherreradelasheras:
This is happening due to use an integer for measure startTime, that depending on the file tpq will lead to decimal values for short duration notes

### DIFF
--- a/src/midifunctor.cpp
+++ b/src/midifunctor.cpp
@@ -798,7 +798,7 @@ FunctorCode GenerateMIDIFunctor::VisitStaffDef(const StaffDef *staffDef)
 
 FunctorCode GenerateMIDIFunctor::VisitSyl(const Syl *syl)
 {
-    const int startTime = m_totalTime + m_lastNote->GetScoreTimeOnset();
+    const double startTime = m_totalTime + m_lastNote->GetScoreTimeOnset();
     const std::string sylText = UTF32to8(syl->GetText());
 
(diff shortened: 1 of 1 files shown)


craigsapp:
This fix will improve the problem, but will not be an exact solution (although rounding to the nearest millisecond might fully hide the problem in the round-off bits of the `double`).   Ideally MIDI ticks should be used only rather than time in seconds for these calculations.   Also, lyrics (`<syl>`) can only be children of `<note>` in MEI, so using the same tick/time value as the parent `<note>` might be good.

fernandoherreradelasheras:
I agree that only relying on ticks would be a cleaner solution but I haven't looked at the whole code to know if that is feasible at the moment (there are a lot of specific cases and conditions relating to this).

Anyway using a double, as the code adding note on / off events is doing, ensures that the time used for the syl, if not right, at least matches the one the note was inserted at:
 
Screenshot from 2024-03-11 23-44-48

So maybe we can get this in as a fix for the "lyrics not matching notes" issue and then later create a feature request to refactor the code to work on ticks only

craigsapp:
Yes, your fix make the timings much more accurate, the only very rare case will be when the timing value is very close to 0.5 milliseconds and it gets rounded up or down instead of the more desired direction.

PR #3628, opened 2024-03-19
fix ties from Dorico

rettinghaus:
Small change to the MusicXML import to catch all ties exported from Dorico, which puts them into a separate `notations` element.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3148,7 +3148,7 @@ void MusicXmlInput::ReadMusicXmlNote(
         }
 
         // ties
-        ReadMusicXmlTies(notations.node(), layer, note, measureNum);
+        ReadMusicXmlTies(node, layer, note, measureNum);
 
         // articulation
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -3834,7 +3834,8 @@ void MusicXmlInput::ReadMusicXmlBeamStart(const pugi::xml_node &node, const pugi
 void MusicXmlInput::ReadMusicXmlTies(
     const pugi::xml_node &node, Layer *layer, Note *note, const std::string &measureNum)
 {
-    for (pugi::xml_node xmlTie : node.children("tied")) {
+    pugi::xpath_node ties = node.select_node("notations[tied]");
+    for (pugi::xml_node xmlTie : ties.node().children("tied")) {
```
lpugin, line 3838: You need to check that `ties` and `ties.node()` are not null

rettinghaus: changed approach

Issue #996, opened 2019-02-20
SVG (float formatting) issue with Java on non US locale Ubuntu


rettinghaus:
Building the Java toolkit fails on Ubuntu 18.10 with swig 3.0.12.
I tried master branch and develop branch.

lpugin:
What it the error?

rettinghaus:
Had to specify the paths to `jni.h` and `jni_md.h`  for g++. I added it to the line in the `build.sh` and now it works. 
But the `output.svg` can't be displayed (although the file looks quite ok).

lpugin:
Can you make a PR with the changes?

rettinghaus:
Will do. 
But the development branch is broken. The svg output with native build only shows a slur, the java output won't work at all (see output files in zip). Any idea?
[importer.zip](https://github.com/rism-ch/verovio/files/2888891/importer.zip)

NB. The java build soutput of the main branch isn't working, too.

lpugin:
> But the development branch is broken.

Can you check with the latest commit? And what command parameter did you use with the native version?

rettinghaus:
Latest commit merged. Just used `./verovio -r ../data ../doc/importer.mei` in tools. Same result.

lpugin:
Should be fixed

rettinghaus:
The SVG output still can't be opened with any viewer I tested.

lpugin:
This is weird. The file `importer-dev-native.svg` in your zip has corrupted dimensions `width="1050,00px" height="322,00px"` and is missing the xml directive. It has a duplicated `@id`, but otherwise it is valid and displays correctly
![image](https://user-images.githubusercontent.com/689412/53249430-56d20e80-36b8-11e9-9e0f-daf109865f46.png)

lpugin:
This is where the dimensions are set
https://github.com/rism-ch/verovio/blob/develop/src/svgdevicecontext.cpp#L83-L93
It is weird that only these values are wrongly formatted.

lpugin:
It all works as expected on Mac OS X (10.14.3), Java 1.8.0_162 and Swig 3.0.12

lpugin:
@rettinghaus any progress on this?

rettinghaus:
No. I tried to play around with the options in in `main.java` without any luck.

lpugin:
Can you confirm that the issue comes from the wrongly formatted dimensions?

rettinghaus:
[...]

PR #3630, opened 2024-03-21
Set locale once in toolkit

DavidBauer1984:
This PR addresses the changes introduced in #3611 . The problem is that setting and resetting the locale a few thousand times is not a good idea, performance-wise. Verovio now spends over 10 % of runtime in `StringFormat` due to this.

heaviest-stack-trace

StringFormat

Instead we add an option `--set-locale`. If included, the locale is changed once in `Toolkit::SetOptions` and reset in the toolkit destructor.

@ammatwain Can you please check whether this still fixes your locale issues? I could not really test it from Qt. Note that the process locale is kept changed during toolkit lifetime. I hope that this does not interfere with your application.

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -652,6 +652,7 @@ public:
     OptionBool m_preserveAnalyticalMarkup;
     OptionBool m_removeIds;
     OptionBool m_scaleToPageSize;
+    OptionBool m_setLocale;
     OptionBool m_showRuntime;
     OptionBool m_shrinkToFit;
     OptionIntMap m_smuflTextFont;
--- a/include/vrv/toolkit.h
+++ b/include/vrv/toolkit.h
@@ -790,7 +790,7 @@ private:
 
     Options *m_options;
 
-    std::locale m_previousLocale;
+    std::optional<std::locale> m_previousLocale;
 
     /**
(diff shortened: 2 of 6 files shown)


DavidBauer1984:
I am still wondering whether we should change the locale at all in Verovio. The problem is that `std::locale::global` is never thread-safe (see [this discussion](https://stackoverflow.com/questions/4057319/is-setlocale-thread-safe-function)) and therefore can cause trouble for people that do batch processing with Verovio (i.e. launch several toolkit instances in parallel threads). 

We could also just make it a responsibility of the caller to make sure that the locale is properly set.

lpugin:
Alternatively we could add an option to do it, which would be `false` by default. Would that make sense?

rettinghaus:
relates to #996

DavidBauer1984:
The option `--set-locale` was added. If enabled, it changes the locale in the Verovio toolkit. Currently, there will be no effect on the command-line version.

ammatwain:
> @ammatwain Can you please check whether this still fixes your locale issues? I could not really test it from Qt. Note that the process locale is kept changed during toolkit lifetime. I hope that this does not interfere with your application.

Perfect @DavidBauer1984,  yours is the most efficient solution!

PR #3634, opened 2024-03-22
fix #3633: output a single midi lyrics event per note with all syls in the note

### LINKED ISSUE #3633: wrong midi output for notes with multiple syllables
Notes with multiple syllables (usually no more than two) connected by elision slurs like this: 
```
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead />
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef>
                  <staffGrp>
                     <staffDef n="1" lines="5" meter.count="4" meter.unit="4">
                        <clef shape="G" line="2" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <pb />
                  <measure right="end" n="1">
                     <staff n="1">
                        <layer n="1">
                           <note dur="4" oct="4" pname="g">
                              <verse n="1">
                                 <syl>el</syl>
                              </verse>
                           </note>
                           <note dur="4" oct="5" pname="c">
                              <verse n="1">
                                 <syl con="d" wordpos="i">a</syl>
                              </verse>
                            </note>
                           <note dur="4" oct="5" pname="c">
                              <verse n="1">
                                 <syl con="b" wordpos="t">gua</syl>
                                 <syl con="d" wordpos="i">a</syl>
                              </verse>
                           </note>
                           <note dur="4" oct="5" pname="c">
                              <verse n="1">
                                 <syl wordpos="t">zul</syl>
                              </verse>
                            </note>
                        </layer>
                     </staff>
                  </measure>
               </section>
            </score>
         </mdiv>
      </body>
   </music>
</mei>
```

are properly rendered in SVG:

lyrics-with-slision-slurs

But the midi output adds an event for each syl element in the note: 

mididump-with-syl-slur

Resulting on the last one replacing the previous one.

As MIDI does not support anything else than a single text event per note we could try to split notes on those situations or just join all text syllables on the same event. This last approach looks simpler and works fine with software like synthethizer V. As separator I think space is the choice as messages are limited to ASCII and dash would be confusing on karaoke-like apps displaying lyrics.

fernandoherreradelasheras:


rettinghaus:
I would expect such addition in `VisitVerse` (or `VisitSyl`).

fernandoherreradelasheras:
> I would expect such addition in `VisitVerse` (or `VisitSyl`).

Ok, thanks! I'll take a look into it

rettinghaus:
@fernandoherreradelasheras any luck?

fernandoherreradelasheras:
Sorry, got this on hold while experimenting with different approaches to multiple verse lyrics. I will get back to this soon

src/midifunctor.cpp
```diff
@@ -799,9 +801,15 @@ FunctorCode GenerateMIDIFunctor::VisitStaffDef(const StaffDef *staffDef)
...
-    const std::string sylText = UTF32to8(syl->GetText());
 
-    m_midiFile->addLyric(m_midiTrack, startTime * m_midiFile->getTPQ(), sylText);
+    m_currentNoteLyrics.append(!m_currentNoteLyrics.empty(), ' ');
+    m_currentNoteLyrics.append(UTF32to8(syl->GetText()));
+
+    const Object *parent = syl->GetParent();
+    assert(parent);
+    if (parent->GetNext(syl, SYL) == NULL) {
+        m_midiFile->addLyric(m_midiTrack, startTime * m_midiFile->getTPQ(), m_currentNoteLyrics);
```
lpugin, line 811: Just move this to a new `GenerateMIDIFunctor::VisitVerseEnd` and remove the lookup here.

lpugin (changes requested):

lpugin:
Please re-open with the requested changes.

PR #3636, opened 2024-03-22
Update clang-format-action

rettinghaus:
This PR updates the `clang-format-action` to the latest version and moves checks to `clang-format` version 18. Previous versions of `clang-format` missed to check for a space after `override`, where we had some inconsistencies.


craigsapp:
Does this mean pushed code is automatically formatted by clang-format, and/or should I make sure that my version of clang-format is up to date? (I have clang-format version 16.0.3 installed). In any case I don't use `override` in `iohumdrum.cpp`.

lpugin:
You need to update to version 18.

craigsapp:
Now updated to `clang-format version 18.1.2`.

PR #3668, opened 2024-05-07
ci(gh-actions): include macos-14 in python wheel workflow (rebased)

musicEnfanthen:


ahankinson (approved):

rettinghaus:
Instead of piling up all the excludes should we rethink the matrix? And why not also build `universal2` on Mac?

PR #3673, opened 2024-05-08
WIP: refactor svg

rettinghaus:
This aims to get rid of the unwanted css attribute `overflow` from the SVG output. 

Instead of extracting the symbols from each font the paths are put in a g element for referencing in the SVG. 
That removes the previous need for a `@viewbox` on every glyph and the later resizing with `@width` and `@hight` (making the resulting SVG way smaller).

The scaling for the glyphs is added during adding the snippets in Verovio. 

Currently it is hard-coded, so scaling with this doesn't work.


fonts/generate.py
```diff
@@ -211,7 +211,7 @@ def generate_css(opts: Namespace) -> bool:
     Et.register_namespace("", SVG_NS["svg"])
     svg_font: Et.ElementTree = Et.parse(str(font_pth))
     font_el: Optional[Et.Element] = svg_font.find(".//svg:defs/svg:font", SVG_NS)
-    if not font_el:
+    if font_el is None:
```
ahankinson, line 214: I would caution against making such changes. The XML libraries are weird about what constitutes “falsey” and what is actually `None`. If you don’t need to make these sorts of changes and are just doing it for your own sense of aesthetics, I would leave them.

rettinghaus: Without this change Python throws: 

> DeprecationWarning: Testing an element's truth value will raise an exception in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.

And in doubt it may help to consult the documentation: we use `find` from `xml.etree.ElementTree` which returns `None` if nothing is found. 
See <https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.find>.

ahankinson: Ok, great! Thanks

lpugin:
Coming back to this, I prefer not to touch it. The way the SVG is supported by various SVG renderers has been carefully adjusted and I don't see the need to change it for SharpVectors. My experience has shown that changing things like this will quite likely break rendering with another SVG renderer.

PR #3674, opened 2024-05-08
Update actions

rettinghaus:
This PR updates used actions in reaction to the numerous deprecation warnings.


lpugin:
Reverting since makes the action fail.

Issue #3574, opened 2024-01-22
Combined display of mensural and cmn notation


craigsapp:
It would be useful to display both mensural an CMN notation in the same score.

The current idea is to use `@tstamp` (and/or `@startid`) to link the mensural notes (& etc.) to link to CMN content.

Here is some basic example data:

<img width="779" alt="Screenshot 2024-01-21 at 23 49 18" src="https://github.com/rism-digital/verovio/assets/3487289/7cdb0a13-b4d3-4e3e-b004-86c274eee1e3">

<details>
<summary>Click to view MEI data for above example </summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt>
    <unpub>This MEI file was created by Verovio's Humdrum converter. When published, this unpub element should be removed, and the enclosing pubStmt element should be properly filled out.</unpub>
   </pubStmt>
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2024-01-21T23:47:45" version="4.2.0-dev-80ec72d">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
  <extMeta>
   <frames xmlns="http://www.humdrum.org/ns/humxml" />
  </extMeta>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m1nwdsx9">
    <score xml:id="s6q5cyl">
     <scoreDef xml:id="ssds15j" midi.bpm="400.000000">
      <staffGrp xml:id="sj8jzz">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="C" line="3" />
        <meterSig xml:id="metersig-L3F1" count="3" unit="1" visible="false" />
        <mensur xml:id="mensur-L4F1" sign="O" />
       </staffDef>
      </staffGrp>
     </scoreDef>
[...]

Issue #3573, opened 2024-01-21
Mensural <dot> placement in score layout


craigsapp:
Mensural dots are currently centered between adjacent notes:

<img width="401" alt="Screenshot 2024-01-21 at 11 51 38" src="https://github.com/humdrum-tools/verovio-humdrum-viewer/assets/3487289/6d547381-aaed-403b-8867-225ad44a568d">

<details>

<summary> Click to view MEI data for above example </summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt>
    <unpub>This MEI file was created by Verovio's Humdrum converter. When published, this unpub element should be removed, and the enclosing pubStmt element should be properly filled out.</unpub>
   </pubStmt>
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2024-01-21T11:51:23" version="4.2.0-dev-80ec72d">
     <name>Verovio</name>
     <p>Transcoded from Humdrum</p>
    </application>
   </appInfo>
  </encodingDesc>
 </meiHead>
 <music decls="#work0_encoded">
  <body>
   <mdiv xml:id="m1756h3a">
    <score xml:id="s14g40td">
     <scoreDef xml:id="slw6g3r" midi.bpm="280.000000">
      <staffGrp xml:id="s1dfdahk">
       <staffDef xml:id="staffdef-L1F1" n="1" notationtype="mensural.white" lines="5">
        <clef xml:id="clef-L2F1" shape="C" line="3" />
       </staffDef>
      </staffGrp>
     </scoreDef>
     <section xml:id="section-L1F1">
      <staff xml:id="staff-L1F1" n="1">
       <layer xml:id="layer-L1F1N1" n="1">
        <mensur xml:id="mensur-L3F1" modusmaior="2" modusminor="2" prolatio="2" tempus="3" sign="O" />
        <note xml:id="note-L4F1" dur="brevis" oct="4" pname="c" />
[...]

PR #3679, opened 2024-05-15
Develop mensural fixes

lpugin:
* [Always initialise ligatures (avoid crash in cmn)](https://github.com/rism-digital/verovio/commit/498e1f16e608ceef2aa6e7794fd25ab9de9ce7a8)
* [Support @dots with mensural duration (mei-all)](https://github.com/rism-digital/verovio/commit/5486e48dadb89ff6082a5fe76616199deb8e5634)

### DIFF
--- a/src/calcdotsfunctor.cpp
+++ b/src/calcdotsfunctor.cpp
@@ -60,10 +60,6 @@ FunctorCode CalcDotsFunctor::VisitChord(Chord *chord)
 
 FunctorCode CalcDotsFunctor::VisitNote(Note *note)
 {
-    // We currently have no dots object with mensural notes
-    if (note->IsMensuralDur()) {
-        return FUNCTOR_SIBLINGS;
-    }
     if (!note->IsVisible()) {
         return FUNCTOR_SIBLINGS;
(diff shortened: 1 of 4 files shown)


craigsapp:
Related to commit https://github.com/rism-digital/verovio/pull/3679/commits/5486e48dadb89ff6082a5fe76616199deb8e5634 (`@dots` in mei-all), note this issue: https://github.com/rism-digital/verovio/issues/3573 where it would be useful for dots not to be centered in score layout.

Can ligatures be displayed in cmn (or on a mensural staff while also displaying cmn on another staff) now?  I will test it out. (Related to issue https://github.com/rism-digital/verovio/issues/3574)

lpugin:
`@dots` support has been added because we have a lot of PAE incipits encoded in mensural notation where dots are actually CWMN dots (including double dots). These will be not centered but instead placed close to the note. I think we still need to have this for `<dot>` since using `@dots` is not to be recommended in my opinion.

Ligature: yes, now they can be displayed in CMN. You can still use `--ligature-as-bracket`

Issue #3627, opened 2024-03-19
Missing SMuFL codes for ornamental neume-components in square notation


martha-thomae:
We are missing some SMuFL symbols for some "ornamental" neumes in square notation: quilisma, oriscus, and liquescent. These are just rendered as regular square neume components (like punctums), when they should be rendered differently. However, these "ornamental neumes" are already supported with SMuFL. In this issue, I am including the MEI code for each of these ornamental neumes and their suggested SMuFL code:

### (i) Quilisma

- MEI Encoding:
  ```xml
  <nc>
    <quilisma/>
  </nc>
  ```

- Proposed SMuFL code: 
**U+E99B** (_chantQuilisma_) in [Medieval and Renaissance plainchant single-note forms](https://w3c.github.io/smufl/latest/tables/medieval-and-renaissance-plainchant-single-note-forms.html)

  <img src="https://github.com/rism-digital/verovio/assets/13948831/210e2f22-c750-45c6-b0ae-afeaa741dbe3" alt="drawing" width="350"/>

### (ii) Oriscus

  - MEI Encoding:
    ```xml
    <nc>
      <oriscus/>
    </nc>
    ```

  - Proposed SMuFL code: 
  **U+EA2A** (_medRenOriscusCMN_) in [Medieval and Renaissance plainchant in CMN](https://w3c.github.io/smufl/latest/tables/medieval-and-renaissance-symbols-in-cmn.html)

    <img src="https://github.com/rism-digital/verovio/assets/13948831/31fce5c3-e885-42a2-87d8-2cf5f7a89c89" alt="drawing" width="350"/>

###  (iii) Liquescent as small punctum
  This liquescent looks like a small square note. It can happen in two instances: **(Case A)** in a neume with a single component but no `@curve`; or **(Case B)** in a neume with at least two neume components (one of them with a `liquescent`). 

  - **Case A:** MEI Encoding (neume with only one `<nc>`, no `@curve` in the `<nc>`)
    ```xml
    <neume>
      <nc>
        <liquescent/>
      </nc>
    </neume>
    ```

  - **Case B:** MEI Encoding (neume with at least two `<nc>`)
    ```xml
    <neume>
      <nc/>
      ...
      <nc>
        <liquescent/>
      </nc>
      ...
      <nc/>
    </neume>
    ```

  - Proposed SMuFL code: 
[...]

PR #3693, opened 2024-05-27
add basic support for oriscus and quilisma

rettinghaus:
This PR adds basic reading/writing support for `oriscus` and `quilisma` elements. 

Also it fixes missing/non functional support for liquescent color.

Relates to #3627.


lpugin (approved):

lpugin:
Are you going to implement the rest? Or just the basic?

rettinghaus:
I had a look, but the existing implementation there is quite different than everything we have elsewhere. So I'm wondering what would be the best approach. We should discuss this first.

lpugin:
OK, I will look into it.

PR #3709, opened 2024-06-18
Develop humdrum

craigsapp:
Recent updates for the `develop-humdrum` branch.


craigsapp:
Note that the new checked only compile for python toolkit without Humdrum code included (so it does not check iohumdrum.cpp main content and humlib code that I can notice).

lpugin:
@craigsapp I can merge once the Appveyor issue is fixed

craigsapp:
Should be fixed now.  I do not understand why `M_E` was undefined (it should be defined in `cmath`).  

>  https://ci.appveyor.com/project/LaurentPugin/verovio-r1t6l/builds/50040712#L330

Is there a link to the appveyor checks that is easy to find for the PR compiling checks?  It seems like it used to be on pages like this.

PR #3714, opened 2024-06-24
Humdrum reader: ottava fixes

gregchapman-dev:
1. Don't start an ottava twice (it will nuke the first one).
2. Add getStartIdForOttava, that works somewhat like getEndIdForOttava.  This picks up ottava start notes that are hidden because of a voice split between the ottava start mark and the first notes in the ottava, much like getEndIdForOttava does for ottava end notes that are hidden because of a voice merge between the last notes in the ottava and the ottava end mark.

### DIFF
--- a/include/vrv/iohumdrum.h
+++ b/include/vrv/iohumdrum.h
@@ -278,6 +278,7 @@ namespace humaux {
         // an ottava line which will be turned off later.  ottavameasure == the
         // starting measure of the ottava mark.
         Note *ottavanotestart;
+        std::string ottavanotestartid;
         Note *ottavanoteend;
         hum::HumNum ottavaendtimestamp;
         Measure *ottavameasure;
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
Here is a test case for getStartIdForOttava (the ottava in the right hand is the one that starts incorrectly without this fix):

[DifficultOttavas.krn.txt](https://github.com/user-attachments/files/15959531/DifficultOttavas.krn.txt)

gregchapman-dev:
@craigsapp what do you think?

craigsapp:
Looks great to me.   I will commit it and check in more detail.

Before the PR, the test data renders as:

Screenshot 2024-06-30 at 21 34 50

where the top line is improperly delayed by an eighth note.

craigsapp:
PR fixes test case:

Screenshot 2024-06-30 at 21 58 18

Ideally the octave line end in the bottom staff would not touch the final barline...

craigsapp (commented):
Looks great

PR #3735, opened 2024-07-30
Fix alignment after staff dragging

yinanazhou:
- Adjust nc spacing
- Adjust `X/YRel` in reset aligners

Refs: https://github.com/DDMAL/Neon/issues/1231

### DIFF
--- a/src/calcligatureorneumeposfunctor.cpp
+++ b/src/calcligatureorneumeposfunctor.cpp
@@ -347,16 +347,19 @@ FunctorCode CalcLigatureOrNeumePosFunctor::VisitNeume(Neume *neume)
             }
         }
 
-        // If the nc overlaps with the previous, move it back from a line width
-        if (overlapWithPrevious) {
-            xRel -= lineWidth;
-        }
+        // xRel remains unset with facsimile
+        if (!m_doc->HasFacsimile()) {
+            // If the nc overlaps with the previous, move it back from a line width
+            if (overlapWithPrevious) {
+                xRel -= lineWidth;
+            }
 
-        nc->SetDrawingXRel(xRel);
-        // The first glyph set the spacing - unless we are starting a ligature, in which case no spacing should be added
-        // between the two nc
-        if (!previousLig) {
-            xRel += m_doc->GetGlyphWidth(nc->m_drawingGlyphs.at(0).m_fontNo, staffSize, false);
+            nc->SetDrawingXRel(xRel);
+            // The first glyph set the spacing - unless we are starting a ligature, in which case no spacing should be
+            // added between the two nc
+            if (!previousLig) {
+                xRel += m_doc->GetGlyphWidth(nc->m_drawingGlyphs.at(0).m_fontNo, staffSize, false);
+            }
         }
 
         previousNc = nc;
--- a/src/page.cpp
+++ b/src/page.cpp
@@ -253,6 +253,9 @@ void Page::LayOutTranscription(bool force)
     CalcAlignmentPitchPosFunctor calcAlignmentPitchPos(doc);
     this->Process(calcAlignmentPitchPos);
 
+    CalcLigatureOrNeumePosFunctor calcLigatureOrNeumePos(doc);
+    this->Process(calcLigatureOrNeumePos);
+
     CalcStemFunctor calcStem(doc);
     this->Process(calcStem);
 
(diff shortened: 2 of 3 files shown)


include/vrv/staff.h
```diff
@@ -144,11 +144,6 @@ class Staff : public Object,
     virtual double GetDrawingRotate() const;
     ///@}
 
-    /**
-     * Adjust drawingStaffSize based on rotate angle
-     */
-    void AdjustDrawingStaffSize();
-
```
lpugin, line 151: Do you really need to remove this? I think this is use outside neon

include/vrv/staff.h
```diff
@@ -228,13 +223,6 @@ class Staff : public Object,
      */
     int GetNearestInterStaffPosition(int y, const Doc *doc, data_STAFFREL place) const;
 
-    /**
-     * Set staff parameters based on
-     * facsimile information (if it
-     * exists).
-     */
-    void SetFromFacsimile(Doc *doc);
-
```
lpugin, line 237: Same here: please explain why you need to remove this

src/calcligatureorneumeposfunctor.cpp
```diff
@@ -347,17 +347,7 @@ FunctorCode CalcLigatureOrNeumePosFunctor::VisitNeume(Neume *neume)
...
-        if (overlapWithPrevious) {
-            xRel -= lineWidth;
-        }
-
         nc->SetDrawingXRel(xRel);
-        // The first glyph set the spacing - unless we are starting a ligature, in which case no spacing should be added
-        // between the two nc
-        if (!previousLig) {
-            xRel += m_doc->GetGlyphWidth(nc->m_drawingGlyphs.at(0).m_fontNo, staffSize, false);
-        }
```
lpugin, line 360: I think what you want is not do this when there is a facsimile. It should not be removed because it is use to calculate the position of the neumes when there is no facsimile

src/page.cpp
```diff
@@ -310,6 +310,13 @@ void Page::ResetAligners()
     AlignVerticallyFunctor alignVertically(doc);
     this->Process(alignVertically);
 
+    if (doc->IsNeumeLines()) {
+        AdjustXRelForTranscriptionFunctor adjustXRelForTranscription;
+        this->Process(adjustXRelForTranscription);
+        AdjustYRelForTranscriptionFunctor adjustYRelForTranscription;
+        this->Process(adjustYRelForTranscription);
+    }
```
lpugin, line 318: This this is transcription (facsimile) related, it is called from `Page::LayOutTranscription`. Could you explain why this needs to be added here?

include/vrv/staff.h
yinanazhou: Oops, I'll revert it then

src/page.cpp
yinanazhou: I need to call `Page::ResetAligners()` after dragging a staff to realign the staff children. The horizontal/vertical alignment is reset for accid so I need this to adjust `X/YRel` for accids so that it wont jump.

lpugin: Could you explain why accids behave any differently than neumes or ncs ? They should not

yinanazhou: Accids don't have a pitch, so they are skipped in `CalcAlignmentPitchPosFunctor`. After reset alignment functors `Page::ResetAligners()`, `X/YRel` of accids get reset to 0, I need to adjust them back

yinanazhou: hmmm I guess I should call `Page::LayOutTranscription` instead of `Page::ResetAligners`?

src/calcligatureorneumeposfunctor.cpp
```diff
@@ -348,14 +348,14 @@ FunctorCode CalcLigatureOrNeumePosFunctor::VisitNeume(Neume *neume)
         }
 
         // If the nc overlaps with the previous, move it back from a line width
-        if (overlapWithPrevious) {
+        if (overlapWithPrevious && !m_doc->HasFacsimile()) {
             xRel -= lineWidth;
         }
 
         nc->SetDrawingXRel(xRel);
```
lpugin, line 355: Since `xRel` is not used with facsimile you should put everything in one block
```
        // xRel remains unset with facsimile
        if (!m_doc->HasFacsimile()) {
            // If the nc overlaps with the previous, move it back from a line width
            if (overlapWithPrevious) {
                xRel -= lineWidth;
            }
            
            nc->SetDrawingXRel(xRel);
            // The first glyph set the spacing - unless we are starting a ligature, in which case no spacing should be added
            // between the two nc
            if (!previousLig) {
                xRel += m_doc->GetGlyphWidth(nc->m_drawingGlyphs.at(0).m_fontNo, staffSize, false);
            }
        }
```

src/page.cpp
lpugin: Why not calling `Page::LayOutTranscription` instead of `Page::ResetAligners`? That would apply `AdjustYRelForTranscriptionFunctor`, wouldn't it?

lpugin:
I resolved a conflict. Please double check.

PR #3743, opened 2024-08-08
Enable takt barlines and import from Volpiano

rettinghaus:
This PR adds reading/writing of visual attributes for `barLine`, enabling to draw single barlines as takt. 
Also it adds basic import for barlines to Volpiano support (including breaks, which are shown as takt barlines).

### Example
`1---dcd---3---fef---7---dcd---4---fef---77---`

image

### DIFF
--- a/src/iovolpiano.cpp
+++ b/src/iovolpiano.cpp
@@ -126,7 +126,7 @@ bool VolpianoInput::Import(const std::string &volpiano)
             layer->AddChild(end);
         }
         else if (ch == '6') {
-            // not supported yet
+            LogWarning("Volpiano '6' barline is not supported");
         }
         else if (ch == '7') {
             BarLine *takt = new BarLine();
--- a/src/iovolpiano.cpp
+++ b/src/iovolpiano.cpp
@@ -120,6 +120,14 @@ bool VolpianoInput::Import(const std::string &volpiano)
             dbl->SetForm(BARRENDITION_dbl);
             layer->AddChild(dbl);
         }
+        else if (ch == '5') {
+            BarLine *end = new BarLine();
+            end->SetForm(BARRENDITION_end);
+            layer->AddChild(end);
+        }
+        else if (ch == '6') {
+            // not supported yet
+        }
         else if (ch == '7') {
             BarLine *takt = new BarLine();
             takt->SetMethod(BARMETHOD_takt);
--- a/src/iovolpiano.cpp
+++ b/src/iovolpiano.cpp
@@ -16,6 +16,7 @@
 
 //----------------------------------------------------------------------------
 
+#include "barline.h"
 #include "clef.h"
 #include "doc.h"
 #include "layer.h"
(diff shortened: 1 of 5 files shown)


src/view_element.cpp
```diff
@@ -420,11 +420,23 @@ void View::DrawBarLine(DeviceContext *dc, LayerElement *element, Layer *layer, S
         barLine->SetEmptyBB();
         return;
     }
+    StaffDef *drawingStaffDef = staff->m_drawingStaffDef;
+    // Determine the method
+    assert(drawingStaffDef);
+    auto [hasMethod, method] = barLine->GetMethod(drawingStaffDef);
+    if (barLine->HasMethod()) {
```
lpugin, line 427: It is not very clear why you need to reassign `method`. Could you add a comment explaining why?

rettinghaus: The `GetMethod(staffDef)` is overruling libmei's `GetMethod()`, so here it's first looking for the bar method on staffDef and then for the local method.

lpugin (approved):

ahankinson (commented):
I think this is missing functions for `5` and `6` too, right?

image

craigsapp:
Can number 7 (takt) be used in CMN or is this specific to neume notation?  Or in any case verovio allows using it with MEI-all?  I came across this last week in a modernized version (from 1877 publication) that is transcribed from earlier notation:

Screenshot 2024-09-02 at 4 43 45 AM

Either it comes from the original notation, or it was added to mark phrase boundaries/verse line.  It does not occur at the colon (`:`) which is a regular barline

ahankinson:
It's all a bit weird, because technically these are more like breath marks / articulations than "bar lines".

craigsapp:
Perhaps, but they are not in all examples, so probably original from the original melodies.  It could be that the original notation did not have barlines, but had takt only for the phrase boundaries (breath marks).

Here is a case where the takt is use to split 6/2 measures in half, so it can't mean breath marks in this song:

Screenshot 2024-09-02 at 5 19 25 AM

paper page 541 in:
https://rosdok.uni-rostock.de/mcrviewer/recordIdentifier/rosdok_ppn738218359/iview2/phys_0643.iview2?logicalDiv=log_0016

Some German can read the Fraktur font to see if its purpose is described in the introduction:-)

ahankinson:
I guess we should also figure out if `<barLine>` is appropriate, or `<divLine>`?

https://music-encoding.org/guidelines/dev/elements/divLine.html

https://de.wikipedia.org/wiki/Pausa_(Neume)

src/iovolpiano.cpp
```diff
@@ -110,6 +111,28 @@ bool VolpianoInput::Import(const std::string &volpiano)
...
+            dbl->SetForm(BARRENDITION_dbl);
+            layer->AddChild(dbl);
+        }
+        else if (ch == '5') {
+            BarLine *end = new BarLine();
+            end->SetForm(BARRENDITION_end);
+            layer->AddChild(end);
+        }
+        else if (ch == '6') {
+            // not supported yet
```
lpugin, line 129:
```suggestion
            LogWarning("Volpiano '6' barline is not supported");
```

ahankinson:
Well, kinda. Volpiano is designed as pseudo-chant, but with everything in modern chant it gets filtered through CMN.

Here's what the _Liber Usualis_ has to say about these symbols:

image

So it may be that the creators of Volpiano thought of them to be _like_ bar lines, but they were probably representing concepts that were closer to the _diviso_ of the LU. 

With Craig's example the takt seem to be more phrase markings than bar lines. Who knows, but we can't create a standard encoding based on a single weird source.

ahankinson:
Preliminary tests indicate that `<divLine>` may work, but it needs a bit of finessing. So we can accept this PR, but the actual implementation may change later.

PR #3745, opened 2024-08-15
Fix dots on unisons

### LINKED ISSUE #3705: Too many dots on unisons
Not sure whether this qualifies as a feature request or a bug report: Dotted unisons in double stem notation sharing the same notehead should only have one dot, not two (cf. Behind Bars, p. 58).
This:
```
<staff n="2">
    <layer n="1">
        <note dots="1" dur.ppq="6" dur="4" oct="3" pname="a" stem.dir="up"/>
    </layer>
    <layer n="2">
        <note dots="1" dur.ppq="6" dur="4" oct="3" pname="a" stem.dir="down"/>
    </layer>
</staff>
```
currently produces this:
dotted_unison
The lower dot should be omitted. The problem only occurs when the notehead is placed on a staff or ledger line, not between lines.

brdvd:
This PR fixes an issue with dots on unisons by properly checking the staff when looking for the unison partner.

Some examples from the issue:

Dots1
Dots2
Dots3
Dots4
Dots5

Closes #3705 .

### DIFF
--- a/include/vrv/elementpart.h
+++ b/include/vrv/elementpart.h
@@ -22,7 +22,7 @@ class TupletNum;
 //----------------------------------------------------------------------------
 
 /**
- * This class models a group of dots as a layer element part and has not direct MEI equivlatent.
+ * This class models a group of dots as a layer element part and has no direct MEI equivalent.
  */
 class Dots : public LayerElement, public AttAugmentDots {
 public:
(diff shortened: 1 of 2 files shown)


rettinghaus (approved):
Thanks for the fix (and finding the typos).

lpugin:
Thanks!

PR #3757, opened 2024-09-02
Integrate latest changes in develop-humdrum

lpugin:
(See commits)


craigsapp:
Should be ok to merge.  You can point out any compiler warnings (or errors) that occur.

PR #3792, opened 2024-09-19
Proportions in CMME to MEI

martha-thomae:
Add support for encoding CMME `<TempoChange>` as an MEI proportion `<proport>` element with `@type = cmme_tempo_change` following the MEI `<mensur>` element, and for encoding CMME `<Proportion>` as a simple MEI `<proport>` element (not following any mensuration sign at all) with `@type = cmme_proportion`.

### DIFF
--- a/src/iocmme.cpp
+++ b/src/iocmme.cpp
@@ -36,6 +36,7 @@
 #include "measure.h"
 #include "mensur.h"
 #include "note.h"
+#include "proport.h"
 #include "rdg.h"
 #include "rest.h"
 #include "score.h"
@@ -315,6 +316,9 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
         }
         else if (name == "OriginalText") {
             CreateOriginalText(eventNode);
+        }
+        else if (name == "Proportion") {
+            CreateProport(eventNode);
         }
         else if (name == "Rest") {
             CreateRest(eventNode);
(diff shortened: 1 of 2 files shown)


src/iocmme.cpp
```diff
@@ -804,6 +826,26 @@ void CmmeInput::CreateOriginalText(pugi::xml_node originalTextNode)
...
+{
+    assert(m_currentContainer);
+
+    /// Proportion part coming from CMME <Proportion>. In this case, create an MEI <proport> element is created alone
+    /// (not following an MEI <mensuration> element)
+    Proport *proport = new Proport();
+    int numVal = this->ChildAsInt(proportNode, "Num");
+    int denVal = this->ChildAsInt(proportNode, "Den");
+    if (numVal != VRV_UNSET) {
+        proport->SetNum(numVal);
```
martha-thomae, line 839: Verovio doesn't currently support the SetNum() method in the Proport object.

src/iocmme.cpp
```diff
@@ -804,6 +826,26 @@ void CmmeInput::CreateOriginalText(pugi::xml_node originalTextNode)
...
+    }
+    if (denVal != VRV_UNSET) {
+        proport->SetNumbase(denVal);
```
martha-thomae, line 842: Verovio doesn't currently support the `SetNumbase()` method in the `Proport` object.

PR #3795, opened 2024-09-20
Add Ellipsis and Lacuna support in CMME import

DILewis:
Ellipsis --> gap
Lacuna --> supplied/spsace

### DIFF
--- a/src/iocmme.cpp
+++ b/src/iocmme.cpp
@@ -633,7 +633,7 @@ void CmmeInput::CreateLacuna(pugi::xml_node lacunaNode)
     assert(m_currentContainer);
     Space *space = new Space();
     Supplied *supplied = new Supplied();
-    supplied->addChild(space);
+    supplied->AddChild(space);
     int num;
     int numbase;
     data_DURATION duration = this->ReadDuration(lacunaNode, num, numbase);
--- a/src/iocmme.cpp
+++ b/src/iocmme.cpp
@@ -310,7 +310,6 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
                 CreateEllipsis();
             }
             else if (eventNode.child("Lacuna")) {
-                pugi::xml_node lacunaNode = eventNode.select_node("./Lacuna").node();
                 CreateLacuna(lacunaNode);
             }
             else {
--- a/src/iocmme.cpp
+++ b/src/iocmme.cpp
@@ -309,7 +309,7 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
             else if (eventNode.child("Ellipsis")) {
                 CreateEllipsis();
             }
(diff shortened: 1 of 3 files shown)


DILewis:
Added Signum Congruentiae

src/iocmme.cpp
```diff
@@ -304,6 +306,16 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
                 pugi::xml_node barlineNode = eventNode.select_node("./Barline").node();
                 CreateBarline(barlineNode);
             }
+            else if (eventNode.select_node("./Ellipsis")) {
```
lpugin, line 309: You don't need an xpath here. Just `evenNode.child("Ellispis")` is enough, which give you the node straight way
```suggestion
            else if (eventNode.child("Ellipsis")) {
```

src/iocmme.cpp
```diff
@@ -304,6 +306,16 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
...
+                CreateEllipsis();
+            }
+            else if (eventNode.select_node("./Lacuna")) {
+                pugi::xml_node lacunaNode = eventNode.select_node("./Lacuna").node();
+                CreateLacuna(lacunaNode);
```
lpugin, line 314:
```suggestion
                pugi::xml_node lacunaNode = eventNode.select_node("./Lacuna").node();
                CreateLacuna(eventNode.child("Lacuna"));
```

src/iocmme.cpp
```diff
@@ -304,6 +306,16 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
                 pugi::xml_node barlineNode = eventNode.select_node("./Barline").node();
                 CreateBarline(barlineNode);
             }
+            else if (eventNode.select_node("./Ellipsis")) {
+                CreateEllipsis();
+            }
+            else if (eventNode.select_node("./Lacuna")) {
```
lpugin, line 312:
```suggestion
            else if (eventNode.child("Lacuna")) {
```

src/iocmme.cpp
lpugin: (The previous line can be removed)

src/iocmme.cpp
```diff
@@ -304,6 +306,16 @@ void CmmeInput::ReadEvents(pugi::xml_node eventsNode)
                 pugi::xml_node barlineNode = eventNode.select_node("./Barline").node();
                 CreateBarline(barlineNode);
             }
+            else if (eventNode.child("Ellipsis")) {
+                CreateEllipsis();
+            }
+            else if (eventNode.child("Lacuna")) {
+                pugi::xml_node lacunaNode = eventNode.select_node("./Lacuna").node();
```
lpugin, line 313:
```suggestion
```

Issue #3796, opened 2024-09-20
CMME conversion


DILewis:
Issues arising from Biblissima-hosted meeting to work on support for the CMME format for mensural notation (see [cmme.org](https://www.cmme.org))

Some things are not supported or less well supported in MEI:

- Double-headed bmol (flat) signs [ignored in conversion]
- Gamma clefs (rare) [ignored in conversion]
- Half coloration (notes that span colored mensural units) [ignored in conversion]
- CMME describes barlines in terms of number of lines and presence of repeat dots separately (so single barline with repeats is possible [currently just using single, double, rptboth]
- OriginalText equivalent not available in the right place
- Signum Congruentiae are common, but not supported [currently using `@type`]

Some Verovio issues:
- Fermatas are not rendered
- We use `<gap>` for `<cmme:Ellipsis>`, but this is not visualised (in CMME, there is a stave gap)
- Breaks in layers have no effect (for obvious reasons), but could be visualised in other ways

We're also doing some less good things:
- Not all annotations are preserved

DILewis:
Also missing from MEI:

- <space> has duration information, but is missing num and numbase, which is needed for mensural durations

DILewis:
For _signum congruentiae_, CMME has `<Offset>`, which is a vertical position. This is used in our sample files, but only to resolve clashes of symbols. Currently we do nothing with this.  
Similarly, we ignore Orientation for the Signum (which indicates whether the symbol is flipped, usually for use below the staff). Signum also takes Side, which gives a horizontal offset left or right, possibly to support symbols that happen between notes or rests. Neither of these latter occur in our sample data.

lpugin:
One additional issue we need to solve is the cases of accidentals within a ligature. This is currently not allowed in MEI. I am added it to the list in the first post of this issue.

PR #3797, opened 2024-09-20
Add support for CMME import

lpugin:
CMME files can be imported with `-f cmme.xml` but should be auto-detected.

A CMME file is imported as an MEI mensural file with one single `mdiv/score`. Sections in CMME are imported as distinct `section` elements within the `score`.

Everything in imported as `mensural` notation, including sections marked as `Plainchant` in CMME since there is no difference in the way the music is encoded in those sections. Where there is less voices in a section than in the rest of the score, empty and invisible staff elements `<staff visibility="false"/>` are added to the section.

The initial `staffDef` elements have no `@clef.*` and no `@keyisg` and these are given in the `layer` instead. Every `staffDef` contains a `mensur` indicating that all level of divisions are binary by default.

There is one [issue](https://github.com/rism-digital/verovio/issues/3796) listing remaining points to tackle.


lpugin:
I would suggest for this to be merged. Several small things that remain to be adjusted will quite likely need some adjustments to MEI. In the meantime it will be easier if we already have this in the next release. More documentation will be added to the reference book once the release is made.

ahankinson (approved):

PR #3806, opened 2024-10-02
Refactoring for using fraction in alignment calculation

lpugin:
Internal refactoring to use a new `Fraction` class for managing the duration calculations and the alignment.

Additional clean-up of the code wit the removal of duration `#define` that are not replaced with the LibMEI `data_DURATION`, making it better type across the codebase.

Proper handling for mensural notation remains to be finalized, but not change should be expect. The few timemap changes triggered in the test suite seem to be due to rounding differences.

Once merge, the fractions will be made available in the Timemap too.


include/vrv/horizontalaligner.h
```diff
@@ -75,7 +124,7 @@ class Alignment : public Object {
      */
     ///@{
     Alignment();
-    Alignment(double time, AlignmentType type = ALIGNMENT_DEFAULT);
+    Alignment(const Fraction &time, AlignmentType type = ALIGNMENT_DEFAULT);
```
craigsapp, line 136: No need to change, but note that `const Fraction &time` is good, but `Fraction time` would work just as well (class contains only two integers) so placing a copy on the stack should not be a problem (and in theory it could be faster since it would not involve setting up pointers behind the scenes).

src/horizontalaligner.cpp
```diff
@@ -33,6 +33,97 @@
...
+
+void Fraction::Reduce()
+{
+    if (m_denominator < 0) { // Keep the denominator positive
+        m_numerator = -m_numerator;
+        m_denominator = -m_denominator;
+    }
+    int gcdVal = std::gcd(abs(m_numerator), abs(m_denominator));
+    m_numerator /= gcdVal;
+    m_denominator /= gcdVal;
```
craigsapp, line 135: Maybe add an if statement to only divide by gcdVal if it is not 1 (to save time of two unnecessary divisions in that case).

src/durationinterface.cpp
```diff
@@ -80,6 +80,8 @@ Fraction DurationInterface::GetInterfaceAlignmentDuration(int num, int numBase)
     if (this->HasNum()) num *= this->GetNum();
     if (this->HasNumbase()) numBase *= this->GetNumbase();
 
+    // Fraction duration(noteDur);
+    // duration = duration * numBase / num;
     Fraction duration = DUR_MAX / pow(2.0, (double)(noteDur - 2.0)) * numBase / num;
```
craigsapp: I am wondering how these calculations behave with the `Fraction` class.  Is `DUR_MAX` of type `Fraction`?    
 I am thinking it is a sort of enumeration behaving link `int`.  Otherwise the calculations will be using `double`.   Preferably the `int` version of `pow(2, x)` would be used instead of the double version `pow(2.0, x)`, but in any case the `pow()` function will be outputting -0.5 for `pow(2, -1)` in any case.  I am wondering what `- 2.0` is doing.  Converting to whole-note units?

craigsapp:
Looks great!  I added one comment on a code line that is confusing, but if you understand it, then it is OK.

craigsapp (approved):
Just the one complex line I commented about to check if that is what you really want.

src/horizontalaligner.cpp
```diff
@@ -33,6 +34,124 @@
...
+        denom = 1;
+    }
+    Reduce();
+}
+
+Fraction::Fraction(data_DURATION duration)
+{
+    duration = vrv::DurationMin(duration, DURATION_1024);
+    duration = vrv::DurationMax(duration, DURATION_NONE);
+    int den = pow(2, (duration + 1));
```
brdvd, line 54: `DURATION_NONE` is defined as -2 which gives pow(2, -1) on the rhs which is 0.5. Now casting to int (which is truncating) gives `den = 0`. I think this edge case still needs clarification.

src/horizontalaligner.cpp
```diff
@@ -33,6 +34,124 @@
...
+    }
+    int num = m_numerator * other.m_denominator;
+    int denom = m_denominator * other.m_numerator;
+    return Fraction(num, denom);
+}
+
+bool Fraction::operator==(const Fraction &other) const
+{
+    return (m_numerator == other.m_numerator) && (m_denominator == other.m_denominator);
+}
```
brdvd, line 95: This is weird compared to the others, because it would mean that 1/2 is not equal to 3/6. Since we now should have C++20 available, this is a neat example to try out the new spaceship operator:

```cpp
bool Fraction::operator<=>(const Fraction &other) const
 {
     return m_numerator * other.m_denominator <=> other.m_numerator * m_denominator;
 }
```

In theory, this should replace the implementation for all five operators `==`, `<`, `<=`, `>`, `>=`.

brdvd (commented):
Huge PR which is promising. I looked a bit into the `Fraction` class and left a few comments.

src/durationinterface.cpp
lpugin: Not sure but I think you were looking at an outdated version. This has been remove and uses now the `Fraction(data_DURATION)` constructor.

src/horizontalaligner.cpp
lpugin: You are right. I will change `DURATION_NONE` to `DURATION_maxima`.

src/horizontalaligner.cpp
lpugin: You a right too. I guess this has not been problematic because fraction are reduced throughout the code. But it should be fixed. I was not ware of the spaceship operator! I'll change to it.

lpugin:
@ahankinson I think this can be merged. I left `Fraction` within `horizontalaligner.h/cpp`. Having a distinct file for each class is a rule of thumb only for MEI elements, and the `Fraction` class is very small. 

The only differences left are for the mensural notation where the spacing is slightly different.

image

ahankinson:
Done!

Issue #3807, opened 2024-10-02
common and cut time signatures are not included on MIDI export


fernandoherreradelasheras:
time signatures with sym="common"  or sym="cut" are not exported to MIDI files. This is happening for initial time signatures in scoreDef/staffDef and for meterSig elements

PR #3808, opened 2024-10-02
Add MIDI time signatures for "common" and "cut" meterSig

fernandoherreradelasheras:
Fix for issue #3807

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -500,8 +500,13 @@ void Doc::ExportMIDI(smf::MidiFile *midiFile)
             if (!meterSig && (scoreDef->HasMeterSigInfo())) {
                 meterSig = vrv_cast<MeterSig *>(scoreDef->GetMeterSig());
             }
-            if (meterSig && meterSig->HasCount() && meterSig->HasUnit()) {
-                midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetUnit());
+            if (meterSig) {
+                if (meterSig->HasSym() && meterSig->GetSym() != METERSIGN_open) {
+                    midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetTotalCount());
+                }
+                else if (meterSig->HasCount() && meterSig->HasUnit()) {
+                    midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetUnit());
+                }
             }
         }
 
--- a/src/midifunctor.cpp
+++ b/src/midifunctor.cpp
@@ -775,8 +775,14 @@ FunctorCode GenerateMIDIFunctor::VisitScoreDef(const ScoreDef *scoreDef)
     // set MIDI time signature
     if (scoreDef->HasMeterSigInfo()) {
         const MeterSig *meterSig = vrv_cast<const MeterSig *>(scoreDef->GetMeterSig());
-        if (meterSig && meterSig->HasCount() && meterSig->HasUnit()) {
-            m_midiFile->addTimeSignature(m_midiTrack, currentTick, meterSig->GetTotalCount(), meterSig->GetUnit());
+        if (meterSig) {
+            if (meterSig->HasSym() && meterSig->GetSym() != METERSIGN_open) {
+                m_midiFile->addTimeSignature(
+                    m_midiTrack, currentTick, meterSig->GetTotalCount(), meterSig->GetTotalCount());
+            }
+            else if (meterSig->HasCount() && meterSig->HasUnit()) {
+                m_midiFile->addTimeSignature(m_midiTrack, currentTick, meterSig->GetTotalCount(), meterSig->GetUnit());
+            }
         }
     }
 


src/doc.cpp
```diff
@@ -500,8 +500,13 @@ void Doc::ExportMIDI(smf::MidiFile *midiFile)
             if (!meterSig && (scoreDef->HasMeterSigInfo())) {
                 meterSig = vrv_cast<MeterSig *>(scoreDef->GetMeterSig());
             }
-            if (meterSig && meterSig->HasCount() && meterSig->HasUnit()) {
-                midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetUnit());
+            if (meterSig) {
+                if (meterSig->HasSym() && meterSig->GetSym() != METERSIGN_open) {
+                    midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetTotalCount());
+                }
+                else if (meterSig->HasCount() && meterSig->HasUnit()) {
```
rettinghaus, line 507: I think `meter.count` and `meter.unit` would overrule any written symbol, so it should be two simple if clauses.

src/midifunctor.cpp
```diff
@@ -775,8 +775,14 @@ FunctorCode GenerateMIDIFunctor::VisitScoreDef(const ScoreDef *scoreDef)
...
     if (scoreDef->HasMeterSigInfo()) {
         const MeterSig *meterSig = vrv_cast<const MeterSig *>(scoreDef->GetMeterSig());
-        if (meterSig && meterSig->HasCount() && meterSig->HasUnit()) {
-            m_midiFile->addTimeSignature(m_midiTrack, currentTick, meterSig->GetTotalCount(), meterSig->GetUnit());
+        if (meterSig) {
+            if (meterSig->HasSym() && meterSig->GetSym() != METERSIGN_open) {
+                m_midiFile->addTimeSignature(
+                    m_midiTrack, currentTick, meterSig->GetTotalCount(), meterSig->GetTotalCount());
+            }
+            else if (meterSig->HasCount() && meterSig->HasUnit()) {
```
rettinghaus, line 783: same as above

rettinghaus (changes requested):
Thanks for this PR, just a little note from my side.

src/doc.cpp
```diff
@@ -500,8 +500,13 @@ void Doc::ExportMIDI(smf::MidiFile *midiFile)
             if (!meterSig && (scoreDef->HasMeterSigInfo())) {
                 meterSig = vrv_cast<MeterSig *>(scoreDef->GetMeterSig());
             }
-            if (meterSig && meterSig->HasCount() && meterSig->HasUnit()) {
-                midiFile->addTimeSignature(midiTrack, 0, meterSig->GetTotalCount(), meterSig->GetUnit());
+            if (meterSig) {
+                if (meterSig->HasSym() && meterSig->GetSym() != METERSIGN_open) {
```
lpugin, line 504:
```suggestion
                if (meterSig->HasSym() && (meterSig->GetSym() != METERSIGN_open)) {
```

src/doc.cpp
lpugin: Would you want to have both added to the midi file? If not then the two conditions should be flipped.

craigsapp: You should not have both added, of course.   I would say if you have an explicit count and unit, those should be used.   If there is no count/unit, then it should be inferred from the symbol, with common-time being 4/4 and cut-time being 2/2. 

I am wondering about this duplication:

```
meterSig->GetTotalCount(), meterSig->GetTotalCount());
```

the total count is the top number, and but this has nothing to do with the bottom number (the unit), so there seems to be an assumption here which is based on "4/4' and "2/2" having the same number on the top and the bottom.  But the top is the count and the bottom is the unit.

fernandoherreradelasheras: Yeah, in the same way that, for common/cut sym, meterSig->HasCount() returns false but meterSig->GetTotalCount() stills gives a valid value (4 or 2), we should do the same for unit and return 4 or 2 when HasUnit() is false. But doing it on GetUnit() would look weird from the semantic point of view... maybe GetEffectiveUnit()?  Or maybe GetTop()/GetBottom() / GetNum()/GetDem()

lpugin: You can simply have
```cpp
int countAndUnit = (meterSig->GetSym() == METERSIGN_common) ? 4 : 2;
```
And see this.

Don't forget to flip the clauses since the priority should be on explicit `@unit` and `@count`

fernandoherreradelasheras: Yeah, my point here was to encapsulate the logic resolving common = 4/4 and cut = 2/2 inside the MeterSig class instead of resolving those details at the document level. Also the ternary operator here will add a 2/2 time signature for a MeterSig with "open" sym (or any other further supported symbol we might introduce).

lpugin: OK. You can add a `MeterSig::GetSymImplicitUnit()`

lpugin (changes requested):
add a `MeterSig::GetSymImplicitUnit()`

PR #3813, opened 2024-10-06
Add D binding

andre2007:
Binding for the D Programming Language is added.

### DIFF
--- a/bindings/d/README.md
+++ b/bindings/d/README.md
@@ -1,0 +1,6 @@
+# Verovio D Binding
+
+Compile application with command `dmd -i app.d`  
+Run application with command `app sample.musicxml`
+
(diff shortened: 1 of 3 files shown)


rettinghaus (commented):
Thanks for this PR. Please add also some words about this to https://github.com/rism-digital/verovio-reference-book

lpugin:
Thanks. Have you consider using [SWIG](https://www.swig.org/Doc4.2/D.html#D)? This is what we use for Python and Java, and it makes it considerably easier to maintain because new methods or method changed do not have to be added or modified in the binding

andre2007:
Yes, an automatic solution makes much more sense 👍. I haven't used SWIG so far, I will need to try it out and check how get the result is with the D integration. I will create a new pull request when I am done, but it will take some time.

PR #3815, opened 2024-10-06
Add --timemap option to pass json options to the command-line

lpugin:
Usage:
```bash
laurent@gromit tools % ./verovio -h json         
Verovio 4.4.0-dev-53bde0a-dirty

Example usage:

 verovio [-s scale] [-r resource-path] [-o outfile] infile

Options (marked as * are repeatable)

Method JSON options for the command-line
 --timemap-options <s>                  The JSON options to be passed when producing the timemap (default: "{}")
```

Example:
```
verovio --timemap-options "{'includeMeasures': true, 'includeRests': true}" inputfile
```

### DIFF
--- a/include/vrv/options.h
+++ b/include/vrv/options.h
@@ -88,7 +88,7 @@ enum option_SMUFLTEXTFONT { SMUFLTEXTFONT_embedded = 0, SMUFLTEXTFONT_linked, SM
 // Option
 //----------------------------------------------------------------------------
 
-enum class OptionsCategory { None, Base, General, Layout, Mensural, Margins, Midi, Selectors, Full };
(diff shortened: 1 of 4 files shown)


craigsapp:
Looks good.  But what about the Fractions option?

lpugin:
When somebody will merge that PR ;-)

craigsapp:
;-)

For implementation I am wondering:

(1) will `qstamp` remain a double, which I highly recommend (i.e., it is required), for backwards compatibility and for avoiding real-time calculation of the double version for real-time highlighting, 

(2) What would the fractional parameter be called, with the two possibilities being `qfrac` which would allow both double and Fractions versions be available at the same time (which I do not think is absolutely necessary, or the Fraction option would change the type of `qstamp` to change from double to Fraction.

PR #3817, opened 2024-10-07
Use spaceship operator for fractions

brdvd:
This PR introduces the C++20 spaceship operator to the new `Fraction` class. There should be no changes in behaviour, I simply wanted to try this out. 

Lessons learned on my side:
- Equality comparison still needs to be defined. It is only default generated [whenever the spaceship operator is default](https://stackoverflow.com/questions/60681266/three-way-comparison-replaces-all-others-comparison-operators-except) (lexicographic order). The latter is not the case for `Fraction`: we define our own spaceship operator.
- The order on fractions should be _strong_, since they are always internally reduced.

### DIFF
--- a/include/vrv/horizontalaligner.h
+++ b/include/vrv/horizontalaligner.h
@@ -88,15 +88,9 @@ public:
 
     /** Equality operator */
     bool operator==(const Fraction &other) const;
-    /** Less than operator */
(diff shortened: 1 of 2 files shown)


lpugin:
Interesting! Thanks for looking into it - and I like the operator's name ;-)

Issue #922, opened 2018-07-23
Att::GetTstamp2() quanitization problem


craigsapp:
In `Att::MeasurebeatToStr()`, the beat value is limited to one fractional digit.  Why is that?

https://github.com/rism-ch/verovio/blob/6fd52221603c4a3fc2d195a7deba9dd430e8e23f/src/att.cpp#L302-L305

This should be removed or at least 4 or 5 digits after the decimal should be given.

This is the cause of issue https://github.com/humdrum-tools/verovio-humdrum-viewer/issues/140 where a trill line endtime is internally changed from 2.75 to 2.8 by this function, causing the trill line to extend too far.

lpugin:
Please change it to 4 decimal.

craigsapp:
Done with commit https://github.com/rism-ch/verovio/commit/1e3c4ba91ccb723538ed6e5b02c1392a55c95ec9

New test example:

```
 <trill xml:id="trill-L5F1" staff="1" startid="#note-L5F1" tstamp2="0m+2.7500" />
```

You do not want to use the stringstream class for such functions?   That will avoid trailing zeros when converting doubles into strings.


```cpp
std::string Att::MeasurebeatToStr(data_MEASUREBEAT data) const
{
    std::stringstream ss;
    ss << data.first << "m+" << data.second;
    return ss.str();
}
```

Example output with such a function:

```
<trill xml:id="trill-L5F1" staff="1" startid="#note-L5F1" tstamp2="0m+2.75" />
```

If rounding for tuplets is important, then this can be done to limit it to a particular number of digits like this:

```
std::string Att::MeasurebeatToStr(data_MEASUREBEAT data) const
{
    std::stringstream ss;
    ss << data.first << "m+" << (int)(data.second*1000000.0)/1000000.0;;
    return ss.str();
}
```

This limits the number of digits to 6 while also avoiding trailing zeros in the fractional part of the number.


The above code will produce the same as the previous example, since there are only two decimal digits:

```
 <trill xml:id="trill-L5F1" staff="1" startid="#note-L5F1" tstamp2="0m+2.75" />
```

PR #3818, opened 2024-10-08
Change decimal formatting

lpugin:
~The PR switches `%.10g` formatting decimal with a precision limited to 4 digits after the `.`.~

[Changing my mind: using `stringstream` with no relevant figures limitation]

This applies to `@tstamp` or `@tstamp2`, for example, and avoids the trailing zeros. See previous discussions [here](https://github.com/rism-digital/verovio/issues/922) and [there](https://github.com/rism-digital/verovio/issues/975)

The output will be `tstamp="1"` or `tstamp="1.75"` instead of `tstamp="1.0000"` or `tstamp="1.7500"` respectively.

~This implementation avoids the use of `stringstream`, which I find over the top. It sets an arbitrary limit to 10 significant numbers - if the value is larger that `999,999`, then the number of digits after the `.` will be reduced. We can adjust the value (or even change to stringstream) if a case where the loss would have an impact can be identified, but this looks quite unrealistic to me.~

### DIFF
--- a/libmei/addons/att.cpp
+++ b/libmei/addons/att.cpp
@@ -32,7 +32,7 @@ std::string Att::StrToStr(std::string str) const
 
 std::string Att::DblToStr(double data) const
 {
-    return StringFormat("%f", data);
+    return StringFormat("%.10g", round(data * 10000.0) / 10000.0);
 }
 
 std::string Att::IntToStr(int data) const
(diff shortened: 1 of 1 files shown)


rettinghaus (approved):
Looks familiar

lpugin:
> Looks familiar

Yes... actually maybe switching to stringstream might be better and solve this issue once for all

rettinghaus (approved):
Looks fine, thanks.

PR #3822, opened 2024-10-11
Add file for Fraction class

lpugin:
* No change expected


rettinghaus (approved):

bindings/iOS/all.h
```diff
@@ -184,6 +190,7 @@
 #import <VerovioFramework/positioninterface.h>
 #import <VerovioFramework/preparedatafunctor.h>
 #import <VerovioFramework/proport.h>
+#import <VerovioFramework/quilisma.h>
```
ahankinson, line 193: Just checking that these are supposed to be here? They don't seem like they're immediately relevant to this PR?

include/vrv/fraction.h
```diff
@@ -0,0 +1,79 @@
+/////////////////////////////////////////////////////////////////////////////
+// Name:        faction.h
```
ahankinson, line 2:
```suggestion
// Name:        fraction.h
```

bindings/iOS/all.h
lpugin: It happens for the `iOS` framework to be outdated, and it was the case here. In any case, it is always updated during releases, but I am also trying to keep it updated whenever new files are added.

ahankinson (approved):
Just a couple questions, but otherwise looks good.

bindings/iOS/all.h
ahankinson: OK!

PR #3824, opened 2024-10-13
fix: make c4 as default for neume notation

yinanazhou:
refs: https://github.com/DDMAL/Neon/issues/1241

### DIFF
--- a/src/clef.cpp
+++ b/src/clef.cpp
@@ -101,11 +101,6 @@ int Clef::GetClefLocOffset(data_NOTATIONTYPE notationType) const
         defaultOct = 3;
         offset = 4;
     }
-    else if (this->GetShape() == CLEFSHAPE_C) {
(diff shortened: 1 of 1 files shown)


lpugin:
This is how [this](https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/neume/neume-002.mei) file looks before and after:

image

I think you are changing the default octave for the C clef, which is not that same as making C4 the default.

yinanazhou:
This is expected. The test file is outdated as it uses C3 as default. I will update it.

lpugin:
OK, thanks. I think you also need to update [neume-001](https://github.com/rism-digital/verovio.org/blob/gh-pages/_tests/neume/neume-001.mei) (from the sixth staff)

lpugin:
Merging following discussion with @yinanazhou. The test file will be updated.

PR #3829, opened 2024-10-15
WIP: Improved Python bindings

ahankinson:
Updates the Python bindings with a few new capabilities:

 - Uses the -fastproxy and -olddeps options. In my testing it made the Python bindings a bit faster, but confirmation would be appreciated 
 - Added type hints for all the methods in the bindings using a .pyi file. 
 - Corrected some typos and inconsistencies


ahankinson:
WIP: Still needs a few things

rettinghaus:
Maybe then mark it as draft?

ahankinson:
Hm, now why didn't I think of that? ;-D

ahankinson:
OK, ready for review now.

setup.py
```diff
@@ -156,7 +156,7 @@ def get_version() -> str:
           'verovio.data.Petaluma': os.listdir('./data/Petaluma'),
           'verovio.data.text': os.listdir('./data/text'),
       },
```
rettinghaus, line 158: We should add the `py.typed` here, see https://peps.python.org/pep-0561/#packaging-type-information.

bindings/python/verovio.pyi
```diff
@@ -0,0 +1,105 @@
...
+    def clearHumdrumBuffer(self) -> None: ...
+    def setInputFrom(self, format: str) -> bool: ...
+    def getInputFrom(self) -> int: ...
+    def getOutputTo(self) -> int: ...
+    def setLocale(self) -> None: ...
+    def resetLocale(self) -> None: ...
+    def initClock(self) -> None: ...
+    def resetClock(self) -> None: ...
+    def getRuntimeInSeconds(self) -> float: ...
+    def logRuntime(self) -> None: ...
```
rettinghaus, line 78: Is there any particular logic to the order? Alphabetical sorting would make hand-editing much easier.

rettinghaus (commented):
Looks good to me.

bindings/python/verovio.pyi
ahankinson: I think it's the order that's in toolkit.h, which is in semi-alphabetical order.

But the ones at the end are kinda questionable, since these are annotated with `@nodocs` and don't appear in the Verovio book, so they're not technically "public."

setup.py
ahankinson: I think it gets copied here:

```py
      package_dir={'verovio': './bindings/python',
                   'verovio.data': './data'},
```

ahankinson: I've added it to the package_data object... not sure what it does, but it's there.

rettinghaus (approved):

PR #3841, opened 2024-10-31
Use bracketSpanLog_FUNC type for bracketSpan@func

lpugin:
Drop two string values added by MusicXML importer In order to use typed bracketSpanLog_FUNC.

If these values are really relevant, they should be added to the `@func` data type


rettinghaus:
Standing for two years https://github.com/music-encoding/music-encoding/issues/1065

ahankinson (approved):

PR #3862, opened 2024-11-18
Use range loops instead of iterator when possible

lpugin:
No change expected since this is only about simplifying code syntax.

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -763,7 +763,7 @@ void Doc::PrepareData()
 
     if (!prepareDelayedTurns.GetDelayedTurns().empty()) {
         for (auto &staves : layerTree.child) {
-            for (auto layers : staves.second.child) {
+            for (auto &layers : staves.second.child) {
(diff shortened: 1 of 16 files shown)


src/doc.cpp
```diff
@@ -769,12 +762,12 @@ void Doc::PrepareData()
     prepareDelayedTurns.SetDataCollectionCompleted();
 
     if (!prepareDelayedTurns.GetDelayedTurns().empty()) {
-        for (staves = layerTree.child.begin(); staves != layerTree.child.end(); ++staves) {
-            for (layers = staves->second.child.begin(); layers != staves->second.child.end(); ++layers) {
+        for (auto &staves : layerTree.child) {
+            for (auto layers : staves.second.child) {
```
lpugin, line 766:
```suggestion
            for (auto &layers : staves.second.child) {
```

ahankinson:
It's b0rken... Something about parts of C++20 not supported on some compilers?

PR #3881, opened 2024-12-09
Apply filters when processing pedals

lpugin:
Fixes #3880

* const_cast should be avoided, but needs refactoring of GetTstampStaves

PR #3882, opened 2024-12-09
Const overload for GetTstampStaves

brdvd:
This MR adds a const overload to `TimePointInterface::GetTstampStaves`. Related to the changes in #3881 .

### DIFF
--- a/include/vrv/timeinterface.h
+++ b/include/vrv/timeinterface.h
@@ -85,7 +85,7 @@ public:
     bool IsOnStaff(int n) const;
 
     /**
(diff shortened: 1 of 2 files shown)


include/vrv/timeinterface.h
```diff
@@ -87,7 +87,10 @@ class TimePointInterface : public Interface, public AttStaffIdent, public AttSta
     /**
      * Return a vector of staves looking at the @staff attribute or at the parent staff or the @startid
```
rettinghaus, line 88:
```suggestion
     * Return a vector of staves looking at the @staff attribute or at the parent staff of the @startid
```

Issue #3892, opened 2024-12-16
Small performance increase obtainable by constructing regular expressions just once.


paul-bayleaf:
Neither bug nor feature request - a suggestion for an optimisation.

Constructing a std::regex takes some work as it has to parse the regex and build a corresponding finite state machine.  Currently Verovio re-constructs each std::regex every time it does a match.  This is unnecessary, a performance saving can be made by just constructing each std::regex once and reusing it for all subsequent matches.  This is easily done by declaring each as "static const".  (constexpr is not available for std::regex). Regexs are used in iohumdrum.cpp, iomusxml.cpp, iopae.cpp and vrv.cpp.

Some timings:

1) There is a modest performance gain converting mei to mei.

Using https://github.com/music-encoding/sample-encodings/blob/main/MEI_5.0/Music/Complete_examples/Bach-JS_BrandenburgConcert_No2_I_BWV1047.mei
verovio -t mei -f mei -o test.mei Bach-JS_BrandenburgConcert_No2_I_BWV1047.mei

Current develop branch 2.496s. With static const std::regex 2.330s.  Performance gain 6.7%

2) Also modest gain converting humdrum to mei

Using https://kern.humdrum.org/cgi-bin/ksdata?l=musedata/bach/brandenburg&file=bwv1047a.krn&f=kern
verovio -t mei -f humdrum -o test.mei bwv1047a.krn

Current develop branch 5.763s.  With static const std::regex 5.527s.  Performance gain 4%

3) There is a bigger performance gain when converting musicxml that contains lyrics to mei, as this uses a lot of regexs.

Using https://www.verovio.org/examples/musicxml/Schubert_An_die_Sonne_D.439.xml
verovio -t mei -f xml -o test.mei Schubert_An_die_Sonne_D.439.xml

Current develop branch 2.118s.  With static const std::regex 1.153s.  Performance gain 45.6%

Although the gains are small in absolute time, they cost nothing to implement. I'll prepare a PR for consideration.

PR #3893, opened 2024-12-16
Implement issue #3892 Small performance increase obtainable by constructing regular expressions just once.

paul-bayleaf:
Make all std::regex definitions "static const" so that they are only constructed once.

### DIFF
--- a/src/iohumdrum.cpp
+++ b/src/iohumdrum.cpp
@@ -8443,13 +8443,13 @@ void HumdrumInput::setInstrumentAbbreviation(ELEMENT *element, const std::string
 
     std::string name8 = name;
     // Substitute b and "-flat" for Unicode flat symbol:
-    std::regex exp1("\\b([ABDEFG])b\\b");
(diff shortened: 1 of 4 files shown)


craigsapp (commented):
The change should be good, but static variables are not necessarily faster if they have to be recalled from slower memory (L2/L3 compared to L1).  Static variables can also affect thread performance.

Relative speeds to get to CPU:

| memory | clock cycles | size | 
| ---- | --- | --- | 
| L1 | 3-5 | 16-64KB | 
| L2 | 10-20 | 256kb-2MB |
| L3 | 30-50 | 4-64mb (shared between cores) |
| RAM | 300-500 | GB |
| SSD | ~100,000 | TB | 
| HD | ~10,000,000 | TB |

So if a variable takes <10 clock cycles to generate, it is not useful to store in L2; or <30 clock cycles if it is stored in L3.

lpugin:
Thanks!

lpugin (approved):

PR #3896, opened 2024-12-17
Add initial support for <annot type="score"> (classes and round trip read/write)

DILewis:
Class definitions, etc for a new type of annotation, 'annotScore', for annotations that we might want to appear in a score application. This PR also includes functionality for reading and writing (hopefully) transparently.
I'm putting some examples to test into [this repo](https://github.com/DILewis/MEI-annot-tests)


src/iomei.cpp
```diff
@@ -8264,6 +8317,13 @@ bool MEIInput::ReadXMLComment(Object *object, pugi::xml_node element)
     return true;
 }
 
+bool MEIInput::IsAnnotScore(pugi::xml_node annot)
+{
+    // If the annotation is a score annotation, it'll have @type="score" (we can also guess)
+    std::string value = annot.attribute("type").value();
+    return value == "score";
```
lpugin, line 8324: Stylistic preference
```suggestion
    return (value == "score");
```

src/iomei.cpp
```diff
@@ -7666,6 +7690,35 @@ bool MEIInput::ReadAnnot(Object *parent, pugi::xml_node annot)
...
+    }
+    this->ReadUnsupportedAttr(annot, vrvAnnotScore);
+    // Unless annot has only text we do not load children because they are preserved in Annot::m_content
+    if (hasNonTextContent) {
+        return true;
+    }
+    else {
+        // vrvAnnot->m_content.remove_children();
+        return this->ReadTextChildren(vrvAnnotScore, annot, vrvAnnotScore);
+    }
```
lpugin, line 7719: Keep only what you use now and remove the commented code. We can easily put things back whenever needed.

DILewis: I've taken cancelled code out of here, but left a few comments to remind myself what's gone

PR #3915, opened 2025-01-14
fix: rotation offset used in neon

yinanazhou:
- Set rotate to 0 if not has rotate
- Add clef offset when using default clef

refs: https://github.com/DDMAL/Neon/issues/1254, https://github.com/DDMAL/Neon/issues/1226

### DIFF
--- a/include/vrv/editortoolkit_neume.h
+++ b/include/vrv/editortoolkit_neume.h
@@ -135,10 +135,10 @@ struct ClosestBB {
         Zone *zoneA = a->GetFacsimileInterface()->GetZone();
         Zone *zoneB = b->GetFacsimileInterface()->GetZone();
 
-        int distA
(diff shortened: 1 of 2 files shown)


src/facsimilefunctor.cpp
```diff
@@ -184,7 +184,10 @@ FunctorCode SyncFromFacsimileFunctor::VisitStaff(Staff *staff)
         m_staffZones[staff] = zone;
 
         // The staff slope is going up. The y left position needs to be adjusted accordingly
-        if (zone->HasRotate() && zone->GetRotate() < 0) {
+        if (!zone->HasRotate()) {
+            zone->SetRotate(0);
```
lpugin, line 188: Why do not need it to actually set it to `0`? I think if you do this, then writing the file will have an explicit `0` value in the MEI, which should not be necessary.

yinanazhou: I see. `@rotate` is used in calculating closest staff. I will revert this commit and make other changes.

PR #3916, opened 2025-01-14
Add /bigobj option

agkhall:
Without this option, [fatal error C1128](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1128) prevents compilation due to the object file format size limit, at least while building Verovio as a Windows shared library using MSVC via x64 Native Tools Command Prompt for VS 2022: 
library
Adding this option seems to have fixed the library, although I haven't actually tested the DLL yet.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -29,6 +29,7 @@ endif()
 
 # Sets the compile-time value for the RESOURCE_DIR macro in the source code.
 add_compile_definitions(RESOURCE_DIR="${CMAKE_INSTALL_PREFIX}/share/verovio")
+add_compile_options(/bigobj)
 
 include_directories(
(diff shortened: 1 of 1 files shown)


craigsapp:
That looks good, here is info on the web that says heavy use of STL or templating may require `/bigobj` option for compmiling Windows exeutables:

> A standard COFF (Common Object File Format) object file can contain up to 65,536 (64K) sections. Each section corresponds to a segment of code, data, or other resources. However, in large projects—especially those with a lot of template or inline code, such as C++ projects with extensive use of STL or heavy templating—this limit can easily be exceeded
> 
> To resolve this issue, you need to instruct the compiler to use the /bigobj option, which allows the object file to exceed the 64K section limit.

Issue #2147, opened 2021-04-21
[QUESTION] Usage of this pointer


DavidBauer1984:
The codebase uses `this->` in many places to access class members and member functions. It is used quite a lot, but not consistently everywhere. The only functional difference I can think of is when there is a local variable with the same name as a class member. However, this is very unlikely, since class members start with "m_" while local variables do not. Even if the member function is virtual, then which function is called during execution does not depend on whether `this` is used or not. So basically, it could be omitted. What is the reason for using it? Is it historical?

lpugin:
Yeah, you are right the codebase is not consistent on that point. It should certainly be omitted for class members since they do start with `m_`. I am less sure what to decide for methods event though you are right it is not relevant for execution. I guess having (i.e., typing) `this->` is a way to trigger the autocomplete and is one reason why it is there in many places. I would suggest:
* omit them for class members
* have it for methods

Does it sound reasonable?

DavidBauer1984:
Yes, sounds good to me. For me it would be helpful to have a clear convention. Maybe we can put this in the coding guidelines.

Here is a related [discussion](https://stackoverflow.com/questions/9590820/should-i-use-this-within-a-class) on StackOverflow. Apparently, some people use it and some do not.

lpugin:
Added here https://github.com/rism-digital/verovio-reference-book/commit/9649c8c6e7ab2ec2e2c68d3adcd65963f09c876e (new documentation in preparation)

PR #3919, opened 2025-01-25
Split Fraction constructors

brdvd:
This PR adds a simplified `Fraction(int)` constructor setting the denominator to 1. In this case we don't have to reduce the fraction and don't need to check for zero denominator.

### DIFF
--- a/include/vrv/fraction.h
+++ b/include/vrv/fraction.h
@@ -20,7 +20,8 @@ class Fraction {
 
 public:
     // Constructors - make them explicit to avoid type conversion
-    explicit Fraction(int num = 0, int denom = 1);
+    explicit Fraction(int num = 0) : m_numerator(num), m_denominator(1) {}
+    explicit Fraction(int num, int denom);
     explicit Fraction(data_DURATION duration);
 
     // Enable implicit conversion constructor for `int`
--- a/src/fraction.cpp
+++ b/src/fraction.cpp
@@ -29,7 +29,7 @@ Fraction::Fraction(int num, int denom)
         denom = 1;
     }
     m_denominator = denom;
-    Reduce();
+    this->Reduce();
 }
 
 Fraction::Fraction(data_DURATION duration)
@@ -39,7 +39,7 @@ Fraction::Fraction(data_DURATION duration)
     int den = pow(2, (duration + 1));
     m_numerator = 8;
     m_denominator = den;
-    Reduce();
+    this->Reduce();
 }
 
 Fraction Fraction::operator+(const Fraction &other) const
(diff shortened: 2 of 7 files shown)


rettinghaus (approved):

src/fraction.cpp
```diff
@@ -122,7 +122,7 @@ void Fraction::Reduce()
         m_numerator = -m_numerator;
         m_denominator = -m_denominator;
     }
-    int gcdVal = std::gcd(abs(m_numerator), abs(m_denominator));
+    const int gcdVal = std::gcd(m_numerator, m_denominator);
```
craigsapp, line 125: This is OK, because `std::gcd()` (introduced in C++17) internally applies `abs()` to the two input arguments (or at least only outputs non-negative values).   There is also a GCC macro `__gcd()` that can return negative values:

```cpp
#include <iostream>
#include <algorithm>     // For std::gcd (C++17 and later)
#include <bits/stdc++.h> // For __gcd    (GCC specific)

int main() {
    int a = -2, b = 4;
    std::cout << std::gcd(a, b) << std::endl;  // +2
    std::cout << __gcd(a, b) << std::endl;     // -2
    return 0;
}
```

include/vrv/fraction.h
```diff
@@ -20,7 +20,8 @@ class Fraction {
 
 public:
     // Constructors - make them explicit to avoid type conversion
-    explicit Fraction(int num = 0, int denom = 1);
+    explicit Fraction(int num = 0) : m_numerator(num), m_denominator(1) {}
+    explicit Fraction(int num, int denom);
```
craigsapp, line 24: Great addition

src/horizontalaligner.cpp
```diff
@@ -516,15 +516,15 @@ void Alignment::Reset()
...
     m_type = ALIGNMENT_DEFAULT;
 
-    ClearGraceAligners();
+    this->ClearGraceAligners();
 }
 
 Alignment::~Alignment()
 {
-    ClearGraceAligners();
+    this->ClearGraceAligners();
```
craigsapp, line 527: Is there a reason for adding `this->` to member-function calls? (Here and elsewhere).

rettinghaus: Style according to the [coding guidelines](https://book.verovio.org/contributing/guidelines.html#class-method-and-member-names).

craigsapp: > Screenshot 2025-01-25 at 13 04 31

I have not been doing this (pun intended), but I will start doing it.   This is perhaps a later addition to the style guide due to the last sentence.

brdvd: It was added to the style guide after [this discussion](#2147 ) (pun intended) which I started after diving into the codebase. After using the convention for quite some time, member function calls without `this->` start looking odd.

PR #3924, opened 2025-01-31
Fix lost IDs on symbolDef

rettinghaus:
This PR fixes a problem with losing the `@xml:id` on `symbolDef` elements when outputting MEI using the `--remove-ids` option. The functor simply ignored the `AltSymInterface`.

It also fixes lost attributes on `graphic` (e.g., `@ulx`) that are not supported yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Support for `verse@place` for lyrics above the staff
 * Support for MIDI octave displacement without `@oct.ges` (@brdvd)
 * Option `--lyric-height-factor` to increase the spacing of the lyrics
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Support for `verse@place` for lyrics above the staff
 * Support for MIDI octave displacement without `@oct.ges` (@brdvd)
 * Option `--lyric-height-factor` to increase the spacing of the lyrics
+* Fix `symbolDef` losing `@xml:id`s (@rettinghaus)
```
lpugin, line 8:
```suggestion
* Fix `symbolDef` losing `@xml:id`s with `--remove-ids` (@rettinghaus)
```

PR #3927, opened 2025-01-31
treat virtual units like doubles

rettinghaus:
This PR brings the writing of virtual unit in liune with doubles, changing this

`<multiRest loc="2" block="true" num="4" width="6.0000vu" />`

into this

`<multiRest loc="2" block="true" num="4" width="6vu" />`

### DIFF
--- a/libmei/addons/att.cpp
+++ b/libmei/addons/att.cpp
@@ -44,7 +44,9 @@ std::string Att::IntToStr(int data) const
 
 std::string Att::VUToStr(data_VU data) const
 {
-    return StringFormat("%fvu", data);
+    std::stringstream sstream;
+    sstream << round(data * 10000.0) / 10000.0 << "vu";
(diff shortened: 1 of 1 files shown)


libmei/addons/att.cpp
```diff
@@ -44,7 +44,9 @@ std::string Att::IntToStr(int data) const
 
 std::string Att::VUToStr(data_VU data) const
 {
-    return StringFormat("%fvu", data);
+    std::stringstream sstream;
+    sstream << round(data * 10000.0) / 10000.0 << "vu";
+    return sstream.str();
```
lpugin, line 49: Why not using `DblToStr`?

rettinghaus:
Like this?

lpugin:
I was thinking `VuToStr` to return `DblToStr(data.GetVu()) + "vu"`

PR #3930, opened 2025-01-31
 Fix more lost IDs 

rettinghaus:
Same fix as in #3924 just for `surface` and `zone`.


rettinghaus:
(has to be adjusted after #3929)

PR #3931, opened 2025-01-31
Score annotations now have borders on the boxes and adjust vertically placement sensibly

DILewis:


src/view_control.cpp
```diff
@@ -471,25 +471,70 @@ void View::DrawAnnotScore(
...
+            dc->DrawPolygon(4, boxOutline);
+            // Drawing the rectangle afterwards because I can't work out how to make the polygon transparent
+            this->DrawFilledRectangle(dc, x1, y, x2, y + boxHeight);
+            break;
+        case SPANNING_END:
+            // Draw a box with the left side open to show it continues from previous system
+            if (!annotScore->GetEnd()->Is(TIMESTAMP_ATTR)) {
+                x2 += annotScore->GetEnd()->GetDrawingRadius(m_doc);
+            }
+            this->DrawFilledRectangle(dc, x1, y, x2, y + boxHeight);
```
lpugin, line 529: Any reason you are not drawing the rectangle afterwards here

DILewis: Only because I did it in that order initially (because I wasn't sure if it was going to draw the rect on top of the line, obscuring it) and switched for the polygon (because I struggled to make the fill NULL). Should I swap them for consistency?

lpugin: Yes, make it consistent. Transparent rectangles are not really used, I think, so we might need to look into this.

lpugin:
Could you add an example of the result?

DILewis:
Using this [CRIM-derived example](https://github.com/DILewis/MEI-annot-tests/blob/main/CRIM_Josquin_Ave-Maria.mei) from my test repo, generates CRIM_Josquin_Ave-Maria 
This isn't very pretty, because the lines and boxes aren't styled. Loading in Firefox and then adding the following CSS through dev tools 
.annotScore rect {
  fill: #CCD;
}
.annotScore polyline, .annotScore polygon {
  stroke: #009;
  stroke-width: 100;
}
gives Screen Shot 2025-01-31 at 17 51 01

lpugin:
You can change the CSS with the `--svg-css` option. What do you think the default should be? A light-grey filled rectangle?

DILewis:
The rendering question bothers me, and I don't know the most idiomatically Verovio approach. I guess options might include:
1. Invisible unless explicitly switched..
2. Light-grey filled rectangle (I think breaking the default greyscale rendering might be weird)
3. Rotating colours / fills (I note that nth-child rules for a client trying to distinguish by colour will fail miserably)
4. 2, but with a visible index for the annotation, so marginalia can be linked. Is there any sense in giving `annot` `@n`?

I think starting with 2 is probably the simplest

craigsapp:
> Rotating colours / fills (I note that nth-child rules for a client trying to distinguish by colour will fail miserably)

No rotating colors...

 but with a visible index for the annotation, so marginalia can be linked. Is there any sense in giving annot @n?

That is a good idea.  An additional idea might be to use `@type` which gets converted to a CSS class when verovio renders to SVG.  Then categories of annotations can be given different styles.

DILewis:
Hi @craigsapp, thanks for the thoughts!
We're not (yet?) talking about boxes around notes (not least because it gets complicated quickly, for example for single voices on polyphonic staves). Our initial thoughts have been that in many cases, (dynamic) client applications would do the last mile of drawing, so giving bounding boxes might be enough, and the wide variety of ways of rendering could be left to software. 
As far as box extents are concerned, this code supports `@tstamp` and `@startid`/`@endid`, so there are options for the encoder, but I do note that the current bars are pretty tight to the symbols. `@plist` is also available for the annotations, but is currently not rendered.
On colour, I do subscribe to the 'no information should be exclusively communicated by colour' mantra, but there's nothing wrong with colouring boxes differently (as you do). I think the mapping should be for users/clients, but it's not clear that there's yet a mechanism to support that. We might need at least to provide a list of the known annotations and their ids as a helper function, but even there, the boxes we're drawing at the moment don't have a single attribute that ties them to an annotation (it's currently either `@id` or `@class='id-xxxx'` depending on whether it's a single box or a continuation across a system).

PR #3932, opened 2025-01-31
treat percentage like doubles

rettinghaus:
This PR brings finally percentage values in line with other output of other doubles.


rettinghaus:
Test fails on `develop` branch, as expected. Fix included here.

PR #3936, opened 2025-02-03
Use inline

rettinghaus:
This PR combines the `PercentToStr` method for other percentage values, and marks several forwards as `inline` to reduce function calls.

### DIFF
--- a/libmei/addons/att.h
+++ b/libmei/addons/att.h
@@ -90,8 +90,11 @@ public:
     std::string MeasurementsignedToStr(data_MEASUREMENTSIGNED data) const;
     data_MEASUREMENTSIGNED StrToMeasurementsigned(const std::string &value, bool logWarning = true) const;
 
-    std::string MeasurementunsignedToStr(data_MEASUREMENTUNSIGNED data) const { return MeasurementsignedToStr(data); }
-    data_MEASUREMENTUNSIGNED StrToMeasurementunsigned(const std::string &value, bool logWarning = true) const
+    inline std::string MeasurementunsignedToStr(data_MEASUREMENTUNSIGNED data) const
+    {
+        return MeasurementsignedToStr(data);
+    }
+    inline data_MEASUREMENTUNSIGNED StrToMeasurementunsigned(const std::string &value, bool logWarning = true) const
     {
         return StrToMeasurementsigned(value, logWarning);
     }
--- a/libmei/addons/att.h
+++ b/libmei/addons/att.h
@@ -132,10 +132,10 @@ public:
     std::string PercentToStr(data_PERCENT data) const;
     data_PERCENT StrToPercent(const std::string &value, bool logWarning = true) const;
 
-    std::string PercentLimitedToStr(data_PERCENT_LIMITED_SIGNED data) const;
+    inline std::string PercentLimitedToStr(data_PERCENT_LIMITED_SIGNED data) const { return PercentToStr(data); }
     data_PERCENT_LIMITED StrToPercentLimited(const std::string &value, bool logWarning = true) const;
 
-    std::string PercentLimitedSignedToStr(data_PERCENT_LIMITED data) const;
+    inline std::string PercentLimitedSignedToStr(data_PERCENT_LIMITED data) const { return PercentToStr(data); }
     data_PERCENT_LIMITED_SIGNED StrToPercentLimitedSigned(const std::string &value, bool logWarning = true) const;
 
     std::string PitchnameToStr(data_PITCHNAME data) const;
--- a/libmei/addons/att.cpp
+++ b/libmei/addons/att.cpp
@@ -668,7 +668,7 @@ data_PERCENT_LIMITED_SIGNED Att::StrToPercentLimitedSigned(const std::string &va
 {
     static const std::regex test("(+|-)?[0-9]+(\\.?[0-9]*)?%");
     if (!std::regex_match(value, test)) {
-        if (logWarning) LogWarning("Unsupported data.PERCENT.LIMITED.SIGNEd '%s'", value.c_str());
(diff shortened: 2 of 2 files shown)


libmei/addons/att.h
```diff
@@ -42,13 +42,13 @@ class Att : public AttConverterBase {
...
     /** Dummy string converter */
-    std::string StrToStr(std::string str) const;
+    std::string StrToStr(const std::string str) const;
 
     /** @name Basic converters for writing */
     ///@{
-    std::string DblToStr(double data) const;
-    std::string IntToStr(int data) const;
-    std::string VUToStr(data_VU data) const;
+    std::string DblToStr(const double data) const;
```
lpugin, line 49: What difference does it make?

libmei/addons/att.h
```diff
@@ -42,13 +42,13 @@ class Att : public AttConverterBase {
     static data_STAFFREL_basic StaffrelToStaffrelBasic(data_STAFFREL staffrel);
 
     /** Dummy string converter */
-    std::string StrToStr(std::string str) const;
+    std::string StrToStr(const std::string str) const;
```
lpugin, line 45: This should be change to a reference

libmei/addons/att.h
rettinghaus: Basically none, except for code consistency.

lpugin: Consistency with what?

craigsapp:
In general for modern compilers you should not use the inline keyword, and instead rely on compiler optimization flags to decide whether the function should be inlined or not automatically.  In particular inlining will increase the size of the executable since the function's code is inserted where there would otherwise be a shared function (so very small functions are not a problem, but these could be determined automatically when optimizing the code).   

Code size versus speed should be considered in particular for the Javascript toolkit version for download over the internet, where size is more important than otherwise for a C++ library version of verovio running on a local computer.

Asking the new ChatGPT 3o model ("Great at coding and logic") about this:

> What is the advantage/disadvantage of using inline keyword in C++?

The `inline` keyword in C++ serves two main purposes, each with its own advantages and disadvantages:

# Advantages

1. **Potential Performance Improvement:**
- **Reduced Function Call Overhead:**
When you mark a function as `inline`, you are suggesting to the compiler that it should replace the function call with the actual function code. This can eliminate the overhead associated with a function call (such as setting up a stack frame), which can be beneficial especially for very small, frequently called functions.
2. **Header File Definitions:**
- **Avoiding Multiple Definitions:**
In C++, if you define a function in a header file that is included in multiple translation units, you risk violating the One Definition Rule (ODR). Declaring the function as `inline` tells the linker that multiple identical definitions across different translation units are acceptable. This makes it safe to define functions in headers.

# Disadvantages

1. **Code Bloat:**
 - **Increased Binary Size:**
If an inline function is large or used in many places, substituting the function code at every call site can lead to a larger binary. This code bloat can negatively impact performance, for example by increasing the pressure on the instruction cache.
2. **Limited Compiler Control:**
 - **A Hint, Not a Command:**
The inline keyword is merely a suggestion to the compiler. Modern compilers perform their own optimizations and may choose to inline functions that aren’t marked inline or ignore the inline suggestion if the function is too complex. Thus, using the keyword does not guarantee that inlining will occur.
3. **Debugging Difficulties:**
 - **Harder to Set Breakpoints:**
When code is inlined, there isn’t a single, distinct function call to break on during debugging. This can make stepping through the code or diagnosing issues more challenging.


# Summary

- **When to Use `inline`:**
 - It is most useful for very small functions where the overhead of a function call is significant compared to the function's work.
It is essential when you want to define a function in a header file that will be included in multiple translation units, thus avoiding multiple definition errors.
- **When to Avoid or Be Cautious with `inline`:**
 - For large functions or functions that are called infrequently, inlining might increase code size without providing a noticeable performance benefit.
 - Relying too much on `inline` can sometimes hinder the compiler's ability to perform more effective optimizations.

In modern C++ development, the compiler's optimization capabilities are quite advanced. Often, it is best to write clear and maintainable code and let the compiler decide which functions to inline based on its optimization heuristics, rather than relying too heavily on the `inline` keyword.

rettinghaus:
@craigsapp I'm not completely sure what you want to tell. 

- Binary size before this PR: 13611448
- Binary size after this PR: 13611192

So it's smaller _and_ faster.
It would be good if you could share those compiler optimization flags.

lpugin:
It does not seem to make a very significant difference size-wise. Probably not more speed-wise.

You can see https://emscripten.org/docs/tools_reference/settings_reference.html#inlining-limit for the emscripten compiler.

brdvd:
If a function is defined inside a _class declaration_, it is automatically considered inline by the compiler. So it should not be necessary to add `inline` for functions like `Att:: MeasurementunsignedToStr`.

It is a different story if the function does not belong to a class. Without `inline` it would violate the One Definition Rule and you get a linker error there. One example is `DurationMin/Max` in `vrv.h`.

I think so far the approach has been to add `inline` defensively, i.e. only in those cases where it is required for successful compilation.

craigsapp:
> So it's smaller and faster.

It is smaller, which is interesting, but you did not provide proof of speed (but I would expect inline code to be faster since it does not need to jump to a new location in memory).  Probably there are no local variables in the function, so the space saved is based on the lack of a pointer to a separate function.

It is fine with me, but note that `inline` is a suggestion that now-a-days is typically left to the compiler when optimizing code.

> It is a different story if the function does not belong to a class. Without inline it would violate the One Definition Rule and you get a linker error there. One example is DurationMin/Max in vrv.h.

That is a good point, although I avoid defining functions in header files, so I have not been aware of that (probably because I got duplicate function compiling errors in the past and instinctively avoid doing that :-).

lpugin:
OK. I removed them https://github.com/rism-digital/verovio/commit/45286357365d72a17475c1c15514603165231a27

PR #3939, opened 2025-02-06
Use a simplified internal structure for Mensur

lpugin:
@martha-thomae could you double check?


martha-thomae:
I doubled checked it now, all seems to be working perfectly! Thank you!

PR #3947, opened 2025-02-09
Fix handling of -p and -a options regarding MEI output

### LINKED ISSUE #3938: Writing mensural impossible without `--all-pages`
**Describe the problem**
Trying to load a Mensural MEI into Verovio and saving it back gives an Error: 

> [Error] MEI output with filter is not possible for mensural music

**To Reproduce**
1. Take a mensural file from Verovio's test suite
2. `verovio -t mei` _filename_
3. See error

**Expected behavior**
The file is written back.

**Input data**
See https://www.verovio.org/test-suite.xhtml?cat=mensural

**Verovio information**
 - develop
 - command-line

 - macOS

brdvd:
This PR applies the `-a` option only for rendering. It keeps the behavior for rendering as before:
- By default, the first page is rendered.
- If `-p` is provided, then only the given page is rendered.
- If `-a` is provided, then all pages are rendered.
- If `-p` and `-a` are provided, then the range from the given page until the end is rendered.

The behavior for MEI output is changed. The option `-a` is ignored here. Instead:
- By default, all pages are written.
- If `-p` is provided, then only the given page is written. This corresponds to setting a page filter and does not work in certain cases, e.g. mensural music.

Closes #3938 .

### DIFF
--- a/tools/main.cpp
+++ b/tools/main.cpp
@@ -363,7 +363,7 @@ int main(int argc, char **argv)
 
     if (outformat == "svg") {
         const int from = page ? *page : 1;
-        const int to = (page && !all_pages) ? *page : toolkit.GetPageCount();
+        const int to = all_pages ? toolkit.GetPageCount() : from;
 
         for (int p = from; p <= to; ++p) {
             std::string cur_outfile = outfile;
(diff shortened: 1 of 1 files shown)


lpugin:
Looks good, thanks

PR #3961, opened 2025-02-20
fix: use `currentColor` for `divLine` stroke (Firefox)

yinanazhou:
refs: https://github.com/DDMAL/Neon/issues/1290, https://github.com/DDMAL/Neon/issues/1290

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -466,7 +466,8 @@ void SvgDeviceContext::StartPage()
         m_currentNode.append_attribute("type") = "text/css";
         std::string css = "g.page-margin{font-family:Times,serif;} "
                           "g.ending, g.fing, g.reh, g.tempo{font-weight:bold;} g.dir, g.dynam, "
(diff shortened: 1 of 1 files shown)


rettinghaus (commented):
This is basically fine, but I'd like to understand why you need this. Verovio is using glyphs for `divLine`, so you shouldn't be needing it.

yinanazhou:
This is because `divLine` is quite thin. It's a bit difficult for the user to see and to click on it. So I've been setting the stroke to be wider. Also when I change the `divLine` color, the stroke will cover the glyph color, so I change stroke color as well.

rettinghaus:
But this is then a Neon specific feature that depends on the font you are using and shouldn't be integrated in Verovio.

yinanazhou:
Neon is using `Bravura`. Do you know which font would work for `divLine`?

lpugin:
@yinanazhou could you try adding the rule as `svgCss` option passed to Verovio in Neon? That should do the trick.

PR #3963, opened 2025-02-22
update midifile to upstream

rettinghaus:


craigsapp:
This update is good, as you probably noticed related to FIFO and LIFO for linking notes that was discussed two weeks ago in the `midifile` issues, although I don't think that the switch to FIFO linking of note ons/offs would affect midifile export from verovio.  

I made some changes to the sorting algorithm (which is used in verovio) which will affect the midifile order of messages occurring at the same time. Initially I wanted to sort note offs to come before note ons (to avoid problems with LIFO linking).  But then I notice that it is possible for zero-length grace notes to cause a stuck note-on if that is done, so I sort note-offs to come after note-ons at the same timestamp.  I may not have done this in the previous sorting method, and if so, there was a chance previously that some zero-length grace notes could have generated a stuck-note problem, that in theory should no longer be possible.

craigsapp:
Also, I think there was a bug with sorting of the end-of-track meta message which looks like it could have caused problems with the last note's duration in a track (shortening the last note).  I have noticed this on VHV and was presuming it was a bug in the Wild West MIDI library, but if this PR fixes that problem, then it was in the `midifile` code instead.

Issue #3975, opened 2025-03-01
svg output has transparent lines when --mm-output is used


fernandoherreradelasheras:
SvgDeviceContext::DrawLine is not adding the stroke attribute when the current pen has not color.

PR #3976, opened 2025-03-01
use currentColor as default for lines when svg is not using global st…

fernandoherreradelasheras:
Fix for #3975

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -814,11 +814,8 @@ void SvgDeviceContext::DrawLine(int x1, int y1, int x2, int y2)
 {
     pugi::xml_node pathChild = AddChild("path");
     pathChild.append_attribute("d") = StringFormat("M%d %d L%d %d", x1, y1, x2, y2).c_str();
-    if (m_penStack.top().HasColor()) {
+    if (m_penStack.top().HasColor() || this->UseGlobalStyling()) {
         pathChild.append_attribute("stroke") = this->GetColor(m_penStack.top().GetColor()).c_str();
-    }
-    else if (!this->UseGlobalStyling()) {
-        pathChild.append_attribute("stroke") = "currentColor";
     }
     if (m_penStack.top().GetWidth() > 1) pathChild.append_attribute("stroke-width") = m_penStack.top().GetWidth();
     this->AppendStrokeLineCap(pathChild, m_penStack.top());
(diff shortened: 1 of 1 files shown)


src/svgdevicecontext.cpp
```diff
@@ -817,6 +817,9 @@ void SvgDeviceContext::DrawLine(int x1, int y1, int x2, int y2)
     if (m_penStack.top().HasColor()) {
         pathChild.append_attribute("stroke") = this->GetColor(m_penStack.top().GetColor()).c_str();
     }
+    else if (!this->UseGlobalStyling()) {
+        pathChild.append_attribute("stroke") = "currentColor";
+    }
```
rettinghaus, line 822:
```suggestion
    if (m_penStack.top().HasColor() || this->UseGlobalStyling()) {
        pathChild.append_attribute("stroke") = this->GetColor(m_penStack.top().GetColor()).c_str();
    }
```

rettinghaus (changes requested):

src/svgdevicecontext.cpp
```diff
@@ -814,7 +814,7 @@ void SvgDeviceContext::DrawLine(int x1, int y1, int x2, int y2)
 {
     pugi::xml_node pathChild = AddChild("path");
     pathChild.append_attribute("d") = StringFormat("M%d %d L%d %d", x1, y1, x2, y2).c_str();
-    if (m_penStack.top().HasColor()) {
+    if (m_penStack.top().HasColor() || this->UseGlobalStyling()) {
```
lpugin, line 817: This does not make sense to me. It seems it would now use `m_penStack.top.GetColor()` even when `m_penStack.top.HasColor()` is `false`

rettinghaus: Your design:

> case (COLOR_NONE): return "currentColor";

lpugin: Your comment seems to imply that *my* design is bad. Would you be able to explain why, and how you would change it? That would be helpful.

rettinghaus: I'm not feeling very chatty today, sorry. Your design is not bad. 
When `--mm-output` is used `UseGlobalStyling` is set to false and we need to add `@stroke`. 

So looking again I see the problem.

```suggestion
    if (m_penStack.top().HasColor() || !this->UseGlobalStyling()) {
```

Do you have a better solution?

rettinghaus (approved):

PR #3984, opened 2025-03-07
@type="score" can use tokens rather than full string

DILewis:
Use regex to deal with cases like `@type="score highlight green"`


src/iomei.cpp
```diff
@@ -8349,9 +8350,12 @@ bool MEIInput::ReadXMLComment(Object *object, pugi::xml_node element)
 
 bool MEIInput::IsAnnotScore(pugi::xml_node annot)
 {
-    // If the annotation is a score annotation, it'll have @type="score" (we can also guess)
+    // If the annotation is a score annotation, it'll have @type="score"
+    // or at least @type as a set of tokens, one of which is 'score'.
+    // N.B. Future versions of MEI might have a dedicated attribute for this
     std::string value = annot.attribute("type").value();
-    return (value == "score");
+    std::regex scoreRegex("(^|\\s)score($|\\s)");
```
paul-bayleaf, line 8357: Recommend using static const std::regex for efficiency, as compiling regex is slow.  See https://github.com/rism-digital/verovio/pull/3893

DILewis:
`static const` added (n.b. this is probably still not optimal for the number of times it's run, but I hope the use of `@type` for annotations is temporary, at which point, this can be improved)

rettinghaus (approved):

PR #4000, opened 2025-03-17
Simplify and split apart checks perform when building the tree

lpugin:
The PR is an internal refactoring with no change expected. 

It checks if a child is allowed for an element based on the ClassId, and removes unnecessary `dynamic_cast` asserts. This is done through a refactoring of `Object::IsSupportedChild` method.

Some additional checks performed before in some overridden version of the methods are now in a new `Object::AddChildAdditionalCheck`

The PR also removes an old temporary fix for https://github.com/MeasuringPolyphony/mp_editor/issues/62.


lpugin:
See https://github.com/rism-digital/verovio/blob/0eb2addc9ac184d52c43865652c1710235919f97/src/object.cpp#L836-L842

You should add a `Measure` object to the section but construct it with `Measure(UNMEASURED)`, and then add the `Staff` to it. When exporting the MEI, there will be not `<measure>` object.

lpugin:
@craigsapp, any chance to look at this? Thanks!

craigsapp:
Do you mean check if it works in the MP Editor, or within a develop-humdrum context (as in VHV)?  I have been using it in VHV for a little bit and have not noticed any problems.

lpugin:
I guess the is or used to be one place in the Humdrum importer in Verovio that does (or used to) add as `Staff` to a `Section`, but I could not find it. So I assume it has been removed and we can go ahead. If by any chance, this breaks the MP editor, we can fix it in Verovio directly.

craigsapp:
I am pretty sure that it will not affect the MP editor.   And I will complain if I otherwise discover any problems. 😉

PR #4004, opened 2025-03-19
remove double semicolons

rettinghaus:
Spring cleaning for @craigsapp

### DIFF
--- a/src/hum/humlib.cpp
+++ b/src/hum/humlib.cpp
@@ -3063,7 +3063,7 @@ int Convert::kernToBase7(const string& kerndata) {
 		return diatonic;
 	}
 	int octave = Convert::kernToOctaveNumber(kerndata);
-	return diatonic + 7 * octave;;
+	return diatonic + 7 * octave;
 }
 
 
@@ -5585,7 +5585,7 @@ void Convert::makeBooleanTrackList(vector<bool>& spinelist,
 	}
 	fill(spinelist.begin(), spinelist.end(), false);
 
-   string buffer = spinestring;;
+   string buffer = spinestring;
 	vector<string> entries;
 	string separator = "[^\\d\\$-]+";
    HumRegex hre;
@@ -19830,7 +19830,7 @@ string HumdrumFileBase::getUriToUrlMapping(const string& uri) {
 			// no files in root directory, but no reperoties either
 			repertoryQ = true;
 		}
-		string output = "http://";;
+		string output = "http://";
 		output += "kern.ccarh.org";
 		output += "/data?";
 		if (repertoryQ) {
@@ -21757,7 +21757,7 @@ bool HumdrumFileBase::analyzeLinks(void) {
 			}
 		}
 	}
-	return isValid();;
+	return isValid();
 }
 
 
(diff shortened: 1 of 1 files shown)


craigsapp:
Thanks!   Although this is a downstream file from the humlib repository, https://github.com/craigsapp/humlib, that is generated automatically from the source code files:

https://github.com/craigsapp/humlib/tree/master/src
https://github.com/craigsapp/humlib/tree/master/include

So you or I could remove the empty statements in that repository in the separate files that then get merged into the single file `humlib.cpp` and `humlib.h` files that then get copied into verovio.

In other words, it is best not to do this merge as the changes will be lost the next time I copy humlib into verovio.

-------------------------------

I know why you checked:

Screenshot 2025-03-19 at 06 13 35

This was brought to my attention the last time I merged, where that specific change generated a merge conflict that could not automatically be resolved, which is strange since I did not change that file in `develop-humdrum`.  All of the other conflicts (which I also did not change as well and came from `develop` were merged automatically).

Usually I merge `develop` into `develop-humdrum` with these commands:

```bash
git checkout develop
git pull origin develop
git checkout develop-humdrum
git merge develop
```

After a lengthly discussion with ChatGPT, it suggested:

```bash
git checkout develop
git pull origin develop  # Ensure you have the latest develop
git checkout develop-humdrum
git pull origin develop-humdrum  # Ensure you have the latest develop-humdrum
git merge -X theirs develop  # Prefer 'develop' changes for most files
git checkout develop-humdrum -- src/iohumdrum.cpp include/vrv/iohumdrum.h  # Restore protected files
git add src/iohumdrum.cpp include/vrv/iohumdrum.h
git commit --no-edit  # Commit without changing the default merge message
```

Although that does not seem ideal if there are cases where `src/iohumdrum.cpp` or `include/vrv/iohumdrum` are changed in a different branch, since this would overwrite those changes.

ChatGPT also suggested:

> If develop often gets merged from feature branches, consider rebasing develop-humdrum instead of merging.
>
> `git rebase develop`
>
> This will replay `develop-humdrum`'s changes on top of the latest `develop`, eliminating merge conflicts caused by historical divergence.
>
> If conflicts occur during rebasing, resolve them and continue with:
> `git rebase --continue`

But:

> Running `git rebase develop` in `develop-humdrum` only changes `develop-humdrum`, not `develop`. If you later merge or force-push it into develop, then develop may be affected.

Are you typically using `rebase` instead of `merge` in your branches (and what about @lpugin or others)?  I am wondering about the second sentence in the above quote.   Perhaps this is the cause of the problem since the history of `develop` can be changed by a rebase, and then when I merge `develop` into `develop-humdrum`, it has to do something more complex due to changes in the history of `develop`.

For this case it was not a problem since it was one character, but last month I had to do lots of merge conflicts manually from a `develop` merge that originated from `martha-thomae/scoringup_branch`.  In complicated cases where I can't tell which conflict is the correct one to keep, I copy the contents of the file from the current `develop` branch version on the github webpage for that file.

lpugin:
I usually try to rebase, but when there are some conflicts I prefer to do a merge.

@craigsapp shall I close this PR, or wait before merging?

craigsapp:
I recomment not merging this pull request, and I will fix the double `;;` from the humlib repotory (otherwise these fixes will be overwritten eventually).

craigsapp:
Fixed in issue https://github.com/rism-digital/verovio/pull/4004

PR #4009, opened 2025-03-26
GetClassName to lowerCamelCase and change ID generator to random first letter

lpugin:
The `Object::GetClassName` is used mostly for logging purposes. For classes implementing an MEI element, the value returned by the method typically matches the MEI element name, except for the initial capital letter (e.g., `Note` for `note` or `ScoreDef` for `scoreDef`).

There was and is no good reason to have an initial capital letter, except to match exactly the class name. However, there are a few places where the class name is also used to match the MEI element name (e.g., in the SVG output), and a conversion to the initial letter to lower case is needed. So it actually makes more sense to match the MEI element name (with the lower case initial), an this PR changes this.

The only place where this change might have an impact is in the Humdrum importer, so @craigsapp should have a close look at it. See for example https://github.com/rism-digital/verovio/blob/4a996c1736b4a7d6aa3e88a2b771ca02ee3d884e/src/iohumdrum.cpp#L5522C1-L5525C10

For such cases, it would actually be better to use the ClassId. So instead doing:
```cpp
ClassId classId = obj->GetClassId();
if (classId != STAFFGRP) {
    continue;
}
```

@craigsapp how do you want to proceed?

### DIFF
--- a/include/vrv/abbr.h
+++ b/include/vrv/abbr.h
@@ -28,7 +28,7 @@ public:
     virtual ~Abbr();
     Object *Clone() const override { return new Abbr(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Abbr"; }
+    std::string GetClassName() const override { return "abbr"; }
     ///@}
 
 private:
--- a/include/vrv/accid.h
+++ b/include/vrv/accid.h
@@ -45,7 +45,7 @@ public:
     virtual ~Accid();
     Object *Clone() const override { return new Accid(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Accid"; }
+    std::string GetClassName() const override { return "accid"; }
     ///@}
 
     /** Override the method since it is align to the staff */
--- a/include/vrv/add.h
+++ b/include/vrv/add.h
@@ -28,7 +28,7 @@ public:
     virtual ~Add();
     Object *Clone() const override { return new Add(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Add"; }
+    std::string GetClassName() const override { return "add"; }
     ///@}
 
 private:
--- a/include/vrv/anchoredtext.h
+++ b/include/vrv/anchoredtext.h
@@ -33,7 +33,7 @@ public:
     virtual ~AnchoredText();
     Object *Clone() const override { return new AnchoredText(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "AnchoredText"; }
+    std::string GetClassName() const override { return "anchoredText"; }
     ///@}
 
     /**
--- a/include/vrv/annot.h
+++ b/include/vrv/annot.h
@@ -32,7 +32,7 @@ public:
     // This fails because of the copy contructor in ObjectListInterface (TextListInterface parent)
     // Object *Clone() const override { return new Annot(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Annot"; }
+    std::string GetClassName() const override { return "annot"; }
     ///@}
 
     /**
--- a/include/vrv/app.h
+++ b/include/vrv/app.h
@@ -28,7 +28,7 @@ public:
     virtual ~App();
     Object *Clone() const override { return new App(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "App"; }
+    std::string GetClassName() const override { return "app"; }
     ///@}
 
     /** Getter for level **/
--- a/include/vrv/arpeg.h
+++ b/include/vrv/arpeg.h
@@ -39,7 +39,7 @@ public:
     virtual ~Arpeg();
     Object *Clone() const override { return new Arpeg(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Arpeg"; }
+    std::string GetClassName() const override { return "arpeg"; }
     ///@}
 
     /**
--- a/include/vrv/artic.h
+++ b/include/vrv/artic.h
@@ -37,7 +37,7 @@ public:
     virtual ~Artic();
     Object *Clone() const override { return new Artic(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Artic"; }
+    std::string GetClassName() const override { return "artic"; }
     ///@}
 
     /** Override the method since alignment is required */
--- a/include/vrv/barline.h
+++ b/include/vrv/barline.h
@@ -43,7 +43,7 @@ public:
     virtual ~BarLine();
     Object *Clone() const override { return new BarLine(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "BarLine"; }
+    std::string GetClassName() const override { return "barLine"; }
     ///@}
 
     /** Override the method since alignment is required */
--- a/include/vrv/beam.h
+++ b/include/vrv/beam.h
@@ -292,7 +292,7 @@ public:
     virtual ~Beam();
     Object *Clone() const override { return new Beam(*this); }
     void Reset() override;
-    std::string GetClassName() const override { return "Beam"; }
+    std::string GetClassName() const override { return "beam"; }
     ///@}
 
     /**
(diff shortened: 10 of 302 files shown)


craigsapp:
That sounds good, and it will be much more efficient changing from `GetClassName()` to `GetClassId()`:

So the first one I find:

```cpp
        std::string name = obj->GetClassName();
        if (name != "StaffGrp") {
            continue;
        }
```

will now become:

```cpp
        Classid classId = obj->GetClassId();
        if (classId != "STAFFGRP) {
            continue;
        }
```

or

```cpp
        if (obj->GetClassId() != STAFFGRP) {
            continue;
        }
```

And when I really need the text form of the class ID, it will still be available (now with a lower case starting letter):

```cpp
    std::string id = object->GetClassName();
    std::transform(id.begin(), id.end(), id.begin(), ::tolower);
```

Here I convert them all to lower case, so no change will be needed.

I will do those updates now and commit to the `develop-humdrum` branch after which I will submit a pull request into `develop`.




`

craigsapp:
> but the overhead was too much for this at it was implemented. There is certainly a better way to implement this if we really want it back, but I am not sure it is necessary.

Probably not necessary.  But here are suggestions for implementing `GetClassName()` as a `static const string`, with `GetClassName()` returning a `const string&` (so no string creation or copying needed at run time).


Simple case:

```cpp
#include <iostream>
#include <string>

class MyClass {
public:
    const std::string& GetClassName() const {
        return m_className;
    }

private:
    static const std::string m_className;
};

const std::string MyClass::m_className = "myClassName";

int main() {
    MyClass obj;
    std::cout << "Class name: " << obj.GetClassName() << std::endl;
    return 0;
}
```

There is a single storage location for the class name shared across all MyClass objects.

How to do this with an inherited base class would be a bit more complex, since the static string could not be stored in Object itself, but in the classes that inherit it.

```cpp
#include <iostream>
#include <string>

// Base class with pure virtual method
class Object {
public:
    virtual ~Object() = default;

    // Pure virtual method
    virtual const std::string& GetClassName() const = 0;
};

class Note : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Note::m_className = "Note";

class Rest : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Rest::m_className = "Rest";

int main() {
    Note obj1;
    Rest obj2;

    std::cout << "Object 1 is a: " << obj1.GetClassName() << std::endl;
    std::cout << "Object 2 is a: " << obj2.GetClassName() << std::endl;

    return 0;
}
```

Or when working with polymorphism:

```cpp
#include <iostream>
#include <string>
#include <vector>
#include <memory>  // for std::unique_ptr

class Object {
public:
    virtual ~Object() = default;
    virtual const std::string& GetClassName() const = 0;
};

class Note : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Note::m_className = "Note";

class Rest : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Rest::m_className = "Rest";

int main() {
    std::vector<std::unique_ptr<Object>> objects;

    objects.emplace_back(std::make_unique<Note>());
    objects.emplace_back(std::make_unique<Rest>());
    objects.emplace_back(std::make_unique<Note>());

    for (const auto& obj : objects) {
        std::cout << "Class name: " << obj->GetClassName() << std::endl;
    }

    return 0;
}
```

Which should output:

```tsv
Class name: Note
Class name: Rest
Class name: Note
```

craigsapp:
And an example that does not use smart pointers (manual memory allocation with `new` and `delete`):

```cpp
#include <iostream>
#include <string>
#include <vector>

class Object {
public:
    virtual ~Object() = default;
    virtual const std::string& GetClassName() const = 0;
};

class Note : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Note::m_className = "Note";

class Rest : public Object {
public:
    const std::string& GetClassName() const override {
        return m_className;
    }
private:
    static const std::string m_className;
};
const std::string Rest::m_className = "Rest";

int main() {
    std::vector<Object*> objects;

    objects.push_back(new Note());
    objects.push_back(new Rest());
    objects.push_back(new Note());

    for (const auto* obj : objects) {
        std::cout << "Class name: " << obj->GetClassName() << std::endl;
    }

    for (auto* obj : objects) {
        delete obj;
    }

    return 0;
}
```

craigsapp:
A static map would be `O(log N)` compared to the virtual static string method which would be `O(1)`, so much more efficient to use the later, particularly as `N` increases.

craigsapp:
Here is a variant of the implementation that moves the static variable to `GetClassName()` so that the variable does not clutter up the class definition:

```cpp
#include <iostream>
#include <string>
#include <vector>

class Object {
public:
    virtual ~Object() = default;
    virtual const std::string& GetClassName() const = 0;
};

class Note : public Object {
public:
    const std::string& GetClassName() const override {
        static const std::string name = "Note";
        return name;
    }
};

class Rest : public Object {
public:
    const std::string& GetClassName() const override {
        static const std::string name = "Rest";
        return name;
    }
};

int main() {
    std::vector<Object*> objects;

    objects.push_back(new Note());
    objects.push_back(new Rest());
    objects.push_back(new Note());

    for (const auto* obj : objects) {
        std::cout << "Class name: " << obj->GetClassName() << std::endl;
    }

    for (auto* obj : objects) {
        delete obj;
    }

    return 0;
}
```

Static class members is the standard way of sharing data across all class members and functions like a global variable only accessible by instances of the class, so I don't see any problem with it as there is no other good way of doing it (other than an external global variable which would be worse).

Similar, but not as efficient would be to output a `string` copy rather than a `const string&`:

```cpp
#include <iostream>
#include <string>
#include <vector>

class Object {
public:
    virtual ~Object() = default;
    virtual std::string GetClassName() const = 0;
};

class Note : public Object {
public:
    std::string GetClassName() const override {
        return "Note";
    }
};

class Rest : public Object {
public:
    std::string GetClassName() const override {
        return "Rest";
    }
};

int main() {
    std::vector<Object*> objects;

    objects.push_back(new Note());
    objects.push_back(new Rest());
    objects.push_back(new Note());

    for (const auto* obj : objects) {
        std::cout << "Class name: " << obj->GetClassName() << std::endl;
    }

    for (auto* obj : objects) {
        delete obj;
    }

    return 0;
}
```

This would not need to involve constructing a class name since it is already available at compile time.

craigsapp:
Here is a more generalized solution, but not very practical.  The name of the class can be accessed from `typeid(obj).name()` which typically returns the name of the class, but it will typically be mangled in C++:

```cpp
#include <iostream>
#include <typeinfo>

class MyClass {
    // Class definition
};

int main() {
    MyClass obj;
    std::cout << "Class name: " << typeid(obj).name() << std::endl;
    return 0;
}
```

which in C++ returns `7MyClass`, which is a mangled version of the class name.   This is not portable due to the mangling, since demangling is not standard in C++.

Here is an example that should work in g++/clang and Windows:

```cpp
#include <iostream>
#include <typeinfo>
#include <string>

#ifdef _WIN32
    #include <windows.h>
    #include <dbghelp.h>
    #pragma comment(lib, "dbghelp.lib")
#else
    #include <cxxabi.h>
    #include <cstdlib>
    #include <memory>
#endif

// Demangle function
std::string demangle(const char* mangled_name) {
    #ifdef _WIN32
        char demangled_name[1024];
        if (UnDecorateSymbolName(mangled_name, demangled_name, sizeof(demangled_name), UNDNAME_COMPLETE)) {
            return std::string(demangled_name);
        } else {
            return std::string(mangled_name); // Return original if demangling fails
        }
    #else
        int status = 0;
        std::unique_ptr<char, void(*)(void*)> res{
            abi::__cxa_demangle(mangled_name, nullptr, nullptr, &status),
            std::free
        };
        return (status == 0) ? res.get() : mangled_name;
    #endif
}

class MyClass {
public:
    virtual ~MyClass() = default;
};

int main() {
    MyClass obj;
    std::cout << "Demangled name: " << demangle(typeid(obj).name()) << std::endl;
    return 0;
}
```

Output:

```tsv
Demangled name: MyClass
```

PR #4012, opened 2025-03-28
Change GetClassName to std::string_view

lpugin:
According to ChatGPT a string_view would be a better option:

image


lpugin:
Wrongly pushed to develop, I'll revert it if needed

craigsapp:
That is interesting as I was not aware of `std::string_view`.  Soon we can get ChatGPT (Deep Seek, etc.) to write verovio code 😜 (`std::string_view` was added to C++17 according to ChatGPT).

PR #4016, opened 2025-03-30
Change Object::GetClassName() to std::string_view

lpugin:
This PR optimises the `Object::GetClassName` with the use of `std::string_view`. No change in behaviour expected.

Also fixes a missing file in the xcode project.

### DIFF
--- a/Verovio.xcodeproj/project.pbxproj
+++ b/Verovio.xcodeproj/project.pbxproj
@@ -311,6 +311,7 @@
 		4D543E221B80AACF004B823C /* view_control.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 4D543E211B80AACF004B823C /* view_control.cpp */; };
 		4D5572401CF3F32A008D06A0 /* octave.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 4D55723F1CF3F32A008D06A0 /* octave.cpp */; };
 		4D5572441CF57B5C008D06A0 /* pedal.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 4D5572431CF57B5C008D06A0 /* pedal.cpp */; };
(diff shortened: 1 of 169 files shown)


paul-bayleaf:
Using a string_view::data() in a function that expects a null terminated string as in

   LogDebug("Something was wrong when searching positioners for %s '%s'", harm->GetClassName().data(),

only works because the string_view in question is known to have been constructed from a string literal.  However, the null is not part of the string_view's range so this use requires dereferencing the null byte outside the range of the string_view.   String_views are not, in general, null terminated, which is why string_view does not provide a c_str() function.

From the C++17 draft

© ISO/IEC N4659
14 [ Note: Unlike basic_string::data() and string literals, data() may return a pointer to a buffer that
is not null-terminated. Therefore it is typically a mistake to pass data() to a function that takes just a const charT* and expects a null-terminated string. — end note 

May I suggest that although the above example will work, I feel it is a risky idiom and out of spec for string_view so is best  avoided.

lpugin:
Thanks for the comment. Yes, I was aware of the difference between `.data` and `.c_str` but was considering it ok since it always comes from the method.

Improving this is not a really needed, so I am going to keep `std::string`.

brdvd:
I still like the idea as it avoids unnecessary string copies. As pointed out by @paul-bayleaf , the only problem seems to be the usage of `GetClassName` in the Log functions. This could be resolved by implementing logging with C++20 string interpolation via `std::format`. I had this on my list, but did not find the time to work on it.

Here is a small example where a `std::string_view` is plugged into a message:

```cpp
#include <iostream>
#include <format>
#include <string_view>

std::string_view getGreeting() {
    return "Hello, World!";
}

int main() {
    std::string_view greeting = getGreeting();
    
    // Use std::string_view with std::format
    std::string formatted = std::format("Greeting message: {}", greeting);
    
    std::cout << formatted << std::endl;

    return 0;
}
```

It would require quite some work, as we need to touch all the Log messages and replace them with the new interpolation syntax. I would be quite happy to have this in Verovio at some point in the future.

PR #4022, opened 2025-04-04
Add g clef to neume notation for neon

yinanazhou:
- Added E900 Bravura font
- Added editor support for g clef in neon

refs: https://github.com/DDMAL/Neon/issues/1308


lpugin:
Looks good. However, the clef also need to be added to the Leipzig font. I'll do that.

Issue #3983, opened 2025-03-06
Add ledger line reference to note


yinanazhou:
As shown in the screenshot below, the highlighted note is on a ledger line:

![Image](https://github.com/user-attachments/assets/dc0d90fa-41ac-4d35-b6ab-8a6f59ee42ac)

But when I hide the note, it would be better to have a reference to the ledger line so that I can hide the ledger line as well:

![Image](https://github.com/user-attachments/assets/506a25fc-440f-4c8d-8a2d-956eafddccde)

Refs: https://github.com/DCMLab/reductive_analysis_app/issues/16

lpugin:
You need to use `--svg-html5`. The reason for which this option is required is because of the use of the `data-*` attribute for the reference.

ahankinson:
More background on how this can be fixed here:

https://github.com/rism-digital/verovio/pull/3861

yinanazhou:
Thanks @ahankinson and @lpugin. I'm working with an app that was developed a few years ago. Using --svg-html5 would require significant changes to switch from `id` to `data-id`. I’d also prefer not to modify the MEI file, so the @visible approach isn’t ideal either. What I’m looking for is more like an option to include `data-related` without using `data-id`. Do you have any suggestions?

lpugin:
@kepper, any ideas?

PR #4028, opened 2025-04-14
fix: add note reference to ledger lines as default

yinanazhou:
- Removed `svgHtml5` condition

refs: https://github.com/rism-digital/verovio/issues/3983, https://github.com/DCMLab/reductive_analysis_app/issues/16

### DIFF
--- a/src/calcledgerlinesfunctor.cpp
+++ b/src/calcledgerlinesfunctor.cpp
@@ -81,7 +81,7 @@ void CalcLedgerLinesFunctor::CalcForLayerElement(
         left -= width / 2;
     }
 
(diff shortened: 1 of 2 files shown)


lpugin:
As pointed out in #3983 this is not really an option. From what I understand from what you need the proper way to fix your issue is not to have a ledger line when the not is hidden (with `@visible="false"`). That should be pretty straight forward. Would this work for you?

lpugin:
I just double checked, and making a note invisible with `@visible="false"` makes the ledger lines disappear. I know this is not what is wanted by this PR, but just to confirm that hiding the ledger lines in `per se` not an issue.

I am closing the PR since there is not clear way to adjust it.

PR #4029, opened 2025-04-15
Add core functionality to parse, render, view score annotations

DILewis:
We take `<annot type="score">` as denoting an annotation about the music (rather than the encoding). This is parsed into a different object – a score annotation – which is treated like a control event. 

Score annotations that affect a region of the score are drawn as a ribbon above affected staves; if the annotation targets a set of elements (with `@plist`), the elements are given a new attribute to point to the annotation (and can be styled with CSS).

Score annotations (like annotations) are also rendered as `g` objects themselves, which list any elements enumerated in `@plist`.

This way, developers of client applications have a range of choices for building visualisations or interactions.


lpugin:
Please try adding `#include <optional>` to object.h. Hopefully that will fix appveyor failing.

I am checking what is happening with the test-suite. If you could drop a minimal explanation to the PR that would be nice because it is touching quite a few things. Thanks!

lpugin:
Nearly there! Only the formatting to fix

ahankinson:
Would it make sense to have an option that you pass to `renderOptions` to en/disable the rendering of annotations?

lpugin:
I think there is still quite much to discuss on how to render (show, style, etc.) them. I would suggest to have that in a separate PR once we are clearer with what we need so we don't had too many options. 

Also, the `@type="score"` is really something that should go away but that needs some adjustments in MEI. It might be better to wait until this happens. For now, removing `"score" in `@type` will hide the annotation, so that can serve as a way not to show them for now.

ahankinson:
Well, this currently feels (to me) a bit like a project-specific thing, which isn't really suitable for wider use. `@type` is not prescribed anywhere except here. It's also unclear what is meant by "this is interpreted as a control event". Given that `<annot>` can take all the time (`@tstamp`) and ID (`@startid`) as well as `@part` and `@partstaff` as targets, I don't have a good sense of what is meant by this targeting "the music"?

Would it make sense to build this as support for `<annot>` (without reducing it to only accepting `@type="score"`)?

lpugin:
`<annot>` as an annotation to the encoding is already supported by Verovio, but it does nothing with is except adding an empty `<g>` in the SVG with the corresponding ID. So this is why there was need to distinguish an annotation of the score as opposed to an annotation of the encoding, and `@type` is currently used for that.

ahankinson:
> distinguish an annotation of the score as opposed to an annotation of the encoding

Could you give an example of what the difference is between the two things? If you're targeting parts, time, staff, etc. then this seems to be an annotation of the score? If you're targeting IDs, then it's an encoding?

lpugin:
I guess the distinction is more for `annot` to act like a TEI [note](https://www.tei-c.org/Vault/P5/4.1.0/doc/tei-p5-doc/en/html/ref-note.html) and `annot type="score"` like a TEI [annotation](https://www.tei-c.org/Vault/P5/4.1.0/doc/tei-p5-doc/en/html/ref-annotation.html) (e.g., a Web Annotation).

include/vrv/annotscore.h
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @name Constructors, destructors, and other standard methods
+     * Reset method reset all attribute classes
+     */
+    ///@{
+    AnnotScore();
+    virtual ~AnnotScore();
+    // Object *Clone() const override { return new AnnotScore(*this); }
+    void Reset() override;
+    std::string GetClassName() const override { return "annotScore"; }
```
lpugin, line 38: Element name should still be `annot`
```suggestion
    std::string GetClassName() const override { return "annot"; }
```

ahankinson:
i like the idea of adding an `@motivation` to `<annot>` straight from TEI: https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-annotation.html . Particularly, I like the fact that the values map to the Web Annotation spec. 

I can open a ticket on the MEI repo if there's broad consensus that this is a good idea here. @DILewis ?

DILewis:
I've also (unsurprisingly) been mulling `@motivation` for a while. It has the advantage of mapping trivially into web annotations, which is helpful (especially for mei-friend, which supports both). 

TEI has a controlled vocabulary of just the core oa motivations. which I think probably wouldn't work for us. If we adopt it, we'll need to think hard about what terms we need. It's possible that we might need a controlled and an uncontrolled attribute? Or to use `@motivation` for pre-defined and `@type` for project-specific motivations.

I'm also not sure whether the distinction Laurent identified (between annotation of encoding and annotation of music) is a motivation. I _think_ it is, but I'd appreciate the input of others.

ahankinson:
> which I think probably wouldn't work for us

Why not? Any use case experiences are useful to capture.

> `@type` for project-specific motivations

The problem with overloading `@type` is that processors have no way of knowing what kind of data to expect there since it's open-ended. If there's a more specific attribute that we can define, that is much better, even if we go with an open list.

For example, `@motivation.auth` could also be added to allow custom annotation vocabularies, with a URI specifying the vocabulary in use? That way it's not limited to just the Web Annotation spec, but it can be varied from project to project.

> I'm also not sure whether the distinction Laurent identified (between annotation of encoding and annotation of music) is a motivation ...

We chatted about this. As I see it, the problem boils down to that `<annot>` can be used within `<notesStmt>` in the header, or it can be used within the musical text itself.

Fundamentally, it seems that the problem is that a "`<note>`" element (in the textual sense as a child of "`<notesStmt>`") is not possible in MEI for obvious reasons, so another name needed to be found, and that name was `<annot>`. However, distinguishing between "annot as file note" and "annot as music annotation" in the different contexts is not straightforward. That's my interpretation of what @lpugin describes as an "encoding" vs a "musical" annotation.

There could be an expectation that `<annot>` with `@motivation` can generally only appear as a "musical" annotation?

Or there could be the "baby with the bathwater" option, and solve the problem of what is a good element name for a child in a "notesStmt". For example, `<annot>` could be removed as a child of `<notesStmt>` in favour of a simple `<p>`. Then it becomes exclusively a "musical" annotation.

DILewis:
> > which I think probably wouldn't work for us
> 
> Why not? Any use case experiences are useful to capture.

The Web Annotation model set is meant to be comprehensive, but generic. Depending on the use cases we need to satisfy, I wouldn't be surprised if we (being MEI, not necessarily Verovio) need some more specific ones. In the original standard, this is achieved by defining new `oa:Motivation`s with `skos:broader` linking them to one or more existing motivations. That's not available once we get into TEI/MEI, so maybe I overstated certainty, but we _may_ need a broader palette.

I should also say that mei-friend currently uses existing Motivations in a way that pretty directly relates them to body type (i.e. describing=text, linking=link, highlighting=nothing). That's kind of implied by the vocabulary itself, but seems to be a sign of a slight problem with the modelling (if it's just about the annotation body type and how to display it, it's not purely a motivation)

> > `@type` for project-specific motivations
> 
> The problem with overloading `@type` is that processors have no way of knowing what kind of data to expect there since it's open-ended. If there's a more specific attribute that we can define, that is much better, even if we go with an open list.

Agreed. `@type` is currently just being used as a hack for building something tangible without changing the schema. It shouldn't stay.
 
> For example, `@motivation.auth` could also be added to allow custom annotation vocabularies, with a URI specifying the vocabulary in use? That way it's not limited to just the Web Annotation spec, but it can be varied from project to project.

I like that (but then I would, because it's a LOD way to do things). It gives us space to try things out without fiddling with the schema each time.

> > I'm also not sure whether the distinction Laurent identified (between annotation of encoding and annotation of music) is a motivation ...
> 
> We chatted about this. As I see it, the problem boils down to that `<annot>` can be used within `<notesStmt>` in the header, or it can be used within the musical text itself.
> 
> Fundamentally, it seems that the problem is that a "`<note>`" element (in the textual sense as a child of "`<notesStmt>`") is not possible in MEI for obvious reasons, so another name needed to be found, and that name was `<annot>`. However, distinguishing between "annot as file note" and "annot as music annotation" in the different contexts is not straightforward. That's my interpretation of what @lpugin describes as an "encoding" vs a "musical" annotation.

There are a few other places where `<annot>` is used where it may not mean the same thing. I'm not expert enough to be certain. For example, in the `<bibl>` element. I guess there's a question of whether an element can or should take its semantics from the element class(es) of its parent or target(s)?

> There could be an expectation that `<annot>` with `@motivation` can generally only appear as a "musical" annotation?

Maybe? If it wouldn't be useful for the note-like case. An alternative might be a list of acceptable motivations for each, but that might also mean new `oa:Motivation` instances.

> Or there could be the "baby with the bathwater" option, and solve the problem of what is a good element name for a child in a "notesStmt". For example, `<annot>` could be removed as a child of `<notesStmt>` in favour of a simple `<p>`. Then it becomes exclusively a "musical" annotation.

Again, depends on whether that works for the other cases (or whether those other cases are 'proper' `<annot>`). But yes, I have no problem with that approach. I also don't know how widely these are used.

ahankinson (approved):
Approved with reservations as noted.

PR #4032, opened 2025-04-19
feat: update changelog to follow keepachangelog

rettinghaus:
This updates the changelog to be compliant with <https://keepachangelog.com/en/1.1.0/>, at least back to version 4.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,628 +1,759 @@
 # Changelog
 
-## [unreleased]
-* Support for `data.HEXNUM` in `note@head.shape`
-* Support for additional tablature features introduced in MEI 5.1 (@paul-bayleaf)
-* Improved support for expansion lists (@wergo)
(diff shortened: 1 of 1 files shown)


lpugin:
I appreciate the effort but I am not convinced this is something useful for this project. We have a release about every two months, with a changelog of usually five to twenty lines, rarely more. They are typically listed from the most important additions to the small fixes, so it is quite quick and easy to have a full overview of what has been done. The suggestion is over the top.

The main problem with the proposed change are the categories, especially between `Added` and `Changed`. It will be often tricky to decide in which category things should be list, which makes them pointless. For example, you listed `Improved timemap with grace notes and arpeggios` as a changed, but this is equally an addition. So at the end of the day, people simply need to read things...

rettinghaus:
Same problem with conventional commits. You need to decide into which category a commit falls. But maybe you're right and this is to much here.

PR #4033, opened 2025-04-21
Improve generation of `git_commit.h` to also work in submodules

henrythasler:
The current method (check for `.git` folder) in `tools/get_git_commit.sh` fails when verovio is added as a submodule to another repository. The bash script was refactored to use git [rev-parse](https://linux.die.net/man/1/git-rev-parse) to check if it's inside a git structure. Furthermore, a timestamp was added as a comment in the header-file to allow to verify that the file is up-to-date.

### DIFF
--- a/tools/get_git_commit.sh
+++ b/tools/get_git_commit.sh
@@ -5,19 +5,20 @@ cd ..
 output="./include/vrv/git_commit.h"
 COMMIT=""
 
-if [ ! -d "./.git" ]; then
-     echo "This is not a git directory and the git commit sha will remain undefined"
-else
+if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
     SHA=$(git describe  --exclude '*' --abbrev=7 --always --dirty)
     if [ -z "$SHA" ]; then
         echo "Undefined git commit version"
     else
         COMMIT="-$SHA"
     fi
+else
+    echo "This is not a git directory and the git commit sha will remain undefined"
 fi
 
 echo "////////////////////////////////////////////////////////" > $output
 echo "/// Git commit version file generated at compilation ///" >> $output
+echo "/// Timestamp: $(date +"%Y-%m-%dT%H:%M:%S%:z")             ///" >> $output
 echo "////////////////////////////////////////////////////////" >> $output
 echo "" >> $output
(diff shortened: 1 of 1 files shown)


tools/get_git_commit.sh
```diff
@@ -5,19 +5,20 @@ cd ..
...
     else
         COMMIT="-$SHA"
     fi
+else
+    echo "This is not a git directory and the git commit sha will remain undefined"
 fi
 
 echo "////////////////////////////////////////////////////////" > $output
 echo "/// Git commit version file generated at compilation ///" >> $output
+echo "/// Timestamp: $(date +"%Y-%m-%dT%H:%M:%S%:z")             ///" >> $output
```
ahankinson, line 21:
```suggestion
echo "/// Timestamp: $(date -u +"%Y-%m-%dT%H:%M:%SZ")             ///" >> $output
```

`%:z` is not correct, since this outputs a literal `:z`. `%Z` will output the timezone (CEST); `date -u` will output the time in UTC. 

Just the letter `Z` at the end will print the literal, which is Zulu time, or the same as UTC. So:

`date -u +"%Y-%m-%dT%H:%M:%SZ"` -> `2025-04-22T11:39:53Z`
`date -u +"%Y-%m-%dT%H:%M:%S%Z" -> `2025-04-22T11:39:53UTC`

I think the suggested format is what is intended.

henrythasler: My idea was to use ISO 8601 date format (see [this comment](https://unix.stackexchange.com/a/629504)) but using UTC is also fine for me.

ahankinson: The SO post said that the date command on macOS doesn’t support that

henrythasler:
Regarding this being functionally equivalent: It seems it's different when you add a repo as submodule because then `.git` is just a file whereas it's a folder if you clone it separately. And if it's a file the current [Bash Conditional Expression](https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions) fails. Using my proposed solution fixes this problem.

image

I have no explanation why git works this way...

henrythasler:
Alignment in the generated file looks good to me:
image

lpugin (approved):

PR #4045, opened 2025-05-05
Add #define to drop editor code

lpugin:
For now the code remains included in all builds by default since the definition is OFF. We might change this in the future if the editor code grows. @mei-friend: do you use the `Toolkit::Edit` function at all?


wergo:
Hi @lpugin, no we do not use Toolkit::Edit in @mei-friend at all.

PR #4047, opened 2025-05-07
Cleanup for output / input classes and functors and in-progress page-based serialization

lpugin:
This PR brings some code clean-up for the SaveFunctor and the Output / Input classes:
* Add a Output::Skip virtual method to keep the SaveFunctor agnostic of the type of output
* Add a Output::Export methods to unify implementation (and remove confusing MEIOutput::GetOutput method)

It also introduces a serialization of the internal page-based MEI, which is basically the page-based MEI together with the hidden elements (editorial markup and mdivs). This will eventually offer way to cache laid-out data for applications that needs it, in which case it would be good to adjust the Verovio customization in order to be able to validate it.

For now, this cannot but used directly from the Toolkit class, but we can make this happen with a dedicated `inputFrom` and `outputTo` option value (`mei-pb-internal`?) , which I will do in a follow-up PR. This will also allow to add round-trip tests for it.


include/vrv/iomei.h
```diff
@@ -252,6 +251,7 @@ class MEIOutput : public Output {
     ///@{
     bool IsScoreBasedMEI() const { return m_scoreBasedMEI; }
     bool IsPageBasedMEI() const { return !m_scoreBasedMEI; }
+    bool IsSerializingMEI() const { return m_serializing; }
```
rettinghaus, line 254: The naming is confusing, maybe better `IsSerialization` or `IsSerialized`?

include/vrv/iomei.h
```diff
@@ -264,6 +264,11 @@ class MEIOutput : public Output {
      */
     void SetIgnoreHeader(bool ignoreHeader) { m_ignoreHeader = ignoreHeader; }
 
+    /**
+     * Setter for the page-based serialization flag (default is false)
+     */
+    void SetSerializing(bool serializing) { m_serializing = serializing; }
```
rettinghaus, line 270: Same.

include/vrv/iobase.h
```diff
@@ -46,6 +51,11 @@ class Output {
      */
     virtual bool WriteObjectEnd(Object *object) { return true; }
 
+    /**
+     * Method for skipping under certain circumstances
+     */
+    virtual bool Skip(Object *object) { return false; }
```
brdvd, line 57: I guess this is intended to be read-only. Maybe add `const` here, e.g.
```cpp
virtual bool Skip(const Object *object) const { ... }
```

src/iomei.cpp
```diff
@@ -182,9 +183,27 @@ MEIOutput::MEIOutput(Doc *doc) : Output(doc)
...
+    if (object->Is(MDIV)) {
+        VisibilityDrawingInterface *interface = object->GetVisibilityDrawingInterface();
+        assert(interface);
+        if (!interface->IsHidden() || this->IsSerializing()) return false;
+        if (this->IsPageBasedMEI() || this->HasFilter()) return true;
+    }
+    else if (object->IsEditorialElement()) {
+        VisibilityDrawingInterface *interface = object->GetVisibilityDrawingInterface();
+        assert(interface);
+        if (!interface->IsHidden() || this - IsSerializing()) return false;
```
brdvd, line 197: Interesting error which actually compiles.
```cpp
this->IsSerializing()
```

src/iomei.cpp
```diff
@@ -1589,6 +1607,10 @@ void MEIOutput::WriteMdiv(pugi::xml_node currentNode, Mdiv *mdiv)
 {
     assert(mdiv);
 
+    if (mdiv->IsHidden() && this->IsSerializing()) {
+        m_currentNode.append_attribute("verovio.serialization") = "hidden";
```
brdvd, line 1611: Maybe we can define `verovio.serialization` as a string constant somewhere. It is used a lot.

src/iomei.cpp
```diff
@@ -7801,7 +7858,7 @@ bool MEIInput::ReadAppChildren(Object *parent, pugi::xml_node parentNode, Editor
         if (first) {
             first->SetVisibility(Visible);
         }
-        else {
+        else if (!m_deSerializing && !parent->Is(SYSTEM)) {
```
brdvd, line 7861: (Minor) I would prefer the spelling `m_deserializing` as `deserializing` is a single word.

brdvd (changes requested):
Looks pretty good. There is one code typo in `MEIOutput::Skip` that needs to be fixed and a few optional improvements.

src/iomei.cpp
lpugin: Yes, this is actually quite a nasty one. I [had](https://github.com/rism-digital/verovio/pull/4047/commits/7a10c416ee6758f9838acf50b9b508df154cd079) fix one but mist this one. Thanks!

src/iomei.cpp
lpugin: I was hesitating. I'll change it, then.

src/iomei.cpp
lpugin: I'll have a define

lpugin:
Additional comment: the PR now also includes a `mei-pb-serialization` input and output option for use from the Toolkit class, which will allow for a test case to be added - to be done after the PR because otherwise it will break the current test suite evaluation.

src/toolkit.cpp
```diff
@@ -293,6 +299,9 @@ FileFormat Toolkit::IdentifyInputFrom(const std::string &data)
         // <score-timewise> == root node for time-wise organization of MusicXML data
         // <opus> == root node for multi-movement/work organization of MusicXML data
 
+        if (std::regex_search(initial, std::regex("<(verovio-serialization)[\\s\\n>]"))) {
```
brdvd, line 302: `[\\s\\n]` can be simplified as white space includes newline.

src/toolkit.cpp
```diff
@@ -518,6 +527,26 @@ bool Toolkit::LoadData(const std::string &data)
     return this->LoadData(data, true);
 }
 
+void Toolkit::SetViewAndEditor()
+{
+    m_view.SetDoc(&m_doc);
+
+#if defined NO_HUMDRUM_SUPPORT
```
brdvd, line 534: Just for my understanding: why is the editor toolkit only set, if we have no humdrum support?

src/toolkit.cpp
```diff
@@ -990,16 +1020,26 @@ std::string Toolkit::GetMEI(const std::string &jsonOptions)
     if (!mdiv.empty()) meioutput.SetMdiv(mdiv);
 
     if (generateFacs) {
-        if (meioutput.HasFilter() || !scoreBased || (m_options->m_breaks.GetValue() != BREAKS_encoded)
-            || m_doc.HasSelection()) {
+        if (meioutput.HasFilter() || m_doc.HasSelection()) {
```
brdvd, line 1023: Now the error message below does not reflect the simplified condition.

src/toolkit.cpp
lpugin: I am not sure either. Maybe there was some dependencies at some point. I have [introduced](https://github.com/rism-digital/verovio/pull/4045) a new define for the editor. I'll have a look and make this consistent.

include/vrv/iobase.h
```diff
@@ -36,6 +36,11 @@ class Output {
     Output() {}
     virtual ~Output();
 
+    /**
+     * Main method for exporting the data and to be overridden in child class
```
ahankinson, line 40:
```suggestion
     * Main method for exporting the data. Overridden in child classes.
```

ahankinson (commented):
Just one question about code re-org, otherwise looks fine.

PR #4053, opened 2025-05-12
update tasks to use up to 8 jobs in parallel

rettinghaus:
Small update to use up to 8 processes when building Verovio within VSCode.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -16,6 +16,9 @@
 			"label": "make",
 			"type": "shell",
 			"command": "make",
+			"args": [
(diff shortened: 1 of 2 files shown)


.vscode/tasks.json
```diff
@@ -16,6 +16,9 @@
 			"label": "make",
 			"type": "shell",
 			"command": "make",
+			"args": [
+				"--jobs=16"
```
lpugin, line 20: We use 8 in other places, so that is more consistent
```suggestion
				"--jobs=8"
```

PR #4057, opened 2025-05-13
Use the className first letter instead of a random letter in the ID generator

lpugin:
* Using a random letter broke the deterministic ID generation based on input CRC

PR #4061, opened 2025-05-15
Revert "Use the className first letter instead of a random letter in the ID generator"

lpugin:
Reverts rism-digital/verovio#4057


lpugin:
Breaks rendering. To be adjusted

PR #4062, opened 2025-05-15
Use Object::s_xmlIDCounter instead of std::rand for ID first letter

lpugin:
This should not break the IDs and also preserved them with `--xml-id-seed`

### DIFF
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -801,7 +801,7 @@ int Object::DeleteChildrenByComparison(Comparison *comparison)
 void Object::GenerateID()
 {
     // A random letter from a-z
-    char letter = 'a' + (std::rand() % 26);
+    char letter = 'a' + (s_xmlIDCounter % 26);
     m_id = letter + Object::GenerateHashID();
 }
 


ahankinson (approved):

src/object.cpp
```diff
@@ -801,7 +801,7 @@ int Object::DeleteChildrenByComparison(Comparison *comparison)
 void Object::GenerateID()
 {
     // A random letter from a-z
-    char letter = 'a' + (std::rand() % 26);
+    char letter = 'a' + (s_xmlIDCounter % 26);
```
craigsapp, line 804: You could remove all digits from the random value for the ID by adapting:

```
static const std::string base62Chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```

in vrv.cpp to:

```
static const std::string base52Chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```

And then using `BaseEncodeLetter(x, 26)` for lower case IDs, or `BaseEncodeLetter(x, 52)` for upper and lower case letter IDs (less desirable).

craigsapp: `char letter = 'a' + (s_xmlIDCounter % 26);` will work, but more complicated computationally 😉

PR #4064, opened 2025-05-15
add dependabot configuration for GitHub Actions

rettinghaus:
I noticed that the GitHub Actions are quite outdated, so I propose to add a dependabot config.


lpugin (approved):
OK. Let's see what this bring.

PR #3586, opened 2024-01-31
Implement additional font options

lpugin:
Allow support for `@fontfam` (currently on `clef` and `meterSig`) to use another font for these elements. To do so, the font needs to be loaded with the `--font-load-all` option.

Example:

image

<details>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
   <meiHead>
      <fileDesc>
         <titleStmt>
            <title>Clef change and spacing</title>
            <respStmt>
               <persName role="editor">Laurent Pugin</persName>
               <persName role="encoder">Craig Sapp</persName>
            </respStmt>
         </titleStmt>
         <pubStmt>
            <date isodate="2021-02-02">2021-02-02</date>
            <pubPlace>
               <ref target="https://github.com/rism-digital/verovio/issues/2018" />
            </pubPlace>
         </pubStmt>
         <seriesStmt>
            <title>Verovio test suite</title>
         </seriesStmt>
         <notesStmt>
            <annot>It is not necessary for clef changes on one staff to be allocated space on other staves </annot>
         </notesStmt>
      </fileDesc>
      <encodingDesc>
         <appInfo>
            <application version="3.2.0" label="2">
               <name>Verovio</name>
            </application>
         </appInfo>
      </encodingDesc>
   </meiHead>
   <music>
      <body>
         <mdiv>
            <score>
               <scoreDef midi.bpm="112">
                  <staffGrp bar.thru="true" symbol="brace">
                     <staffDef n="1" lines="5">
                        <clef shape="G" line="2" fontname="Petaluma" />
                        <keySig pname="a" mode="major" sig="3s" />
                        <meterSig count="3" unit="8" sym="common" fontname="Petaluma" enclose="brack" />
                     </staffDef>
                     <staffDef n="2" lines="5">
                        <clef shape="G" line="2" />
                        <keySig pname="a" mode="major" sig="3s" />
                        <meterSig count="3" unit="8" />
                     </staffDef>
                  </staffGrp>
               </scoreDef>
               <section>
                  <measure n="15">
                     <staff n="1">
[...]

PR #4068, opened 2025-05-19
Fix loading custom fonts in JS and support base64 strings

lpugin:
This PR fixes an issue with the font name not being properly detected when a custom font is passed as base64 string to the JS toolkit.

It also allows for the font to be passed as base64 string directly and not has url. So in addition to be able to do
```json
{
    "font": "GoldenAge",
    "fontAddCustom": ["https://www.verovio.org/examples/fonts/custom/GoldenAge.txt"],
    "fontFallback": "Bravura"
}
```
one can pass the content of `https://www.verovio.org/examples/fonts/custom/GoldenAge.txt` as value directly. This can be useful in [server side](https://github.com/rism-digital/verovio/pull/3586) uses.

### DIFF
--- a/emscripten/npm/src/verovio-toolkit.js
+++ b/emscripten/npm/src/verovio-toolkit.js
@@ -216,7 +216,6 @@ export class VerovioToolkit {
         let filesInBase64 = [];
         // Get all the files and convert them to a base64 string if necessary
         for ( const file of files ) {
-            const file = files[i];
             // The file in already passed as base64 string - nothing to do
             if (!/^https?:\/\//.test( file )) {
                 filesInBase64.push( file );
--- a/emscripten/npm/src/verovio-toolkit.js
+++ b/emscripten/npm/src/verovio-toolkit.js
@@ -216,7 +216,7 @@ export class VerovioToolkit {
         let filesInBase64 = [];
         // Get all the files and convert them to a base64 string if necessary
         for ( const file of files ) {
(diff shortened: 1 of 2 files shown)


emscripten/npm/src/verovio-toolkit.js
```diff
@@ -212,23 +212,28 @@ export class VerovioToolkit {
         if (!options.hasOwnProperty('fontAddCustom')) {
             return options;
         }
-        const filenames = options['fontAddCustom'];
+        const files = options['fontAddCustom'];
         let filesInBase64 = [];
-        // Get all the files and convert them to a base64 string
-        for (let i = 0; i < filenames.length; i++ ) {
+        // Get all the files and convert them to a base64 string if necessary
+        for ( let i = 0; i < files.length; i++ ) {
```
ahankinson, line 218:
```suggestion
        for ( const file of files ) {
```

Using a `for ... of` removes the need to do the next line since you otherwise don't need the iterator `i`.

emscripten/npm/src/verovio-toolkit.js
```diff
@@ -216,7 +216,6 @@ export class VerovioToolkit {
         let filesInBase64 = [];
         // Get all the files and convert them to a base64 string if necessary
         for ( const file of files ) {
-            const file = files[i];
             // The file in already passed as base64 string - nothing to do
```
craigsapp, line 219: "The file in already" &rarr; "The file is already".

PR #4074, opened 2025-05-22
Preserve ID in annot even with `--remove-ids`

lpugin:
Tiny PR that preserves `annot@xml:id` even with the `--remove-ids` option is enabled since these IDs are quite likely to be referenced from the outside. No change expected - at least from my point of view. Pinging @DILewis

### DIFF
--- a/src/iomei.cpp
+++ b/src/iomei.cpp
@@ -1497,7 +1497,7 @@ std::string MEIOutput::IDToMeiStr(Object *element)
 void MEIOutput::WriteXmlId(pugi::xml_node currentNode, Object *object)
 {
     // Keep IDs for annot elements
-    if (m_removeIds && !object->Is({ANNOT, ANNOTSCORE}) && !m_referredObjects.contains(object)) {
+    if (m_removeIds && !object->Is({ ANNOT, ANNOTSCORE }) && !m_referredObjects.contains(object)) {
         return;
     }
     currentNode.append_attribute("xml:id") = IDToMeiStr(object).c_str();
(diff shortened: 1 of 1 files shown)


ahankinson:
What's the use case for the `--remove-ids` option in the first place?

lpugin:
Good question. This option is there when you want to output the MEI with no IDs. It will only preserve IDs that are referenced. That is useful when you want to reuse (copy / paste) some parts of the encoding, which would otherwise duplicate the IDs.

rettinghaus:
I was thinking that we could make use of this in the test suite, to also check if Verovio preserves the structure of the input files.

lpugin:
> I was thinking that we could make use of this in the test suite, to also check if Verovio preserves the structure of the input files.

I've done it. I'll be adding a round-trip test.

rettinghaus:
However I do not see the point in this exception. Why would you choose to `--remove-ids` if you need to reference them? Or better asked: Why would all the other IDs interfere?

DILewis:
I'm kind of on the fence about this, or rather, I'm not automatically enthusiastic. Although client apps and stand-off annotations are more likely to use `annot` `xml:ids` than random other element ids, I'm not sure it's a difference in kind (in the way that other protected ids actually break the encoding). There's no way that Verovio can know what elements are pointed to at a given time. 

So, I feel like it depends on the use cases for `--remove-ids` and am happy to defer to those who actually use that...

lpugin:
OK, let's drop it then. I don't need it and it seems to bring more confusion than anything else.

lpugin:
@DILewis I'll remove this ids from all `annot` in the examples here https://github.com/rism-digital/verovio.org/tree/gh-pages/_tests/annot

lpugin:
For the record https://github.com/rism-digital/verovio.org/commit/4754ae1ee5416f70592256e0788e2270989a5c91

PR #4075, opened 2025-05-23
Add round-trip test

lpugin:
This PR adds a round-trip evaluation and the test suite units. It loads the MEI and exports it and performs and XML diff.

Test suite examples that are expected to modify the MEI when loaded are flagged with a `./appInfo@type"skip-round-trip"` and are ignored from the evaluation. 

For now, test suite examples failing the round-trip are added to the `log.md`. Eventually, we could make the action fail.

### DIFF
--- a/.github/workflows/tests_build.yml
+++ b/.github/workflows/tests_build.yml
@@ -106,6 +106,7 @@ jobs:
         working-directory: ${{ github.workspace }}/${{ env.DEV_DIR }}/doc
         run: |
           python3 ./test-suite-diff.py ${{ github.workspace }}/${{ env.TEMP_DIR }}/${{ env.PR_DIR }}/ ${{ github.workspace }}/${{ env.TEMP_DIR }}/${{ env.DEV_DIR }}/ ${{ github.workspace }}/${{ env.OUTPUT_DIR }}/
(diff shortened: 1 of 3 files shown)


lpugin:
We need to check that Verovio preserves it. Maybe we can change this once it is in place?

ahankinson:
Doesn't Verovio preserve the existing PIs, like

`<?xml-model href="https://music-encoding.org/schema/5.1/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>`

lpugin:
Kind of. I just checked and it currently does not preserve additional instructions. We need to check what pugixml allows for that.

ahankinson:
OK -- I've got a bunch of other changes. Would you prefer a PR on this one, or just commit directly?

lpugin:
Just commit

rettinghaus:
Why take an extra round running through all files again instead of integrating it into to the exisisting scrit?

lpugin:
I don't think this is a problem.

lpugin:
If you want to rewrite it, I am not against it. But maybe we can test it like this first?

rettinghaus:
No problem.

ahankinson:
OK, I've made the changes. There should be no difference in the output.

lpugin:
You changed the resource path and now I have
```
[Error] Failed to load font and glyph bounding boxes
[Error] Bravura font could not be loaded.
[Error] Failed to load font and glyph bounding boxes
[Error] Leipzig font could not be loaded.
[Error] Text font could not be initialized.
```
Why did you change it?

ahankinson:
ah, shoot. I did that by mistake and didn't mean to check it in. sorry. 

(I changed it because I was running from the `docs` directory.)

lpugin:
Looks good to be merged.

PR #4076, opened 2025-05-23
Enable test-suite-roundtrip.py in the GH action

lpugin:
Follow up of #4075


lpugin:
Looks like it is working as expected https://github.com/rism-digital/verovio/actions/runs/15212203575/job/42788889090

PR #4081, opened 2025-06-03
Annotscore code improvements

brdvd:
This PR adds some code polishing to the changes from #4029 .

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -12,6 +12,7 @@
 #include <functional>
 #include <iterator>
 #include <map>
+#include <memory>
 #include <string>
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


rettinghaus (approved):
Looks great! Thanks!

include/vrv/object.h
```diff
@@ -859,9 +858,9 @@ class Object : public BoundingBox {
     bool m_isExpansion;
 
     /**
-     * An optional list of back-links to plist referring objects
+     * List of back-links to plist referring objects
      */
-    std::optional<ListOfConstObjects> m_plistReferences;
```
lpugin, line 864: Why remove this? We now have a std::list on all objects, when this is nearly never used.

lpugin: This is weird. Yes, the use of `optional` is for limiting memory consumption. This is what I get from ChatGPT

> Does std::optional<std::vector<T>> save memory?
>
> Yes, it can save memory when the vector is empty, because:
>
>    A default-constructed std::vector<T> still allocates some memory (it may allocate a small capacity even when empty).
>
>    std::optional<std::vector<T>> avoids this allocation when the vector is not needed by being in a disengaged state (std::nullopt).
>
> However, when the vector is used (emplace or operator=), it will allocate memory like a normal std::vector<T>.

brdvd: If you want to optimize for memory, I would suggest to use `std::unique_ptr<ListOfConstObjects>` which should definitely be smaller that a `std::list`.

lpugin: Since this is within a C++ class, as far as I understand it should be fine.

Issue #3876, opened 2024-12-04
Test build of all bindings on ci


rettinghaus:
It would be good to have test builds with GitHub Actions for all supported bindings, so that we don't miss breaking changes in any of those.

lpugin:
Any plans to work on this @rettinghaus ?

lpugin:
I guess this can be closed

PR #4086, opened 2025-06-09
Build Java toolkit in CI

rettinghaus:
This PR updates the `pom.xml` following maven naming conventions and sets up a CI step to build the Java toolkit and uploads the artifact.

First part of #3876

NB: We should think about pushing the artifacts to Maven Central.


rettinghaus:
see https://github.com/rettinghaus/verovio/actions/runs/15535606446

lpugin (approved):

PR #4096, opened 2025-06-16
Fix unused warnings

lpugin:
I prefer to have them fixed like this than with #4094


lpugin:
> What about if you want to know the basic meaning of the unused parameters in the future?

What do you mean by basic meaning? 

`clang-format` would cope with commented parameter names, but I am not sure this is really useful.

craigsapp:
I mean basic meaning inferred by the name of the parameter.

PR #4097, opened 2025-06-17
Drop light version and reduce initial memory

lpugin:
This PR reducing the initial stack/heap memory allocation in the JS toolkit from 256/512 to 64/128. There was some issue with the memory growth options in some browsers, and testing showed that this is no longer the case.

The PR also drop the light memory allocation version, which is no longer needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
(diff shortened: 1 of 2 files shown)


ahankinson (approved):

craigsapp:
Is this related to Karen's problem?

PR #4113, opened 2025-07-07
set metcon for pickup measures

### LINKED ISSUE #1568: mrest enhancement in MusicXML-to-MEI converter
In the following example MusicXML file, there are two main problems related to the whole-measure rest and to inferred tuplet spaces. In this issue I will discuss the whole-measure rest, and I will add another for the implicit `<space>s` that are needed in the last measure.

Intended rendering (in Sibelius):

Screen Shot 2020-07-21 at 9 39 21 PM

Actual rendering after import of MusicXML into verovio:

Screen Shot 2020-07-21 at 9 44 59 PM

The whole-measure rest in the first measure of the top staff of the piano part is being converted into a whole-measure rest, but with an explicit duration of a whole note rather than the duration of the 2/8 meter, which means the duration of the rest should be a quarter note rather than a whole note.

In the MusicXML file, there are 256 ticks per quarter note:

```xml
    <divisions>256</divisions>
```

The problem whole-measure rest:

```xml
   <note default-x="118">
    <rest />
    <duration>256</duration>
    <instrument id="P2-I1" />
    <voice>1</voice>
    <type>whole</type>
    <staff>1</staff>
   </note>
```

Notice that the `<type>` is `whole`, but the `<duration>` is 256, which is equivalent to a quarter note.  In the MusicXML-to-MEI conversion:

```xml
            <rest xml:id="rest-0000001621851033" dur.ppq="256" dur="1" />
```

`@dur` is set to "1" (a whole note duration).  This is incorrect since the meter is 2/8, so the duration of the rest should be a quarter note.  When converting `@dur` to "4", the rhythm becomes correct, and there is no gap at the end of the first measure:

Screen Shot 2020-07-21 at 10 02 30 PM

Of course, this rest should be encoded as an `<mRest>` since the duration of the measure matches the time signature:

```
            <mRest xml:id="rest-0000001621851033" />
```
 
Using this gives the correct display for that measure in verovio:

Screen Shot 2020-07-21 at 10 08 11 PM

for 4/4, 2/2, 1/1, 8/8 meters, it may be difficult to identify an mRest since the duration is the same as a whole note, but in MusicXML, it should be easy to detect such cases as in this example by comparing  the `<type>` to the `<duration>`.  In this case the type is `whole`, which has an expected duration of 1024.  But since it is instead 256, that is one clue that this is an mRest rather than a rest.  In addition the duration of the meter is one quarter note or 256 divisions, which reinforces that this is an mRest rather than a rest.

---------

I will also point out a conceptual error in the MusicXML importer as well as the Sibelius MuisicXML export:  The `<duration>` element of `<note>s` in MusicXML maps **_exactly_** to `note@dur.ges` in MEI (plus `note@dots.ges` and also incorporates `tuplet@num` and `tuplet@numbase`).  There is absolutely no difference between them other than that MusicXML version includes the tuplet ratio in <duration>.  It is conceptually incorrect to map MusicXML `note/duration` to `note@dur.ppq`; however, Sibelius MusicXML has the same conceptual error as can be noted by the power-of-two value that is always generated for `<divisions>` in the MusicXML file (causing any tuplets to be approximately represented in `<duration>`.

A problem with mapping `<duration>` to `note@ppq` is that this implies that the performance duration of the MIDI note is that value, but rather the value in the MusicXML file is the time from the start of that note to the start of the next note (this is called the inter-onset interval (IOI) in cognitive psychology).  So `<duration>` in MusicXML is actually an IOI rather than a duration.  This is also true for `note@dur` in MEI.  In MIDI, the PPQ is typically a physical performance time, and so mapping to `note@ppq` is implying that that `<duration>` is not an IOI but rather a performance time, which is incorrect.   Consider a note with a staccato articulation:  the `note@dur` of the note remains the same, but when mapping to MIDI, there is no articulation parameter, and instead the duration of the note is shortened, so if this note/rest were staccato, I would expected `note@ppq` to be 1/2 of the IOI duration in `note@dur`.

MusicXML derives the divisions/duration/type system from Musedata, where it also maps exactly to MEI `@dur`/`@dur.ges` system:   Both the duration and type is always required in Musedata, as it is in MusicXML, while in MEI `@dur.ges` is only given when it disagrees with `@dur`. Musedata and Musicxml use different units for `@dur`/`@dur.ges`, while MEI uses the same units (making it clearer that they are closely related to each other).  When there is a difference between the visual duration of the note and the alignment duration they will be different in a manner similar to MEI `@dur` and `@dur.ges`.  This system was specifically to encode echoes of mensural notation in the music of J.S. Bach.  I will also point out that this is also analogous to the rhythm-encoding system in SCORE, where P7 of note gives the "duration" of the note for alignment purposes, and P6 (notehead shape)/P9 (flags/augmentation dots) give the visual appearance of the note.

Here is the documentation for `<divisions>` in MusicXML

https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-divisions.htm

> Musical notation duration is commonly represented as fractions. The divisions element indicates how many divisions per quarter note are used to indicate a note's duration. For example, if duration = 1 and divisions = 2, this is an eighth note duration. Duration and divisions are used directly for generating sound output, **_so they must be chosen to take tuplets into account_**. Using a divisions element lets us use just one number to represent a duration for each note in the score, while retaining the full power of a fractional representation. If maximum compatibility with Standard MIDI 1.0 files is important, do not have the divisions value exceed 16383.

The emphasis is mine.  Note that the MusicXML export from Sibelius does not take tuplets into account, since divisions of `256` per quarter note is `2*2*2*2*2*2*2` with no factors of `3` to deal with integer divisions of the triplets encoded in this example.   There is also some conceptual mixing of "musical notation duration" and "performance duration" since MIDI files can represent both or a mixture of the two, but when discussion "fractions" it is clear that these values are not intended to be used for performance durations but rather printed music proportional notation durations.

Documentation for duration element:

https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-duration.htm

> Duration is a positive number specified in division units. This is the intended duration vs. notated duration (for instance, swing eighths vs. even eighths, or differences in dotted notes in Baroque-era music). Differences in duration specific to an interpretation or performance should use the note element's attack and release attributes.

There is also a bit of conceptual mixing of performance and notation, but this also relates to IOI duration rather than sounding duration.

rettinghaus:
The rendering still shows a small gap, but the timemap seems to fine with this.

Bildschirmfoto 2025-07-07 um 15 23 04

closes #1568

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1719,6 +1719,13 @@ bool MusicXmlInput::ReadMusicXmlMeasure(
         }
     }
 
+    // set metcon to false for pickup measures
+    int endDuration = m_ppq;
+    for (const int &num : m_meterCount) endDuration *= num;
+    if (m_durTotal != endDuration) {
+        measure->SetMetcon(BOOLEAN_false);
+    }
+
     this->MatchTies(true);
     if (!m_tieStack.empty()) this->MatchTies(false);
     for (auto openTie : m_tieStack) {


lpugin:
Thanks. Could you post the timemap with rests ?

rettinghaus:
```json
[
	{
		"measureOn": "j1cba0n7",
		"on": [
			"v1suadt9" 
		],
		"qstamp": 0,
		"restsOn": [
			"z997z5h",
			"e1t1xgfe" 
		],
		"tempo": 120,
		"tstamp": 0 
	},
	{
		"off": [
			"v1suadt9" 
		],
		"on": [
			"wjv7z2" 
		],
		"qstamp": 1,
		"tstamp": 500 
	},
	{
		"off": [
			"wjv7z2" 
		],
		"on": [
			"ymms5us" 
		],
		"qstamp": 2,
		"tstamp": 1000 
	},
	{
		"off": [
			"ymms5us" 
		],
		"qstamp": 3,
		"tstamp": 1500 
	},
	{
		"measureOn": "z1cubh9x",
		"on": [
			"k1cpvc8" 
		],
		"qstamp": 4,
		"restsOff": [
			"z997z5h",
			"e1t1xgfe" 
		],
		"restsOn": [
			"l12k9or4",
			"q1ww9mrj" 
		],
		"tstamp": 2000 
	},
	{
		"on": [
			"mhppyn8" 
		],
		"qstamp": 5,
		"restsOff": [
			"l12k9or4" 
		],
		"tstamp": 2500 
	},
	{
		"off": [
			"mhppyn8" 
		],
		"on": [
			"n1ymb8cw" 
		],
		"qstamp": 6,
		"tstamp": 3000 
	},
	{
		"off": [
			"n1ymb8cw" 
		],
		"on": [
			"p1ff1ka" 
		],
		"qstamp": 7,
		"tstamp": 3500 
	},
	{
		"measureOn": "lixpcid",
		"off": [
			"k1cpvc8",
			"p1ff1ka" 
		],
		"on": [
			"b5nlouz" 
		],
		"qstamp": 8,
		"restsOff": [
			"q1ww9mrj" 
		],
		"restsOn": [
			"wl5qik2",
			"cl3ewgg" 
		],
		"tstamp": 4000 
	},
	{
		"off": [
			"b5nlouz" 
		],
		"qstamp": 12,
		"restsOff": [
			"wl5qik2",
			"cl3ewgg" 
		],
		"tstamp": 6000 
	} 
] 
```

rettinghaus:
Even when I use the generated MEI.

[tests.zip](https://github.com/user-attachments/files/21106842/tests.zip)

PR #4129, opened 2025-07-18
Remove android binding and add swift binding

lpugin:
The android binding has not been maintain. I'll replace it with a repository with a minimal app instead.

The java package name was still `org.rismch.verovio`. It is updated to `org.verovio.lib`.


rettinghaus (approved):

samuelbradshaw:
cc: @henrythasler (https://github.com/rism-digital/verovio/discussions/4080)

PR #4131, opened 2025-07-21
Humlib update.

craigsapp:
This PR in particular should fix compiler warnings, as it was tested with the compile flags `-Wshorten-64-to-32` and `-Wsign-compare`.

In GNU Makefile, I now use:

```
# MacOS specific warnings
ifeq ($(shell uname -s),Darwin)
  PREFLAGS += -Wshorten-64-to-32
  PREFLAGS += -Wsign-compare
endif
```


rettinghaus (approved):
LGTM

PR #4134, opened 2025-07-24
Removing unused emscripten options

rettinghaus:


rettinghaus:
I would like to have an MEI-only build. How about merging all these options into one?

lpugin:
Why do you need this?

rettinghaus:
Nothing specific and no need to have it upstream.

lpugin:
So are you suggesting we remove all importers from the JS build? We use Plaine and Easie, others use MusicXML. We went through that discussion already https://github.com/rism-digital/verovio/discussions/2791

PR #4136, opened 2025-07-26
Add support for Liberation serif text font

lpugin:
This PR adds support for embedding or linking the open font Liberation serif font. This improves the rendering of text (lyrics, directives) on devices on which the Times font is not available. Layout will still be calculated based on Times, but because Liberation is meant to have the same metrics, the layout result is expected to be much better than with another fallback font with different metrics. 

The use of Liberation must be enabled explicitly with the new `--font-text-liberation` option. The font will be embedded in the SVG as WOFF2 in the same way SMuFL text fonts are. The music font option `--smufl-text-font` option with the value `linked` will also make the Liberation pointing to `https://www.verovio.org/`. We might consider renaming that option since Liberation is obviously not a SMuFL font.

The PR also makes the embedded footer use path instead of text in order to avoid font size differences.

The two screenshots below show the rendering on Android with and without `--font-text-liberation` set. The option avoids all the collisions in the lyrics.

image

image


henrythasler:
I wonder how you manage to render WOFF2 embedded in an SVG on Android? Just a WebView or something more involved?

lpugin:
Yes, I use WebView. I'll publish the code of the Android app once the release is made.

PR #4160, opened 2025-08-19
Adjust noteOff timings

### LINKED ISSUE #4150: MIDI export with consecutive identical notes confuses players
**Describe the problem**
Exporting a score with consecutive identical notes to MIDI results in an erroneous rendering. Specifically, the subsequent identical notes have wrong durations.

**To Reproduce**
1. Export the attached MEI file to MIDI (also attached for convenience)
2. Hear that the C4 notes after the first are being sounded incorrectly
3. The C5 note sounds correctly

**Expected behavior**
The C4 notes should all sound with equal durations.

**Input data**

**Verovio information**
 - Verovio 5.4.0-dev-6ca5e2c
 - Command-line

[consecutive-notes.zip](https://github.com/user-attachments/files/21688649/consecutive-notes.zip)

Image

infojunkie:
This PR fixes #4150 by reducing the noteOff onset by 1 MIDI tick. This helps MIDI players to correctly sort the events, removing the ambiguity of dealing with simultaneous noteOn / noteOff events for the same note.

### DIFF
--- a/src/midifunctor.cpp
+++ b/src/midifunctor.cpp
@@ -741,7 +741,7 @@ FunctorCode GenerateMIDIFunctor::VisitLayerEnd(const Layer *layer)
     // stop all previously held notes
     for (auto &held : m_heldNotes) {
         if (held.m_pitch > 0) {
-            m_midiFile->addNoteOff(m_midiTrack, held.m_stopTime * m_midiFile->getTPQ(), m_midiChannel, held.m_pitch);
+            m_midiFile->addNoteOff(m_midiTrack, held.m_stopTime * m_midiFile->getTPQ() - 1, m_midiChannel, held.m_pitch);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/midifunctor.cpp
```diff
@@ -828,7 +828,7 @@ FunctorCode GenerateMIDIFunctor::VisitNote(const Note *note)
             const double stopTime = startTime + midiNote.duration;
 
             m_midiFile->addNoteOn(m_midiTrack, startTime * tpq, channel, midiNote.pitch, velocity);
-            m_midiFile->addNoteOff(m_midiTrack, stopTime * tpq, channel, midiNote.pitch);
+            m_midiFile->addNoteOff(m_midiTrack, stopTime * tpq - 1, channel, midiNote.pitch);
```
ahankinson, line 831: Can `stopTime * tpq` ever be 0? Is it a problem if you have `0 - 1`?

infojunkie: Good point! I think it's highly unlikely since we're dealing with linearly increasing time, but in the absence of unit tests, better be safe than sorry. I added `std::max(0.0)` in front of each expression.

PR #4172, opened 2025-09-03
Detailed warning for unmatched time-spanning elements

wergo:
Verovio used to report to the console the number of unmatched time-spanning elements with `@startid` and `@endid`, where one or both of the elements pointed to are missing, such as: 

```
[Warning] 2 time spanning element(s) with startid and endid could not be matched.
```

This PR makes Verovio report in detail on each such element found, which helps finding errors in the MEI encoding. 
The console output may look like: 

```
[Warning] 2 time spanning element(s) with startid and endid could not be matched:
[Warning] Unmatched tie tie-002054303300: @startId: #note-000509215993, @endId: #note-000079104669
[Warning] Unmatched slur slur-002054305342: @startId: #note-000509365422, @endId: #note-000654252235
```

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -863,7 +863,7 @@ void Doc::PrepareData()
 
     // Something must be wrong in the encoding because a TimeSpanningInterface was left open
     for (const auto &obj : prepareStaffCurrentTimeSpanning.GetTimeSpanningElements()) {
-        LogWarning("Time spanning element could not be set as running: <%s xml:id='%s' ...>",
+        LogWarning("Time spanning element '%s' with @xml:id '%s' could not be set as running.",
             obj->GetClassName().c_str(), obj->GetID().c_str());
     }
 


wergo:
Thanks @rettinghaus ! Do we even need to report the number? I reduced it even more. What do you think?

wergo:
We should probably update the warnings for the plist and the linking interfaces as well?

rettinghaus (approved):
lgtm, thanks

src/doc.cpp
```diff
@@ -862,9 +862,9 @@ void Doc::PrepareData()
     root->Process(prepareStaffCurrentTimeSpanning);
 
     // Something must be wrong in the encoding because a TimeSpanningInterface was left open
-    if (!prepareStaffCurrentTimeSpanning.GetTimeSpanningElements().empty()) {
-        LogDebug("%d time spanning elements could not be set as running",
-            prepareStaffCurrentTimeSpanning.GetTimeSpanningElements().size());
+    for (const auto &obj : prepareStaffCurrentTimeSpanning.GetTimeSpanningElements()) {
+        LogWarning("Time spanning element could not be set as running: <%s xml:id='%s' ...>",
```
lpugin, line 866:
```suggestion
        LogWarning("Time spanning element '%s' with @xml:id '%s' could not be set as running.",
```

samuelbradshaw:
Thank you for this improvement! I needed to debug some time spanning element warnings today, and it saved a lot of time to switch to the develop branch temporarily to get the IDs of the mismatched elements.

PR #4173, opened 2025-09-03
Import meter visibility

rettinghaus:
This extends the MusicXML support for invisible time signatures. 
Also it makes meterSigGrp part of AttVisibility.


lpugin (commented):
Looks good. Could you drop the line added to the changelog? I prefer bringing there only changes that are (more) significant for the users. Thanks

Issue #4174, opened 2025-09-03
Deprecated "unload" event listener in Chrome


samuelbradshaw:
**Describe the problem**
In Chrome, I get this error message in the console when I load verovio-toolkit-wasm:
[Violation] Permissions policy violation: unload is not allowed in this document.

This is low priority, as it does not appear to prevent Verovio from working, but it would be nice to get it cleaned up.

**Verovio information**
 - Version of Verovio: 5.5.0
 - Binding: JavaScript

**Environment information (as appropriate)**
 - OS: macOS 15.6.1 Sequoia
 - Browser: Chrome

**Screenshots**
<img width="529" height="113" alt="Image" src="https://github.com/user-attachments/assets/1e49e622-82fd-4ba4-a2b7-430bbc1756fc" />

**Additional context**
Here's Chrome's explainer page about the deprecation:
https://chromestatus.com/feature/5579556305502208

And the relevant MDN page (with deprecation warning):
https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event

In Verovio's code, the unload event listener is added here:
https://github.com/rism-digital/verovio/blob/2b392defd7dde378d1a78322701753712ab6b9bf/emscripten/npm/src/verovio-toolkit.js#L234
If I'm understanding correctly, this event listener can simply be removed.

PR #4175, opened 2025-09-03
Remove event listener for deprecated unload event

samuelbradshaw:
This resolves https://github.com/rism-digital/verovio/issues/4174.

### DIFF
--- a/emscripten/npm/src/verovio-toolkit.js
+++ b/emscripten/npm/src/verovio-toolkit.js
@@ -226,14 +226,3 @@ export class VerovioToolkit {
 
 // A pointer to the object - only one instance can be created for now
 VerovioToolkit.instances = [];
-
-
-// If the window object is defined (if we are not within a WebWorker)...
-if ((typeof window !== "undefined") && (window.addEventListener)) {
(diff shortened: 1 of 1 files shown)


ahankinson:
Does this have the possibility of creating a memory leak? Currently it destroys the TK instance when the page is "unloaded" -- but there doesn't seem to be any other place where the instance is destroyed, leading me to question whether it will stick around indefinitely. 

The Usage Notes on the MDN page: https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes

says that "visibility change" or "pagehide" are preferred alternatives, but I'm not sure how that would work -- if it would destroy the instance of Verovio but not re-create it if the page is navigated to again.

lpugin:
Having dug into this a bit more it seems we can drop it completely. Sorry for going back a forth, but please revert it to https://github.com/rism-digital/verovio/pull/4175/commits/e1ef1310484e0b6facd9e78fc6f2a32411824ec9. Thanks!

samuelbradshaw:
Sounds good! Done.

PR #4180, opened 2025-09-09
Clarify option message for avoiding cue notes in MIDI playback

wergo:
The option message for avoiding/skipping cue notes in MIDI playback suggested the opposite: `"MIDI playback of cue notes" `would play them, when set to `false` (thus not ticked in mei-friend), but skip them, when `true`/ticked.

The new formulation makes this clear. 

PS: I would prefer to invert the entire option and transform the variable into `m_midiWithCue`.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -1816,7 +1816,7 @@ Options::Options()
     m_midi.SetCategory(OptionsCategory::Midi);
     m_grps.push_back(&m_midi);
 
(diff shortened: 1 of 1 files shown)


lpugin:
Thanks. I am aware that a negative option is confusing. Changing it would mean changing the default behaviour, which is also problematic - even though I would be happy to consider it. In any case, your suggestion improves clarify already.

PR #4201, opened 2025-11-06
add --svg-content-bounding-boxes command line option

fernandoherreradelasheras:
This enables the code for drawing bounding boxes in the SVG (class "content-bounding-box") enclosing the whole content of the elements.

Such bounding boxes are useful for apps handling click on notes. Using the "self" bounding boxes, restricted to the note head path with a very small clickable area, makes click really hard.

### DIFF
--- a/include/vrv/svgdevicecontext.h
+++ b/include/vrv/svgdevicecontext.h
@@ -212,6 +212,14 @@ public:
     void SetSvgBoundingBoxes(bool svgBoundingBoxes) { m_svgBoundingBoxes = svgBoundingBoxes; }
 
     /**
+     * Setting m_svgContentBoundingBoxes flag (false by default)
+     */
+    void SetSvgContentBoundingBoxes(bool svgContentBoundingBoxes)
+    {
+        m_svgContentBoundingBoxes = svgContentBoundingBoxes;
+    }
+
+    /**
      * Setting m_svgViewBox flag (false by default)
      */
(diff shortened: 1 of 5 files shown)


include/vrv/svgdevicecontext.h
```diff
@@ -211,6 +211,14 @@ class SvgDeviceContext : public DeviceContext {
      */
     void SetSvgBoundingBoxes(bool svgBoundingBoxes) { m_svgBoundingBoxes = svgBoundingBoxes; }
 
+    /**
+     * Setting m_svgContentBoundingBoxes flag (false by default)
+     */
+    void SetSvgContentBoundingBoxes(bool svgContentBoundingBoxes)
```
ahankinson: How is this different from the `SetSvgBoundingBoxes` option?

fernandoherreradelasheras: For notes, the boundig-box encloses only the note head:

Screenshot From 2025-11-06 10-46-22

while the content-bounding-box will enclose the whole area determined by the outer points of the note:

Screenshot From 2025-11-06 10-46-40

In a layer, for example, the bounding-box will be empty but the content-bounding-box will cover the whole area containing all the notes

ahankinson: But the outer `<g>` (id "j5wwj5q" on your screenshot) covers the whole symbol, right? Here's what that looks like for me in a very simple test:

image

PR #4207, opened 2025-11-17
Add support for ossias

rettinghaus:
This is a first implementation for rendering `ossia` with `ostaff`.

Here is a test file:

[wetterfahne.mei.zip](https://github.com/user-attachments/files/23583834/wetterfahne.mei.zip)


rettinghaus:
I think both are on the same commit.

lpugin:
Weird. It shows changes in trill and others that have nothing to do with ossia

rettinghaus:
Opening a new one helped.

PR #4209, opened 2025-11-20
Add support for rest enclosings

rettinghaus:
This PR adds support for parenthesis and brackets around rests.

Bildschirmfoto 2025-11-20 um 20 12 34

Bildschirmfoto 2025-11-20 um 20 13 12

(rendered with Bravura font)

NB: Does not take dots into account.


lpugin:
They should be inside in my opinion

PR #4212, opened 2025-11-26
Fix and uniformise SVG device context pen related attributes

lpugin:
Changes undetected because of the threshold introduced in the test suite.


lpugin:
image

image

rettinghaus (approved):

PR #4217, opened 2025-11-27
Add support for enclosings on mordents and turns

rettinghaus:
Bildschirmfoto 2025-11-27 um 16 36 07
Bildschirmfoto 2025-11-27 um 16 37 05
Bildschirmfoto 2025-11-27 um 16 37 23

### DIFF
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -2857,7 +2857,7 @@ void View::DrawTrill(DeviceContext *dc, Trill *trill, Measure *measure, System *
             this->DrawSmuflCode(dc, x - xCorrEncl, y + trillHeight / 2, enclosingFront, staffSize, false);
         }
 
-        // Upper and lower accidentals are currently exclusive, but sould both be allowed at the same time.
+        // Upper and lower accidentals are currently exclusive, but should both be allowed at the same time.
         if (trill->HasAccidlower()) {
             int accidXShift = (alignment == HORIZONTALALIGNMENT_center) ? 0 : trillWidth / 2;
             char32_t accid = Accid::GetAccidGlyph(trill->GetAccidlower());


src/rest.cpp
```diff
@@ -327,14 +327,11 @@ char32_t Rest::GetRestGlyph(const data_DURATION duration) const
 
 std::pair<char32_t, char32_t> Rest::GetEnclosingGlyphs() const
 {
-    if (this->HasEnclose()) {
```
lpugin: same here

src/view_control.cpp
```diff
@@ -2471,7 +2476,17 @@ void View::DrawMordent(DeviceContext *dc, Mordent *mordent, Measure *measure, Sy
                 accidY += (mordentHeight - m_doc->GetGlyphBottom(accid, staffSize / 2, true)
                     + m_doc->GetDrawingUnit(staffSize * 2 / 3));
             }
+            dc->StartCustomGraphic("accid upper");
```
lpugin:
```suggestion
            dc->StartCustomGraphic("accidUpper");
```

src/view_control.cpp
```diff
@@ -2858,8 +2880,10 @@ void View::DrawTrill(DeviceContext *dc, Trill *trill, Measure *measure, System *
             accidStr.push_back(accid);
             int accidY = y + trillHeight - m_doc->GetGlyphBottom(accid, staffSize / 2, true)
                 + m_doc->GetDrawingUnit(staffSize * 2 / 3);
+            dc->StartCustomGraphic("accid upper");
```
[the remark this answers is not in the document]
lpugin: That's exacly the point. Verovio's principle is to produce SVG where the tree of `g` have a class matching the MEI element name. So when selecting elements with `accid`, one would not expect this element to be selected because it is not an `accid` element in the MEI tree. So I am fine adding a custom graphic (even thought that would be debatable) but giving it an `accid` class is a bad idea.

PR #4250, opened 2025-12-30
MusicXML expansion

infojunkie:
This PR creates a default `expansion` element during MusicXML import. The code interprets the repeats and jumps in the score to produce this expansion.

The code builds upon the existing logic that handles endings to be inserted at the end of MusicXML import. Whereas the existing logic only adds `Ending` groups to handle alternate endings, the new logic adds `Section` groups whenever a group of measures is involved in a repeat or a jump. At the end of the MusicXML import, the logic makes one more pass over these sections to generate the `expansion` element according to the sequence dictated by the standard rules of expansion.

The implementation only affects the class `MusicXmlInput` in iomusxml.h and iomusxml.cpp. 

In the class header, the `EndingInfo` struct gets wrapped into a larger `SectionInfo` that captures all the data needed to produce the expansion. In addition to `EndingInfo`, `SectionInfo` includes `RepeatInfo`, `JumpInfo` and `FineInfo` structs. The original list of `EndingInfo` instances and their associated measures is now replaced with a list of `SectionInfo` instances called `m_sections`. Additional temporary state is needed to populate this list, namely `m_sectionStart`, `m_sectionStop`, `m_jumpInfo`, `m_fineInfo`.

In the class implementation, multiple code blocks are added:
- A new method `ReadMusicXmlSound` parses the jumps and repeats that affect the subsequent creation of section. It is called during the parsing of measures at `ReadMusicXmlMeasure`.  
- The existing method `ReadMusicXmlBarLine` is also augmented with logic to create section infos. 
- The logic in method `ReadMusicXml` that creates `Ending` instances and moves measures into them is replaced with a larger block that also creates `Section` instances as needed. 
- This same method also calls new method `CreateExpansion` which creates the `Expansion` element with a sequence defined by the repeats and jumps associated with the sections above.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1205,7 +1205,7 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
         // go through measures move them info <section> / <ending>
         for (Measure *measure : measures) {
             // also move preceding non-measure siblings
-            // keep stacking them in reverse then detach them in score order
+            // keep stacking them in reverse order then transfer them in score order
             Object *sibling = measure;
             std::list<Object *> siblings;
             while ((sibling = section->GetPrevious(sibling))) {
@@ -1349,11 +1349,15 @@ void MusicXmlInput::CreateExpansion(Section *section)
             // gather all endings, by creating a map from ending number to ending iterator
             std::map<int, decltype(iter)> endings;
             auto begIter = iter;
+            std::optional<decltype(iter)> rptNestedIter;
             while (iter != m_sections.end() && iter->first.m_classId == ENDING) {
                 for (auto i : parseInts(iter->first.m_endingInfo.m_number)) {
                     endings.insert({ i, iter });
                 }
                 endIter = iter++;
+
+                // remember a nested repeat start
+                if (endIter != begIter && endIter->first.m_repeatStart) rptNestedIter = endIter;
 
                 // increment visited count of subsequent sections
                 if (endIter != begIter) endIter->first.m_visited++;
@@ -1379,6 +1383,9 @@ void MusicXmlInput::CreateExpansion(Section *section)
                 std::string endref = "#" + ending->second->first.m_target->GetID();
                 expansion->GetPlistInterface()->AddRefAllowDuplicate(endref);
             }
+
+            // set the repetition to the (latest) nested one
+            if (rptNestedIter) rptIter = *rptNestedIter;
         }
 
         // fine
@@ -2152,7 +2159,7 @@ void MusicXmlInput::ReadMusicXmlBarLine(pugi::xml_node node, Measure *measure, c
 
     // start or end section
     if (measure->GetLeft() == BARRENDITION_rptstart) {
-        m_sectionStart = musicxml::SectionInfo(true);
+        if (!m_sectionStart || !m_sectionStart->m_repeatStart) m_sectionStart = musicxml::SectionInfo(true);
     }
     if (measure->GetRight() == BARRENDITION_rptend) {
         m_sectionStop = musicxml::SectionInfo(musicxml::RepeatInfo(repeatTimes, repeatAfterJump));
@@ -2164,14 +2171,13 @@ void MusicXmlInput::ReadMusicXmlBarLine(pugi::xml_node node, Measure *measure, c
         std::string endingNumber = ending.attribute("number").as_string();
         std::string endingType = ending.attribute("type").as_string();
         std::string endingText = ending.text().as_string();
-        // LogInfo("ending number/type/text: %s/%s/%s.", endingNumber.c_str(), endingType.c_str(),
-        // endingText.c_str());
         if (endingType == "start") {
             // check for corresponding stop points
             std::string xpath = StringFormat("following::ending[@number='%s'][@type != 'start']", endingNumber.c_str());
             pugi::xpath_node endingEnd = node.select_node(xpath.c_str());
             if (endingEnd) {
-                m_sectionStart = musicxml::EndingInfo(endingNumber, endingType, endingText);
+                if (!m_sectionStart) m_sectionStart = musicxml::SectionInfo();
+                m_sectionStart->merge(musicxml::EndingInfo(endingNumber, endingType, endingText));
             }
         }
         else if (endingType == "stop" || endingType == "discontinue") {
@@ -2179,7 +2185,8 @@ void MusicXmlInput::ReadMusicXmlBarLine(pugi::xml_node node, Measure *measure, c
                 LogWarning("MusicXML import: Dangling ending tag skipped");
             }
             else {
-                m_sectionStop = musicxml::EndingInfo(endingNumber, endingType, endingText);
+                if (!m_sectionStop) m_sectionStop = musicxml::SectionInfo();
+                m_sectionStop->merge(musicxml::EndingInfo(endingNumber, endingType, endingText));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


infojunkie:
Example 2: [Salma ya salama (J. Barnel)](https://github.com/user-attachments/files/24387166/salma-ya-salama.zip)

salma-ya-salama

```xml
<expansion plist="#fjbf1bf #fjbf1bf #glio898 #h1x875sx #glio898 #ii7e32y #j1neniec #knxg16m #knxg16m #lbf8lnw #m2lm4ef #nguxhkc #m2lm4ef #ozgghfv #glio898 #ii7e32y #j1neniec #knxg16m" xml:id="expansion-default" />
```

infojunkie:
Thanks for testing, and good to hear it's working well! For your questions:

> Is there a simple way to clean up the markings that indicate endings, codas, repeats, etc. when the score is expanded? I'm assuming this would need to be done by Verovio, after the expansion?

I am not aware of any option in Verovio to remove repeat / jump signs from the score at expansion time... I did work on something similar as a XSD transformation that applies to MusicXML scores - see ["unroll" at `musicxml-midi`](https://github.com/infojunkie/musicxml-midi?tab=readme-ov-file#convert-a-musicxml-score-to-midi). It might help, but I know that it will not handle the lyrics correctly.  This is a feature I am interested in, but it would be a separate piece of work.

> Do you have any thoughts on feasibility of expanding multiple verses in the lyrics [..]

Typically, this is done using the MusicXML [`repeat` element along with the `@times` attribute](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/repeat/). But in your first example, I can't think of a way to obtain the sequence you want without duplicating the chorus after the 3x |: chorus + verse :| block.  I'd be very interested to learn how an experienced engraver would notate that.

src/iomusxml.cpp
```diff
@@ -1244,6 +1272,133 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
     return true;
 }
 
+// vibe-coded on Claude AI using prompt:
+//
+// in C++, write a concise function to parse a string of comma separated ints with optional spaces into a list (array or
+// vector) of ints
+//
+// verified at https://onlinegdb.com/blxCanxpAW
+std::vector<int> parseInts(const std::string &str)
```
lpugin: MINOR use a `std::list` instead

src/iomusxml.cpp
```diff
@@ -1244,6 +1272,133 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+
+            // remember last section
+            seciter = enditer = iter++;
+        }
+        else /* ENDING */ {
+            // gather all endings, by creating a map from ending number to ending iterator
+            std::map<int, decltype(iter)> endings;
+            auto begiter = iter;
+            while (iter != m_sections.end() && iter->first.m_classId == ENDING) {
+                auto is = parseInts(iter->first.m_endingInfo.m_number);
```
lpugin: Avoid `auto` when not an iterator or in lambda

src/iomusxml.cpp
```diff
@@ -1155,43 +1170,53 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+        // <section> / <ending>
+        for (Measure *measure : measures) {
+            // remove other measures from <section> that are not already removed above (first measure)
+            if (measure->GetID() != measures.front()->GetID()) {
+                int idx = section->GetChildIndex(measure);
+                section->DetachChild(idx);
             }
-            LogDebug(logString.c_str());
+            target->AddChild(measure); // add <measure> to sub-element
         }
```
lpugin: It would be clearer to use `InsertBefore` to insert the `target` before the first measure, and then move all measures, rather than using `ReplaceChild` and then skip the first one.

src/iomusxml.cpp
```diff
@@ -1244,6 +1268,132 @@ bool MusicXmlInput::ReadMusicXml(pugi::xml_node root)
...
+    }
+    return result;
+}
+
+void MusicXmlInput::CreateExpansion(Section *section)
+{
+    assert(section);
+
+    Expansion *expansion = new Expansion();
+    expansion->SetID("expansion-default");
```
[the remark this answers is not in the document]
infojunkie: @lpugin the idea of adding new flag(s) for the first expansion specifically sounds wasteful to me... I suggest the following, in line with @samuelbradshaw 's thinking:

- Augment `--expand` to accept "magic" selectors other than expansion ID, such as an integer to denote the N-th expansion, or `first`, etc.
- Add a `--list-expansions` flag to list all expansions, which would be generally useful (unless there's an existing way to make an XPath query into the score)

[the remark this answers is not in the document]
lpugin: I am not fond of the idea of having "magic" IDs. We could make `--expand 1` work because this is not a valid `xml:id`, but then it is confusing because the type for the option still is a string... Let's just add `--expand-first` for now. @infojunkie can you do it as part of this PR? Otherwise I can add it right after merging.

Adding `--list-expansions` needs more thinking.

samuelbradshaw:
> Just an idea, if the logic to calculate the missing expansion logic is based on the MEI document, then it would work for MusicXML (implictly) and for MEI (without the export/import).

I agree that this would be helpful. Most of the scores I'm working with start as MusicXML, but sometimes I'm modifying an MEI document that's already been converted, and it would be nice to generate a new expansion after those modifications. I logged a feature request for this: https://github.com/rism-digital/verovio/issues/4258

samuelbradshaw:
It looks like children in the root `<section>` element that aren't measures aren't being properly handled.

For example, in this score:
https://assets.churchofjesuschrist.org/d2/46/d246mgmulwavz3im4yv2qeqoiuhhyvsaimfwcvtx/this_is_the_christ.mxl

The root `<section>` element has a mix of measures, scoreDefs (for time signature change and key change), page breaks, and section breaks. The measures are being moved into sections, but the other elements aren't, which causes those elements to move to just before the first `<ending>`.

I would expect those elements to move into sections, with the measures that immediately follow them, to maintain the order of the score.

infojunkie:
@lpugin if you don't mind, I'll keep working on the reported issues here and submit the PR again when it's ready.

infojunkie:
> I've added the relevant repeat structures to the MusicXML and performed experiments with both MuseScore and the algorithm here: Both do NOT produce the final A.

The MusicXML I had produced was actually incorrect because it was missing the D.C. al Fine `sound` element. Here's [the correct one](https://github.com/user-attachments/files/24557536/he_is_born_the_divine_christ_child.zip), and the algorithm here does produce the expected ABABABA sequence. Nothing more to do about this case.

I will investigate @samuelbradshaw 's latest report next.

infojunkie:
After a bit of debugging, I'm pretty sure @samuelbradshaw and @lpugin 's reports concern the same bug. In a nutshell, non-measure elements are not moved into sections / endings.

The same bug already existed in the original code, but was never triggered by the existing test files. For example, this modified [45d-Repeats-Nested-Alternatives](https://github.com/user-attachments/files/24604874/45d-Repeats-Nested-Alternatives.zip) adds a time signature to the **2nd measure of the 2nd ending**. In the SVG, the time signature appears at the start of the 3rd ending, because it was not moved from its original place when the endings were created. 

The same bug now applies to created sections as well. I'll fix both cases.

infojunkie:
The only remaining visual change comes from `45f-Repeats-InvalidEndings`, which is an explicitly invalid MusicXML score with a duplicated second ending (bars 3 and 4 both having the same ending number). 

Original algorithm selects ending from bar 4 and applies it to bar 3:
45f-Repeats-InvalidEndings before

New algorithm selects ending from bar 3 and ignores duplicate ending from bar 4:
45f-Repeats-InvalidEndings after

@lpugin is that ok to keep the new behaviour? It can be argued that it is "more correct" given an invalid MusicXML.

Issue #4252, opened 2026-01-01
Reduce specificity of inserted CSS styles


samuelbradshaw:
**Is your feature request related to a problem? Please describe.**
I ran into an issue when adding colored rectangles in an SVG score. I wanted to create a rectangle with a green stroke, but styling it with `.ch-chord-rect { stroke: green; }` didn't work. I had to use `.ch-chord-rect { stroke: green !important; }`. In debugging, I found that the issue was caused by the CSS rules inserted into the SVG by Verovio. Because the CSS rules are prefixed with IDs (as in `#f5cb8tx rect {stroke:currentColor}`), they have [high specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity) and are difficult to override.

**Describe the solution you'd like**
It would be helpful if the specificity of the inserted CSS rules could be reduced, and if unnecessary rules could be removed.

**Additional context**
The current SVG looks like this:

```svg
<svg width="677px" height="1300px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" overflow="visible" id="f5cb8tx">
    <desc>Engraved by Verovio 5.7.0-7135340</desc>
    <defs>…</defs>
    <style type="text/css">#f5cb8tx g.page-margin {font-family:Times,serif;}#f5cb8tx g.ending, #f5cb8tx g.fing, #f5cb8tx g.reh, #f5cb8tx g.tempo {font-weight:bold;}#f5cb8tx g.dir, #f5cb8tx g.dynam, #f5cb8tx g.mNum {font-style:italic;}#f5cb8tx g.label {font-weight:normal;}#f5cb8tx ellipse, #f5cb8tx path, #f5cb8tx polygon, #f5cb8tx polyline, #f5cb8tx rect {stroke:currentColor}</style>
    <svg class="definition-scale" color="black" viewBox="0 0 16920 32500">
        <g class="page-margin" transform="translate(40, 0)">…</g>
    </svg>
</svg>
```
This is my proposal:
- Use `.definition-scale` or another class name for scoping the CSS rules, instead of an ID (classes have lower specificity than IDs)
- Remove `g` from the CSS rules (element names increase specificity)
[...]

PR #4254, opened 2026-01-01
Improve css

rettinghaus:
This PR picks up on some things proposed in #4252 and simplyfies the common CSS rules. 

* The `PrefixCssRules` method now just groups all rules into a [ruleset](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Syntax/Introduction#css_rulesets) instead of prefixing every single rule
* g element selector now applies to a ruleset
* `font-family` has been made a SVG presentation attribute
* `currentColor` is now the default color (inheriting it from HTML)

Also there is a small fix for (beam) polygons that were always carrying a `fill-opacity` attribute.

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -510,6 +510,7 @@ void SvgDeviceContext::StartPage()
     m_svgNodeStack.push_back(m_currentNode);
     m_currentNode.append_attribute("class") = "definition-scale";
     m_currentNode.append_attribute("color") = "currentColor";
+    m_currentNode.append_attribute("fill") = "currentColor";
     m_currentNode.append_attribute("font-family") = resources->GetTextFont() + ", serif";
     if (this->GetFacsimile()) {
         m_currentNode.append_attribute("viewBox")


src/svgdevicecontext.cpp
```diff
@@ -478,21 +478,21 @@ void SvgDeviceContext::StartPage()
...
-        std::string css = "g.page-margin{font-family:" + resources->GetTextFont()
-            + ",serif;} "
-              "g.ending, g.fing, g.reh, g.tempo{font-weight:bold;} g.dir, g.dynam, "
-              "g.mNum{font-style:italic;} g.label{font-weight:normal;} ellipse, path, polygon, polyline, "
-              "rect{stroke:currentColor} ";
+        std::string css = "g { "
+                          ".ending, .fing, .reh, .tempo {font-weight:bold;} "
+                          ".dir, .dynam, .mNum {font-style:italic;} .label {font-weight:normal;} "
+                          "} "
+                          "ellipse, path, polygon, polyline, rect {stroke:currentColor} ";
```
samuelbradshaw: If I'm reading correctly, this would produce the following CSS (formatted for clarity):
```
g {
  .ending, .fing, .reh, .tempo { font-weight: bold; } 
  .dir, .dynam, .mNum { font-style: italic; }
  .label { font-weight: normal; } 
} 
ellipse, path, polygon, polyline, rect { stroke: currentColor }
```
Unfortunately, nesting CSS rule sets is relatively new in vanilla CSS, and it's only beginning to see wide support in browsers (see [CSS nesting](https://caniuse.com/?search=css+nesting)).

PR #4255, opened 2026-01-01
fix opacity for oblique polygons

rettinghaus:
Small fix for (beam) polygons that were always carrying a fill-opacity attribute.


rettinghaus:
Sure, by default the opacity for Brush is set to 1.0, and never unset (i.e. at -1.0). The drawing method for oblique polygons is the only one where no brush is set (as it is not needed). I hesitated to change the definition of the Brush itself, so this was the most logical change to me.

PR #4263, opened 2026-01-19
do not scope default css

### LINKED ISSUE #4252: Reduce specificity of inserted CSS styles

rettinghaus:
No changes in rendering expected

Closes #4252

### DIFF
--- a/src/svgdevicecontext.cpp
+++ b/src/svgdevicecontext.cpp
@@ -492,7 +492,6 @@ void SvgDeviceContext::StartPage()
         // bounding box css - for debugging
         // css += " g.bounding-box{stroke:red; stroke-width:10} "
         //        "g.content-bounding-box{stroke:blue; stroke-width:10}";
-        this->PrefixCssRules(css);
         m_currentNode.text().set(css);
         m_currentNode = m_svgNodeStack.back();
     }


rettinghaus:
Actually it does, because `font-family` is not in the css anymore.

lpugin:
OK, I'll merge that for now. We can add the `--css-document-scoping` if that is necessary.

PR #4266, opened 2026-01-21
Fix #4265 crashes

### LINKED ISSUE #4265: New MusicXML expansion fails
With the examples in #4264 I found following issues:

`at()` call without checking size:

https://github.com/rism-digital/verovio/blob/834e25ec7b96576bec79034531ec0c08acd3e356/src/iomusxml.cpp#L1412

missing check for `m_target`:

https://github.com/rism-digital/verovio/blob/834e25ec7b96576bec79034531ec0c08acd3e356/src/iomusxml.cpp#L1384

@infojunkie could you fix this please?

infojunkie:
This PR fixes the crashes reported in #4265 and adds a warning when a referenced label is not found.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -1365,11 +1365,9 @@ void MusicXmlInput::CreateExpansion(Section *section)
             }
 
             // when jumping back, keep only last ending
-            if (jumpBack) {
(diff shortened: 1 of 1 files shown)


samuelbradshaw:
Thank you!

PR #4267, opened 2026-01-22
Adjust default for MIDI and time-related outputs

lpugin:
This PR is a follow-up of [this](https://github.com/rism-digital/verovio/discussions/3358) discussion

The options `--expand-first` and `--expand-generate` are dropped in favour of the two options `--expand-always` and `--expand-never`.

The new behaviour is that output to SVG and MEI (all variants) do not process expansions by default. Expansion can be triggered (the equivalent of `--expand-first` or `--expand-generate`) with `--expand-always`.

Oppositely, the output for MIDI and time-related outputs (timemap and exansionmap) processes expansions by default (the equivalent of `--expand-first` or `--expand-generate`). It can be disabled with `--expand-never`.

`--expand xmlid` remains unchanged. The MusicXML importer also remains unchanged, except that `--expand-first` is no longer needed for the MIDI output.

One new feature is that expanded MIDI output can be produced without having to reload the file in the toolkit. This means that, with the default options, loading this [example](https://github.com/rism-digital/verovio/issues/235) will produce the un-expanded version when rendering to SVG but the expanded version when rendering to MIDI and the timemap

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -1625,7 +1625,7 @@ void Doc::ExpandExpansions()
     std::string expansionId = this->GetOptions()->m_expand.GetValue();
     bool expandSelected = (!expansionId.empty());
 
-    // We have no --expand xml:id, so generate an expansion or use the first one (generation will be skipped)
+    // We have no --expand xmlid, so generate an expansion or use the first one (generation will be skipped)
     if (!expandSelected) {
         ListOfObjects scores = this->FindAllDescendantsByType(SCORE);
         for (Object *object : scores) {


samuelbradshaw:
To clarify, is this correct?:
- ` ` (default) will expand nothing
- `--expand xmlid` will expand the MIDI, but not MEI and SVG
- `--expand xmlid --expand-always` will expand MIDI, MEI, and SVG
- `--expand xmlid --expand-never` will expand nothing

Or is it this?:
- ` ` (default) will expand the MIDI, but not MEI and SVG
- `--expand-always` will expand MIDI, MEI, and SVG
- `--expand-never` will expand nothing
- `--expand xmlid` specifies which expansion will be expanded, if your score has multiple expansions, but if you only have one expansion, this isn't needed

PR #4278, opened 2026-02-06
Custom tuning

infojunkie:
This PR adds custom tuning / microtonal support to Verovio's MIDI export. 

The overall mode of operation is as follows:

- An [Ableton tuning definition](https://www.ableton.com/en/live-manual/12/using-tuning-systems/) is specified via either command-line, JavaScript option or embedded in a (valid) MusicXML document. The rationale for choosing this particular format would need detailed explanation, and I will have the pleasure of presenting this work at MEI 2026.
- The tuning definition is parsed by a header-only C++ library called [`tuning-library`](https://github.com/surge-synthesizer/tuning-library) which has been integrated into the build here.
- An instance of new class `CustomTuning` holds the tuning and is added to the corresponding `ScoreDef`.
- The `CustomTuning` instance builds a "note map" that maps the tones of the tuning to MEI-compatible note names and accidentals.
- During MIDI export, 2 things happen:
1. A MIDI MTS (MIDI Tuning Standard) SysEx message is generated with the given tuning definition
2. The `GenerateMIDIFunctor::GetMIDIPitch` method is extended to use the `CustomTuning` note map created above to select the correct (retuned) MIDI key for the outgoing note.

Some details are worth mentioning up-front:

- No invalid MEI or MusicXML extensions are used. To embed tunings within MusicXML, the element [`sound/play/other-play[@type = "tuning-ableton"]`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/other-play/) is used. This allows several tunings to be declared at different locations of the score.  

- The MEI `@tune.temper` attribute is also supported as a tuning type in MusicXML using the same mechanism above (in this case, `sound/play/other-play[@type = "tuning-mei"]`.

- The code does not rely on editing `fonts/supported.xml` to support more SMuFL glyphs. In order to produce a robust note map that supports MEI documents and MusicXML documents including SMuFL, the note map converts all accidentals to SMuFL symbols. To ensure that all SMuFL accidentals are available for mapping, regardless of whether the current loaded font supports those symbols, the class `CustomMapping` uses a copy of the [official `glpyhnames.json`](https://github.com/w3c/smufl/blob/gh-pages/metadata/glyphnames.json) to map symbol names to codepoints and vice-versa. 

- The Emscripten build has been adjusted to support C++ exceptions which are needed by `tuning-library`.

- A [long-standing bug in `midifile`](https://github.com/craigsapp/midifile/pull/103) is fixed to generate valid MIDI SysEx messages (and hence the MTS messages which concern us here).

- Previously, the MusicXML importer used the element `note/pitch/alter` to infer the gestural accidental. This approach does not generalize for microtonality, and has been replaced with another one where the written accidental for each pitch-class is "carried-over" as per the rules of solfege. Key signatures (standard and custom) and multiple accidentals per pitch-class are also supported. This new approach works the same whether or not custom tunings are used.

- I've included the entire `tuning-library` package in this PR, but only 4 C++ headers are actually included in the compilation.  If needed, I can remove all other files are add a README to point to the original repo.

I will add usage examples and test files in the comments below. A [demo version of the code can be used online](https://blog.karimratib.me/demos/musicxml/?sheet=data/sagittal.musicxml).

Note to @lpugin : I suggest to merge squash these changes if / when the time comes, because I've been working on this PR for many months and went through many iterations.

### DIFF
--- a/src/doc.cpp
+++ b/src/doc.cpp
@@ -472,7 +472,8 @@ void Doc::ExportMIDI(smf::MidiFile *midiFile)
         std::ifstream f(midiTuningFile.c_str());
         if (f.good()) {
             const std::string ext(".ascl");
-            if (midiTuningFile.size() >= ext.size() && midiTuningFile.substr(midiTuningFile.size() - ext.size()) == ext) {
+            if (midiTuningFile.size() >= ext.size()
+                && midiTuningFile.substr(midiTuningFile.size() - ext.size()) == ext) {
                 std::stringstream buffer;
                 buffer << f.rdbuf();
                 tuningDef = buffer.str();


infojunkie:
Example 4: [MEI microtonal score with external custom tuning](https://github.com/user-attachments/files/25197614/maqam-rast-without-tuning.zip).

Generated by:
```bash
./verovio -r ../data/ -t midi /path/to/maqam-rast-without-tuning.mei --tuning /path/to/24-edo-mei.ascl
```

lpugin (changes requested):
Thanks for the changes.

I think you can remove the `.clang-format` ./tuning-library.

What about the .gitattributes? Why do we need this?

Concerning the option, it should be rename `--tuningFile`.

infojunkie:
Got this response today from Ableton:

> [..] Please feel free to do what you'd like with the ASCL files. No license or permission is necessary.

Issue #4121, opened 2025-07-10
Enable rendering of CMN notes within `<tabGrp>`


reinierdevalk:
**Is your feature request related to a problem? Please describe.**
This request relates to the E-LAUTE diplomatic transcriptions (notehead notation). These transcriptions show only solid, stemless, noteheads on a CMN staff, with the tablature rhythm flags placed above the staff (see example below). Encoding this currently is a little bit hacky, because we use `<dir>`s for the rhythm flags. 

<img width="928" height="125" alt="Image" src="https://github.com/user-attachments/assets/7ff4f460-c237-4af5-acba-02504f8f0807" />

**Describe the solution you'd like**
Our new proposed solution is to use `<tabGrp>`s (and not `<chord>`s) on the CMN staff, which is allowed in MEI. These `<tabGrp>`s will still contain `<note>`s, but now with attributes `@pname` and `@oct` (and not `@tab.course` and `@tab.fret`). However, when we tried that we noticed that Verovio is **not** rendering these notes. We are therefore requesting these notes to be rendered.

This approach will additionally be extremely useful because it will also enable the use of beamed tablature rhythm flags. 

**Additional context**
The result should look similar to the PNG above, but now with original tablature rhythm flags -- like this: 

<img width="1856" height="249" alt="Image" src="https://github.com/user-attachments/assets/e549bd2a-b9e5-4b41-8051-c9fd3fce075a" />

We made a minimal example to show how we want to encode this: https://gist.githubusercontent.com/janjusolja/07d42be15211cf11461781a41c604e25/raw/b5b6a083f813685ef8b3c37e060c4e0d5d811e45/notehead_example.mei (example created by @janjusolja)

Thank you!

lpugin:
I am not really convinced with this proposal. It it now "just" the rendering of CMN notes, it is CMN (e.g., staves, pitches, etc.). It seems to me that using `dir` is closer to what it actually is. I get the point about tablature beamed flag, which seems to be the only limitation we currently have with `dir`. Does anybody have an idea how we could solve that?

reinierdevalk:
[...]

PR #4285, opened 2026-02-22
Implement issue #4121 CMN notes within tabgrp

paul-bayleaf:
Implements issue #4121 CMN notes within tabgrp

**Important:** This needs libmei to be updated to MEI 6 so that NOTATIONTYPE_tab_staff_like is included.  As a temporary kludge I have edited attconverted.cpp and atttypes.h but these will, of course, be overwritten whenever libmei is re-generated.  Please tell me how to proceed, thanks.

Examples of tab.staff-like notation

Example. Taken from issue #4121 but rendered with this PR.
notehead_example mei

Example. A full piece, the MEI has been generated with a modified luteconv - please excuse the missing key signature and enharmonic note spelling, this information not available in the original tablature and luteconv is not equipped to synthesize it.
JBesard_fantasia_18 tab mei


reinierdevalk:
Thanks @paul-bayleaf, this is starting to look very nice! 

You are probably well aware of this, but the stems for the beamed rhythm symbols, as well as the beams themselves, are much lighter (thinner) than the stems and flags for the unbeamed rhythm symbols. Is this something that will be addressed too?

PR #4298, opened 2026-03-12
Merge tab into develop

lpugin:
@paul-bayleaf I am planning to go ahead with this an make a release. Let me know if you have additional changes in the pipeline.


paul-bayleaf:
@lpugin @reinierdevalk asked "You are probably well aware of this, but the stems for the beamed rhythm symbols, as well as the beams themselves, are much lighter (thinner) than the stems and flags for the unbeamed rhythm symbols. Is this something that will be addressed too?"  This is something I have started to look at but I've not got anything the pipeline yet.

PR #4299, opened 2026-03-12
Drop Verovio light build from CI

rettinghaus:
If I remember correctly we dropped the "light" build, so also removing it from the CI.


craigsapp:
yes, a "light" version is not useful on the command line. If necessary one can build a `nohumdrum` which is still lightish. Since the C++ code requires minimum of about c++21, it needs to be compiled/run on a fairly new (and a resource hungry computer/OS).

PR #4300, opened 2026-03-13
New golang bindings

ahankinson:
Adds a golang binding for Verovio. Mostly self-contained. Includes a CI action to build and run the tests. It uses the c_wrapper as the primary interface to the library. 

A small change was made to the CMakeLists file that now extracts the VERSION from the vrvdefs.h file. This is because there is a new pkg-config file that is used by golang to configure the library, and VERSION is part of that package config. 

Includes a README with some examples of how to build and use the library. 

I've been testing it for a couple days and it seems to work fine.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -243,6 +243,43 @@ jobs:
           name: java-build-${{ github.run_id }}
           path: ${{ github.workspace }}/${{ env.TEMP_DIR }}
 
+
+  ##################################
+  # Build and test the Go bindings #
+  ##################################
+  build_go:
+    # skip job if commit message contains "[skip-ci]"
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
+
+    name: Build Go bindings
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout main repo
+        uses: actions/checkout@v6.0.2
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci_build.yml
```diff
@@ -243,6 +243,43 @@ jobs:
...
+
+      - name: Run Go tests
+        working-directory: ${{ github.workspace }}/bindings/go
+        env:
+          PKG_CONFIG_PATH: ${{ github.workspace }}/build/go
+          LD_LIBRARY_PATH: ${{ github.workspace }}/build/go/install/lib
+          VEROVIO_RESOURCE_PATH: ${{ github.workspace }}/build/go/install/share/verovio
+          GOCACHE: ${{ github.workspace }}/.gocache
+        run: go test ./...
+
```
rettinghaus, line 282: Maybe also upload and provide the artifact?

ahankinson: There isn't an artifact? At least, not one that's useful. golang needs access to the source for a library so it can create a static binary of the go library. But for bindings it also needs access to the shared c library (e.g., verovio.dylib) but this is the same as a normal build.

rettinghaus: Ah. I'm not familiar with Go.

lpugin (commented):
Looks good. The main comment is that the README content should be moved to the Reference book under https://book.verovio.org/installing-or-building-from-sources/

You should open a PR to the develop branch of the book repo and it will be merged once we have made the next release.

PR #4304, opened 2026-03-24
Update fonts

rettinghaus:
This PR updates Leland and Gootville fonts.

Leland updated the font to actually use canonical glyph names, so for generating the files in these cases we need to find the codepoint in `supported_glyphs` first.


ahankinson:
I had Codex review this PR:

```
generate_css still filters glyphs by gname[-4:], so SMuFL-named glyphs are removed from the webfont subset.

  - Code path: fonts/generate.py:223
  - In this PR, Leland.svg now contains many non-uniXXXX glyph names, but this filter was not updated (unlike the fixes added later in __write_xml_glyphs / __write_bb_xml).
  - Verified impact from PR artifacts:
      - develop Leland.css embedded font: 333 glyphs
      - pr-4304 Leland.css embedded font: 47 glyphs
      - (Gootville is fine: 203 -> 210 glyphs)
  - This is a functional regression: most Leland symbols will be missing in web output.
```

I verified that Leland.css is significantly smaller now, so it seems like this is a regression.

PR #4316, opened 2026-04-07
First version of GABC format importer

davidrizo:
## Summary

- Adds a GABC (Gregorian chant notation) importer (`src/iogabc.cpp`, `include/vrv/iogabc.h`)
- Implements the `GABCInput` class following the existing `Input` base pattern
- Wires up format detection and input selection in `src/toolkit.cpp`
- Adds `NO_GABC_SUPPORT` CMake option in `cmake/CMakeLists.txt`

## What is implemented

- Clefs: c1–c4, f2–f4, cb3 (partial)
- Note pitches a–m (normal) and A–M (inclinatum/diamond noteheads)
- Note suffixes: liquescent (~><), quilisma (w/W), oriscus (o/O), accidentals (x/#/y), episema (_), augmentation dots (./..)
- Barlines (`:`) and custos (`Z`/`z`/`z0`)
- Syllable/word structure with lyrics mapped to `Syllable` + `Syl` + `Neume` + `Nc` MEI elements

## Known limitations / TODOs

- GABC auto-detection not yet implemented (format must be specified explicitly)
- `cb3` clef flat sign not added to `StaffDef`
- Multiple clefs in one staff (`c1@c4` syntax) not supported
- Episema not yet stored in MEI output
- `StaffDef lines` hardcoded to 4

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fonts/supported.xml
```diff
@@ -15,7 +15,7 @@
         <glyph glyph-code="E003" smufl-name="bracketTop" />
         <glyph glyph-code="E004" smufl-name="bracketBottom" />
         <!--<glyph glyph-code="E005" smufl-name="reversedBracketTop" />-->
-        <!--<glyph glyph-code="E006" smufl-name="reversedBracketBottom" />-->
+        <!--<glyph glyph-code="E006" smufl-name="reversedBracketBottom" />-->EA29
```
rettinghaus:
```suggestion
        <!--<glyph glyph-code="E006" smufl-name="reversedBracketBottom" />-->
```

davidrizo:
Note this version is not yet final, but I wanted to have a first version, partially worked, uploaded

lpugin:
Thanks, I'll have a look.

rettinghaus:
I think that it would be good to split up this PR into the new rendering method and the importer part. 

NB: The new gylph `medRenStrophicusCMN` should also be present in Leipzig font.

lpugin (changes requested):
A few global comments on style

* Make all methods in GABCInput UpperCamelCase
* Separate member methods (public/private) and member variables (public/private)
* Make static functions in .cpp private static methods in GABCInput

Also, all added .h and .cpp files need to be added to the xcode project. Header files need to be public in VerovioFramework scheme (Target Membership)

include/vrv/iogabc.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef __VRV_IOGABC_H__
+#define __VRV_IOGABC_H__
+
+#include <string>
+#include <vector>
+
+//----------------------------------------------------------------------------
+
+#include "iobase.h"
+#include "pugixml.hpp"
```
lpugin: That's not needed, is it?

include/vrv/iogabc.h
```diff
@@ -0,0 +1,72 @@
...
+#include "vrvdef.h"
+
+namespace vrv {
+class Layer;
```
lpugin, line 22: Add empty line before

include/vrv/iogabc.h
```diff
@@ -0,0 +1,72 @@
...
+    // We have an internal structure with the equivalences of pitches to the G2.
+    // In order to compute the pitch depending on the clef, we apply an offset.
+    // To speed up, when the clef is found, this offset is updated
+    // By default, if clef not present, we use G2 as reference
+    int currentClefPitchOffset;
+
+    Layer *layer;
+
+    void addAccidental(Syllable *syllable, data_ACCIDENTAL_WRITTEN accid, data_PITCHNAME pname, data_OCTAVE oct);
+    void addEpisema(Nc *nc, const std::string form);
```
lpugin: Should be passed as reference parameter (&)

src/iogabc.cpp
```diff
@@ -0,0 +1,710 @@
...
+    if (idx < 0 || idx >= input.size()) {
+        return '\0';
+    }
+    return input[idx];
+}
+
+inline void debug(std::string message)
+{
+    std::cerr << "DEBUG: " << message << "\n";
+}
```
lpugin: Use `LogDebug`

src/iogabc.cpp
```diff
@@ -0,0 +1,710 @@
...
+
+//////////////////////////////////////////////////////////////////////////
+// Note: we've developed this parser thinking of the speed and the ease for others to understand.
+// For easiness to maintain, and due to the simplicity of the encoding,
+// we've avoided here formal parsing techniques or regular expressions.
+// Following other importers, this parser suposes the input is valid.
+//////////////////////////////////////////////////////////////////////////
+// Helper functions
+// It returns '\0' if the index is out of bounds,
+inline char getCharAt(std::string input, int index)
```
lpugin: Pass input as reference parameter

davidrizo:
Sorry for the delay in resolving this PR.

I've fixed the style issues @lpugin.

According to the @rettinghaus suggestion, I've removed by now the medRenStrophicusCMN, to be added later when the GABC import code is finished.

lpugin:
Please put the medRenStrophicusCMN back, it is simpler to have it here.

cmake/CMakeLists.txt
```diff
@@ -190,7 +190,7 @@ set(all_SRC
     ${midi_SRC}
     ../src/json/jsonxx.cc
     ../src/pugi/pugixml.cpp
-        ../include/vrv/iogabc.h
+    ../include/vrv/iogabc.h
```
lpugin, line 193: That should not be needed.

src/iogabc.cpp
```diff
@@ -0,0 +1,705 @@
...
+    assert(index >= 0);
+
+    auto idx = static_cast<std::size_t>(index);
+    if (idx >= input.size()) {
+        return '\0';
+    }
+    return input[idx];
+}
+
+void GABCInput::Debug(const std::string &message)
```
lpugin, line 78: Use LogDebug

src/iogabc.cpp
```diff
@@ -0,0 +1,705 @@
...
+}
+
+std::optional<GABCInput::PitchOctaveType> GABCInput::FindPitch(char ch, int clefPitchOffset)
+{
+    if (ch < 'a' || ch > 'm') {
+        return std::nullopt;
+    }
+
+    // Base c2 mapping: a..m = e3..c5.
+    constexpr int c2BaseDiatonicIndexForA = 3 * 7 + 2; // E3
```
lpugin, line 395:
```suggestion
    const int c2BaseDiatonicIndexForA = 3 * 7 + 2; // E3
```

src/iogabc.cpp
```diff
@@ -0,0 +1,705 @@
...
+
+    Neume *neume = new Neume();
+    Nc *previousNC = nullptr;
+    bool ligateNext = false; // set after OBLIQUE_LIGATURE to mark the following nc as ligated
+    bool lastWasStrophicus = false; // true when last nc was a strophicus
+    data_PITCHNAME strophicusPname = PITCHNAME_NONE;
+    int strophicusOct = 0;
+
+    int currentIndex = 0;
+    while (static_cast<std::size_t>(currentIndex) < music.size()) {
```
lpugin, line 414:
```suggestion
    while (currentIndex < (int)music.size()) {
```

include/vrv/iogabc.h
davidrizo: Line 17 is not required, iobase.h contains the Input class

include/vrv/iogabc.h
davidrizo: I understand that it is preferable to use objects passed by reference rather than pointers:
void addAccidental(Syllable &syllable, data_ACCIDENTAL_WRITTEN accid, data_PITCHNAME pname, data_OCTAVE oct);
    void addEpisema(Nc &nc, const std::string form);

src/iogabc.cpp
davidrizo: Done. I've removed some working logs.... Recall, this is not the final version.

src/iogabc.cpp
davidrizo: I guess you want it for improving the performance. I've added it as a const reference.

cmake/CMakeLists.txt
davidrizo: Sure, removed

src/iogabc.cpp
davidrizo: Removed

src/iogabc.cpp
davidrizo: `constexpr` is used for this kind of things, just like an old `#define`. Do you still prefer `const`?

davidrizo:
> Please put the medRenStrophicusCMN back, it is simpler to have it here.

@lpugin In order to add it correctly, I have to add the glyph to the Leipzig font. It's on other repository. Should I change that font with a PR to that repository, change it here by modifying the original font, or just change the XML related files by using, e.g., the Bravura glyphs?

lpugin (approved):

lpugin:
Let me merge this since this is still into a separate develop branch.

lpugin:
@davidrizo I did some cleanup. You probably want to pull it before making further changes

PR #4317, opened 2026-04-16
Some small improvements

rettinghaus:
This brings a pile of some small improvements:

* Unfiy function naming in `cpp.py` and add some basic docstrings
* Update Python projects to drop Python 3.9 , which reached eol
* Update poetry lock files
* Format JSON files


libmei/tools/cpp.py
```diff
@@ -643,20 +643,22 @@ def vrv_get_att_config_default(module: str, gp: str, att: str) -> Optional[str]:
...
+def vrv_get_formatted_type(datatype: str) -> str:
+    """Get the formatted datatype."""
     if datatype in DATATYPES["mapped"]:
         return DATATYPES["mapped"][datatype]
     return datatype.replace(".", "_")
 
 
-def vrv_getformattedvallist(att: str, vallist: str) -> str:
+def vrv_get_formatted_vallist(att: str, vallist: str) -> str:
+    """Get the formatted vallist."""
```
ahankinson: Docstrings should be meaningful. These seem to just repeat the name of the function?

ahankinson (commented):
🤷 it's ok, I guess? Seems like a lot of effort for very little effect.

The cpp changes might be a bit more useful, but they should also be in their own PR so that they could be reverted if needed.

libmei/tools/cpp.py
rettinghaus: I agree, but the exisiting ones did the same.

rettinghaus: Ok, did another round on touching up docstrings.

PR #4320, opened 2026-04-22
avoid copy on string manipulation

rettinghaus:


src/ioabc.cpp
```diff
@@ -652,7 +652,7 @@ void ABCInput::ParseKey(std::string &keyString)
             std::string modeString(&keyString[i]);
             // capitalization is ignored for the modes
             // and in fact only the first three letters of each mode are parsed
-            modeString = modeString.substr(0, 3);
+            modeString.resize(3);
```
craigsapp, line 655: I don't understand this line: If modeString is "A" what happens? resize(3) will add two uninitialized chars after the "A", then the loop after it will process these uninitialized chars.

src/ioabc.cpp
```diff
@@ -1183,7 +1183,7 @@ void ABCInput::ReadInformationField(const char &dataKey, std::string value)
     if (dataKey == '%' || dataKey == '\0')
```
craigsapp, line 1183: general verovio style?: 

```
 if (dataKey == '%' || dataKey == '\0') return;
```

Or

```
 if (dataKey == '%' || dataKey == '\0') {
    return;
}
```

Are you using AI? 😜

src/ioabc.cpp
```diff
@@ -1183,7 +1183,7 @@ void ABCInput::ReadInformationField(const char &dataKey, std::string value)
...
         return;
     else if (value.find('%') != std::string::npos) {
-        value = value.substr(0, value.find('%'));
+        value.resize(value.find('%'));
```
craigsapp, line 1186: Preferably store value.find('%') in a variable to reuse on lines 1185 (if-statement) and 1186 (truncate string).i.e.: no need to `value.find('%')` twice (could take just as long than copy it is fixing 😜 )

PR #4326, opened 2026-05-04
Unify string conversion in libmei

rettinghaus:
Getting rid of string manipulation and casting.

### DIFF
--- a/libmei/addons/att.cpp
+++ b/libmei/addons/att.cpp
@@ -51,12 +51,12 @@ std::string Att::VUToStr(data_VU data) const
 
 double Att::StrToDbl(const std::string &value) const
 {
-    return atof(value.c_str());
+    return stof(value);
 }
 
 int Att::StrToInt(const std::string &value) const
 {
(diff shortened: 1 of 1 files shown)


libmei/addons/att.cpp
```diff
@@ -700,7 +701,7 @@ data_PERCENT Att::StrToPercent(const std::string &value, bool logWarning) const
         if (logWarning) LogWarning("Unsupported data.PERCENT '%s'", value.c_str());
         return 0;
     }
-    return atof(value.substr(0, value.find("%")).c_str());
+    return stof(value);
```
lpugin, line 704: Silently ignoring exceptions is a bad practice

rettinghaus: What do you mean? We do a regex test before.

lpugin:
Can you check the change yielded in the test-suite?

rettinghaus:
There is a change in a tie test, that is not really indistinguishable.

PR #4328, opened 2026-05-08
New: always set breaks: none for PAE

ahankinson:
This provides a significant processing speedup since it skips all the castoff and layout decisions.

Benchmarking shows that this provides a 2.73x speedup in incipit parsing.

I experimented with only setting breaks: none if the user didn't explicitly set breaks in the options, but I don't think it's possible to detect between unset and `BREAKS_auto`. Suggestions on how to do this are welcome!

### DIFF
--- a/src/toolkit.cpp
+++ b/src/toolkit.cpp
@@ -843,6 +843,10 @@ bool Toolkit::LoadData(const std::string &data, bool resetLogBuffer)
     // might have been ignored because of the --breaks auto option.
     // Regardless, we won't do layout if the --breaks none option was set.
     int breaks = m_options->m_breaks.GetValue();
+    if (inputFrom == PAE) {
+        // PAE loading intentionally skips cast-off/layout for faster incipit processing.
+        breaks = BREAKS_none;
+    }
 
     // When loading page-based MEI, the layout is marked as done
     // In this case, we do not cast-off the document (breaks is expected to be not set)


rettinghaus (approved):

lpugin:
I think this is problematic because long incipits will always be rendered on one line. Am I wrong? @rettinghaus since you approved it, can you comment?

rettinghaus:
I think it makes sense for incipits to be restricted to one line. How long can an incipit be? Is there a specific use case that gives you a headache, @lpugin?

@ahankinson One possibility would be to add `BREAKS_unset` to the enum and check for that. In this case set `BREAKS_none` for PAE, and `BREAKS_auto` for everything alse.

lpugin:
Where is it said that PAE is only for incipits?

rettinghaus:
Provocative question. Is there a case where PAE is used for more?

IAML states:

> The Plaine & Easie Code is a library standard that enables entering music incipits in modern or mensural notation.

lpugin:
So what? Is it forbidden to have an incipit on more than one line?

ahankinson:
Setting it to UNSET and then checking the value prior to setting it to AUTO sounds good? Any reason why that wouldn’t work?

lpugin:
Is it a problem to set it to `none` when indexing?

ahankinson:
I just figured that 99% of the time, PAE incipits won’t be long enough to require the whole layout process to be applied to them, so I was optimizing for the most likely scenario.

lpugin:
Sure. But we have nowhere the concept of an `unset` option, and `--breaks` is an important one, so I am not convinced the proposal is a good idea. I'll think about it

lpugin:
To be rediscussed if necessary.

Issue #4331, opened 2026-05-13
`<barLine/>` in `staffDef@notationtype="tab.lute.german|french|italian"` gets rendered with `path@stroke-width="47"`, not 27


Brathenning:
**Describe the problem**
With `staffDef@notationtype="tab.lute.german|french|italian"` `<barLine>` is almost twice as thick ( `path@stroke-width="47"`, not 27) as the bar line of `<measure>` and also too much to the right.

<img width="344" height="236" alt="Image" src="https://github.com/user-attachments/assets/9dce88dd-a4c7-4036-ae58-25d9278490f5" />

<img width="356" height="285" alt="Image" src="https://github.com/user-attachments/assets/4d2475f8-efe0-4b94-b624-651cfe996d37" />

<img width="348" height="299" alt="Image" src="https://github.com/user-attachments/assets/2b3957b5-f099-4bf0-9c83-dec292cc57cc" />

**To Reproduce**
1. Load the data attached
2. Render it
3. See error

**Expected behavior**
`<barLine>` should be as thick as the other bar lines and further to the right.

<img width="336" height="206" alt="Image" src="https://github.com/user-attachments/assets/dc44acd9-22de-4489-92fa-b3195684d406" />

(manually adjusted `path` in the svg)

**Input data**
```xml
<?xml version='1.0' encoding='UTF-8'?>
<?xml-model href="https://music-encoding.org/schema/5.1/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.1/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.1">
	<meiHead>
		<fileDesc>
			<titleStmt>
				<title>Minimal GLT example</title>
			</titleStmt>
			<pubStmt/>
		</fileDesc>
		<encodingDesc>
			<appInfo/>
		</encodingDesc>
	</meiHead>
	<music xml:id="m8m56cx2">
      <body xml:id="b73ctgr2">
         <mdiv xml:id="mzwi59s2" n="1">
            <score xml:id="seqzfpg1">
               <scoreDef xml:id="s9eahny1" mnum.visible="false">
                  <staffGrp xml:id="srkc6t21">
[...]

PR #4334, opened 2026-05-14
Adjust m_drawingStaffSize for German tablature.  Correct rounding. Fix issue  #4331

paul-bayleaf:
Adjust m_drawingStaffSize for German tablature.  Correct rounding. Fix  issue #4331.  Bar line width now 27 in svg.

Also, the lack of rounding in staff.cpp was causing a 1 pixel error.


lpugin:
Thanks!

PR #4336, opened 2026-05-16
Implement key signature parsing for v2 (WIP)

lpugin:
We need to find a solution for storing the fact that a `$n` is supplied (`$[n]`). With other key signatures (`x` and `b`), individual accidental are marked as enclosed, so we can infer that the full key signature is supplied of not by checking if all accidentals are enclosed. However, since we have no accidental with that one, the information is currently lost.

We should not use editorial markup and also not `keySig@type`. @ahankinson Any thoughts how this can be represented in MEI-Basic?

### DIFF
--- a/include/vrv/iopae.h
+++ b/include/vrv/iopae.h
@@ -627,7 +627,8 @@ private:
      * @name Methods that parse sub string instantiate corresponding objects
      */
     ///@{
-    bool ParseKeySig(KeySig *keySig, const std::string &paeStr, pae::Token &token);
+    bool ParseKeySigV1(KeySig *keySig, const std::string &paeStr, pae::Token &token);
+    bool ParseKeySigV2(KeySig *keySig, const std::string &paeStr, pae::Token &token);
     bool ParseClef(Clef *clef, const std::string &paeStr, pae::Token &token, bool scoreDef = false);
     bool ParseMeterSig(MeterSig *meterSig, const std::string &paeStr, pae::Token &token);
     bool ParseMensur(Mensur *mensur, const std::string &paeStr, pae::Token &token);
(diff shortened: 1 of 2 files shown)


lpugin:
Merging into develop-pae2 now that the PR is there for the record.

ahankinson:
Is there a reason it has to use MEI Basic, and not full MEI?

As far as I can see, Basic doesn’t support any “supplied” markup for anything, so that would be a problem for supplied clefs and other supplied keysig accidentals too?

ahankinson:
If $n is provided then staffDef@keysig=0 would make sense to encode the presence of the key signature as 0 on the circle of fifths, but with no impact on showing any accidentals. This is available in Basic as far as I can tell?

This would also be round-tripable since $n = keysig 0, but no keysig in PAE would omit any keysig attribute in MEI.

lpugin:
MEI-Basic would give a better interoperability (e.g., MuseScore), so it would be good. It would be a shame to have to switch to MEI full just for supplied `$n` key signatures.

I have to check, but I don't think `staffDef@keysig` in addition to `keySig` (if I understand correctly) would work because internally Verovio promotes attributes to element, so we would end up with two keySig elements.

ahankinson:
> It would be a shame to have to switch to MEI full just for supplied $n key signatures.

how are other supplied accidentals in PAE encoded in MEI Basic?

ahankinson:
`keySig` isn’t available in MEI Basic as far as I can see?

https://music-encoding.org/guidelines/dev/mei-basic/elements.html

lpugin:
Yes, that's right. MEI full then.

lpugin:
For now I added `@type="pae-v2-supplied"`. It will make round tripping possible, and it is also there on `$x` and `$b` key signatures.

https://github.com/rism-digital/verovio/commit/6f6b49e5176c260d09b2b795180cc96dd4039636

It can be changed to `keySig@enclose` if this is one day added to MEI.

Issue #4340, opened 2026-05-22
Importing from MusicXML


rettinghaus:
Importing from MusicXML I get tons of these errors:

> [Error] Custom tuning: Invalid or missing file glyphnames.json
[Error] Custom tuning: SMuFL glyph U+E260 not found in glyph table

Also there are nowe strange artifacts with lonely 

`<accid/>`

tags.

lpugin:
I don't

lpugin:
Provide some data to test please.

rettinghaus:
[test.musicxml.zip](https://github.com/user-attachments/files/28157904/test.musicxml.zip)

rettinghaus:
First note from conversion:

```xml
<note xml:id="ggbtzm8" dur.ppq="8" dur="breve" oct="4" pname="g">
   <accid xml:id="hth78ra" />
   <verse xml:id="i1gi6d3g" n="1">
      <syl xml:id="j15qxwld" con="s" wordpos="s">Der Herr ist mein</syl>
   </verse>
   <verse xml:id="l1e470lr" n="2">
      <syl xml:id="m1c8mmwk" con="s" wordpos="s">Er erquicket meine</syl>
   </verse>
   <verse xml:id="o1yzxu50" n="3">
      <syl xml:id="p1akx2g3" con="s" wordpos="s">Und ob ich schon wanderte im</syl>
   </verse>
   <verse xml:id="r4wvuwv" n="4">
      <syl xml:id="s1m0a09p" con="s" wordpos="s">Du bereitest vor mir</syl>
   </verse>
   <verse xml:id="u950cl4" n="5">
      <syl xml:id="v1hasdzs" con="d" wordpos="i">Gutes und Barm</syl>
   </verse>
   <verse xml:id="x12guqb1" n="6">
      <syl xml:id="y1p4fkm3" con="s" wordpos="s">Ehre sei dem Vater</syl>
   </verse>
</note>
```

rettinghaus:
In other cases every "normal" accidental carries a `glyph.auth="smufl"`.

lpugin:
Why don't you provide a bit more information on the version and the binding / platform ?

rettinghaus:
macOS, CLI, latest stable, homebrew version.

Same for local build of latest devlop.

rettinghaus:
### MusicXML:

```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #4341, opened 2026-05-22
[MusicXML] fix accid import

rettinghaus:
With Release 6.2.1 the errors on import are gone, but the problem of empty accid elements or useless/unneeded glyph.* attributes remained. This PR provides a fix for a cleaner conversion to MEI.

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3165,7 +3165,7 @@ void MusicXmlInput::ReadMusicXmlNote(
                             // the MEI equivalent of the written accidental. The custom tuning will always choose
                             // the SMuFL glyph over the gestural or written accidentals.
                             accid->SetAccidGes(Att::AccidentalWrittenToGestural(current.m_accid));
-                            if (accid->HasAccid()) {
+                            if (!current.m_glyphName.empty()) {
                                 accid->IsAttribute(false);
                                 if (!current.m_glyphName.empty()) {
                                     accid->SetGlyphName(current.m_glyphName);
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3158,23 +3158,26 @@ void MusicXmlInput::ReadMusicXmlNote(
                     try {
                         for (const auto &current : m_currentAccids.at(note->GetPname())) {
                             Accid *accid = new Accid();
-                            note->AddChild(accid);
-                            accid->IsAttribute(false);
+                            accid->IsAttribute(true);
 
                             // to make sure the new *gestural* accidental conforms to the carried-over *written*
                             // accidental, we translate the latter to a SMuFL glyph and set the gestural accidental to
                             // the MEI equivalent of the written accidental. The custom tuning will always choose
                             // the SMuFL glyph over the gestural or written accidentals.
                             accid->SetAccidGes(Att::AccidentalWrittenToGestural(current.m_accid));
-                            if (!current.m_glyphName.empty()) {
-                                accid->SetGlyphName(current.m_glyphName);
-                                accid->SetGlyphAuth(current.m_glyphAuth);
+                            if (accid->HasAccid()) {
+                                accid->IsAttribute(false);
+                                if (!current.m_glyphName.empty()) {
+                                    accid->SetGlyphName(current.m_glyphName);
+                                    accid->SetGlyphAuth(current.m_glyphAuth);
+                                }
+                                else if (current.m_accid != ACCIDENTAL_WRITTEN_NONE) {
+                                    char32_t glyph = Accid::GetAccidGlyph(current.m_accid);
+                                    accid->SetGlyphName(CustomTuning::GetGlyphName(glyph, m_doc));
+                                    accid->SetGlyphAuth("smufl");
+                                }
                             }
-                            else if (current.m_accid != ACCIDENTAL_WRITTEN_NONE) {
-                                char32_t glyph = Accid::GetAccidGlyph(current.m_accid);
-                                accid->SetGlyphName(CustomTuning::GetGlyphName(glyph, m_doc));
-                                accid->SetGlyphAuth("smufl");
-                            }
+                            note->AddChild(accid);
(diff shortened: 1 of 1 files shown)


lpugin:
@infojunkie please have a look

infojunkie:
In the absence of tests, I will need to re-run all the validations that were done for the custom tuning feature.

I will likely be unable to review this until after MEC 2026 is finished.

To provide some context, can @rettinghaus show what he means by:

> the problem of empty accid elements or useless/unneeded glyph.* attributes

With examples please?

lpugin:
> I will likely be unable to review this until after MEC 2026 is finished.

That's OK

> With examples please?

There is an example [here](https://github.com/rism-digital/verovio/issues/4340#issuecomment-4521338316) (and some more background in the issue)

infojunkie:
Thanks, that should be enough. I'll update this ticket soon as I have time to review.

infojunkie:
Sorry for the delay. I tested this PR against the test files I'd created for the microtonal PR, and one test failed - I'm still debugging where the failure is coming from. Thanks for your patience.

src/iomusxml.cpp
```diff
@@ -3158,23 +3158,26 @@ void MusicXmlInput::ReadMusicXmlNote(
...
                             accid->SetAccidGes(Att::AccidentalWrittenToGestural(current.m_accid));
-                            if (!current.m_glyphName.empty()) {
-                                accid->SetGlyphName(current.m_glyphName);
-                                accid->SetGlyphAuth(current.m_glyphAuth);
-                            }
-                            else if (current.m_accid != ACCIDENTAL_WRITTEN_NONE) {
-                                char32_t glyph = Accid::GetAccidGlyph(current.m_accid);
-                                accid->SetGlyphName(CustomTuning::GetGlyphName(glyph, m_doc));
-                                accid->SetGlyphAuth("smufl");
+                            if (accid->HasAccid()) {
```
infojunkie: This line will always be false because `accid` is instantiated just above. The correct condition to preserve the carried-over accidentals should be:

```c++
if (!current.m_glyphName.empty() || current.m_accid != ACCIDENTAL_WRITTEN_NONE) {
```

This line will explicitly add `glyph.smufl` and `glyph.auth` that are needed to properly map the notes to a MIDI tuning. It can happen that these glyphs are redundant compared to the standard set of gestural accidentals, but at least now empty accids will never be created.

We can do further work to avoid adding redundant glyphs, so if this desired I can keep working on this PR. I would need access to this fork or make my own fork. Let me know what you prefer @rettinghaus and @lpugin .

lpugin: Thanks. I can merge this can you can make another PR?

lpugin: Otherwise just make another PR that replaces that one.

infojunkie: This PR as it stands breaks some microtonal scores so it would be best to avoid merging it altogether, or merge it after the change I suggest above. In either case I need to do further work.

rettinghaus: @infojunkie your suggested fix still convolutes purely gestural accidentals with visible information `glyph.auth="smufl" glyph.name="accidentalFlat"`.

infojunkie: > your suggested fix still convolutes purely gestural accidentals with visible information

I agree. That's due to a limitation of the MEI gestural accidentals which accommodate neither the full set of written accidentals, nor SMuFL accidentals which are a superset of both. The gestural domain is therefore incapable of expressing the full set of accidentals that might be defined in a tuning. Please refer to [this active MEI issue](https://github.com/music-encoding/music-encoding/issues/1167) which will hopefully address this limitation.

In the meantime, the microtonal feature relies on the ability to add SMuFL glyphs during MusicXML import. Whereas we can make special cases for gestural accidentals such as "f" <=> "accidentalFlat", the cases where the incoming accidentals are not part of gestural accidentals would be broken unless we add my fix above.

You can test it youself with this minimal example:

[github-4340-3.zip](https://github.com/user-attachments/files/28602998/github-4340-3.zip)

When exporting the MusicXML to MEI, you can see the difference between the `develop` branch (`github-4340-3.blessed.mei`) and this PR (`github-4340-3.mei`).

rettinghaus: Sorry, but I still don't understand completely what you are trying to do here. Gestural accidentals are (very) basically there to reach the black keys on a keyboard and to match the indication of the key signature. If there is no accidental written and the sounding pitch doesn't match the written this should be somehow defined in the `instrDef` or at least be an inherant information on the note itself, and not something on an added invisble accidental.

infojunkie: I tried to explain above that injecting those smufl glyphs is a way to persist the exact incoming accidentals that are lost during the written-to-gestural translation. The exact accidentals are needed because the custom tuning relies on them to map score notes to their tuning degrees and MIDI keys.

This is a [huge PR that was merged 3 months ago after 6 months of work and one month of reviews](https://github.com/rism-digital/verovio/pull/4278). I suggest that you go through its code to better understand the logic. You can also read the [short paper I presented at MEC 2026](https://docs.google.com/document/d/1JCFh__g4e_MI3X3N4WCMky05f9mkB1QD6DAa0G6MUcA/edit?usp=sharing) to understand the context of those changes.

In any case, now that this code exists in the main branch, I think the relevant point here is to avoid regressing the functionality. I've supplied a minimal example to show that this present PR does introduce a regression, and I've suggested a fix to avoid the regression, while honoring the spirit of this PR. I am also willing to go further and add logic to the custom tuning code to narrow down even more the injection of glyphs, for cases where today's gestural accidentals would be enough for the custom tuning.

Beyond that, I don't think I can help further until the [work of modeling custom tunings within MEI](https://github.com/music-encoding/music-encoding/issues/1167) is finalized and ready for implementation in Verovio.

lpugin: We should definitely fix the empty accidental issue, and I hope #4356 does it.

PR #4342, opened 2026-05-23
Improve render performance

sonovice:
This improves rendering performance by reducing repeated work in layout and drawing hot paths.

The changes are split into small commits so they can be reviewed independently. No output changes expected.

Benchmark corpus: 22 full scores from the Verovio `gh-pages` examples, mixing MEI and MusicXML, rendering all pages to SVG with a Release CLI:

```sh
verovio -r data --log-level error -f mei-or-musicxml -x 1 -t svg -a -o /tmp/out input
```

| Optimization | Commit | Real | Incremental | Total speedup | Instructions | Instruction reduction | Max RSS |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Baseline `develop` | `86c6476f` | 10.91 s | - | - | 163.62 B | - | 381.0 MiB |
| Lazy custom tuning | `773d48a6` | 9.50 s | 12.9% | 12.9% | 139.77 B | 14.6% | 369.8 MiB |
| Trim horizontal passes | `4bdb8edb` | 8.89 s | 6.4% | 18.5% | 126.66 B | 22.6% | 355.4 MiB |
| Simplify generated measure numbers | `85e08de2` | 8.80 s | 1.0% | 19.3% | 126.36 B | 22.8% | 365.6 MiB |
| Find page boundaries directly | `d4e8e263` | 8.73 s | 0.8% | 20.0% | 126.31 B | 22.8% | 373.5 MiB |
| Reduce draw dispatch overhead | `e36440bb` | 8.28 s | 5.1% | 24.1% | 118.08 B | 27.8% | 368.3 MiB |
| Reduce repeated page drawing traversals | `140ab7b3` | 7.95 s | 4.0% | 27.1% | 113.74 B | 30.5% | 348.7 MiB |

Final result on this corpus: 10.91 s -> 7.95 s, a 27.1% wall-time reduction. Retired instructions are down 30.5%. Peak RSS did not increase in this run.

Validation:

* Full-piece benchmark corpus: 412 baseline SVGs vs 412 final SVGs, normalized mismatches 0
* Verovio `_tests` + `musicxmlTestSuite`: 763 baseline SVGs vs 763 final SVGs, normalized mismatches 0
* Only normalization was the version-specific text inside the SVG `<desc>` element
* `ctest` was run against the final CMake build, but this build exposes no tests

The benchmarked full-score set was:

* `editor/brahms.mei`
* `examples/downloads/Ahle_Jesu_meines_Herzens_Freud.mei`
* `examples/downloads/Bach_Chorale_bwv.363.mei`
* `examples/downloads/Beethoven_op.18.mei`
* `examples/downloads/Beethoven_op.18_1.mei`
* `examples/downloads/Brahms_StringQuartet_Op51_No1.mei`
* `examples/downloads/Chopin_Etude_op.10_no.9.mei`
* `examples/downloads/Guami_Canzona_24.mei`
* `examples/downloads/Hummel_Concerto_for_trumpet.mei`
* `examples/downloads/Krebs_Trio_Eb_2.mei`
* `examples/downloads/Marenzio_Sesto_a_5_01.mei`
* `examples/downloads/Schubert_Lindenbaum.mei`
* `examples/musicxml/Bach_Brandburg_Concerto_BWV.1050_M1.xml`
* `examples/musicxml/Bach_Invention_No_1_BVW.772.xml`
* `examples/musicxml/Bach_Nun_komm_der_Heiden_Heiland_BWV.659.xml`
* `examples/musicxml/Chopin_Etude_Op.10_No.12.xml`
* `examples/musicxml/Costeley_Je_vois_de_glissantes_eaux.xml`
* `examples/musicxml/Haydn_String_Quartet_in_C_Major_Op.74_No.1_M2.xml`
* `examples/musicxml/Schubert_An_die_Sonne_D.439.xml`
* `examples/musicxml/Schubert_Staendchen_D.923.xml`
* `examples/musicxml/Vivaldi_Concerto_No.4_in_F_Minor_Winter.xml`
* `examples/musicxml/Waldteufel_Op.138_Les_Patineurs.xml`

### DIFF
--- a/include/vrv/customtuning.h
+++ b/include/vrv/customtuning.h
@@ -12,7 +12,9 @@
 
 //----------------------------------------------------------------------------
 
+#include <cassert>
 #include <map>
+#include <memory>
 
 //----------------------------------------------------------------------------
 
@@ -29,14 +31,26 @@ class CustomTuning {
 public:
     CustomTuning() {}
     CustomTuning(const std::string &tuningDef, Doc *doc, bool useMusicXmlAccidentals);
+    CustomTuning(const CustomTuning &other);
+    CustomTuning(CustomTuning &&other) noexcept = default;
+    CustomTuning &operator=(const CustomTuning &other);
+    CustomTuning &operator=(CustomTuning &&other) noexcept = default;
 
     /**
      * @name Getters and validity checkers
      */
     ///@{
-    bool IsValid() const { return m_tuning.notationMapping.count > 0; }
-    Tunings::Tuning &GetTuning() { return m_tuning; }
-    const Tunings::Tuning &GetTuning() const { return m_tuning; }
+    bool IsValid() const { return m_tuning && m_tuning->notationMapping.count > 0; }
+    Tunings::Tuning &GetTuning()
+    {
+        assert(m_tuning);
+        return *m_tuning;
+    }
+    const Tunings::Tuning &GetTuning() const
+    {
+        assert(m_tuning);
+        return *m_tuning;
+    }
     std::map<std::string, std::string> &GetNoteMap() { return m_noteMap; }
     const std::map<std::string, std::string> &GetNoteMap() const { return m_noteMap; }
     ///@}
@@ -75,7 +89,7 @@ public:
     //
 
 private:
-    Tunings::Tuning m_tuning;
+    std::unique_ptr<Tunings::Tuning> m_tuning;
     std::map<std::string, std::string> m_noteMap;
 
     //----------------//
--- a/include/vrv/customtuning.h
+++ b/include/vrv/customtuning.h
@@ -14,7 +14,6 @@
 
 #include <cassert>
 #include <map>
-#include <memory>
 
 //----------------------------------------------------------------------------
 
@@ -31,10 +30,11 @@ class CustomTuning {
 public:
     CustomTuning() {}
     CustomTuning(const std::string &tuningDef, Doc *doc, bool useMusicXmlAccidentals);
+    ~CustomTuning();
     CustomTuning(const CustomTuning &other);
-    CustomTuning(CustomTuning &&other) noexcept = default;
+    CustomTuning(CustomTuning &&other) noexcept;
     CustomTuning &operator=(const CustomTuning &other);
-    CustomTuning &operator=(CustomTuning &&other) noexcept = default;
+    CustomTuning &operator=(CustomTuning &&other) noexcept;
 
     /**
      * @name Getters and validity checkers
@@ -89,7 +89,7 @@ public:
     //
 
 private:
-    std::unique_ptr<Tunings::Tuning> m_tuning;
+    Tunings::Tuning *m_tuning = NULL;
     std::map<std::string, std::string> m_noteMap;
 
     //----------------//
--- a/src/adjustxposfunctor.cpp
+++ b/src/adjustxposfunctor.cpp
@@ -30,6 +30,7 @@ AdjustXPosFunctor::AdjustXPosFunctor(Doc *doc) : DocFunctor(doc)
     m_cumulatedXShift = 0;
     m_staffN = 0;
     m_staffSize = 100;
+    m_hasExcludedElements = false;
     m_rightBarLinesOnly = false;
     m_measure = NULL;
 }
@@ -94,6 +95,12 @@ FunctorCode AdjustXPosFunctor::VisitLayerElement(LayerElement *layerElement)
 {
     if (layerElement->IsScoreDefElement()) return FUNCTOR_SIBLINGS;
 
+    // Track excluded elements while we are already traversing the page so the caller can avoid a separate scan.
+    if (!m_excludes.empty() && layerElement->Is(m_excludes)) {
+        m_hasExcludedElements = true;
+        return FUNCTOR_CONTINUE;
+    }
+
     // we should have processed aligned before
     assert(layerElement->GetAlignment());
 
@@ -101,11 +108,6 @@ FunctorCode AdjustXPosFunctor::VisitLayerElement(LayerElement *layerElement)
         // if nothing to do with this type of element
         // this happens for example with Artic where only ArticPart children are aligned
         return FUNCTOR_SIBLINGS;
-    }
-
-    // If we have a list of types to exclude and it is one of them, stop it
-    if (!m_excludes.empty() && layerElement->Is(m_excludes)) {
-        return FUNCTOR_CONTINUE;
     }
 
     // If we have a list of types to include and it is not one of them, stop it
@@ -160,7 +162,7 @@ FunctorCode AdjustXPosFunctor::VisitLayerElement(LayerElement *layerElement)
     Alignment *nextAlignment = vrv_cast<Alignment *>(
         layerElement->GetAlignment()->GetParent()->GetNext(layerElement->GetAlignment(), ALIGNMENT));
     AlignmentType next = nextAlignment ? nextAlignment->GetType() : ALIGNMENT_DEFAULT;
-    if (layerElement->Is({ DOTS, FLAG }) && currentReference->HasMultipleLayer()
+    if ((layerElement->Is(DOTS) || layerElement->Is(FLAG)) && currentReference->HasMultipleLayer()
         && (next != ALIGNMENT_MEASURE_RIGHT_BARLINE)) {
         const int additionalOffset = selfRight - m_upcomingMinPos;
         if (additionalOffset > m_currentAlignment.m_offset) {
@@ -221,6 +223,8 @@ FunctorCode AdjustXPosFunctor::VisitMeasure(Measure *measure)
     Filters filters;
     Filters *previousFilters = this->SetFilters(&filters);
 
+    m_measureTieEndpoints = measure->GetInternalTieEndpoints();
+
     for (auto staffN : m_staffNs) {
         m_minPos = 0;
         m_upcomingMinPos = VRV_UNSET;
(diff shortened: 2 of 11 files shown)


lpugin:
Thanks! I will review each step, and it will take a bit of time.

include/vrv/customtuning.h
```diff
@@ -75,7 +89,7 @@ class CustomTuning {
     //
 
 private:
-    Tunings::Tuning m_tuning;
+    std::unique_ptr<Tunings::Tuning> m_tuning;
```
lpugin: Can't we have just a `Tuning *` ? We usually do not use `unique_ptr`and the memory management should be clear.

src/adjustxposfunctor.cpp
```diff
@@ -160,7 +162,7 @@ FunctorCode AdjustXPosFunctor::VisitLayerElement(LayerElement *layerElement)
     Alignment *nextAlignment = vrv_cast<Alignment *>(
         layerElement->GetAlignment()->GetParent()->GetNext(layerElement->GetAlignment(), ALIGNMENT));
     AlignmentType next = nextAlignment ? nextAlignment->GetType() : ALIGNMENT_DEFAULT;
-    if (layerElement->Is({ DOTS, FLAG }) && currentReference->HasMultipleLayer()
+    if ((layerElement->Is(DOTS) || layerElement->Is(FLAG)) && currentReference->HasMultipleLayer()
```
lpugin, line 165: Why changing this?

src/adjustxposfunctor.cpp
```diff
@@ -402,8 +405,8 @@ std::pair<int, int> AdjustXPosFunctor::CalculateXPosOffset(LayerElement *layerEl
         if (!overlap) {
             // if last element of the tuplet is rest, make sure there is sufficient distance between it and next
             // note/chord (for ledger lines)
-            if (layerElement->Is({ NOTE, CHORD }) && !layerElement->GetFirstAncestor(TUPLET) && bboxElement->Is(REST)
-                && bboxElement->GetFirstAncestor(TUPLET)) {
+            if ((layerElement->Is(NOTE) || layerElement->Is(CHORD)) && !layerElement->GetFirstAncestor(TUPLET)
```
lpugin, line 408: Same here. Why changing `layerElement->Is({ NOTE, CHORD })` ?

src/view_control.cpp
```diff
@@ -79,8 +114,7 @@ void View::DrawControlElement(DeviceContext *dc, ControlElement *element, Measur
     this->StartOffset(dc, element, 100);
 
     // For dir, dynam, fermata, and harm, we do not consider the @tstamp2 for rendering
-    if (element->Is({ ANNOTSCORE, BEAMSPAN, BRACKETSPAN, FIGURE, GLISS, HAIRPIN, LV, OCTAVE, PHRASE, PITCHINFLECTION,
-            SLUR, TIE })) {
+    if (IsPostponedTimeSpanningControlElement(element->GetClassId())) {
```
lpugin, line 117: What is the impact of that change? Cant' we optimize `BoundingBox::Is(const std::vector<ClassId> &classIds)`

src/view_control.cpp
```diff
@@ -335,98 +371,78 @@ void View::DrawTimeSpanningElement(DeviceContext *dc, Object *element, System *s
...
-        else if (element->Is(TIE)) {
-            // For ties we limit support to one value in @staff
-            if (!isFirst) continue;
-            // cast to Slur check in DrawTie
-            this->DrawTie(dc, dynamic_cast<Tie *>(element), x1, x2, staff, spanningType, graphic);
-        }
-        else if (element->Is(TRILL)) {
-            // cast to Trill check in DrawTrill
-            this->DrawTrillExtension(dc, dynamic_cast<Trill *>(element), x1, x2, staff, spanningType, graphic);
+        switch (classId) {
```
lpugin, line 379: I suspect the gain here to be mostly due to the change form `dynamic_cast` to `vrv_cast`, so only changing that should be enough

For the record, the `classId` mechanism has been added to Verovio precisely because RTTI was too slow. See [this](https://github.com/rism-digital/verovio/commit/db6ab53c27432c5d5326afda121b6db8fead14b8) commit more than ten years ago. 

There are only a few places where we still rely on it and where we need a `dynamic_cast`. That place is **not** one of them and has probably been overlooked when changing for RTTI to classIds.

src/view_element.cpp
```diff
@@ -891,20 +885,23 @@ void View::DrawDurationElement(DeviceContext *dc, LayerElement *element, Layer *
...
-    else if (dynamic_cast<Note *>(element)) {
-        dc->StartGraphic(element, "", element->GetID());
-        this->DrawNote(dc, element, layer, staff, measure);
-        dc->EndGraphic(element, this);
-    }
-    else if (dynamic_cast<Rest *>(element)) {
-        dc->StartGraphic(element, "", element->GetID());
-        this->DrawRest(dc, element, layer, staff, measure);
-        dc->EndGraphic(element, this);
+    switch (element->GetClassId()) {
```
lpugin, line 888: Same here, only changing the `dynamic_cast` to `vrv_cast` should be enough.

lpugin (commented):
I don't find d4e8e2633a0a813654d51217af028a5faf21a74d impact is significant enough. That can be dropped.

Also, could you re-evaluate 140ab7b3e5cef103b654c4db65aa3811ccc2d63f with only the changes from `dynamic_cast` to `vrv_cast`? I find the other changes make the code quite more convoluted and more difficult to understand, so it would be good to see the impact separately.

src/adjustxposfunctor.cpp
sonovice: This was only a micro-optimization for a hot path. Is({ ... }) builds a temporary vector for the two-class check, while the direct Is(DOTS) || Is(FLAG) form avoids that small allocation/construction cost. There is no intended behavior change. I can revert this one (and the others) if you don't like that.

src/adjustxposfunctor.cpp
sonovice: Same reason as above.

include/vrv/customtuning.h
sonovice: Good point. I changed this to keep the lazy allocation but use a raw `Tunings::Tuning *` with explicit destructor/copy/move handling.

sonovice:
I tested a simplified variant from `85e08de24` with only the known-safe drawing RTTI removals and without `d4e8e263` or `140ab7b3`.

Results on the 22-file PR corpus, 3 runs each:

- Parent through `85e08de24`: 12.64 s average
- Simplified `vrv_cast`-only variant: 12.28 s average
- Full PR before cleanup commits: 11.68 s average

My read is that the `vrv_cast` cleanup is worth keeping on its own, but the full drawing/traversal changes still have measurable additional impact. If the readability concern is the deciding factor, I can split/drop the convoluted traversal parts and keep only the simpler RTTI cleanup.

src/view_control.cpp
sonovice: I re-ran a reduced variant to separate the RTTI removal from the larger dispatch changes. The `vrv_cast`-only drawing variant improved the 22-file corpus from 12.64 s to 12.28 s on average, so you are right that RTTI removal accounts for a real part of the gain.

The full PR was still faster in the same run, averaging 11.68 s. So the helper/switch/direct-dispatch changes are not purely noise, but the extra gain is smaller than the readability cost may justify. If you prefer, I can keep only the `vrv_cast`/known-class checks and drop the helper/switch parts.

src/view_control.cpp
sonovice: Agreed. That path does not need RTTI when the class branch has already established the concrete type. I tested a variant that keeps the old `if/else` dispatch structure and only changes the known-safe RTTI casts to `vrv_cast` or direct interface returns. It rendered the same 412 SVGs with 0 normalized mismatches and improved the corpus from 12.64 s to 12.28 s on average.

src/view_element.cpp
sonovice: Agreed that the RTTI check is the important part here. One caveat: in Release builds `vrv_cast` is `static_cast`, so it cannot directly replace `dynamic_cast` as the branch predicate. In the simplified test variant I used `element->Is(CHORD)`, `element->Is(NOTE)`, and `element->Is(REST)` for the predicates and kept the old dispatch shape. That rendered identically.

ahankinson:
It would be good to have a statement about where and how AI was used in this PR.

src/view_element.cpp
lpugin: Yes, `dynamic_cast` cannot be replaced blindly with `vrv_cast`. The later is always (and must only be) used following a `classId` check. That's been all done with this kind of optimisation in mind.

Note that for debugging purposes, you can change the definition of `vrv_cast` from `static_cast` to `dynamic_cast`.

infojunkie:
I want to note that the existing test suite does not exercise the `CustomTuning` functionality. So this PR would merge untested code unless someone manually tests it.

I am planning to add custom tuning tests, but they would need to be added to https://github.com/rism-digital/verovio.org and there's no existing template to follow, since we're not comparing visual output in this case. So I don't expect this to be a trivial undertaking.

src/view_control.cpp
lpugin: I came up with an optimisation possibility for `BoundingBox::Is` now `IsAnyOf` [here](https://github.com/rism-digital/verovio/pull/4352)

lpugin:
@sonovice let me know if you need any clarifications on the change made in-between.

sonovice:
I'm sorry for not responding. On family vacation right now and back in the last week of June.

lpugin:
No problem. Enjoy your holiday!

lpugin:
@sonovice since there are some conflicts, and some improvements have been made, I am closing the PR. Please do open a new one if needed. Thanks for looking into this.

PR #4349, opened 2026-06-05
Fixed: Python build process. 

### LINKED ISSUE #4026: Calling setup.py directly is no longer recommended
See:
 - https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
 - https://packaging.python.org/en/latest/guides/modernize-setup-py-project/
 - https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/

But also:
 - https://packaging.python.org/en/latest/discussions/setup-py-deprecated/ (answer: not completely)

TL;DR, calling setup.py directly (`python setup.py ...` )is no longer recommended, although there isn't a real clear path forward. The hang-up seems to be (unsurprisingly) the building of the C++ parts. But there have been movements in the latest versions of Python to remove or replace various bits of the "traditional" build system. 

Verovio only seems to call setup.py directly for building the source distributions. However, it also recommends calling setup.py directly in the book: https://book.verovio.org/installing-or-building-from-sources/python.html#building-the-toolkit-without-cmake

This might be a good opportunity to move to the new `pyproject.toml` standard, https://packaging.python.org/en/latest/guides/writing-pyproject-toml/. 

ChatGPT is suggesting we look at [scikit-build-core])https://github.com/scikit-build/scikit-build-core) as a build backend, which seems to be a way of using CMake in building the Python extensions. This might be good, since it would mean that there would be a single build system, instead of two different ones for Python.

So the path forward seems to be:

- [ ] Move what we can to pyproject.toml
- [ ] (Interim?) continue to use the existing setup.py as the build backend in the pyproject file?
- [ ] Integrate scikit-build-core
- [ ] Change the GH actions to use the new setup?

This seems like a significant amount of work, and since everything is working now it's not urgent. But it's probably an opportunity to update to pyproject.toml and unify the build system on CMake.

ahankinson:
This PR replaces the current Python build setup with a newer one that no longer requires `setup.py`. 

This new process requires setting up the GitHub project as a Trusted Publisher for PyPI, since it no longer requires repo secrets. This simplifies the publishing process considerably, but until it is set up, the build will fail. 

`uv` is now recommended for creating and managing isolated build environments; it can be installed with `brew install uv` on Mac.

Most of the changes were done using AI (codex gpt 5.5) but audited by me and hand-corrected where necessary. I also manually verified the build process locally on my Mac (macOS 26 Tahoe). 

Here is the full description of the new build process, and more details on how this PR will change the Python builds.  This documentation can be added to the Verovio Book once the PR is accepted.

This PR fixes #4026 

---

The Python package is built from the repository root with `pyproject.toml`.
The legacy `setup.py` build path has been ported to `pyproject.toml` and removed.

## Local wheel build

Install the Python version you want to build with. For example, to use
Python 3.12:

```sh
uv python install 3.12
```

For another supported Python version, replace `3.12` with the version to test,
for example `3.10`, `3.11`, `3.13`, or `3.14`.

Create a virtual environment from that Python version:

```sh
uv venv --python 3.12 .venv
```

Install the development build tools into the virtual environment:

```sh
uv pip install --python .venv/bin/python ".[dev]"
```

Activate the virtual environment:

```sh
source .venv/bin/activate
```

The `dev` extra installs local build tools on all supported Python versions.
`cibuildwheel` is only installed when the development environment uses
Python 3.11 or newer, because current cibuildwheel releases no longer run on
Python 3.10. This does not prevent building CPython 3.10 wheels; cibuildwheel
can still build `cp310` wheels when cibuildwheel itself runs on Python 3.11 or
newer.

Build a wheel from the repository root:

```sh
python -m build --wheel
```

Build a source distribution:

```sh
python -m build --sdist
```

Validate built distributions:

```sh
twine check dist/*
```

The build uses an isolated PEP 517 environment. Build dependencies such as
`scikit-build-core`, `swig`, and `ninja` are installed into that isolated build
environment as needed.

## How the package is built

`pyproject.toml` uses `scikit-build-core` as the build backend. It configures
the existing CMake build with `BUILD_AS_PYTHON=ON`, and CMake builds the
`verovio._verovio` extension module from the C++ sources and the SWIG interface.

SWIG-generated files are build outputs. In particular, `verovio.py` and the
SWIG wrapper source are generated into the CMake build/package output and should
not be maintained as tracked Python source.

The wheel installs:

- the compiled `_verovio` extension module
- the generated `verovio.py` SWIG proxy
- `__init__.py`
- `py.typed` and `verovio.pyi`
- the Verovio resource data under `verovio/data`

The `data/` directory is included by the CMake install rules, not by
`pyproject.toml` package discovery. The Python build sets
`wheel.packages = []`, so `scikit-build-core` packages the files installed by
CMake. For the Python package, CMake installs matching resource files from the
repository `data/` directory into `verovio/data` in the wheel. The package
initializer then sets the default toolkit resource path to that installed
directory with `importlib.resources`.

## Versioning

The Python package version is dynamic. `tools/python_metadata.py` reads the
base version from `include/vrv/vrvdef.h`:

- `VERSION_MAJOR`
- `VERSION_MINOR`
- `VERSION_REVISION`
- `VERSION_DEV`

When `VERSION_DEV` is true, the package version receives a PEP 440 development
suffix, for example `6.3.0.dev33`.

CI may set `VEROVIO_BUILD_VERSION` to force a specific build version. This is
used for TestPyPI builds so repeated development uploads get unique versions.

The Python package version is separate from the runtime toolkit version returned
by `verovio.toolkit().getVersion()`. The runtime version is compiled from the
C++ version defines and may include git commit information from
`include/vrv/git_commit.h`.

## GitHub Actions

The Python wheel workflow builds:

- one source distribution
- Linux wheels with `cibuildwheel`
- macOS wheels with `cibuildwheel`
- Windows wheels with `cibuildwheel`

The matrix runs platform builds in parallel. The workflow validates all
distributions with `twine check` before publishing.

Publishing uses PyPI Trusted Publishing:

- pushes to `develop-pypi` publish to TestPyPI
- release tags matching `v*` or `version-*` publish to PyPI

PyPI and TestPyPI must be configured with trusted publishers for this repository
and the corresponding GitHub Actions environments.

## Source builds from sdist

Building a wheel from an sdist should work without a git checkout. In that case
the Python package version comes from `PKG-INFO`, and CMake still compiles the
extension module from the included C++ sources.

If the source tree has no `.git` metadata, the runtime toolkit version may omit
the git commit suffix. This affects provenance detail only; it should not affect
whether the wheel builds or imports.


ahankinson:
Working on the test suite build now.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Programming Language :: C++",
+    "Operating System :: OS Independent",
+]
+
+[project.optional-dependencies]
+dev = [
+    "build>=1.2",
+    "cibuildwheel>=3.4; python_version >= '3.11'",
```
rettinghaus, line 21: If you switch for the build system to scikit, why declaring these development dependencies?

ahankinson: They’re optional to help the CI build. You can ignore them if you want

PR #4365, opened 2026-06-16
MIDI diffs

infojunkie:
This PR adds MIDI diffing as part of the `test-suite-diff.py` script.

The only added dependency is `py_midicsv` which is used to produce a text CSV representation of the MIDI files, only if `filecmp` reports binary differences between the raw MIDI files. The core Python module `difflib` generates the familiar diff format on the generated CSVs. 

The diff output is written to a different log file called `log_midi.md`. The GitHub workflow in `test_build.yml` now checks for the presence of either `log.md` and `log_midi.md` directly in the "Show the list of changes detected" step.

You can see an example of the output at:
- https://github.com/rism-digital/verovio/actions/runs/27781688489 for the warning if MIDI files are different
- https://github.com/rism-digital/verovio/actions/runs/27781688489/job/82207822005 for the detailed output of `log_midi.md` (which I have now cleaned up to remove the extra blank line between each diff line).

### DIFF
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -43,7 +43,7 @@ def start_webpage():
 
 def end_webpage(html, body, htmlOutFile):
     js = """
-    $(document).ready(function(){ 
+    $(document).ready(function(){
         $(".nav-tabs a").click(function(e){
             e.preventDefault();
             $(this).tab('show');
@@ -130,6 +130,8 @@ if __name__ == "__main__":
             pngFile2 = os.path.join(path_in2, item1, name + '.png')
             jsonFile1 = os.path.join(path_in1, item1, name + '.json')
             jsonFile2 = os.path.join(path_in2, item1, name + '.json')
+            midiFile1 = os.path.join(path_in1, item1, name + '.mid')
+            midiFile2 = os.path.join(path_in2, item1, name + '.mid')
             pngFileOut = os.path.join(path_out, item1, name + '.png')
             pngFile1Out = os.path.join(path_out, item1, name + '.after.png')
             pngFile2Out = os.path.join(path_out, item1, name + '.before.png')
@@ -139,6 +141,11 @@ if __name__ == "__main__":
             timeMap2 = json.load(open(jsonFile2, 'r'))
             if jsondiff(timeMap1, timeMap2) != 1:
                 print(f'::warning title=changes in time map detected::{name} produced a different time map')
+
+            midi1 = open(midiFile1, 'rb').read()
+            midi2 = open(midiFile2, 'rb').read()
+            if (midi1 != midi2):
+                print(f'::warning title=changes in MIDI detected::{name} produced a different MIDI file')
 
             diffValue = pngdiff(pngFile1, pngFile2, delete_diff_file=True)
             if (diffValue > (args.threshold / 100.0)):
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -151,7 +151,7 @@ if __name__ == "__main__":
                 csv1 = pm.midi_to_csv(midiFile1)
                 csv2 = pm.midi_to_csv(midiFile2)
                 diff = difflib.unified_diff(csv1, csv2, fromfile=midiFile1, tofile=midiFile2, n=0)
-                log_midi.append("****** {}:\n{}".format(item1, '\n'.join(list(diff))))
+                log_midi.append("****** {}:\n{}".format(item1, ''.join(list(diff))))
 
             diffValue = pngdiff(pngFile1, pngFile2, delete_diff_file=True)
             if (diffValue > (args.threshold / 100.0)):
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -151,7 +151,8 @@ if __name__ == "__main__":
                 csv1 = pm.midi_to_csv(midiFile1)
                 csv2 = pm.midi_to_csv(midiFile2)
                 diff = difflib.unified_diff(csv1, csv2, fromfile=midiFile1, tofile=midiFile2, n=0)
-                log_midi.append("****** {}:\n{}".format(item1, ''.join(list(diff))))
+                diff_str = ''.join(diff)
+                log_midi.append(f'****** {item1}:\n{diff_str}')
 
             diffValue = pngdiff(pngFile1, pngFile2, delete_diff_file=True)
             if (diffValue > (args.threshold / 100.0)):
--- a/doc/test-suite-diff.py
+++ b/doc/test-suite-diff.py
@@ -69,7 +69,7 @@ def end_webpage(html, body, htmlOutFile):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description='Examine diffrences in renderings of MEI scores')
+        description='Examine differences in renderings of MEI scores')
     parser.add_argument('input_dir1')
     parser.add_argument('input_dir2')
     parser.add_argument('output_dir')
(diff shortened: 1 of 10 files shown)


infojunkie:
@lpugin sorry to bother you, but can you tell what I'm missing for the Python bindings:

```
  File "/home/runner/work/verovio/verovio/dev-dir/bindings/python/../../doc/test-suite.py", line 42, in <module>
    tk = verovio.toolkit(False)
         ^^^^^^^^^^^^^^^
AttributeError: module 'verovio' has no attribute 'toolkit'
```

https://github.com/rism-digital/verovio/actions/runs/27594856261/job/81582961646?pr=4365#step:10:384

Thanks!

ahankinson:
I think the test runner needs to be updated to the new python build system. I will look into that today

ahankinson:
@infojunkie you should be able to bring your branch up to date with develop and it should fix the building issue.

infojunkie:
Thanks @ahankinson  works great!

@lpugin Looks like there are no issues rendering to MIDI and diffing as part of the test suite. If you're ok with that, I will continue working on this PR by doing the following:

- Providing a visual MIDI diff by converting the MIDIs to a JSON structure and diffing the JSON in case of discrepancy
- Triggering an error instead of warning in case of MIDI diff

lpugin:
I don't think diffing the JSON will bring much since that will be hard to understand. The test suite is primarily to raise changes, not so much for debugging them.

We don't raise errors when there is a difference because that is then blocking intended modifications to be merged. What you can add is a step that shows the list of modified MIDI, and that step will be skipped where there are none. That's makes it easy to check. We have the same for visual changes:

image

ahankinson:
Perhaps the [filecmp](https://docs.python.org/3/library/filecmp.html) module is better suited for this task?

infojunkie:
Thanks both, I'll concoct something and push it here.

craigsapp:
I would convert to binasc (See Midifile librray).  Binasc converts the MIDI file into human readable text (and exactly equivalent in meaning) which can then diffed (one MIDI Message per line). I see not point in diffing hex codes...

Example slide:  The blue box is the MIDI output from Binas (probably binasc -m option in the binasc example program).


Screenshot 2026-06-17 at 1 33 27 PM

infojunkie:
Thanks @craigsapp - agreed, binary diff is useless. I am currently testing [midiff](https://github.com/Teskann/midiff) to see if it can yield insights into unexpected changes. I will check your recommendation too.

infojunkie:
@lpugin @ahankinson I've noticed that the GitHub workflow runs `test-suite.py` once from the `develop` branch, which currently does not generate the MIDI files. While this makes total sense in order to generate the original output, it means that this PR cannot function correctly because there are no MIDI files generated on the `develop` side. 

I've created a [second, smaller PR](https://github.com/rism-digital/verovio/pull/4372) to only add MIDI generation to `test-suite.py` without any diffing code. Merging that second PR before this one would ensure that MIDI files are generated on both sides of the comparison.

If you have a different suggestion to get around this catch-22, happy to do it.

infojunkie:
The test build is now correctly failing until it finds MIDI files in the dev branch.

infojunkie:
@lpugin @ahankinson now the test diff works for MIDI and reports on the differences in the step "Show the list of changes detected":

https://github.com/rism-digital/verovio/actions/runs/27781688489/job/82207822005

(This example is a bit extreme because I rigged the code to add an extra SysEx to every exported MIDI file).

If this looks ok, I'll add some microtonal scores to the test suite and revert all the rigging.

infojunkie:
> Thanks @craigsapp - agreed, binary diff is useless. I am currently testing [midiff](https://github.com/Teskann/midiff) to see if it can yield insights into unexpected changes. I will check your recommendation too.

I ended up looking at the `midiff` code and only used the internal `py_midicsv` which along with the core `difflib` is completely enough to generate a good-looking diff (imo).

craigsapp:
If you implement mididiff (and maintain it), however you choose is fine with me 😜. I see a small problem: random bytes could be inserted. Also, binasc will have difficulty parsing an incorrect MIDI file. But if the Midifile library writes the file, it is difficult to get syntax errors.

infojunkie:
> If you implement mididiff (and maintain it), 

I chose NOT to use `midiff` because it's overkill. The actual code is 5 lines. The heavy lifting is done by `py_midicsv` and `difflib`:
```python
           if not filecmp.cmp(midiFile1, midiFile2, shallow=False):
                print(f'::warning title=changes in MIDI detected::{name} produced a different MIDI file')
                csv1 = pm.midi_to_csv(midiFile1)
                csv2 = pm.midi_to_csv(midiFile2)
                diff = difflib.unified_diff(csv1, csv2, fromfile=midiFile1, tofile=midiFile2, n=0)
                log_midi.append("****** {}:\n{}".format(item1, '\n'.join(list(diff))))
```

Here's how the diff looks like:
```shell
****** github-4340:
--- /home/kratib/src/music/verovio/test/github-4340-3.blessed.mid
+++ /home/kratib/src/music/verovio/test/github-4340-3.mid
@@ -10,4 +10,5 @@
-2, 0, Note_on_c, 0, 64, 90
-2, 119, Note_on_c, 0, 64, 0
-2, 120, Note_on_c, 0, 64, 90
-2, 239, Note_on_c, 0, 64, 0
+2, 0, System_exclusive, 20, 7F, 7F, 08, 09, 03, 7F, 7F, 40, 00, 39, 5F, 42, 40, 3C, 1F, 45, 00, 3E, 5F, 47, 40, 41, 20, 3A, 7F, 43, 60, 3D, 3F, 46, 20, F7
+2, 0, Note_on_c, 0, 65, 90
+2, 119, Note_on_c, 0, 65, 0
+2, 120, Note_on_c, 0, 65, 90
+2, 239, Note_on_c, 0, 65, 0
```

craigsapp:
OK mididiff, midif, or py_midicsv (they are all the same to me).  Another thought is that I convert pitch bend to  a float from -1.0 to +1.0 if the parser leaves it as an 14-bit integer, it will be harder to parse.

infojunkie:
> I convert pitch bend to a float from -1.0 to +1.0 if the parser leaves it as an 14-bit integer, it will be harder to parse.

Yes, most if not all MIDI messages need extra formatting to be fully legible. It's a matter of balancing many factors but I'm happy with this one for the time being.

On a tangential note, you may be interested in the MIDI structure explorer that I've developed for the excellent [Kaitai Struct IDE](https://ide.kaitai.io/) : https://github.com/infojunkie/musicxml-midi/blob/main/src/ksy/standard_midi_file.ksy . Kaitai Struct is a declarative language to parse binary file formats. I've extended their MIDI parser to support message formatters including CC, SysEx, GM, etc.

doc/test-suite-diff.py
```diff
@@ -140,6 +146,13 @@ def end_webpage(html, body, htmlOutFile):
             if jsondiff(timeMap1, timeMap2) != 1:
                 print(f'::warning title=changes in time map detected::{name} produced a different time map')
 
+            if not filecmp.cmp(midiFile1, midiFile2, shallow=False):
+                print(f'::warning title=changes in MIDI detected::{name} produced a different MIDI file')
+                csv1 = pm.midi_to_csv(midiFile1)
+                csv2 = pm.midi_to_csv(midiFile2)
+                diff = difflib.unified_diff(csv1, csv2, fromfile=midiFile1, tofile=midiFile2, n=0)
+                log_midi.append("****** {}:\n{}".format(item1, ''.join(list(diff))))
```
ahankinson: You might as well make this an f-string too.

lpugin:
One question. Why is the PR showing changes in the test suite?

image

infojunkie:
I had to change the workflow to detect 2 log files, not just one. Instead of duplicating the existing 2 steps that detected and displayed a single file (log.md), I collapsed the detection and display of the 2 log files in a single step.

lpugin:
Ok. When there was no change, the step used to be skipped, which was quite handy. This should remain the same.

infojunkie:
@lpugin I reinstated the workflow step to check for log files.

Screenshot from 2026-06-22 08-24-31

PR #4366, opened 2026-06-16
Fixed: CI tests building

ahankinson:
This updates the testing GitHub action to use the new Python build process for both the PR and the develop branch.

Additionally, it creates separate virtual environments for the develop and PR branches, rather than reinstalling the PR in the same venv as the develop build. This helps keep builds isolated.

### DIFF
--- a/.github/workflows/tests_build.yml
+++ b/.github/workflows/tests_build.yml
@@ -68,7 +68,12 @@ jobs:
           ls -alh ${{ env.OUTPUT_DIR }}
           sudo apt-get install wireguard
           sudo echo "${{ secrets.VPN_CONFIGURATION }}" > wg0.conf
-          python3 -m pip install build diffimg jsondiff lxml xmldiff cairosvg
+          python3 -m venv .venv-dev
+          python3 -m venv .venv-pr
+          .venv-dev/bin/python -m pip install --upgrade pip
+          .venv-pr/bin/python -m pip install --upgrade pip
+          .venv-dev/bin/python -m pip install build diffimg jsondiff lxml xmldiff cairosvg
+          .venv-pr/bin/python -m pip install build diffimg jsondiff lxml xmldiff cairosvg
 
       - name: Checkout the dev branch
         uses: actions/checkout@v6.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_build.yml
```diff
@@ -83,13 +88,12 @@ jobs:
...
       - name: Build Python toolkit and run the tests for the dev branch
-        working-directory: ${{ github.workspace }}/${{ env.DEV_DIR }}/bindings
+        working-directory: ${{ github.workspace }}/${{ env.DEV_DIR }}
         run: |
-          cmake ../cmake -DNO_HUMDRUM_SUPPORT=ON -DNO_ABC_SUPPORT=ON -DNO_PAE_SUPPORT=ON -DBUILD_AS_PYTHON=ON -DVRV_DYNAMIC_CAST=ON -B python
-          cd python
-          make -j8
-          python3 ../../doc/test-suite.py ${{ github.workspace }}/${{env.GH_PAGES_DIR}}/_tests ${{ github.workspace }}/${{ env.TEMP_DIR }}/${{ env.DEV_DIR }}/
-          python3 ../../doc/test-suite.py ${{ github.workspace }}/${{env.GH_PAGES_DIR}}/musicxmlTestSuite ${{ github.workspace }}/${{ env.TEMP_DIR }}/${{ env.DEV_DIR }}/
+          ${{ github.workspace }}/.venv-dev/bin/python -m build --wheel
```
rettinghaus, line 93: Instead of checking out the whole repos and building the wheel locally we could simplify everything by installing directly from the branches with

`pip install git+https://github.com/rism-digital/verovio@${{github.base_ref}}`

and 

`pip install git+https://github.com/rism-digital/verovio@${{github.head_ref}}`

ahankinson: My understanding is that pip would anyway check out the ref and then build it? So that would be adding a layer of indirection?

rettinghaus: It should create a shallow copy with `depth=1`. 

It would be even faster with curl

```shell
curl -L https://github.com{{ github.head_ref }}.zip -o verovio.zip
pip install verovio.zip --no-cache-dir
```

ahankinson: Feel free to open a PR with your suggestions.

PR #4370, opened 2026-06-16
Add permissions to workflow

rettinghaus:
CodeQL warned me that 'Workflow does not contain permissions'.


ahankinson:
Why does it need permissions?

rettinghaus:
> If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write. Often these permissions do not adhere to the principle of least privilege and can be reduced to read-only, leaving the write permission only to a specific types as issues: write or pull-requests: write.

PR #4375, opened 2026-06-22
Add pitch position query API for rendered layouts

### LINKED ISSUE #4189: Feature Request: Query a pitch graphical position of a layout given score position and staff
**Is your feature request related to a problem? Please describe.**

The ability to query the graphical position of a pitch on a given layout can enhance interactivity with a rendered score (think feedback, annotations etc). 

**Describe the solution you'd like**

For a given layout (meaning SVGs are already rendered):

_Input Spec_
- A MIDI Pitch (Int)
- A Score Position (Double) in verovio score-time format from the beginning of the score
- A staff number (if appropriate)

_Output Spec_
- a graphical coordinate pair `(x, y)` relative to the SVG Page

_Considerations_
- The "score position" can lie between two existing rendered notes (and not necessarily on a note itself). So the `x` position should be interpolated between the two surrounding pitch events.
- The "MIDI Pitch" parameter might not correpond to an existing pitch on that position. The method should thus calculate the `y` position based on score context on that position: _key, clef, accidental context_. 

**Additional context**

A similar method exists in the [GuidoLib](https://github.com/grame-cncm/guidolib) library using Visitor pattern (mostly to fetch context and calculate `y`). 
We would be happy to contribute on the development and provide tests.

arshiacont:
## Summary

Closes #4189.

This PR adds a toolkit API for querying the graphical SVG page position of a pitch at a given score-time position and staff.

The new API is exposed as:

- C++: `Toolkit::GetPitchPosition(double scoreTime, double midiPitch, int staffN)`
- C wrapper / WASM export: `vrvToolkit_getPitchPosition`
- JavaScript: `VerovioToolkit.getPitchPosition(scoreTime, midiPitch, staff = 1)`

The method returns a JSON object containing the computed SVG page coordinates and context information, including:

```json
{
  "x": 0,
  "y": 0,
  "loc": 0,
  "midi": 60,
  "staff": 1,
  "measureId": "...",
  "page": 1,
  "system": 1,
  "scoreTime": 0,
  "interpolated": true,
  "pname": 1,
  "accid": null,
  "accidSemitone": 0,
  "oct": 4
}
```

## Motivation

Issue #4189 requests a way to query the graphical position of an arbitrary pitch in an already rendered layout. This is useful for interactive applications that need to place feedback, annotations, overlays, cursors, or other UI elements on top of rendered Verovio SVG output.

The requested inputs are:

- MIDI pitch
- score-time position from the beginning of the score
- staff number

The requested output is an `(x, y)` coordinate relative to the SVG page.

## Implementation Details

This PR adds score-time-based measure lookup and layout interpolation support:

- Adds `MeasureScoreTimeComparison` to find the measure enclosing a requested score-time value.
- Adds `Measure::EnclosesScoreTime(...)` using the measure timemap onset/offset data.
- Adds `Measure::GetXAtScoreTime(...)`, which interpolates the x position between surrounding measure alignments.
- Ensures the timemap is calculated before resolving measure score-time boundaries.
- Handles exact measure-boundary queries by advancing to the next measure where appropriate, so onset queries resolve to the expected bar.
- Maps logical measures to drawing-page layout measures, including cases where rendered layout measures use `@corresp`.

Pitch-to-y placement is context-aware:

- Uses the current clef context via the active layer.
- Uses key signature accidentals as the initial accidental context.
- Updates accidental context from preceding notes in the staff up to the queried score-time.
- Chooses a contextual pitch spelling for the requested MIDI pitch, preferring the active key signature direction where relevant.
- Supports fractional MIDI pitches by interpolating vertically toward the next diatonic step.
- Accounts for active octave transposition lines on the target staff.

The returned coordinates are converted to SVG page coordinates:

- The x coordinate includes page margin offset.
- The y coordinate is converted from Verovio's internal drawing coordinate system to SVG page coordinates.
- The response also includes page number, system number, measure id, staff, pitch spelling, accidental, octave, and whether x interpolation was used.

The WASM and npm bindings are updated by:

- Adding `_vrvToolkit_getPitchPosition` to `emscripten/exports.txt`.
- Adding the C wrapper declaration and implementation.
- Adding the Emscripten proxy binding.
- Adding `VerovioToolkit.getPitchPosition(...)` to the npm toolkit wrapper.

## Tests

Added npm-side harnesses for the new API:

- `emscripten/npm/test/get-pitch-position.mjs`
  - Builds a minimal MEI example.
  - Queries `getPitchPosition`.
  - Checks measure/page/system/staff values.
  - Verifies fractional MIDI pitch changes the y coordinate.

- `emscripten/npm/test/get-pitch-position-xml.mjs`
  - Exercises the API on a MusicXML fixture.
  - Compares returned positions against rendered SVG context.
  - Prints note/timemap/SVG diagnostics useful for regression debugging.

## Verification

Verified after rebasing on current `origin/develop`:

```bash
cd emscripten
./buildNpmPackage
```

```bash
node emscripten/npm/test/get-pitch-position.mjs
node emscripten/npm/test/get-pitch-position-xml.mjs
```

```bash
cd emscripten/npm
npm pack --dry-run
```

```bash
git diff --check origin/develop..HEAD
```

All commands completed successfully.

### DIFF
--- a/emscripten/exports.txt
+++ b/emscripten/exports.txt
@@ -23,6 +23,7 @@ _vrvToolkit_getPageCount
 _vrvToolkit_getPageWithElement
 _vrvToolkit_getTimeForElement
 _vrvToolkit_getTimesForElement
+_vrvToolkit_getPitchPosition
 _vrvToolkit_getVersion
 _vrvToolkit_loadData
 _vrvToolkit_loadZipDataBase64
--- a/emscripten/npm/src/emscripten-proxy.js
+++ b/emscripten/npm/src/emscripten-proxy.js
@@ -78,7 +78,7 @@ function getToolkitFunction(VerovioModule, method) {
     // char *getTimesForElement(Toolkit *ic, const char *xmlId)
     mapping.getTimesForElement = VerovioModule.cwrap("vrvToolkit_getTimesForElement", "string", ["number", "string"]);
 
-    // char *getPitchPosition(Toolkit *ic, double scoreTime, int midiPitch, int staff)
+    // char *getPitchPosition(Toolkit *ic, double scoreTime, double midiPitch, int staff)
     mapping.getPitchPosition = VerovioModule.cwrap("vrvToolkit_getPitchPosition", "string", ["number", "number", "number", "number"]);
 
     // char *getMIDIValuesForElement(Toolkit *ic, const char *xmlId)
--- a/emscripten/npm/src/emscripten-proxy.js
+++ b/emscripten/npm/src/emscripten-proxy.js
@@ -78,6 +78,9 @@ function getToolkitFunction(VerovioModule, method) {
     // char *getTimesForElement(Toolkit *ic, const char *xmlId)
     mapping.getTimesForElement = VerovioModule.cwrap("vrvToolkit_getTimesForElement", "string", ["number", "string"]);
 
+    // char *getPitchPosition(Toolkit *ic, double scoreTime, int midiPitch, int staff)
+    mapping.getPitchPosition = VerovioModule.cwrap("vrvToolkit_getPitchPosition", "string", ["number", "number", "number", "number"]);
+
     // char *getMIDIValuesForElement(Toolkit *ic, const char *xmlId)
     mapping.getMIDIValuesForElement = VerovioModule.cwrap("vrvToolkit_getMIDIValuesForElement", "string", ["number", "string"]);
 
--- a/emscripten/npm/src/verovio-toolkit.js
+++ b/emscripten/npm/src/verovio-toolkit.js
@@ -112,6 +112,10 @@ export class VerovioToolkit {
         return JSON.parse(this.proxy.getTimesForElement(this.ptr, xmlId));
     }
 
+    getPitchPosition(scoreTime, midiPitch, staff = 1) {
+        return JSON.parse(this.proxy.getPitchPosition(this.ptr, scoreTime, midiPitch, staff));
+    }
+
     getVersion() {
         return this.proxy.getVersion(this.ptr);
     }
--- a/emscripten/npm/test/get-pitch-position-xml.mjs
+++ b/emscripten/npm/test/get-pitch-position-xml.mjs
@@ -61,32 +61,6 @@ const findNoteheadTranslate = (chunk) => {
   };
 };
 
-const parseStaffLineBounds = (measureChunk) => {
-  if (!measureChunk) return null;
-  const matches = [
-    ...measureChunk.matchAll(/M([\d.]+)\s+([\d.]+)\s+L([\d.]+)\s+([\d.]+)/g),
-  ];
-  if (!matches.length) return null;
-  const points = matches.flatMap((match) => [
-    {
-      x: Number.parseFloat(match[1]),
-      y: Number.parseFloat(match[2]),
-    },
-    {
-      x: Number.parseFloat(match[3]),
-      y: Number.parseFloat(match[4]),
-    },
-  ]);
-  const xs = points.map((p) => p.x);
-  const ys = points.map((p) => p.y);
-  return {
-    minX: Math.min(...xs),
-    maxX: Math.max(...xs),
-    minY: Math.min(...ys),
-    maxY: Math.max(...ys),
-  };
-};
-
 const main = async () => {
   const xml = readFile(TEST_XML_URL);
   const module = await createVerovioModule();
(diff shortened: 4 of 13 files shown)


arshiacont:
@lpugin Your review is appreciated on this and don't hesitate to guide since this is my first contribution to the core.

We have been using this in production since a few weeks and various scores now. So no rush on our side.

lpugin:
Sorry for the delay. I'll look at it next week. In the meantime it would be very useful to see it being used in your system. Can you point to a place where we can see it? I am interested in particular to see how you deal with clef changes and with the interpolation for the space between a bar line and the first note in a measure.

arshiacont:
@lpugin you can test it LIVE on https://metronautapp.com (even in Preview -- but let me know if you need full access)
Or you can see sample runs here: https://platform.antescofo.com/performance-analyzer/  (where all note overlays are calling this method)

arshiacont:
@lpugin if you want to see the Interpolations in action: Declare yourself as Violin on Metronaut, then open any piece. Tap on "Tempo" button and choose "Feedback AI" module and then press PLAY.
In "monophonic" mode, we draw lines continuously on the score in real-time. So they constantly interpolate between events (and should come out "smooth").
You should be able to test that with no-account no-subscription but let me know if you need a Premium access. Just send an email.

lpugin:
Sorry, I cannot find it. The only thing I can see is following notes / rests

https://github.com/user-attachments/assets/44132d7f-114a-4304-b835-c6043a52f81c

lpugin:
Some initial feedback - sorry for the delay.

The main problem with the proposed approach is that it duplicates things that can already be done with other methods. So these should be used to build it up. This means that you would not get everything in one call, but that is to be expected. For instance, the `GetElementsAtTime` method should be used as a starting point to get the measure ID. That means that you should use milliseconds instead of score time. Once you have the measure ID, you can use the `GetPageWithElement` to get the page no. So there is no reason to have these returned in your method.

I am not sure why you return `pname`, `oct`, `accid` since what you are looking for is a position. Also, why returning `staff` and `midi` since these are the value you are passing as parameter. Could you explain?

As I commented in the initial issue, having a generic method to get x and y position. But I think the goal here is to have a cursor-like playback position, so just the x and y would not work, or at least not be enough.

I am currently working on the editor API, which uses the `Edit` method, mostly to modify the content. However, one idea is to use it to get additional information, such as the "context" for an element in the score. Calling `Toolkit::Edit({ "action": "context", "param": { "elementId": "n123" } })` followed by `Toolkit::EditResponse()` returns a JSON object with the context of the "n123" element (ancestors, attributes, siblings, children, referring elements, referenced elements). FYI, this is in the https://github.com/rism-digital/verovio/tree/develop-edit-context branch.

We can do something similar with here, but the answer should be reduced to the minimal information, and getting the measure ID and the page should be done in distinct calls. So it would be something like `Toolkit::Edit({ "action": "playbackPosition", "param": { "time": "1000" } })` followed by `Toolkit::EditResponse()`. I'll think about the pitch position once I have a clearer idea what you need it for.

arshiacont:
> Sorry, I cannot find it. The only thing I can see is following notes / rests
> Screen.Recording.2026-07-08.at.14.24.47.mov

@lpugin tap on the "tempo" button in the bottom toolbar and then choose "Feedback AI" (and enable Microphone). Then try more or less performing the solo line.

lpugin:
Sorry, but I cannot get it to work. I have the microphone enabled and "Feedback AI" checked, but I still see the same cursor. Could you post a video? That would be easier, I think.

arshiacont:
@lpugin better than a video: see sample results on this page (use the drop-down to see change score-performance): https://platform.antescofo.com/performance-analyzer/ The overlay is using this branch.

For the Metronaut Website: when you "play" in the "feedback AI" mode you actually need to play/sing something for the pitch detector to hear something and then draw. :) 

PS: I'll get back to your CodeReview soon. Currently on a one-week break away.

arshiacont:
@lpugin Thanks for the detailed review. I checked the actual consumer in Metronaut and agree that the original response contained unnecessary information. I pushed a change reducing it to only `x`, `y`, `page`, and `system`.

The use case is different from a playback cursor: Metronaut draws a continuous trace of a performed pitch, including fractional pitches and pitches that _do not_ correspond to any notated element. The query is called continuously, and sometimes twice per sample—once for the performed pitch and once for the expected pitch.

`page` is needed because the returned coordinates are relative to a rendered page and the consumer must select the corresponding SVG overlay. `system` is used to stop connecting trace points across a system break. The other values—`measureId`, `staff`, `midi`, `scoreTime`, `pname`, `oct`, `accid`, `loc`, and `interpolated`—were either already known by the caller or unused, so I removed them.

Regarding `GetElementsAtTime`: the consumer’s source of truth is score position in quarter notes (qstamp), produced by a score follower or similar. Converting that to Verovio’s nominal MIDI milliseconds would add another mapping and could conflate performance/audio timing with notated timing. Also, `GetElementsAtTime` can locate existing notation elements, but it cannot calculate the vertical position of an arbitrary or fractional MIDI pitch.
I therefore kept this as one query rather than requiring several JavaScript/WASM calls for every trace point. 
I agree that the internal measure/page lookup should ideally share existing implementation rather than duplicate traversal. I would be happy to extract a shared internal helper if that is the preferred direction, while keeping the score-position input and compact result.

PR #4382, opened 2026-06-29
[MusicXML] fix breath import

### LINKED ISSUE #4381: MusicXML breath-mark timestamp can extend measure duration and corrupt timemap
We found a MusicXML import regression where `<breath-mark>` can be converted to an out-of-measure MEI `@tstamp`, which then changes the generated timemap and shifts later measure starts.

This affects applications that rely on `renderToTimemap()`.

## Reproduction

I will attach the `score.xml` file that reproduces the issue.

Using the attached MusicXML, render the timemap with:

```js
tk.loadData(scoreXml);
const timemap = tk.renderToTimemap({
  includeMeasures: true,
  includeRests: true,
});
```

In the attached score, around measure 9, the MusicXML has a breath mark on a dotted half note in 4/4:

[score-verovio-breathmark-bug.xml](https://github.com/user-attachments/files/29475336/score-verovio-breathmark-bug.xml)

```xml
<note>
  <duration>18</duration>
  <type>half</type>
  <dot/>
  <notations>
    <articulations>
      <breath-mark default-y="21" relative-x="142"/>
    </articulations>
  </notations>
</note>
```

The imported MEI currently gets a breath timestamp past the end of the measure, for example:

```xml
<breath staff="1" tstamp="5.9"/>
```

In 4/4, that is outside the bar. As a result, the timemap shifts the next measure start:

```txt
measure 9:  qstamp = 32
measure 10: qstamp = 36.89990234375
measure 11: qstamp = 40.89990234375
```

Expected timemap:

```txt
measure 9:  qstamp = 32
measure 10: qstamp = 36
measure 11: qstamp = 40
```

## Likely cause

This appears related to commit `44c85a89878f6b1ca3536ae7142edb3ed7a215e6` (`improve breath position`).

Current code computes the breath timestamp with:

```cpp
breath->SetTstamp((double)(m_durTotal + duration) * (double)m_meterUnit / (double)(4 * m_ppq) - 0.1);
```

But this happens after:

```cpp
if (!nextIsChord) m_durTotal += duration;
```

So `duration` is effectively counted twice. This works accidentally for some short notes, but fails for longer notes such as a dotted half.

## Proposed fix

Capture the current element end before advancing `m_durTotal`, then use that for the breath timestamp:

```cpp
const int elementEndTime = m_durTotal + duration;

if (!nextIsChord) m_durTotal += duration;

// ...

breath->SetTstamp(
    (double)elementEndTime * (double)m_meterUnit / (double)(4 * m_ppq) + 1.0 - 0.1
);
```

With this change, the breath mark stays inside the measure and the timemap returns the correct measure starts.

rettinghaus:
Improved version of #4373 taking also caesurae into account.

closes #4381

### DIFF
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3617,9 +3617,6 @@ void MusicXmlInput::ReadMusicXmlNote(
         }
     }
 
-    // add duration to measure time
-    if (!nextIsChord) m_durTotal += duration;
-
     m_ID = "#" + element->GetID();
 
     // breath marks
@@ -3631,7 +3628,7 @@ void MusicXmlInput::ReadMusicXmlNote(
         breath->SetPlace(
             breath->AttPlacementRelStaff::StrToStaffrel(xmlBreath.node().attribute("placement").as_string()));
         breath->SetColor(xmlBreath.node().attribute("color").as_string());
-        breath->SetTstamp((double)(m_durTotal + duration) * (double)m_meterUnit / (double)(4 * m_ppq) - 0.1);
+        breath->SetTstamp((double)(m_durTotal + duration) * (double)m_meterUnit / (double)(4 * m_ppq) + 0.9);
     }
 
     // caesura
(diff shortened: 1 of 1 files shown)


src/iomusxml.cpp
```diff
@@ -3631,7 +3628,7 @@ void MusicXmlInput::ReadMusicXmlNote(
         breath->SetPlace(
             breath->AttPlacementRelStaff::StrToStaffrel(xmlBreath.node().attribute("placement").as_string()));
         breath->SetColor(xmlBreath.node().attribute("color").as_string());
-        breath->SetTstamp((double)(m_durTotal + duration) * (double)m_meterUnit / (double)(4 * m_ppq) - 0.1);
+        breath->SetTstamp((double)(m_durTotal + duration) * (double)m_meterUnit / (double)(4 * m_ppq) + 0.9);
```
craigsapp, line 3631: I don't understand why you are replacing - 0.1 to +0.9?

If I understand, you should calculate the timestamp. and then limit to a maximum of `@units` + 1.0 if it overflows?

Otherwise you are assuming that the note has  a duration of more than one `unit`.

rettinghaus: The 1.0 offset was missing, and 1.0-0.1 is 0.9.

Issue #4380, opened 2026-06-29
ties with accidental doesn't seem to be managed properly from musicXML => MEI


mparisot-antescofo:
**Describe the problem**
We edit our scores in musicXML, we had one with a tie between 2 notes with an accidental on the starting note but not the ending note.

In that case the tie is not displayed and after checking, the conversion to MEI doesn't seem to find the end of the tie.

What's strange is that in the original score if we add another note with an accidental in the measure where the tie ends, then it works. Even if the accidental is not on the ending note if fixes the issue for both notes...

**To Reproduce**
1. Load the data attached

**Expected behavior**

I would expect to have the tie ending on the proper note whether it has an accidental or not.

**Input data**

I'm joining several files:
[flat-tie-err.xml](https://github.com/user-attachments/files/29468457/flat-tie-err.xml): the music XML without the accidental on the ending note of the time
[flat-tie-err-mei.xml](https://github.com/user-attachments/files/29465817/flat-tie-err-mei.xml): the resulting MEI with a tie that has no end

[flat-tie-fixed.xml](https://github.com/user-attachments/files/29468537/flat-tie-fixed.xml): the same musicXML as above but with the accidental on the ending note
[flat-tie-fixed-mei.xml](https://github.com/user-attachments/files/29465818/flat-tie-fixed-mei.xml): the resulting MEI, not that when generating the SVG the accidental is repeated on the ending note which we don't want if we can avoid it

[flat-tie-fixed-2.xml](https://github.com/user-attachments/files/29465838/flat-tie-fixed-2.xml): the music XML with 2 ties, both start on a note with an accidental, but only the first tie ends on a note with an accidetal
[flat-tie-fixed-2-mei.xml](https://github.com/user-attachments/files/29465819/flat-tie-fixed-2-mei.xml): the resulting MEI that shows both ties and the one without the accidental on the ending note doesn't repeat the flat which is what we want while the one with the accidental on the ending note repeats the flat on the ending note.

**Verovio information**
 - 6.2.0
 - directly in the last version of the online Verovio Editor

rettinghaus:
The first file contains two accidentals and works fine.

mparisot-antescofo:
Sorry about that, looks like I mixed my files and should have tripple checked after sending them. all seems to be in order now

i'm not sure if you prefer files like that or directly copy paste of the content in code tags to make it easier for you

rettinghaus:
[...]

PR #4383, opened 2026-06-29
Use `pitch/alter` for gestural accidentals

infojunkie:
This PR reinstates the logic of using MusicXML `pitch/alter` elements to guess the `@accid.ges` value. This logic had been replaced with the approach of carrying over accidentals, but it turns out that it is still needed in cases where tied notes straddle measure boundaries. 

In a strict sense, this case could also be handled by expanding the rules of carry-over accidentals to handle ties. For the sake of reinstating a code block that was previously functioning correctly within the scope of conventional accidentals, and to provide a safety net for MusicXML documents that might be depending on this functionality, we are restoring this functionality, while making sure it integrates within the broader framework of microtonality. 

Ref. #4380 

This PR also avoids injecting redundant SMuFL glyphs in the case of conventional accidentals, as was flagged by @rettinghaus in #4340 . Specifically, it detects that the gestural accidental and the written accidental are strictly equal as a condition to skip injecting the SMuFL glyph.

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -615,6 +615,11 @@ private:
     static void MidiToPitch(int midi, std::string &step, int &alter, int &octave);
     ///@}
 
+    /*
+     * @name Helper to create a string from pitch/alter values.
+     */
+    static std::string PitchAlterToString(data_PITCHNAME pname, float alter);
+
 public:
     //
 private:
@@ -697,8 +702,10 @@ private:
     std::map<Measure *, int> m_measureCounts;
     /* measure rests */
     std::map<int, int> m_multiRests;
-    /* a map of current accidental for each pitch class */
+    /* a map of pitch classes to their current accidental(s) */
     std::map<data_PITCHNAME, std::vector<musicxml::Accidental>> m_currentAccids;
+    /* a map of pitch/alter values to their corresponding accidental(s) */
+    std::map<std::string, std::vector<musicxml::Accidental>> m_alterAccids;
     /* current key signature */
     KeySig *m_currentKeySig = NULL;
     /* A flag indicating we had a clef change */
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3140,6 +3140,7 @@ void MusicXmlInput::ReadMusicXmlNote(
         pugi::xml_node pitch = node.child("pitch");
         if (pitch && !isTablature) {
             const std::string stepStr = pitch.child("step").text().as_string();
+            const float alterVal = pitch.child("alter").text().as_float();
             const int octaveNum = pitch.child("octave").text().as_int();
             if (!stepStr.empty()) note->SetPname(ConvertStepToPitchName(stepStr));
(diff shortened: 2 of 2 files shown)


infojunkie:
@rettinghaus I'd appreciate your review here, especially for the cases that you reported earlier. I am reviewing the reported MIDI diffs.

infojunkie:
> I am reviewing the reported MIDI diffs.

I reviewed the MIDI diffs reported here. There is actually an improvement in MIDI rendering accuracy over the current version, because the MusicXML import now handles cases where `pitch/alter` is set but no accidental is notated.

Notably, the microtonal test `004-maqam-rast.musicxml` now renders the measures 44 onward correctly. In this section, the `pitch/alter` for E and B is set to -0.5 (as per the Rast scale) without explicit accidentals. Previously, this was ignored by the MusicXML import. Reinstating the original code verbatim would have led to a situation where the deduced accidental would not match the tuning definition. In this PR, the code remembers associations between `pitch/alter` and notated accidentals, and thus integrates more throroughly the logic of carried-over accidentals with `pitch/alter` deduction.

craigsapp:
> There is actually an improvement in MIDI rendering accuracy over the current version, because the MusicXML import now handles cases where pitch/alter is set but no accidental is notated.

Can you explain this further?  I don't see how the `note/accidental` is used to generate MIDI data: only the `note/pitch/alter` information is needed; the `accidental` is used only to visually display an accidental. One possibility is how MusicXML deals with microtonal pitch.

Also the basic process: how is MIDI rendering involved in MusicXML-to-MEI importer? It should be independent: first MusicXML-to-MEI conversion, and then MEI-to-MIDI convertion (so how does MusicXML-to-MIDI involved in that process)?

> 004-maqam-rast.musicxml

Link? (So I don't have to go find it).

infojunkie:
> Link? (So I don't have to go find it).

https://github.com/rism-digital/verovio.org/tree/gh-pages/_tests/midi

> how is MIDI rendering involved in MusicXML-to-MEI importer? 

MIDI export is improved because MusicXML import is improved, resulting in a more correct MEI representation of the MusicXML and therefore a more accurate MIDI export.

> Can you explain this further? [..] the accidental is used only to visually display an accidental

`pitch/alter` is useful as a signal for sharps and flats, but not for microtonal music. In order to write a MIDI exporter that supports microtonality, the code needs to integrate both signals. I presented the microtonal part at MEC 2026, so I won't repeat that long exposition. This present PR integrates the `pitch/alter` part.

rettinghaus:
Lookd better.  

Still, now it produces many schematron errors: 
>`The value of @accid.ges should not duplicate the value of @accid.`

Preferable is to add accid.ges only, when it differs from accid. If there is no visible accidental I'd also prefer to keep it as attribute to the note.

infojunkie:
Thanks @rettinghaus for the review. Can you please give me the steps to run Schematron and an example doc that generates the warning?

rettinghaus:
Your latest change looks fine. Let me do some more testing.

rettinghaus (approved):
Much better, thank you.

Issue #4388, opened 2026-07-06
Bug: selection renders the score-initial clef instead of the running clef at the selection start


kuanyui:
**Describe the problem**

A selection made with `Toolkit::Select` + `{"measureRange": ...}` (the documented way to render an excerpt: https://book.verovio.org/interactive-notation/content-selection.html) renders the wrong clef at the selection start whenever a clef change occurs anywhere BEFORE the selected range: the selection opens with the score-initial clef, while its content is laid out against the correct running clef - a clef change inside the selection even renders as a change relative to a clef the reader never sees. Found while excerpting Liszt's Mazeppa (m105 moves the upper staff to the F clef; a m106-108 excerpt opens in G).

The running state itself survives: `Doc::ReactivateSelection` copies the selection system's drawing `ScoreDef`, and the copied `StaffDef`s carry the correct `m_currentClef` (verified by instrumentation). The loss is in `ScoreDef::ResetFromDrawingValues` (src/scoredef.cpp):

```cpp
Clef *clef = vrv_cast<Clef *>(staffDef->FindDescendantByType(CLEF));
if (clef) clef->ReplaceWithCopyOf(staffDef->GetCurrentClef());
```

`Object::ReplaceWithCopyOf` assigns through `Object::operator=` (src/object.cpp), which copies the base members and CLONES of the children - but no derived attribute values. A `Clef`'s shape / line / octave displacement live in its att-class members (`AttClefShape`, `AttLineLoc`, `AttOctaveDisplacement`), so nothing is transferred and the clef child keeps the initial clef. The `KeySig` / `Mensur` / `MeterSig(Grp)` copies in the same function look equally affected (not verified here).

Present in `version-6.2.1` and character-identical on current `develop`. I could not find an option that affects this (checked the options reference and `options.h`) nor an existing report.

**To Reproduce**

`repro.cpp` (public API only) loads the attached 3-measure score - m1 treble, m2 changes to bass, m3 stays bass - renders the full score and a `{"measureRange": "3-3"}` selection, and prints the first clef glyph of each SVG:

[repro.cpp](https://github.com/user-attachments/files/29712450/repro.cpp)

[minimal.musicxml.xml](https://github.com/user-attachments/files/29712459/minimal.musicxml.xml)

```
$ ./repro /path/to/verovio/data minimal.musicxml
full render first clef:      #E050  (expected #E050 - the score starts in treble)
[...]

PR #4389, opened 2026-07-06
Fix: Selection renders the score-initial clef instead of the running one at the selection start.

kuanyui:
A selection whose measure range is preceded by a mid-piece clef change opened with the score-initial clef. `ScoreDef::ResetFromDrawingValues` wrote the running clef into the scoreDef's clef child via `Object::ReplaceWithCopyOf`, whose `Object::operator=` copies base members and cloned children but no derived attribute values - so the clef child kept its old shape and line. Copy the clef-identity attributes (shape, line, dis, dis.place) explicitly instead.

No regression on the official test suites: `_tests` (629 files) and `musicxmlTestSuite` (150 files) render byte-identical SVGs before and after (the only caller of `ResetFromDrawingValues` is the selection path).

Fixes issue  https://github.com/rism-digital/verovio/issues/4388 . (includes repro, minimal score and before/after screenshots)

### DIFF
--- a/src/scoredef.cpp
+++ b/src/scoredef.cpp
@@ -496,7 +496,14 @@ void ScoreDef::ResetFromDrawingValues()
             meterSigGrp->ReplaceWithCopyOf(staffDef->GetCurrentMeterSigGrp());
         }
         else if (meterSig) {
-            meterSig->ReplaceWithCopyOf(staffDef->GetCurrentMeterSig());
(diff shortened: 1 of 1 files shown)


lpugin:
I made a more generic fix here https://github.com/rism-digital/verovio/commit/4cd6d2447fc1908c543fadf54e04b923fc1fb3ad
Thanks for looking into it.

Issue #4364, opened 2026-06-12
`@con` on `<nc>` does not affect neume rendering (needed for Hufnagel encoding)


zih-syuan:
We're implementing Hufnagel neume support in [Neon](https://github.com/DDMAL/Neon) (DDMAL; see [DDMAL/Neon#1360](https://github.com/DDMAL/Neon/issues/1360)). Hufnagel Clivis A should be encoded using `@con="e"` on `<nc>`, but `@con` on `<nc>` does not currently appear to affect Verovio's neume rendering.

As a workaround [we're currently using `@ligated="true"`](https://github.com/DDMAL/Neon/issues/1366), which triggers the correct visual rendering but is semantically incorrect — *ligatura* is a square notation concept that doesn't apply to Hufnagel. 

`@con="e"` is an existing MEI attribute value, used for extended connections between neume components. It is documented for Hispanic notation, and we are applying the same connection semantics to encode Hufnagel connected strokes.

**Describe the solution**

`@con="e"` on a `<nc>` should trigger a connected rendering with the preceding `<nc>`, using glyphs in the U+E9B9–E9BC range — the same rendering currently produced by `@ligated`. This should work within compound neumes, not only standalone clivis. (Related: #3724 implemented basic neume connection rendering; this request extends that to cover `@con`.)

The Hufnagel font is still under development, but the issue here is independent of the final glyph design: `@ligated` currently triggers interval-specific connected rendering, while `@con="e"` does not.

**Additional context**

Minimal examples (complete MEI file attached):

**Clivis A:**
```xml
<neume>
  <nc oct="3" pname="a" tilt="s"/>
  <nc oct="3" pname="g" tilt="n" con="e"/>
</neume>
```
Expected: virga + UDV rendered as connected clivis shape (U+E9B9–E9BC)

**Torculus (punctum + clivis A):**
```xml
<neume>
  <nc oct="3" pname="a"/>
  <nc oct="3" pname="b" tilt="s"/>
  <nc oct="3" pname="g" tilt="n" con="e"/>
</neume>
```
Expected: first `<nc>` independent, last two rendered as connected clivis shape

[Hufnag.toy.example.zip](https://github.com/user-attachments/files/28884732/Hufnag.toy.example.zip) includes a minimal MEI example and a reference image showing the desired Hufnagel connected rendering.

zih-syuan:
I have a draft Verovio-side branch in the DDMAL fork:

[...]

PR #4391, opened 2026-07-08
feat(neume): support notation subtypes

zih-syuan:
This PR adds support for neume notation subtypes in `staffDef@notationtype`.

It recognizes:

- `neume.square`
- `neume.hufnagel`

alongside the existing generic `neume` value.

The existing `neume` behavior is preserved for backward compatibility. The new subtype values are treated as part of the same neume notation family, so they follow the existing neume rendering/editor paths.

## Motivation

This is preparatory work for #4364.

The issue concerns Hufnagel-specific `@con` rendering on `<nc>`. Before implementing that behavior, Verovio needs to distinguish generic neume notation from more specific neume notation types, especially `neume.square` and `neume.hufnagel`.

This PR does not implement `@con` rendering yet. It only adds notation type recognition and routes the new neume subtype values through the existing neume code paths.

## What changed

- Added internal notation type values for `neume.square` and `neume.hufnagel`.
- Added string conversion support for reading/writing these notation type values.
- Extended `IsNeumeType()` so `neume`, `neume.square`, and `neume.hufnagel` are treated as the neume notation family.
- Updated existing neume-specific checks to use `IsNeumeType()` where the current neume behavior should apply to all neume notation types.

## What this PR does not do

- Does not implement Hufnagel `@con` rendering.
- Does not change `@ligated` behavior.
- Does not change font selection.
- Does not bundle or require a Hufnagel font.
- Does not remove or change the existing generic `notationtype="neume"` behavior.

## Testing

Built locally with:

```bash
cmake -S cmake -B build-neume-notationtypes -DCMAKE_BUILD_TYPE=Debug
cmake --build build-neume-notationtypes -j 4
```

Also tested minimal MEI files with:

- `notationtype="neume"`
- `notationtype="neume.square"`
- `notationtype="neume.hufnagel"`

All three rendered successfully to SVG.

## Notes

CLA pending.

### DIFF
--- a/libmei/dist/attconverter.cpp
+++ b/libmei/dist/attconverter.cpp
@@ -2937,6 +2937,8 @@ std::string AttConverterBase::NotationtypeToStr(data_NOTATIONTYPE data) const
         case NOTATIONTYPE_mensural_black: value = "mensural.black"; break;
         case NOTATIONTYPE_mensural_white: value = "mensural.white"; break;
         case NOTATIONTYPE_neume: value = "neume"; break;
+        case NOTATIONTYPE_neume_square: value = "neume.square"; break;
+        case NOTATIONTYPE_neume_hufnagel: value = "neume.hufnagel"; break;
         case NOTATIONTYPE_tab: value = "tab"; break;
         case NOTATIONTYPE_tab_staff_like: value = "tab.staff-like"; break;
         case NOTATIONTYPE_tab_guitar: value = "tab.guitar"; break;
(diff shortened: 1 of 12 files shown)


lpugin:
This looks good. One change you need to do is to have the notation types changed in the `./libmei/mei/mei-verovio.xml`. 
See a previous example here https://github.com/rism-digital/verovio/blob/1a2585cb4125941dbc50f46546204d6711c2bf0d/libmei/mei/mei-verovio.xml#L562-L600

Then you need to re-generate the files following https://book.verovio.org/contributing/generate-code-with-libmei.html

Then the files in `libmei/dist` should be the same as what you have, but these should not be edited by hand.

Can you have a try? Please ask if you need help.

zih-syuan:
Thanks! One clarification: is `mei-verovio_compiled.odd` expected to be regenerated from `mei-verovio.xml`, or should it be edited directly as part of this change?

I see the LibMEI generation step uses `mei-verovio_compiled.odd` as input for regenerating `libmei/dist`, but I’m not sure what the preferred workflow is for the compiled ODD itself.

lpugin:
Only `mei-verovio.xml` should be edited. `mei-verovio_compiled.odd` will be generated by the `ant` script.

zih-syuan:
Thanks! I added the notation type customization to `libmei/mei/mei-verovio.xml` and verified that the compiled ODD contains both `neume.square` and `neume.hufnagel`. I also regenerated LibMEI from the ODD and confirmed that it produces no additional changes in `libmei/dist`.

lpugin:
Looks good. Let me know when this is ready to be reviewed / merged

zih-syuan:
It is ready for review now. Thank you!

PR #4396, opened 2026-07-12
Resolve measures by position instead of number in MusicXML import. 

### LINKED ISSUE #3902: Bug Report: Duplicate <measure number="1"> Causes Unexpected Behavior in Verovio
Hello,

While using Verovio to render music scores, I encountered an issue when the XML contains two <measure number="1"> elements. In such cases, all data within the second <measure> is inherited by the first <measure>. This behavior leads to some unexpected bugs in the rendered score.

Could you please look into this issue? Let me know if you need additional details or examples to reproduce the problem.

Thank you for your assistance!

kuanyui:
> [!WARNING]
> DISCLOSURE: This PR heavily relies on LLM agent.

This PR is mean to fix issue #3902, screenshots, examples of this PR are in https://github.com/rism-digital/verovio/issues/3902#issuecomment-4950156346

The MusicXML importer resolved "which measure does this element belong to" through the measure `number` attribute, first match from the start of the section. MusicXML does not require `number` to be unique within a part - the spec's `text` attribute documentation begins "If measure numbers are not unique within a part, this can cause problems for conversions between partwise and timewise formats" - and a file holding several movements or songs restarts it (MuseScore emits exactly that for a section break with "start with measure one"). When numbers repeat, elements of later movements land in movement 1 or disappear, in three places:

- the `m_controlElements` attach loop: every octave-shift, direction, dynamic, pedal, tie, fermata, tempo... of a later movement is attached to the FIRST measure carrying that number, so movement 1 collects foreign octave lines (with the `octave '...' is not encoded
  in the measure of its start` warning);
- `ProcessClefChangeQueue`: later movements' clef changes are inserted into movement 1's same-numbered measures, and when the mis-resolved measure does not contain the original staff object, `AddClefs` silently drops them;
- `OpenSlur`: a pending slur stop is matched to a slur start by (slur number, measure number), so a dangling stop from one movement captures the same-numbered start of a later movement and the slur is drawn backwards across the movement boundary.

The importer already tracks each created measure's document position in `m_measureCounts`, and 1f7a7d6c6 ("Handle non-unique measure numbers") already uses exactly that count in `AddMeasure` to disambiguate duplicate numbers when merging parts. This change completes that migration for the remaining number-string sites: the control-element list, the clef-change queue and the slur stacks are keyed by the count instead, resolved through a new `FindMeasureByCount`. Warnings that used to print the (ambiguous) number now print the 1-based document position, and the `measureNum` parameters this change orphaned are dropped (`ReadMusicXmlBackup`/`Forward` never used theirs before this change, so they are left untouched).

Fixes #3902 - verified on that thread's own attachment (three measures all numbered 1): its six text directions and dynamic all imported into the first measure (7/0/0), and now distribute to their own measures (2/3/2), matching the intended rendering shown there via the Humdrum path.

No regression on the official suites (SVG output compared with a fixed `--xml-id-seed`, importer warnings compared as well):

- `musicxmlTestSuite` (all 150 files): byte-identical except `12ad-Clefs-Extreme-Octave` - that file itself contains two measures numbered 5. Before, both head clef changes piled up at the end of measure 4 and the final measure lost its clef; now each is engraved as the customary cautionary clef at the end of its own preceding measure.
- `_tests` (623 MEI files): byte-identical (the change is confined to the MusicXML importer).

On the real-world file from the report - a three-movement piano concerto exported as one file, numbering restarting twice, 14 parts, 644 measures - 1014 control events were attached to a wrong measure before the change and 0 after, and the piano part's clef changes, previously all confined to movement 1 (movements 2-3 kept none), now reach every movement. Through the `measureRange` selection feature the orphaned octave elements used to render as a stack of dashed `(8)` lines towering over unrelated measures (screenshots in the issue thread); those selections are clean after the change.

Minimal examples are in the issue thread ( https://github.com/rism-digital/verovio/issues/3902#issuecomment-4950156346 ): `minimal.musicxml` (measures numbered 1,2,1,2 - an octave line and a clef change in the second "movement" teleport into the first) and `slur-minimal.musicxml` (a dangling slur stop in the first movement captures the second movement's slur). `control-unique-numbers.musicxml` is the identical music with unique numbers; it renders identically before and after the change.

---

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -67,27 +67,27 @@ class Trill;
 namespace musicxml {
 
     struct OpenSlur {
-        OpenSlur(const std::string &measureNum, short int number, curvature_CURVEDIR curvedir)
+        OpenSlur(int measureCount, short int number, curvature_CURVEDIR curvedir)
         {
-            m_measureNum = measureNum;
+            m_measureCount = measureCount;
             m_number = number;
             m_curvedir = curvedir;
         }
 
-        std::string m_measureNum;
+        int m_measureCount;
         short int m_number;
         curvature_CURVEDIR m_curvedir;
     };
 
     struct CloseSlur {
-        CloseSlur(const std::string &measureNum, short int number, curvature_CURVEDIR curvedir)
+        CloseSlur(int measureCount, short int number, curvature_CURVEDIR curvedir)
         {
-            m_measureNum = measureNum;
+            m_measureCount = measureCount;
             m_number = number;
             m_curvedir = curvedir;
         }
 
-        std::string m_measureNum;
+        int m_measureCount;
         short int m_number;
         curvature_CURVEDIR m_curvedir;
     };
@@ -218,10 +218,9 @@ namespace musicxml {
     };
 
     struct ClefChange {
-        ClefChange(const std::string &measureNum, Staff *staff, Layer *layer, Clef *clef, const int &scoreOnset,
-            bool afterBarline)
+        ClefChange(int measureCount, Staff *staff, Layer *layer, Clef *clef, const int &scoreOnset, bool afterBarline)
         {
-            m_measureNum = measureNum;
+            m_measureCount = measureCount;
             m_staff = staff;
             m_layer = layer;
             m_clef = clef;
@@ -229,7 +228,7 @@ namespace musicxml {
             m_afterBarline = afterBarline;
         }
 
-        std::string m_measureNum;
+        int m_measureCount;
         Staff *m_staff;
         Layer *m_layer;
         Clef *m_clef;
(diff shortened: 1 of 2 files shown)


lpugin:
Can you explain why it is necessary to change the type from `string` to `int`? It is a token in MusicXML, so converting to an `int` is problematic.

kuanyui:
You are right that `number` is a token and that parsing it as an integer would be wrong - the change does neither. The `int` is a different quantity, not a conversion: the measure's document position within the part, i.e. the index `ReadMusicXmlPart` already records for every measure it creates in `m_measureCounts` (`std::map<Measure *, int>`). The token is never parsed (there is no `stoi` / `as_int` on it anywhere in the diff) and keeps flowing verbatim, as a string, into `@n` through `measure->SetN(measureNum)` - that path is untouched. A non-numeric number like "X1" imports exactly as before.

What the change addresses is which quantity identifies a measure internally. The spec calls `number` "the attribute that identifies the measure" and scopes that identification to grouping across parts ("Going from partwise to timewise, measures are grouped via this attribute. In partwise files, it should be the same for measures in different parts that share the same left barline"); within a part it does not require uniqueness, and the `text` attribute documentation begins with exactly this failure mode: "If measure numbers are not unique within a part, this can cause problems for conversions between partwise and timewise formats." Assembling a score from partwise input is such a conversion, and multi-movement exports are the non-unique case (MuseScore restarts numbering at a section break with "start with measure one"): several measures carry `number="1"`, and the first-match `AttNNumberLikeComparison` lookup resolves all of them to the first one - that is #3902. "Which of the measures numbered 1" is information the string does not contain, so no treatment of the string can recover it; the document position is unambiguous, and the importer already owns it. (Resolving duplicates by encounter order does not work either: the flat control-element list restarts its measure sequence at every part boundary, which is indistinguishable from a duplicated number.)

The position key is also the importer's existing idiom for this job, so the change completes a migration rather than starting one:

- `OpenSpanner` (hairpins, brackets) and `OpenDashes` already store `int m_measureCount`, fed from `m_measureCounts.at(measure)`, and pair across measures by count difference;
- since 1f7a7d6c6 ("Handle non-unique measure numbers"), `AddMeasure` uses the same count to pick the right measure when merging parts whose numbers repeat.

The PR moves the remaining token-keyed lookups (the control-element attach loop, the clef-change queue, slur stop/start pairing) onto that same key - these are exactly the sites that misfire on files with repeated numbers. For files whose numbers are unique nothing changes: both official suites render byte-identical except `12ad-Clefs-Extreme-Octave`, which itself numbers two measures 5.

Incidentally, the string keyspace was already not purely the token: before this change, the multirest barline site passed a stringified internal index - `ReadMusicXmlBarLine(*it, measureIter->first, std::to_string(lastElementIter->first))` - so a fermata on a multirest's closing barline was keyed by a 0-based position that the later `@n` lookup could only match by accident. With the key typed `int`, a token/position mix-up no longer compiles.

The one user-visible change is the wording of the import warnings, which now print the 1-based document position instead of the token, because when this bug fires the token is exactly the ambiguous part (in the attachment on #3902, "measure 1" names three different measures). Happy to make them print both, e.g. `measure 315 (number "1")`, if you prefer.

> [!NOTE]
> DISCLOSURE: This reply heavily relies on the assistance of LLM agents.

rettinghaus:
I'm still not convinced. Multiple pieces shouldn't be cramped into one MusicXML file. They should be stored as separate files in the zip container together:

```xml
<container>
  <rootfiles>
    <rootfile full-path="movement1.musicxml" media-type="application/vnd.recordare.musicxml+xml"/>
    <rootfile full-path="movement2.musicxml" media-type="application/vnd.recordare.musicxml+xml"/>
  </rootfiles>
</container>
```

Until now, the golden rule has been not to attempt to correct faulty export data from other programs.

kuanyui:
I agree with the golden rule - and it is the reason the change is shaped the way it is: it does not correct anything. The association between an element and its measure is stated explicitly by the file's structure (every direction, attributes and slur stop sits inside its `<measure>` element); the importer reads it correctly and then loses it in its own deferred bookkeeping, by re-finding the parent through a key that cannot tell two same-numbered measures apart. The fix only preserves what the file already said, and no data is altered on the way: the duplicated numbers themselves still land in `@n` exactly as exported. The minimal from [the issue thread](https://github.com/rism-digital/verovio/issues/3902#issuecomment-4950156346) after the fix - the measure numbers still read 1, 2, 1, 2, untouched:

**Fixed**
import-fixed

On such files the current behavior is not a faithful pass-through of a questionable export; it introduces faults the exporter never wrote: directions and octave lines are attached to a different movement than the one they are encoded in, clef changes are silently dropped, and a slur ends on a note the file never connected it to. The same minimal today - the second movement's bass clef teleported to the end of document measure 1, a change nothing in the file asked for:

**Buggy**
import-buggy

Under a strict reading of the rule, that is the side that needs fixing.

On packaging, agreed that separate pieces are cleaner as separate files. But the container documentation assigns multiple rootfiles to alternate renditions of the same work - "Additional `<rootfile>` elements can describe alternate versions such as PDF and audio files" ([W3 MusixXML 4.0 - Compressed MXL Files](https://www.w3.org/2021/06/musicxml40/tutorial/compressed-mxl-files/)) - with the `<opus>` document as the multi-movement device; and in practice the importer receives one score-partwise document at a time regardless of packaging (as far as I can see, verovio reads the first rootfile of a container - [src/toolkit.cpp#L462-L467](https://github.com/rism-digital/verovio/blob/46a4df5256c8326d9d2d6783ff6e4048312dea34/src/toolkit.cpp#L462-L467) - and opus documents have no import path). Within one document, duplicate numbers are also not confined to multi-movement exports: the official musicxmlTestSuite's own `12ad-Clefs-Extreme-Octave` numbers two measures 5 - a single piece - and it is the one file in the suites whose rendering this change affects: its second clef change engraves correctly now.

The project has also already made this call once: 1f7a7d6c67d03126aa4300bdd75c29d2fa1e00f9 ("Handle non-unique measure numbers") makes `AddMeasure` disambiguate duplicate numbers by document position when merging parts. Today the importer is split - measure creation tolerates duplicate numbers, element attachment does not - and this change closes that gap, with no output change for any file whose numbers are unique. If it would help, I am happy to add a one-time import warning when duplicate measure numbers are encountered, so users of strict encodings still notice them.

> [!NOTE]
> DISCLOSURE: This reply heavily relies on the assistance of LLM agents.

lpugin:
I agree with @rettinghaus that having multiple movements in one single MusicXML file will be a problem. But this PR actually fixes the way measures are addressed in the importer, so that seems fine to me.

lpugin:
@kuanyui please sign and send the [CLA](https://rism.digital/resources/verovio-cla.pdf). Thanks

lpugin (approved):

kuanyui:
@lpugin Signed and sent. Thanks.

PR #4400, opened 2026-07-16
Use pitch/alter to derive accid.ges

### LINKED ISSUE #4397: MusicXML import: carried-over accidental leaks across staves/octaves, producing wrong @accid.ges and breaking tie matching
**Describe the problem**

During MusicXML import, the carried-over accidentals table (`m_currentAccids` in `iomusxml.cpp`) is keyed by pitch letter (`note->GetPname()`) only — octave, staff, and voice are not part of the key. An accidental written in one staff/octave therefore "carries over" onto any later note in the same measure that shares the letter, even on a different staff and in a different octave.

This has two user-visible consequences:

1. **Wrong sounding pitch.** A note with no `<accidental>` and no `<alter>` (i.e., a natural — MusicXML pitch is absolute) receives a phantom `@accid.ges` copied from an unrelated note. MIDI output plays the wrong pitch.
2. **Broken ties.** Tie matching compares sounding pitches (`Note::IsEnharmonicWith` → `GetMIDIPitch`), so the poisoned tie-stop note no longer matches its open tie start — neither in the fast path in `ReadMusicXmlTies` nor in `CloseTie`. The tie is reported as "left open", or worse: since `MatchTies` has no distance limit, the open tie can later be closed onto a same-letter note **many measures away**, drawing a giant arc across the system.

**To Reproduce**

Load the attached 2-measure file (`tie-accid-mre.musicxml`, also inlined below). It contains, on a piano grand staff:

- Measure 1, staff 2: a D3 whole note with `<tie type="start">`.
- Measure 2, staff 1: a D♯4 quarter (explicit `<accidental>sharp</accidental>`, `<alter>1</alter>`) — a different staff and a different octave.
- Measure 2, staff 2: the D3 whole note with `<tie type="stop">` (no `<alter>`, no `<accidental>` — a plain D natural).

Observed (verovio npm 6.1.0 and 6.2.0, WASM toolkit):

```
[Warning] MusicXML import: There are 1 ties left open
```

and in the resulting MEI the tie has no `@endid`, while the measure-2 D3 has acquired a sharp it does not have in the source:

```xml
<tie xml:id="xx9tx60" startid="#vf9ex5e" />
<note xml:id="ot4xelk" dur.ppq="16" dur="1" oct="3" pname="d">
  <accid xml:id="p58d57m" accid.ges="s" glyph.auth="smufl" glyph.name="accidentalSharp" />
</note>
```

Removing the (musically unrelated) D♯4 from staff 1 makes the tie import correctly.

<details>
<summary><code>tie-accid-mre.musicxml</code></summary>

```xml
<score-partwise version="3.1">
  <part-list>
    <score-part id="P1"><part-name>Piano</part-name></score-part>
  </part-list>
  <part id="P1">
    <measure number="1">
      <attributes>
        <divisions>4</divisions>
        <time><beats>4</beats><beat-type>4</beat-type></time>
        <staves>2</staves>
        <clef number="1"><sign>G</sign><line>2</line></clef>
        <clef number="2"><sign>F</sign><line>4</line></clef>
      </attributes>
      <note>
        <rest/>
        <duration>16</duration>
        <voice>1</voice>
        <staff>1</staff>
      </note>
      <backup><duration>16</duration></backup>
      <note>
        <pitch><step>D</step><octave>3</octave></pitch>
        <duration>16</duration>
        <tie type="start"/>
        <voice>2</voice>
        <type>whole</type>
        <staff>2</staff>
        <notations><tied type="start"/></notations>
      </note>
    </measure>
    <measure number="2">
      <note>
        <pitch><step>D</step><alter>1</alter><octave>4</octave></pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <accidental>sharp</accidental>
        <staff>1</staff>
      </note>
      <note>
        <rest/>
        <duration>12</duration>
        <voice>1</voice>
        <staff>1</staff>
      </note>
      <backup><duration>16</duration></backup>
      <note>
        <pitch><step>D</step><octave>3</octave></pitch>
        <duration>16</duration>
        <tie type="stop"/>
        <voice>2</voice>
        <type>whole</type>
        <staff>2</staff>
        <notations><tied type="stop"/></notations>
      </note>
    </measure>
  </part>
</score-partwise>
```

</details>

**Expected behavior**

The measure-2 D3 keeps its natural sounding pitch (`<alter>` absent ⇒ no chromatic alteration; MusicXML pitch is not contextual), and the tie D3→D3 is closed. A written accidental should carry over only to notes on the same staff position (same letter **and** octave, same staff), per standard notation rules.

**Additional context**

- This looks closely related to #4380 / PR #4383, but is not covered by that fix: the reinstated `pitch/alter` override only runs `if (alterVal != 0.0)`, so a note with **no** `<alter>` (our case) still falls through to `m_currentAccids.at(note->GetPname())` and inherits the leaked accidental. Reading the current `develop` source, the repro above should still fail there.
- Real-world impact: in a Sibelius-exported piano piece, a D♯4 in the right hand poisons the left hand's D3 tie stop at the start of the measure; the tie opened in the previous measure then gets matched by `MatchTies` to the next clean D3 — **29 measures later** — rendering a tie arc spanning half the system. Happy to attach that score (or a truncated excerpt) as well.
- Possible directions: key the carry-over table by (staff, letter, octave) instead of letter alone; and/or treat "no `<alter>` and no `<accidental>`" as an authoritative natural that overrides (or clears) the carried-over entry, extending the approach of #4383 — whose description already anticipates "expanding the rules of carry-over accidentals".

Versions: verovio 6.1.0 and 6.2.0 (npm, WASM), macOS. Import via `loadData()` + `getMEI()`.

infojunkie:
To fix #4397 

This PR reverses the order of priority to rely primarily on `pitch/alter` to deduce the `@accid.ges`. In order to preserve microtonal accidentals, the alter-to-accidentals memoization that was added in the previous PR is kept. The mechanism of accidental carry-over is completely removed because it introduced too many regressions.

The existing tests seem to be passing, but I want to manually examine the MIDI differences and the MEI produced from the MusicXML. I am also thinking of adding unit tests to this repo that would specifically validate the MusicXML-to-MEI import.

### DIFF
--- a/include/vrv/iomusxml.h
+++ b/include/vrv/iomusxml.h
@@ -702,12 +702,8 @@ private:
     std::map<Measure *, int> m_measureCounts;
     /* measure rests */
     std::map<int, int> m_multiRests;
-    /* a map of pitch classes to their current accidental(s) */
-    std::map<data_PITCHNAME, std::vector<musicxml::Accidental>> m_currentAccids;
     /* a map of pitch/alter values to their corresponding accidental(s) */
     std::map<std::string, std::vector<musicxml::Accidental>> m_alterAccids;
-    /* current key signature */
-    KeySig *m_currentKeySig = NULL;
     /* A flag indicating we had a clef change */
     int m_clefChanged = 0;
 
--- a/src/iomusxml.cpp
+++ b/src/iomusxml.cpp
@@ -3174,33 +3174,33 @@ void MusicXmlInput::ReadMusicXmlNote(
                     try {
                         for (const auto &current : currentAccids) {
                             // Avoid adding empty accidentals
-                            if (current.m_accid == ACCIDENTAL_WRITTEN_NONE && current.m_glyphName.empty()) continue;
+                            if ((current.m_accid == ACCIDENTAL_WRITTEN_NONE || current.m_accid == ACCIDENTAL_WRITTEN_n) && current.m_glyphName.empty()) continue;
 
                             Accid *accid = new Accid();
                             note->AddChild(accid);
                             accid->SetAccidGes(Att::AccidentalWrittenToGestural(current.m_accid));
+                            accid->IsAttribute(true);
 
                             // Because gestural accidentals do not map 1:1 to written accidentals, we may be losing
                             // information if we rely only on gestural accidentals to look up tuning tones.
                             // Instead, we set the accidental's SMuFL glyph name to whatever was carried over. The
                             // custom tuning will always choose the SMuFL glyph over the written or gestural
                             // accidentals. SPECIAL CASE: When the gestural accidental is the same as the written
-                            // accidental (and not NONE), we are sure that the gestural accidental will not lose
+                            // accidental we are sure that the gestural accidental will not lose
                             // information, and therefore we don't need to explicitly set the SMuFL glyph. This ensures
                             // that scores that only feature "regular" accidentals will never have redundant SMuFL
                             // glyphs.
-                            if (current.m_accid == ACCIDENTAL_WRITTEN_NONE
-                                || Att::AccidentalGesturalToWritten(accid->GetAccidGes()) != current.m_accid) {
-                                if (!current.m_glyphName.empty()) {
-                                    accid->SetGlyphName(current.m_glyphName);
-                                    accid->SetGlyphAuth(current.m_glyphAuth);
-                                }
-                                // We have a current.m_accid
-                                else {
-                                    char32_t glyph = Accid::GetAccidGlyph(current.m_accid);
-                                    accid->SetGlyphName(CustomTuning::GetGlyphName(glyph, m_doc));
-                                    accid->SetGlyphAuth("smufl");
-                                }
+                            if (!current.m_glyphName.empty()) {
+                                accid->SetGlyphName(current.m_glyphName);
+                                accid->SetGlyphAuth(current.m_glyphAuth);
+                                accid->IsAttribute(false);
+                            }
+                            // We have a current.m_accid, avoid adding it if it's the same as the gestural accidental.
+                            else if (Att::AccidentalGesturalToWritten(accid->GetAccidGes()) != current.m_accid) {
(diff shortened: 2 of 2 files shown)


infojunkie:
Reverting to `pitch/alter` to derive `@accid.ges` undermines the microtonal support as it was previously formulated.

This is somewhat expected, because whereas notated accidentals were previously used to distinguish between the different tuning tones in the score, now the `pitch/alter` information is needed to make this distinction.

It is possible to salvage the microtonal work by editing the microtonal MusicXML documents manually and adding `pitch/alter` entries that help the importer distinguish between the tones. The actual values of `pitch/alter` are overridden by the entries of the tuning definition - they just need to be unique per accidental.

(In a future PR, I will add the feature to create a custom tuning directly from the `pitch/alter` values, thereby bypassing the need for an external tuning definition altogether.)

I will keep testing more to make sure the above approach holds for all test cases, before I mark this PR as ready for review. I will also prepare a PR with the edited microtonal MusicXML test documents in `verovio.org`.

lpugin:
I re-run the test-suite evaluation after having merge the PR in verovio.org but we still have some errors 
```
[8](https://github.com/rism-digital/verovio/actions/runs/29893115049/job/89930921085#step:17:69)
2 round-trip change(s) detected
****** 003-keys-and-accidentals-advanced.mei: 1104 ******
****** 005-maqam-rast-external-tuning.mei: 6896 ******
```

infojunkie:
Thanks will investigate.

infojunkie:
> I re-run the test-suite evaluation after having merge the PR in verovio.org but we still have some errors

These errors have been there since June 29 (the earliest date at which they appear):
https://github.com/rism-digital/verovio/actions/runs/28391892556/job/84120807292

However, it's so close to the date when I added these test MEI files to `verovio.org` that it might have been the case since the beginning.

Regardless, I think it's a good idea to fix them in this PR.

infojunkie:
Fixed the round-trip differences.

PR #4401, opened 2026-07-17
Support `pt` and `vu` MEI font sizes

### LINKED ISSUE #4399: Numeric MEI font sizes render as drawing units
## Problem

Numeric MEI font sizes are specified in printer points, but Verovio currently passes the numeric value directly to its internal drawing-size field. For example, `<rend fontsize="10pt">Text</rend>` becomes `font-size="10px"` in SVG and renders extremely small.

MEI defines numeric `fontsize` values as printer points (1/72 inch). Verovio uses 100 drawing units per millimeter, so 10pt should become 353 drawing units after rounding.

## Minimal example

```xml
<dir tstamp="1"><rend fontsize="10pt">Text</rend></dir>
```

## Cause

The numeric branches in `View::DrawRend`, `View::DrawSymbol`, and `View::DrawMNum` call `SetPointSize()` with the unconverted MEI value. Percentage and named sizes already scale correctly.

## Proposed fix

Convert numeric point sizes at the rendering boundary with:

```text
points × 25.4 / 72 × 100
```

Keep parsed and serialized MEI values unchanged.

sonovice:
## Summary

Fix numeric MEI font sizes so their declared units are interpreted at rendering time instead of treating numeric values as raw Verovio drawing units.

- track whether `data.FONTSIZENUMERIC` uses `pt` or `vu`, preserving the unit during MEI round trips
- convert point values to drawing units using `points × 25.4 / 72 × 100`
- resolve `vu` values against the active staff's drawing unit and propagate the staff size through text rendering
- apply conversion consistently to `<rend>`, `<symbol>`, and `<mNum>`
- keep percentage and named sizes relative to the current base font size

This makes `fontsize="10pt"` render at 353 drawing units instead of 10, while also adding staff-scaled support for `vu` values.

Fixes #4399

## Testing

- configured and built the command-line toolkit successfully with CMake
- passed the ClangFormat 22 dry-run check for all changed C++ files
- verified an SVG render where `10pt` produces `353px`
- verified staff-scaled rendering of a `2vu` value
- verified MEI round trips preserve both `pt` and `vu` suffixes

## Disclosure

The entire process for this contribution—including issue analysis, implementation, testing, and preparation of this pull request—was completed with the assistance of large language models (LLMs).

### DIFF
--- a/include/vrv/textelement.h
+++ b/include/vrv/textelement.h
@@ -100,6 +100,7 @@ public:
         m_verticalShift = false;
         m_alignment = HORIZONTALALIGNMENT_left;
         m_pointSize = 0;
+        m_staffSize = 100;
         m_actualWidth = 0;
         m_enclose = TEXTRENDITION_NONE;
         m_textEnclose = ENCLOSURE_NONE;
@@ -118,6 +119,8 @@ public:
     bool m_verticalShift;
     data_HORIZONTALALIGNMENT m_alignment;
     int m_pointSize;
+    /** Staff-size percentage used for virtual units; 100 is the unscaled fallback when no staff context exists. */
+    int m_staffSize;
     std::vector<TextElement *> m_enclosedRend;
     data_TEXTRENDITION m_enclose;
     data_ENCLOSURE m_textEnclose;
--- a/include/vrv/view.h
+++ b/include/vrv/view.h
@@ -721,9 +721,15 @@ protected:
     ScoreDef m_drawingScoreDef;
 
 private:
+    /** Convert a numeric MEI font size to Verovio drawing units. */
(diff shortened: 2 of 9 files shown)


lpugin:
@sonovice please sign and send the [CLA](https://rism.digital/resources/verovio-cla.pdf). Thanks

lpugin (approved):

sonovice:
@lpugin sent to your @rism.digital address. 👍

lpugin:
@sonovice it seems we haven't received the CLA. Please double check.

sonovice:
Apologies, I used your direct email address instead of `info@rism.digital`. Just resent to the correct address.

PR #4403, opened 2026-07-17
Render ending labels in preference to numbers

sonovice:
## Summary

Prefer an ending's `@label` over its `@n` value when choosing the text to render.

`@n` provides the ending number, while `@label` supplies explicit display text. When both attributes are present, the label should therefore take precedence so that the encoded presentation is respected. Endings without a label continue to render their `@n` value as before.

## Testing

- built the command-line toolkit successfully with CMake
- passed `git diff --check`
- passed the ClangFormat 22 dry-run check

## Disclosure

The process for this contribution was completed with the assistance of large language models (LLMs).

### DIFF
--- a/src/view_control.cpp
+++ b/src/view_control.cpp
@@ -3177,7 +3177,7 @@ void View::DrawEnding(DeviceContext *dc, Ending *ending, System *system)
 
         const int unit = m_doc->GetDrawingUnit(staffSize);
         if (ending->HasN() || ending->HasLabel()) {
-            const std::string endingText = (ending->HasN()) ? ending->GetN() : ending->GetLabel();
+            const std::string endingText = (ending->HasLabel()) ? ending->GetLabel() : ending->GetN();
             std::stringstream strStream;
             // Maybe we want to add ( ) after system breaks? Or . as a styling options?
             if ((spanningType == SPANNING_END) || (spanningType == SPANNING_MIDDLE)) strStream << "(";


ahankinson:
In MEI, `@label` is generally not supposed to represent document content.

“ Captures text to be used to generate a label for the element to which it’s attached, a "tool tip" or prefatory text, for example. Should not be used to record document content. Value is plain text.”

https://music-encoding.org/guidelines/v5/elements/measure.html

So I don’t think it should render in place of `@n`

sonovice:
@ahankinson Thanks for the super quick check. I am aware that the MEI specs do not consider `@label` to be visible. However, Verovio already shows what's in `@label` if `@n` is missing so this is merely a precedence fix. 

Unfortunately, the MEI specs lack a way of expressing text in such situations since `@n` is not allowed to contain whitespaces.

ahankinson:
As a fallback it’s fine, but `@n` should definitely have precedence.

lpugin:
I agree that if we support both, then showing `@label` when both are given makes more sense then reverse. `@n` is the common one one would expect. But the main point is that it is more restrictive because is it a number-like, so being able to provide a label instead would be useful.

ahankinson:
Is there an example of such a label you could provide?

sonovice:
image

lpugin:
Some background to the use of `@label`. I added it when working on the MuseScore exporter. `@n` was too restrictive, and `@label` gave a bit more flexibility since it allows spaces. I did not really think about the case where both would be given, and looking at it now I think I put the priorities wrong.

ahankinson:
But MEI clearly states that `@label` shouldn't be used for source contents, which seems to be the case in the example shown. There seems to be a use case for allowing source-derived labels of some sort that is less restrictive than currently allowed by `@n`, but as it stands, using `@label` for this kind of contents, and then prioritizing it over `@n`, seems to be contrary to what is specified in MEI.

lpugin:
I agree that using `@label` is not really correct. Maybe one day we will have `label` elements for that - or something along these lines. Until then, giving precedence to `@label` is actually one step in that direction - assuming you would not give precedence to `@n` over a `label` element, would you?

ahankinson:
Well, given that `@n` is expected to be source content, while `@label` is supposed to be a "tool tip" or something, then yeah, I would give preference to `@n` over `@label`. 

Like I said, I think there is a valid use case here. But an attribute is not a great place to store text like this because, apart from whitespace, this text could also hold line endings and other things which can be tricky to include (e.g., Windows vs. Unix line endings, font sizes or faces).  So a 'proper' label text should be specced; Something that can take the text formatting elements (e.g., `<rend>`)

The only solution I can see here as it is, is adding a `<dir>` to the child measure element in the `<ending>`.

lpugin:
I am OK with this change. Follow-up discussion should be moved to the MEI repository since this is where it should be improved further in the first place.

ahankinson:
Done: https://github.com/music-encoding/music-encoding/issues/1818

lpugin:
Great, thanks.

PR #4404, opened 2026-07-17
[WIP] Add runtime custom font registration and rendering

sonovice:
## Status

This is an early WIP for discussion, especially around the public API, dependency footprint, packaging, and migration path.

The branch is currently stacked on #4401 because runtime text shaping needs its MEI font-size unit support. Until that PR lands and this branch is rebased, the comparison also contains that commit.

## Motivation

MEI can select a font family with `fontname`, but applications need a way to provide the corresponding font data to Verovio. Relying on fonts installed on the host is not sufficient for browser/WASM use, server-side rendering, mobile applications, or other embedded environments, and it makes output depend on the machine performing the render.

At present, Verovio lays out text and music using precomputed metrics for a selected set of bundled SMuFL fonts and, for text, Times only. Using another text font is therefore not actually possible with correct layout. Replacing the emitted font through CSS changes the visible glyphs, but Verovio has already positioned and wrapped the text using the original metrics. With a different font's widths, advances, ascenders, and descenders, text, enclosures, and surrounding notation no longer align correctly.

Supporting arbitrary fonts consequently requires real shaping and measurement with the actual font data before layout, not just a way to substitute a CSS font family after rendering.

The goal of this PR is to make font rendering self-contained and deterministic while allowing an application to use its own licensed or project-specific text and SMuFL fonts. The same MEI should render with the same glyph outlines wherever Verovio runs, provided that the application registers the same font files.

## Approach

This introduces a runtime font store that accepts static OpenType fonts and uses HarfBuzz for text shaping. OTF, TTF, WOFF1, and WOFF2 input is supported. Registered fonts are used for MEI text at the element or nested text-run level, including family, weight, and style selection. When a requested bold or italic face is unavailable, a synthetic face is used rather than silently reverting to the regular face.

Registration returns the font's canonical family name. Callers may additionally assign an exact, case-sensitive alias, which is useful when existing MEI refers to a project name rather than the name stored in the font. For example, a font whose canonical family is `Quicksand` can be registered under `QS` for `<rend fontname="QS">`. Aliases are family-wide and may exist independently in the text and music namespaces.

Music fonts are registered with their SMuFL metadata. Missing music glyphs fall back to Bravura; missing text clusters fall back to the corresponding Tinos face and then `.notdef`. Bravura and the regular, italic, bold, and bold-italic Tinos faces provide the bundled defaults. SVG output contains deduplicated glyph paths and `use` elements, so viewers do not need the fonts installed.

Registering a font or alias after an MEI document has been loaded invalidates layout, and the next render performs layout again with the newly available metrics. Re-registering the same bytes and alias is idempotent; alias collisions fail without changing the registry.

## APIs and bindings

All registration paths support optional aliases and return the canonical family name (or an empty string on failure):

- C++ accepts byte buffers and files through `RegisterTextFont` / `RegisterMusicFont`, their Base64 counterparts, and file-based variants.
- JavaScript accepts `Uint8Array` or `ArrayBuffer`; browsers fetch or otherwise obtain the bytes explicitly because WebAssembly cannot access arbitrary locally installed fonts.
- Python accepts `bytes`, Base64 strings, or filenames.
- Go accepts `[]byte` and exposes alias-specific methods such as `RegisterTextFontWithAlias`, with corresponding music, Base64, and file methods.
- Swift accepts `Data` and uses an `alias:` overload, with Base64 and file variants as well.
- Java exposes the corresponding byte-array, Base64, and file overloads through SWIG.

For example, a browser application can register a font before loading or rendering its MEI:

```js
const response = await fetch("/fonts/Quicksand-Regular.woff2");
const bytes = new Uint8Array(await response.arrayBuffer());
const family = toolkit.registerTextFont(bytes, "QS");
// family === "Quicksand"
```

Equivalent binding calls include:

```python
family = toolkit.registerTextFont(font_bytes, "QS")
```

```go
family := toolkit.RegisterTextFontWithAlias(fontBytes, "QS")
```

```swift
let family = toolkit.registerTextFont(fontData, alias: "QS")
```

The CLI options are repeatable:

```sh
verovio --font-add-text MyText.woff2 \
  --font-add-music MyMusic.otf input.mei
```

Explicit aliases can be supplied with:

```sh
verovio --font-add-text-as "QS=/fonts/Quicksand-Regular.woff2" \
  --font-add-music-as "HouseMusic=/fonts/MyMusic.otf" input.mei
```

The alias forms are also exposed as the `fontAddTextAs` and `fontAddMusicAs` string-array Toolkit options. Text and music defaults remain selectable with `fontText` and `font`; `fontFallback` selects the music fallback.

For WebAssembly builds, the bundled runtime resources are packaged in an adjacent `verovio.data` sidecar. Emscripten's `locateFile` hook can be used when an asset pipeline serves that file from another location.

## Scope and compatibility

This initial implementation intentionally supports static outline fonts only. Font collections, variable fonts, color/bitmap fonts, AAT, and bidirectional layout are outside the current scope. Font input is limited to 32 MiB and decoded data to 64 MiB.

`fontAddCustom` remains as a deprecated ZIP adapter for one compatibility release. `fontTextLiberation` and `smuflTextFont` are deprecated as well. Families that are no longer bundled can still be used by registering them at runtime.

The implementation vendors the HarfBuzz shaping sources and the WOFF2/Brotli decoding subset used by Verovio. I would particularly appreciate maintainer feedback on that integration and on how the sidecar resources should fit the release packages.

## Testing

- Clean CMake Release build with `BUILD_FONT_TESTS=ON`
- `font-runtime-registration`: registration formats, aliases and collisions, style selection/synthesis, fallback behavior, shaping, and late-registration relayout
- `font-runtime-cli-alias`: text and music aliases supplied through the CLI

Both runtime-font tests pass locally on macOS with AppleClang 21.

### Performance and delivery size

The implementation was benchmarked in controlled alternating baseline/head runs against `origin/develop`, using 30-sample batches. The native and WASM cold/warm timing results were:

| Runtime / fixture | Cold: before → after (delta) | Warm: before → after (delta) |
| --- | ---: | ---: |
| Native music-only | 5.549 → 5.061 ms (-8.80%) | 0.302337 → 0.000239 ms (-99.92%) |
| Native text-heavy | 5.024 → 4.738 ms (-5.69%) | 0.144355 → 0.001044 ms (-99.28%) |
| Native mixed-pages | 6.912 → 6.346 ms (-8.18%) | 0.151341 → 0.000234 ms (-99.85%) |
| Native lyrics | 5.124 → 4.718 ms (-7.93%) | 0.162780 → 0.000218 ms (-99.87%) |
| WASM music-only | 7.443 → 7.038 ms (-5.45%) | 0.197100 → 0.018818 ms (-90.45%) |
| WASM text-heavy | 6.823 → 6.725 ms (-1.44%) | 0.114220 → 0.061410 ms (-46.23%) |
| WASM mixed-pages | 8.561 → 8.058 ms (-5.87%) | 0.108730 → 0.018261 ms (-83.20%) |
| WASM lyrics | 7.160 → 6.564 ms (-8.32%) | 0.122325 → 0.015232 ms (-87.55%) |

The final native and production no-Humdrum WASM delivery measurements are:

| Artifact | Raw: before → after (delta) | Gzip: before → after (delta) | Brotli: before → after (delta) |
| --- | ---: | ---: | ---: |
| Stripped native CLI | 6,555,080 → 6,308,984 bytes (-3.75%) | 2,335,088 → 2,357,504 bytes (+0.96%) | 1,656,127 → 1,684,494 bytes (+1.71%) |
| Installed native bundle | 12,206,080 → 8,611,840 bytes (-29.45%) | 3,521,334 → 3,630,628 bytes (+3.10%) | 2,659,425 → 2,905,659 bytes (+9.26%) |
| WASM JavaScript + `verovio.data` | 7,013,380 → 4,431,485 bytes (-36.81%) | 2,274,929 → 2,237,721 bytes (-1.64%) | 1,768,938 → 1,958,890 bytes (+10.74%) |

The final synthetic bold/italic follow-up was also measured against its immediate checkpoint. Raw CLI size changed from 7,259,208 to 7,260,600 bytes (+0.019%), and gzip size from 2,437,809 to 2,441,144 bytes (+0.137%). Cold timing changed from 5.440 to 5.305 ms (-2.49%) for music-only, 4.931 to 4.978 ms (+0.95%) for text-heavy, and 6.817 to 7.044 ms (+3.33%) for mixed pages. One 1,000-render text-heavy cache-hit batch measured 0.808 versus 0.962 microseconds per call (+19.07%); a 100,000-render diagnostic measured 0.717 versus 0.730 microseconds per call (+1.81%).

The process for this contribution was completed with the assistance of large language models (LLMs).


sonovice:
One more thought on the defaults: I chose `Bravura` due to SMuFL glyph coverage and `Tinos` for metric compatibility with `Times`. However, since we would be potentially free to choose, it might make sense to go with `Academico` for the text since it was designed to go together with `Bravura`.

No hard preference, of course. I just thought it might be a logical choice.

sonovice:
@lpugin Is this something you would like to add at some point? Or should I just close it?

lpugin:
I want to test I further, so you can leave it open for now.

Issue #3192, opened 2023-01-07
 ClassId for 'xxx' not found


Duanle:
**Describe the problem**
After generating SVG for several times with 'svgAdditionalAttribute', this exception will occur

**Verovio information**
 - Version of Verovio [Verovio 3.15.0-dev-920f376]
 - Binding [NDK]

**Environment information (as appropriate)**
 - OS: [android]

**Screenshots**
![WX20230107-130016](https://user-images.githubusercontent.com/6520982/211132772-f5efd269-de99-4bc0-91a8-b1b24801f2be.png)

lpugin:
Your description of the problem is not detailed enough and you did not provide input data. I will close the issue unless you make the effort to make the issue understandable and that it can be reproduced.

Duanle:
**Describe the problem:**
I call verovio to render svg through jni on an android device. On my android device, this problem is not limited to a certain xml. This problem can occur with every xml. This problem does not occur when I first call 'toolkit. RenderToSVGFile()' for rendering after I start the app and load '. so' lib. However, after several times of rendering (re rendering the current xml or switching xml), this problem occurs. At this time, if I kill the app and restart it, I repeat the process again, which results in no 'dur' tags such as' note 'in svg.
The same version of verovio source code, which I compiled on the Mac, does not have this problem

**This is my jni code:**
`
  
extern "C" JNIEXPORT jobjectArray JNICALL
renderStaffFiles(JNIEnv *env, jobject, jstring jResourcePath, jstring jInputFilePath,jstring jOptionJson) {
    const char *resourcePath = env->GetStringUTFChars(jResourcePath, NULL);
    const char *inputFilePath = env->GetStringUTFChars(jInputFilePath, NULL);
    const char *optionJson = env->GetStringUTFChars(jOptionJson, NULL);

    std::string fileSuffix;
    std::string fileName;
    std::string folderPath;

    getInputFileSuffix(inputFilePath, fileSuffix);
    getInputFileName(inputFilePath, fileName);
    getInputFolderPath(inputFilePath, folderPath);

    vrv::Toolkit toolkit(false);
    toolkit.SetResourcePath(resourcePath);
    toolkit.SetOptions(optionJson);
    toolkit.LoadFile(inputFilePath);

    int count = toolkit.GetPageCount();
    auto jArray = (jobjectArray) env->NewObjectArray(count, env->FindClass("java/lang/String"), env->NewStringUTF(""));

[...]

PR #4418, opened 2026-07-31
Make object factory registrations available across threads

ro-ag:
The object factory's registrars run during process startup, but its registries were marked `thread_local`. As a result, toolkits running on another thread could see an empty registry.

One visible symptom was:

    ClassId for 'staff' not found

This also prevented options such as:

    {"svgAdditionalAttribute": ["staff@n"]}

from adding `data-n` attributes when SVG rendering happened on a worker thread.

This change moves the registry to process-wide, lazily initialized storage and protects it with a shared mutex. Constructor callbacks are copied while holding the read lock and invoked after releasing it, so object construction is not serialized.

I reproduced the failure before the change and checked the fix with:

- a worker-thread class lookup
- an end-to-end worker-thread SVG render using `staff@n`
- an eight-thread stress test repeatedly resolving and constructing staff objects
- a complete debug shared-library build
- clang-format and diff checks

Related to #3192.

### DIFF
--- a/include/vrv/object.h
+++ b/include/vrv/object.h
@@ -1114,10 +1114,6 @@ public:
      * Get the correspondings ClassIds from the vector of MEI element string names
      */
     void GetClassIds(const std::vector<std::string> &classStrings, std::vector<ClassId> &classIds);
-
-public:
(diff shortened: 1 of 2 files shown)


lpugin:
Replaced by #4420. Please test.

Issue #4430, opened 2026-08-15
[MusicXML] Crash on `senza-misura` with symbol


infojunkie:
Given the [attached file from the LilyPond MusicXML test suite](https://github.com/user-attachments/files/31106914/11h-TimeSignatures-SenzaMisura.zip), running Verovio 6.3.0-dev-fb5c4db gives:

```shell
$ verovio 11h-TimeSignatures-SenzaMisura.xml
verovio: /path/to/verovio/src/doc.cpp:1877: int vrv::Doc::GetGlyphWidth(char32_t, int, bool) const: Assertion `glyph' failed.
Aborted (core dumped)
```

Removing the `<senza-misura>` value "X" at line 86 element avoids the crash.

infojunkie:
For context, I am currently working on syncing the [musicxmlTestSuite](https://github.com/w3c-cg/musicxmlTestSuite) (which Michael Cuthbert donated to the W3C Music Notation Community Group) with the latest advances from the [LilyPond MusicXML test suite](https://gitlab.com/lilypond/lilypond/-/tree/master/input/regression/musicxml?ref_type=heads), from which the former was originally forked. Since Verovio uses these files for regression, I want to make sure that the updated suite won't break Verovio.

None of the other test files in the suite crash Verovio.

infojunkie:
Looking at the MEI schema, I _think_ a `<symbolDef>` should be created with an SVG `<text>` element containing the `<senza-misura>` text. This new symbol would be added to the `meterSig` during MusicXML import at https://github.com/rism-digital/verovio/blob/fb5c4db7e9a9d3214a204c4a83c2d2e011782a52/src/iomusxml.cpp#L1799-L1801

But there's another issue of handling empty values. As an experiment, I transformed the MusicXML to MEI and got the following:
```xml
<meterSig xml:id="k1eqr2yz" sym="open" />
```
which presumably causes the crash because there's no glyph associated with the time signature. Adding a supported SMuFL glyph avoids the crash:
```xml
<meterSig xml:id="k1eqr2yz" sym="open" glyph.name="timeSig0" />
```
whereas adding a non-supported SMuFL glyph also causes a crash:
```xml
<meterSig xml:id="k1eqr2yz" sym="open" glyph.name="timeSigX" />
```
which suggests that the SVG export code for time signatures should guard against empty glyphs.

craigsapp:
[...]

PR #4431, opened 2026-08-16
Guard against empty glyph code

infojunkie:
To fix the crashing part of #4430. This PR ensures that an empty glyph code does not get processed while drawing a meter signature.


lpugin:
Replaced by a more generic fix in #4432

PR #4434, opened 2026-08-18
Fix page lookup for timed elements

### LINKED ISSUE #4395: getElementsAtTime().page broken with default options
Starting with verovio 6.0.0, when default options are used, getElementsAtTime().page is always 1. The returned `notes` / `chords` / `rests` are correct. 

If using `expandAlways` (realizing repeats and expansions for all outputs) **or** `expandNever` (expanding nothing regardless of the output), the reported page is correct. Only when none of these options are set (expanding repeats and expansions only in MIDI and timemap outputs) the page reported is broken.

The html test case below loads Verovio 6.2.0 from the CDN, renders a two-page score, and prints
`.page` at three timestamps for three option settings. At 120 bpm each quarter note is 500 ms,
so page 1 = 0–4000 ms and page 2 = 4000–8000 ms; expected `.page` for `[500, 4500, 7500]` ms
is `1, 2, 2`.

<details>
<summary>Minimal test case (click to expand)</summary>

```html
<!DOCTYPE html>
<meta charset="utf-8">
<script src="https://www.verovio.org/javascript/6.2.0/verovio-toolkit-wasm.js"></script>
<pre id="out">loading…</pre>
<script>
const mei = `<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0"><music><body><mdiv><score>
<scoreDef><staffGrp><staffDef n="1" lines="5" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4"/></staffGrp></scoreDef>
<section>
<measure n="1"><staff n="1"><layer n="1"><note dur="4" oct="4" pname="c"/><note dur="4" oct="4" pname="d"/><note dur="4" oct="4" pname="e"/><note dur="4" oct="4" pname="f"/></layer></staff></measure>
<measure n="2"><staff n="1"><layer n="1"><note dur="4" oct="4" pname="g"/><note dur="4" oct="4" pname="a"/><note dur="4" oct="4" pname="b"/><note dur="4" oct="5" pname="c"/></layer></staff></measure>
<pb/>
<measure n="3"><staff n="1"><layer n="1"><note dur="4" oct="5" pname="c"/><note dur="4" oct="4" pname="b"/><note dur="4" oct="4" pname="a"/><note dur="4" oct="4" pname="g"/></layer></staff></measure>
<measure n="4"><staff n="1"><layer n="1"><note dur="4" oct="4" pname="f"/><note dur="4" oct="4" pname="e"/><note dur="4" oct="4" pname="d"/><note dur="4" oct="4" pname="c"/></layer></staff></measure>
</section></score></mdiv></body></music></mei>`;

verovio.module.onRuntimeInitialized = () => {
  const lines = [`verovio ${new verovio.toolkit().getVersion()}  (expected .page = 1,2,2)`, ""];
  for (const [name, opts] of [
    ["(no options)",       {}],
    ["expandAlways:true",  { expandAlways: true }],
    ["expandNever:true",   { expandNever: true }],
  ]) {
    const tk = new verovio.toolkit();
    tk.setOptions(Object.assign({ breaks: "encoded" }, opts));
    tk.loadData(mei);
    const page = [500, 4500, 7500].map(ms => tk.getElementsAtTime(ms).page);
    lines.push(`${name.padEnd(20)} .page = [${page.join(",")}]` + (page.join(",") !== "1,2,2" ? "   <-- WRONG" : ""));
  }
  out.textContent = lines.join("\n");
};
</script>
```

</details>

I think this is happening because wiith default options, `SetMidiDoc()` builds now a separate `Doc` for MIDI/timemap that only gets`PrepareData()` and is never cast off into pages. So every measure sits on the single
un-paginated page, `GetIdx()` is always `0`, and `pageNo` is always `1`.

When `expandAlways` or `expandNever` is set, `SetMidiDoc()` instead points `m_midiDoc` at the
main (paginated) `m_doc`, so the page is correct.

aribradshaw:
Fixes #4395.

## What changed

When the default expansion behavior creates a separate MIDI document, `GetElementsAtTime()` now maps the timed measure back to its notated ID and asks the rendered document for that element's page. When the MIDI and rendered documents are the same object, it retains the direct ancestor lookup.

This prevents the unpaginated MIDI document from reporting page 1 for every timestamp while preserving the existing behavior for `expandAlways` and `expandNever`.

A Go binding regression test loads an encoded two-page score and verifies that a timestamp after the page break reports page 2.

## Validation

- `cmake --build build/python --config Release`
- Local Python binding reproduction: pages `1,2,2` at timestamps before and after the encoded page break
- `gofmt` on the Go regression test
- `git diff --check`

The Go test is included in the repository's Ubuntu CI job. The local Windows Go runner could not execute cgo because GCC is not installed, so upstream CI will provide the independent binding-suite result.

### DIFF
--- a/bindings/go/verovio_test.go
+++ b/bindings/go/verovio_test.go
@@ -82,3 +82,13 @@ func TestLoadDataAndRenderSVG(t *testing.T) {
 		t.Fatal("expected SVG output")
 	}
 }
+
+func TestElementsAtTimeUsesRenderedPage(t *testing.T) {
+	tk := NewToolkitWithResourcePath("../../data")
+	defer tk.Close()
+	tk.SetOptions(`{"breaks":"encoded"}`)
+	tk.LoadData(`<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0"><music><body><mdiv><score><scoreDef><staffGrp><staffDef n="1" lines="5" clef.shape="G" clef.line="2" meter.count="4" meter.unit="4"/></staffGrp></scoreDef><section><measure n="1"><staff n="1"><layer n="1"><note dur="4" oct="4" pname="c"/><note dur="4" oct="4" pname="d"/><note dur="4" oct="4" pname="e"/><note dur="4" oct="4" pname="f"/></layer></staff></measure><measure n="2"><staff n="1"><layer n="1"><note dur="4" oct="4" pname="g"/><note dur="4" oct="4" pname="a"/><note dur="4" oct="4" pname="b"/><note dur="4" oct="5" pname="c"/></layer></staff></measure><pb/><measure n="3"><staff n="1"><layer n="1"><note dur="4" oct="5" pname="c"/><note dur="4" oct="4" pname="b"/><note dur="4" oct="4" pname="a"/><note dur="4" oct="4" pname="g"/></layer></staff></measure></section></score></mdiv></body></music></mei>`)
+	if got := tk.ElementsAtTime(4500); !strings.Contains(got, `"page":2`) {
+		t.Fatalf("expected page 2, got %s", got)
+	}
+}
--- a/bindings/go/verovio_test.go
+++ b/bindings/go/verovio_test.go
@@ -82,13 +82,3 @@ func TestLoadDataAndRenderSVG(t *testing.T) {
 		t.Fatal("expected SVG output")
 	}
 }
-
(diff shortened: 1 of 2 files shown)


lpugin (changes requested):
Looks good. Please drop the test in the go directory.

lpugin (approved):

Issue #4302, opened 2026-03-18
Integrate or update the Typescript type definitions


infojunkie:
**Is your feature request related to a problem? Please describe.**
The Verovio Typescript definition at https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/verovio dates back from version 5.0 and needs to be updated.

**Describe the solution you'd like**
Either update the definitions there, or add them to the npm package here. I would suggest the latter to ensure that the type definitions no longer go out of sync.

**Additional context**
I'll be happy to submit a PR for that once the approach is agreed upon.

ahankinson:
It looks like @bbloomf added the existing ones. Any thoughts?

bbloomf:
I agree that including them here would make the most sense, and that ideally they could get generated as part of the build process.

infojunkie:
There [may be a way to generate automatic d.ts from Emscripten](https://stackoverflow.com/questions/76691590/how-can-i-generate-typescript-types-for-an-emscripten-project), but initially I was thinking of manual updates at point releases.

lpugin:
It looks like generating generating everything automatically will not be easy. It might be possible for the Verovio options. @bbloomf could you tell more about how you get these?

bbloomf:
@lpugin  I had put together a script years ago to parse them from verovioToolkit.getAvailableOptions().

Here is the code I used, just plain javascript:

```
const createVerovioModule = require('verovio/wasm').default;
const { VerovioToolkit } = require('verovio/esm');

const fs = require("fs");
const typeMap = {
    bool: "boolean",
    "std::string": "string",
    int: "number",
    double: "number",
    "std::string-list": "string[]",
    array: "string[]",
};
async function main() {
    const VerovioModule = await createVerovioModule();
    const tk = new VerovioToolkit(VerovioModule);
    const availableOptions = tk.getAvailableOptions();
    console.info(tk.getVersion());
    const types = new Set();
    try {
        const optionSegments = Object.entries(availableOptions.groups).map(
            ([key, group]) =>
                `    /**${"*".repeat(group.name.length)}**
     * ${group.name} *
     **${"*".repeat(group.name.length)}**/
` +
                Object.entries(group.options)
                    .map(
                        ([
                            key,
                            {
                                cmdOnly,
                                shortOption,
                                title,
                                type,
                                description,
                                values,
                                ...option
                            },
                        ]) =>
                            cmdOnly
                                ? ""
                                : `
    /**
[...]

PR #4440, opened 2026-09-01
Integrate TypeScript type definitions

infojunkie:
This PR integrates the TypeScript definitions for Verovio into the npm package.

Addresses #4302


infojunkie:
This first commit integrates a copy of the original `@types/verovio` package into the npm build. The important changes are:

- A folder `emscripten/npm/src/types` that is a copy of the types definitions at `@types/verovio`
- An updated `emscripten/npm/package.json` that now includes type definitions
- A new npm build step that copies the type definitions to `emscripten/npm/dist` 
- A new folder `emscripten/npm/test` that tests TypeScript usage, for both CommonJS and ESM scenarios

My next steps would be:

- Update the type definitions to match the latest Verovio interface (manually or automatically or a bit of both)
- Integrate the tests above into the GitHub CI steps

@lpugin @ahankinson please let me know if you have comments at this stage before I invest more time into the above. Thanks!

emscripten/npm/package.json
```diff
@@ -3,20 +3,36 @@
   "version": "6.4.0-alpha",
   "description": "This is the stable version of the verovio package",
   "main": "dist/verovio-toolkit-wasm.js",
+  "types": "dist/types/index.d.ts",
   "exports": {
-    ".": "./dist/verovio-toolkit-wasm.js",
+    ".": {
+      "import": "./dist/verovio-toolkit-wasm.js",
+      "require": "./dist/verovio-toolkit-wasm.js",
+      "default": "./dist/verovio-toolkit-wasm.js",
```
lpugin: Why is this change neeeded

lpugin:
This looks good to me. Updating will be a mix of automatic and hand coding.

emscripten/npm/package.json
infojunkie: Not sure which change specifically you meant, but I removed some redundant declarations. Without the remaining `"types"` entry, the CJS compilation fails.

lpugin:
It would be good to have a way to distinguish files that are generated and files hand coded. A comment at the beginning of generated files would be a start. See https://github.com/rism-digital/verovio/blob/ddac95fa234ce3fa66c5827030a326d8242f948a/libmei/dist/attclasses.h#L10-L13

infojunkie:
@lpugin are you interested in running the TypeScript test script https://github.com/rism-digital/verovio/pull/4440/changes#diff-5ed6b88f6db589453fda94202c3019b74c9326771c1a16c104e9e47195098a4bR8 during CI ? It will catch unexpected changes in the automatic generation of the interfaces as well as problems building the npm package.

As far as I can tell, the npm package is not built during CI, so that whole process would need to be added as a step.

infojunkie:
@lpugin I've also added documentation to `EngravingDefaults` based on how they relate to registered options. I am not sure I understand the conversion of "MEI units" to "staff spaces" - could you please advise on that?

emscripten/npm/generate-verovio-options.cjs
```diff
@@ -62,13 +62,17 @@ function generateEngravingDefaults(availableOptions) {
...
+        const optionEntry = optionTypes.get(optionKey);
+        if (!optionEntry) {
             problems.push(`${jsonKey}: option ${optionKey} not found in runtime options`);
             continue;
         }
-        const tsType = typeMap[type] || type;
-        lines.push(`    ${jsonKey}: ${tsType};`);
+        const type = typeMap[optionEntry.type] || optionEntry.type;
+        lines.push(`    /**
+     * ${optionEntry.description.replace('MEI units', 'staff spaces').trim()}
```
lpugin: Why do you want this and not keep MEI units?

lpugin:
> Are you interested in running the TypeScript test script? It will catch unexpected changes in the automatic generation of the interfaces as well as problems building the npm package?

Changes can occur when new options are added or options are changed. So as long as changes are not blocking this can be good to have.

emscripten/npm/generate-verovio-options.cjs
infojunkie: Here's what I see in `Options::sync()` 

https://github.com/rism-digital/verovio/blob/develop/src/options.cpp#L2143-L2150

My understanding is that there's a unit conversion happening between engravingDefaults and registered options.

infojunkie:
@lpugin to update the `EditorAction.d.ts` with the latest, I need your help because I am not familiar at all with the editor status (`EditorStatus`) and response (`EditorResponse`) schemas.

Update: I made good progress with the editor types, based on my reading of the `editortoolkit_*` C++ files.

infojunkie:
@lpugin I think I need your guidance here for the best approach to run the TypeScript tests during CI.

I see that [`ci_build` does build the JS artifacts](https://github.com/rism-digital/verovio/blob/develop/.github/workflows/ci_build.yml#L316), but not the npm package if I'm not mistaken. 

What would be the best approach to do so?

- Re-run `buildNpmPackage` from scratch in `test_build` which would significantly increase the build time, or
- Access the built JS artifacts in `ci_build` (e.g. via `actions/download-artifact`) to manually build the npm package in `test_build`. 
- Something else?

Thanks for your advice!

infojunkie:
@lpugin  I added a temporary workflow `npm-ci-module` that copies steps from `ci_build` to demonstrate building the npm module and successfully running the npm test. After successful runs, I merged the new job `build_npm` back to `ci_build`. You can see it running here https://github.com/rism-digital/verovio/actions/runs/35526552900/job/106120661080

emscripten/npm/src/types/EditorTypes.d.ts
```diff
@@ -0,0 +1,285 @@
...
+        element: string;
+        secondaryId?: string;
+    }
+} & (
+    | { insertMode: true; insertion: EditorStatusInsertion; }
+    | { insertMode: false; insertion?: undefined; }
+)
+export type EditorStatus =
+  | { status: EditorStatusStatus; } & (EditorStatusInfo | EditorStatusMessage) // If status exists, expect extra fields
+  | { status?: undefined; } & EditorStatusInfo; // If status is undefined, only expect info
```
lpugin, line 254: This is what I would expect
```
type EditStatus {
    chainedId: string;
    canUndo: boolean;
    canRedo: boolean;
    isMensuralMusicOnly: boolean;
    insertMode: boolean;
    selection: {
        id: string;
        element: string;
        secondaryId?: string;
    } | null;
    insertion: {
        chordMode: boolean;
        restMode: boolean;
        pname: string;
        oct: number;
        dots: number;
        accid: string;
        accidImplicit: boolean;
    } | null;
    invalidLayout: boolean | null;
}
```

emscripten/npm/src/types/EditorTypes.d.ts
```diff
@@ -0,0 +1,285 @@
...
+    context: EditorResponseObjectContext;
+    object: {} | {
+        attributes: {
+            [key: string]: string;
+        };
+        text: string;
+    }
+    referringElements: EditorResponseReferencesContext[];
+    referencedElements: EditorResponseReferencesContext[];
+}
```
lpugin: This is what I would expect
```
type TreeNodeData {
    id: string;
    element: string;
    children?: TreeNodeData[];
    isLeaf?: boolean;
    text?: string;
    attributes?: Record<string, unknown>;
};

type ReferenceObject extends TreeNodeData {
    referenceAttribute: string;
}

type EditResponseContent {
    ancestors: TreeNodeData[];
    context: TreeNodeData;
    object: TreeNodeData;
    referencedElements: ReferenceObject[];
    referringElements: ReferenceObject[];
}
```

emscripten/npm/src/types/EditorTypes.d.ts
infojunkie: There are several differences between what I wrote and what you suggest, so I'll try to articulate the differences and explain my approach:

- Regarding `selection?` vs `selection: {...} | null` : The first says that `selection` key can be left out of the object altogether, while the second says that it _is_ present in the structure and equals null. When I look at [the code for "selection"](https://github.com/infojunkie/verovio/blob/44119792adedd918efed83a7351c7841a98e20c4/src/editortoolkit_shared.cpp#L590), I see that it is not always defined, hence the choice of `selection?`. 
- Same for the other `?` vs `| null` occurrences.
- Regarding the relationship between `insertMode` and `insertion`: `insertion` is [only defined if `insertMode` is true](https://github.com/infojunkie/verovio/blob/44119792adedd918efed83a7351c7841a98e20c4/src/editortoolkit_shared.cpp#L592). Hence it is a stricter definition to relate them via the union type operator.

emscripten/npm/src/types/EditorTypes.d.ts
infojunkie: I've renamed and restructured the types here. Two comments from me:

- To extend `type` definitions, `extend` is not valid and instead needs the intersection operator `&`
- Same approach as [above comment](https://github.com/rism-digital/verovio/pull/4440#discussion_r4067730534) regarding tying `isLeaf` with the presence or absence of `children`

ahankinson:
The `verovio-tests.cts` and `verovio-tests.mts` files have the same logic in them. I understand why it's necessary to have a different 'wrapper' around them but is there some way that the implementation could be consolidated?

infojunkie:
> I understand why it's necessary to have a different 'wrapper' around them but is there some way that the implementation could be consolidated?

In addition to validating the TypeScript definitions, I think one important role of those 2 test files is to supply real-life examples of TypeScript code that uses Verovio. (When I approach a new library, I typically read the tests first to understand how to use it.) As such, we want the test files to be as straighforward as possible. 

From the validation point of view, the test logic itself is less relevant than the ability to invoke the functions and correctly transpile the code in both ESM and CJS modes.

The only ways I could think of unifying the test logic are not fulfilling the above points imo:

- Write the test logic in a separate `verovio-tests.ts` file that is injected as a first step during `npm test` into `verovio-tests.mts` and `verovio-tests.cts`. The problem with this is that IDE syntax highlighting would not function (since the imports are in a different file than their usage in the code.)

- Perform complicated conditional logic to `import` or `require` the dependencies in a single `verovio-tests.ts` file. This would obscure the real-world usage of the library, and I'm not even sure it's feasible.

I'm open to other suggestions.

PS. I did unify the code of the 2 test files to make sure this part is even doable - the only difference is the dependency import section. There is one ESM construct (`import type`) that is not expressible using `require` as far as I can tell, so I had to work around it in the `verovio-tests.cts` header.

